PR #105, opened 2018-11-02
Add progress bars to some commands

ChrisBoehmCA:
* Progress bars were introduced for the following commands: 
   * `zowe zos-jobs submit data-set`
   * `zowe zos-jobs submit local-file`
   * `zowe zos-files download all-members`
   * `zowe zos-files upload dir-to-pds`


packages/zosfiles/src/cli/ZosFilesBase.handler.ts
```diff
@@ -37,6 +37,7 @@ export abstract class ZosFilesBaseHandler implements ICommandHandler {
 
         const response = await this.processWithSession(commandParameters, session, profile);
 
+        commandParameters.response.progress.endBar(); // end any progress bars
```
tucker01: Does there need to be a similar statement in ZosmfBaseHandler? Also, do we terminate any outstanding progress bars in the event of an error in the command processor?

tucker01: If it's being done under error conditions (and success) in Imperative, I think we're covered. Thanks.

packages/zosfiles/src/api/methods/download/Download.ts
```diff
@@ -168,6 +169,13 @@ export class Download {
              * @param mem - an object with a "member" field containing the name of the data set member
              */
             const createDownloadPromise = (mem: { member: string }) => {
+                // update the progress bar if any
+                if (options.task != null) {
+                    downloadsInitiated++;
```
tucker01: I think we need to re-arrange this to increment when complete?

PR #123, opened 2018-11-27
Fix download data set in binary mode

ChrisBoehmCA:
We were converting binary data to string before saving it. This fixes the issue in the master branch

### DIFF
--- a/packages/zosfiles/src/api/methods/download/Download.ts
+++ b/packages/zosfiles/src/api/methods/download/Download.ts
@@ -81,7 +81,7 @@ export class Download {
             }
 
             // Get contents of the data set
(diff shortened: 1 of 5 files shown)


packages/zosfiles/src/api/methods/download/Download.ts
```diff
@@ -96,7 +96,7 @@ export class Download {
             IO.createDirsSyncFromFilePath(destination);
 
             if ((!options.binary) && (content.toString().length > 0)) {
```
AHumanFromCA: Since we're in this area, I would suggest doing `content.byteLength` over `content.toString().length`.

PR #130, opened 2018-12-07
export utils to be able to use it in plugins

nurra01:
Export utils so plugins are able to use it.


ChrisBoehmCA:
Also, please sign off any commits to imperative or zowe-cli so that the DCO check can pass (use -s flag on the git commit command)

PR #162, opened 2018-12-17
workflow start

muzikovam:
System tests are in progress. Creating pull request so you can have any requests meanwhile system tests are being done so then I don't have to change that much :)

### DIFF
--- a/packages/index.ts
+++ b/packages/index.ts
@@ -18,4 +18,5 @@ export * from "./zosjobs";
 export * from "./zostso";
 export * from "./zosmf";
 export * from "./provisioning";
+export * from "./workflows";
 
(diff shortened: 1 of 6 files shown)


AHumanFromCA:
Hi @muzikovam,

Thanks for the PR, could you provide some additional information about the functionality:

* What commands were added
* Some simple example output of the commands

Some other things:

* Could you please update your branch from master so we can verify it is working with the latest code
* All future commits must be presented with a sign-off. We can retroactively approve the commits without the signoff (so you don't have to change the history) but all future commits must have a signoff tag applied. I personally use this tool to help out with commit signing: https://github.com/coderanger/dco

PR #168, opened 2019-01-07
Add npm audit to build

### LINKED ISSUE #136: Add npm audit back to CI/CD pipeline
re-implement merge objects library

Child of #134
5 estimate

Notes:
This code needs copied over to Zowe - CLI.  It was already added to imperative.

ChrisBoehmCA:
Add npm audit to the build to check for vulnerabilities in dependencies

Fix #136

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,1 @@
 registry=https://gizaartifactory.jfrog.io/gizaartifactory/api/npm/npm-release/
-package-lock=false
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,1 @@
-@brightside:registry=https://api.bintray.com/npm/ca/brightside
 registry=https://gizaartifactory.jfrog.io/gizaartifactory/api/npm/npm-release/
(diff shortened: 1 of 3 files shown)


.npmrc
```diff
@@ -1,2 +1,2 @@
-registry=https://gizaartifactory.jfrog.io/gizaartifactory/api/npm/npm-release/
-package-lock=false
\ No newline at end of file
+@brightside:registry=https://api.bintray.com/npm/ca/brightside
```
AHumanFromCA: Could you explain why this change was made?

Without pulling from the gizaartifactory site, with this change any pushes to master of imperative are not immediately available for use by zowe. A manual or overnight build will need to take place in imperative after it is deployed to gizaartifactory possibly adding to development time.

ChrisBoehmCA: Sure. It might not be the right way to go, but what happened was that I had to add package-lock.json for audit to function, which recorded bintray as the location of @brightside/imperative.  The build was then unable to install imperative due to an error saying the tarball data was corrupted. This error was only able to be resolved by adding this line to .npmrc. I was told by @MarkAckert that this at least wouldn't interfere with the publish step since npmrc is removed before publishing in the pipeline, but I understand your concern

Maybe I could redo package-lock.json so that it pulls imperative from gizaartifactory by unsetting my @brightside:registry and then installing. I will try that

ChrisBoehmCA: Done

AHumanFromCA: Sounds good, thanks

Issue #176, opened 2019-01-16
Installing dependencies fails


abastyr:
## Steps to reproduce:

1. Clone the zowe-cli repository (master branch)
2. Open the terminal and go to the repository's folder
3. npm config set @brightside:registry https://api.bintray.com/npm/ca/brightside
4. npm install --registry https://registry.npmjs.org

## Output

```log
npm WARN tarball tarball data for @brightside/imperative@2.0.0-next.201901081528
sha1-s25vNlZA5oZXQ5DGtUZ3WFOi/UY=) seems to be corrupted. Trying one more time.
npm WARN tar ENOENT: no such file or directory, lstat 'C:\Users\basad01\sandbox\
sues\zowe\zowe-cli\node_modules\.staging\@brightside\imperative-4d23ec5e\lib\cmd
rc\doc\option'
npm WARN tar ENOENT: no such file or directory, lstat 'C:\Users\basad01\sandbox\
sues\zowe\zowe-cli\node_modules\.staging\@brightside\imperative-4d23ec5e\lib'
npm ERR! cb() never called!

npm ERR! This is an error with npm itself. Please report this error at:
npm ERR!     <https://npm.community>

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\basad01\AppData\Roaming\npm-cache\_logs

\2019-01-16T14_42_4
672Z-debug.log
```

[2019-01-16T14_42_48_672Z-debug.log](https://github.com/zowe/zowe-cli/files/2764577/2019-01-16T14_42_48_672Z-debug.log)

abastyr:
It seems I have found the workaround.

Running the

```sh
npm update --registry https://registry.npmjs.org
```

and then running the npm install... again looks like it installed the dependencies properly now.

ChrisBoehmCA:
Sorry for the issue. We are investigating the cause. I think it might be related to package-lock.json

AHumanFromCA:
It could be because of the package-lock having the registry for @brightside be different than what it was created with.

One thing that could be the problem is the package integrety

```
"@brightside/imperative": {
      "version": "2.0.0-next.201901081528",
      "resolved": "https://gizaartifactory.jfrog.io/gizaartifactory/api/npm/npm-release/@brightside/imperative/-/@brightside/imperative-2.0.0-next.201901081528.tgz",
      "integrity": "sha1-s25vNlZA5oZXQ5DGtUZ3WFOi/UY="
```

[...]

PR #177, opened 2019-01-16
Remove package-lock.json to avoid dependency installation issue

ChrisBoehmCA:
package-lock.json should still be generated during the npm install step of the Jenkins pipeline, but this should allow developers to install dependencies. We think the issue was caused by users trying to install `@brightside/imperative` via bintray which mismatches the integrity check generated by installing from giza artifactory

Fix https://github.com/zowe/zowe-cli/issues/176


ChrisBoehmCA:
Should I .gitignore package-lock.json?

PR #185, opened 2019-01-23
Workflows create api

abastyr:


__tests__/__resources__/properties/default_properties.yaml
```diff
@@ -83,3 +83,17 @@ provisioning:
...
+#-----------------------------------------------------------------------------#
+workflows:
+  # The name of the workflow you want to test
+  workflowName: workflow-name
+  # The system where you want to run your workflow (must exist and be active within your zOSMF)
+  system: system
+  # Owner of the workflow (userID that exists in zOSMF)
+  owner: owner
+    # Where on USS, in which DSNAME/MEMBER is your workflow definition file.
+  workflowDefinitionFile: path
```
ChrisBoehmCA: Is it possible to source the workflow definition file with the zowe CLI project and automatically upload and delete it during the tests? If so, you could probably eliminate all of these properties and make the tests system-independent. Might be cool

ChrisBoehmCA: If this is not possible, that's okay. Just please share the details of how to fill in these values for our internal systems with the Darkside team so that we can run your tests (via email)

muzikovam: Hi, I changed the workflow path and name to be randomly generated. Owner is zOSMF user. Only one that has to stay is system, because there can be multiple systems within one zOSMF thus user might use different system than primary.

ChrisBoehmCA: Thank you! I appreciate it.  Looks like there's a few places in the delete system tests that still need updated. Could you delete the workflow at the end of the Create tests as well to clean things up, so that we don't accumulate too many randomized files in our test directories? Ideally system tests should return the system to the state it was in before the test was run (when possible) and each system test should be independent of other system tests

muzikovam: Every workflow is deleted in afterEach section (lines 105 - 108). In the failing test there is no workflow created so it's not needed there therefore it is not needed in afterAll :) Actually that would fail.

ChrisBoehmCA: Sounds good!

PR #190, opened 2019-01-31
Directory to uss command

nurra01:
directory-to-uss command should help users to upload a local directory to USS.

PR #191, opened 2019-01-31
Directory to uss

nurra01:
This command should allow users to upload local dir to USS


ChrisBoehmCA:
Please see #190 for history

ChrisBoehmCA:
I think maybe something went wrong with this PR. It's showing commits since November.  It looks like you have modified all commits since then with your signoff. I don't think we want to modify the whole history of the project for this PR. Can you please create a new branch with just your changes, signed off so that we don't modify the history?

zFernand0:
Hi @nurra01,
I would advice the following steps to make the process as painless as possible.
1. Update your fork ([nurra01:master](https://github.com/nurra01/zowe-cli/tree/master))
2. Branch off of your master branch into a new branch (let's say `new-dtu`)
3. Cherry pick every commit from your `directory-to-uss` branch that only add/enhance functionality.
  - Approximately from January 22nd (which seems to be when the dirToUss functions were added)

PR #193, opened 2019-02-01
Dir to uss

nurra01:
This command will allow users upload local dir to USS dir.


nurra01:
#190 and #191 for history
Resolved commits conflicts how it was suggested in pull request #191

ChrisBoehmCA:
Thanks for fixing the history issue. Everything seems to be working well for me. I was able to recursively upload a directory to USS with certain files being uploaded in binary and others in text mode

ChrisBoehmCA:
One question about how --binary-files or --ascii-files work. If I have this file:

```
one/two/Hello.class
```

Are they all transferred in binary mode by just using --binary-files Hello.class? That seemed to work instead of doing --binary-files "one/two/Hello.class". In other words, is it just the filename that's checked and doesn't include any directory names?  If so, maybe we should describe this in the help text somehow.

Alexandru-Dumitru:
> One question about how --binary-files or --ascii-files work. If I have this file:
> 
> ```
> one/two/Hello.class
> ```
> Is that file transferred in binary mode by just using --binary-files Hello.class? That seemed to work instead of doing --binary-files "one/two/Hello.class". In other words, is it just the filename that's checked and doesn't include any directory names? If so, maybe we should describe this in the help text somehow.

Hi @ChrisBoehmCA. Your observation is correct. We are comparing just the file names. We have updated the description to better depict this. We thought having full file specification (including path) would make the parameter too long, so we limited just to file names. I know that raises the question, "what if i have 2 files with the same name, different sub-directories, and I want to transfer them with different modes", well, this case is not covered, but we didn't saw it as a common use case from the start.

PR #203, opened 2019-02-05
Us559527 create delete workflow cli

uzuko01:
Added CLI for Create and Delete workflow APIs

bright zos-workflows | wf create data-set | ds
bright zos-workflows | wf create uss-file | uf 

bright zos-workflows delete workflow-key | wk


ChrisBoehmCA:
Hi, 
Thanks for the changes! The commands looks pretty cool. I was just wondering about the wording of the commands. You have: 
```
bright zos-workflows | wf create data-set | ds
bright zos-workflows | wf create uss-file | uf
```
However these commands don't create data sets or USS files, but create workflows from existing data sets or USS files. So I feel the wording might be a little misleading. Maybe something like create `from-data-set`? 

Any thoughts @jelaplan / @BrandonJenkins14 ?

uzuko01:
@ChrisBoehmCA Hey Chris, yeah, sure I added them just for an example, actually, still waiting for John to give us the structure he wants to see.

ChrisBoehmCA:
We'll hold off on merging until UX has input on the syntax

jelaplan:
Thanks for clueing me in. 

I agree with Chris B about the confusion with creating a file vs. creating a workflow. 

I don't know much about this project. It seems the goal is to create and delete workflows.  My first thought is to show
bright zos-workflows create worklow 
but maybe that is problematic given bright tso issue command and we've seen someone struggle with that. but, we have the files example where 
bright zos-files create data-set-c and that is maybe a better comparison. 

Will we eventually support creating zOSMF templates too? Seems the create verb begs the object to be something you are creating. The implicit thing is awkward. 

Here is what I suggest. Feel free to argue with me if you disagree

bright zos-workflows create workflow-from-uss-file 
bright wf create wfuf

bright zos-workflows create workflow-from-data-set 
bright wf create wfds

bright zos-workflows create workflow-from-local-file 
bright wf create wflf

Jan sent me a list of all the commands. I took a quick look at this and I have further concerns. I'll try to get to a review of that in the new week. My understanding of workflows is pretty weak so I may need to talk to the team and Jan. This group seems like the most complex we've ever created. I see there are four positional arguments and there are options that are vague but I'd guess can contain arrays or other complex list of data, such as the --variables option. The examples given are hard to parse and don't provide any variation. :(

uzuko01:
@jelaplan I have changed the structure if the command, please let me know what you think.
Here is how it looks like now : 

bright zos-workflows create workflow-from-data-set < workflowName > [options]

 COMMAND NAME
 ------------

   workflow-from-data-set | wfds

 POSITIONAL ARGUMENTS
 --------------------

   workflowName          (string)

      Name of the workflow instance to create

 REQUIRED OPTIONS
 ----------------

   --data-set  | --ds (string)

      Data set containing workflow definiton.

   --system-name  | --sn (string)

      System where the workflow will run.

 OPTIONS
 -------

   --owner  | --ow (string)

      User ID of the workflow owner. This user can perform the workflow steps or
      delegate the steps to other users.

   --variables-input-file  | --if (string)

      Specifies an optional properties file that you can use to pre-specify values for
      one or more of the variables that are defined in the workflow definition file.

   --variables  | --vs (string)

      A list of one or more variables for the workflow. The variables that you specify
      here take precedence over the variables that are specified in the workflow
      variable input file.

   --assign-to-owner  | --ao (boolean)

      Indicates whether the workflow steps are assigned to the workflow owner.

   --access-type  | --at (string)

      Specifies the access type for the workflow. Public, Restricted or Private.

      Allowed values: Public, Restricted, Private

   --delete-completed  | --dc (boolean)

      Whether the successfully completed jobs to be deleted from the JES spool.

 EXAMPLES
 --------

   - Create a workflow with name "testworkflow" using data set
   "TESTID.WKFLOW" containing workflow definition xml, on system "TESTM1":

      $ zowe zos-workflows create workflow-from-data-set "testworkflow" --data-set "TESTID.WKFLOW" --system-name "TESTM1"

   - Create a workflow with name "testworkflow" using data set
   "TESTID.WKFLOW" containing workflow definition xml, on system "TESTM1" with
   owner "MYSYSID" and delete succesfully completed jobs:

      $ zowe zos-workflows create workflow-from-data-set "testworkflow" --data-set "TESTID.WKFLOW" --system-name "TESTM1" --owner "MYSYSID" --delete-completed

   - Create a workflow with name "testworkflow" using data set
   "TESTID.WKFLOW" containing workflow definition xml, on system "TESTM1" with
   variable values in the member PROPERTIES of data set TESTID.DATA:

      $ zowe zos-workflows create workflow-from-data-set "testworkflow" --data-set "TESTID.WKFLOW" --system-name "TESTM1" --variables-input-file TESTID.DATA(PROPERTIES)

   - Create a workflow with name "testworkflow" using data set
   "TESTID.WKFLOW" containing workflow definition xml, on system "TESTM1" with
   variable DUMMYVAR value DUMMYVAL and assign it to the owner:

      $ zowe zos-workflows create workflow-from-data-set "testworkflow" --data-set "TESTID.WKFLOW" --system-name "TESTM1" --variables DUMMYVAR=DUMMYVAL --assign-to-owner

muzikovam:
Why isn't the workflow owner required? Or why isn't there the default? (I see the old version now in the code, so I have no idea if you have the default already there)

ChrisBoehmCA:
Still kind of waiting for @jelaplan's input on the revised version, but a few things I've noticed: 

```
  Whether the successfully completed jobs to be deleted from the JES spool.
```

This should probably be rephrased to something like  "Indicates whether successfully completed jobs for this workflow should be deleted from the JES spool" 

Also please surround this with quotes in the example:  `TESTID.DATA(PROPERTIES)`. Many shells will require quotes around parentheses

uzuko01:
@ChrisBoehmCA I will do the changes you suggested. Also wanted to speak with @jelaplan how to refer to the user id from zosmf profile as default owner. : )

jelaplan:
Kostadin
Sorry for taking so long. Your work looks great. Thanks for making the changes we discussed on the phone and it is great to see the rich set of examples. I'd take Chris B's suggestions about text changes. 

I would love to see someone actually start workflows this way. It is apparent that the user would need to be fairly knowledgeable about creating workflow definitions and the nuances of things like the variables-input-file and the variables option. That is something both Mirek and I would want to see .

The variables option might have a more descriptive name. An option is a variable is a parameter, etc. 
It reminds me of the JCLCheck-options on @lam.ly@broadcom.com 's [JCLCheck plugin](https://docs.google.com/document/d/1ngHrRFXAJKYGXqLSGvLI9Inv_7GEcek_r1_hsiegRhQ/edit#heading=h.in7pywm4ehce) where the user has to enter a list of values.  In that case, they use a space separated array. In the case of the wf variables, the user has to put in a key, value pair. I think it is ok as it is but I think it would be useful to offer the user a link to some doc that explains the variables that are available. I talked with Ly about the idea of having a related documentation section in JCLCheck where we could put URLs. That would also be appropriate for the profiles section to link to docs about order of precedence and environmental variables. So, I'll follow up with that idea with @MikeBauerCA and Jason. 

@ChrisBoehmCA  I'm not sure what you mean by how to refer to the user id from zosmf profile as default owner. In my phone call with the team, the owner is different than the --user and they said their would be profile data like user, pass, port, etc. I feel like the owner option is sort of complicated and hard to grasp at first breath and it involves running the workflow beyond creating it. So, inheriting a value from some other place might make it even more complicated to understand. I know that from our Pre-sales dry run of the Sales Workshop, those guys had a lot of trouble grasping the order of precedence and even realizing that the same values could be set three times. After a few days of training it came into focus but it takes some learning.

uzuko01:
@jelaplan Thanks for the review. I will go ahead and make the changes that Chris suggested.
As for the owner option it will be a required every time, if you feel like this will be easier for the users to work with I am fine with it : )

ChrisBoehmCA:
Thanks, just waiting on your changes to approve

packages/workflows/src/cli/create/Create.common.options.ts
```diff
@@ -0,0 +1,144 @@
...
+        required: true
+    },
+
+    /**
+     * Property file containing key-value pairs as workflow variables.
+     * @type {ICommandOptionDefinition}
+     */
+    inputFile: {
+        name: "variables-input-file",
+        aliases: ["if"],
```
muzikovam: wif

packages/workflows/src/cli/create/Create.common.options.ts
```diff
@@ -0,0 +1,144 @@
...
+        required: false
+    },
+
+    /**
+     * Indicates whether the workflow steps are assigned to the workflow owner.
+     * @type {ICommandOptionDefinition}
+     */
+    assignToOwner: {
+        name: "assign-to-owner",
+        aliases: ["ao"],
```
muzikovam: ato

packages/workflows/src/cli/create/Create.common.options.ts
uzuko01: vif ?

muzikovam: oh, yes

AHumanFromCA:
I have a question about the merge process. Should we merge this PR using the old jenkinsfile or should we merge this after #219 which has changes for our new versioning scheme. I'm indifferent to the decision.

ChrisBoehmCA:
How about we merge #219 first to get a chance to test it out with these changes?

AHumanFromCA:
Sounds good, I'll wait for master to finish building then will merge #219 in.

After that is all built, I'll merge master in here and if it passes will merge this PR to master.

Make sense?

ChrisBoehmCA:
Makes sense to me

PR #205, opened 2019-02-06
Added List API plus its unit and system tests

melvl01:
I added API, that list zOSMF workflows from registry. There are six optional parameters that helps filter workflows per user's preference. System tests used create and delete API.


melvl01:
Will fix it in my branch

Issue #142, opened 2018-12-07
Implement CD Automated Scripts


JTonda:
A. Daily
B. beta (trigger every two weeks) - update ts doc
C. Latest (button -> future auto)

Child of #135
Estimate 5

Depends on #139

NOTES:

CD portion for daily/beta/latest

Daily is master branch
beta is beta branch
latest is latest branch

- zowe `@daily` should use imperative `@daily` in pkg.json
- zowe `@beta` should use imperative `@beta` in pkg.json
- zowe `@latest` should use imperative `@latest` in pkg.json

Code repo tasks

- ~Enhance **setup** step (from #139)~
       Assigned: Abandoned
       Status: Not Started
    - ~Fail if version was manually changed (allow for an override if we need to for some reason)~
    - Instead of failing if the version was manually changed, the pipeline will continue as normal. On deploy branches, the pipeline will use the input step to ask for which of the versions should be used to publish. If no response is received within 15 minutes, the build will try to deploy to the version in package.json. If it fails, just rerun the build with the desired version and the publish will happen.
- [x] Enhance **install** step (from #139)
       Assigned: @AHumanFromCA 
       Status: Complete
    - Accept packages to keep npm tags in line. (e.g zowe is in latest branch so it needs imperative latest (optional)
    - [x] Enforce integrity of PRs so tests can run against the newer versions.
    - [x] Create stage will need enhanced so it can allow a commit. Ensures no new commits were present on the branch.
- [x] Enhance **build** step (from #139)
       Assigned: @ChrisBoehmCA 
       Status: Complete
    - Do an `npm pack` instead of just `tar lib`
- [x] All steps in CI build are only triggered when a build happens on a release branch defined in setup
       Assigned: @AHumanFromCA 
       Status: Complete
- [x] ~Publish code doc step (this will probably be a custom step instead of a default one)~ Used the create stage function
       Assigned: ~No one~
       Status: Abandoned
    - ~[ ] Accepts command that builds documentation for code. (`npm run doc` is default)~
- [x] Deploy
       Assigned: @AHumanFromCA
       Status: Complete
    - [x] Use publish config to determine where to publish to
    - [x] Ask for user input for the version, give choices and default to package version if no response.
    - [x] Restrict which level can be bumped for protected branches.
        - Builds are not restricted but the email and input limit the available choices
- [x] Send emails out to admin group when ci/cd build operates on release branch.
       Assigned: No one
       Status: Not Started
- [x] Sonar Task?
[...]

PR #219, opened 2019-02-15
Enhancement/142

AHumanFromCA:
Implement shared pipeline as part of #142

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,12 +154,12 @@
     "coverageDirectory": "<rootDir>/__tests__/__results__/unit/coverage"
   },
   "jestSonar": {
-    "reportPath": "__tests__/__results__/jest-sonar",
-    "additionalResultsProcessors": "jest-sonar-reporter"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -93,10 +93,9 @@
     "jest-cli": "^22.4.3",
     "jest-environment-node": "^22.4.3",
     "jest-environment-node-debug": "^2.0.0",
-    "jest-html-reporter": "^0.5.7",
-    "jest-junit": "^3.6.0",
+    "jest-junit": "5.2.0",
```
zFernand0: Vulnerability introduced
See build here: https://wash.zowe.org:8443/blue/organizations/jenkins/brightside/detail/PR-219/2/pipeline

AHumanFromCA: Updated version.

Note:

jest-junit will now warn that they deprecated support for `testResultsProcessor`. Unfortunately, we can't move everything into the `results` array due to no support from jest-sonar-reporter. See 3dmind/jest-sonar-reporter#29

PR #228, opened 2019-02-20
add FAQ, fix links

BrandonJenkins14:
Added FAQs section to contain info about installing as root. There is only one FAQ, but we can grow this list as we see common themes in Slack/elsewhere.

I tried to summarize the info without losing critical information. Please confirm if info is accurate/sufficient to help the next user who comes along. 

(also fixed a few old docs-site links) 

Signed-off-by: BrandonJenkins14 <brandon.jenkins@broadcom.com>


zFernand0:
@dkelosky @AHumanFromCA,
I don't see a problem with this. It reflects the already discussed behavior.
Is there anything that should be added?

PR #230, opened 2019-02-21
Start CLI

muzikovam:
Added CLI for Start workflow APIs

bright zos-workflows start workflow-full | wf
bright zos-workflows start workflow-step | ws

Discussed with UX


zFernand0:
There seems to be conflicting files that need addressed.
@muzikovam, could you please take a look?

ChrisBoehmCA:
There's still a small conflict that needs to be resolved

PR #232, opened 2019-02-22
zos-workflows list api enhancement

uzuko01:
Add a method in List Workflows API to retrieve wf key by given name.


ChrisBoehmCA:
Would system tests be beneficial for this function?

uzuko01:
@ChrisBoehmCA  I do not think so, the method I added does nothing with the system, just uses an already existing one.

ChrisBoehmCA:
Maybe I'm misunderstanding, but doesn't the call to ` await this.listWorkflows(session, zOSMFVersion, workflowName);` contact z/OSMF?  If so, it seems like there should be a system test for this function

uzuko01:
@ChrisBoehmCA, but there are already system tests for this.listWorkflows()

ChrisBoehmCA:
OK.

PR #236, opened 2019-02-22
Enhancement/130

PeteSwauger:
Performance timing added
Typescript 3


Jenkinsfile
```diff
@@ -9,163 +9,896 @@
 *                                                                                 *
 */
 
-@Library('shared-pipelines@v1.0.2') import org.zowe.pipelines.nodejs.NodeJSPipeline
+// The following property need to be set for the HTML report @TODO figure out how to get this nicely on jenkins
+//System.setProperty("hudson.model.DirectoryBrowserSupport.CSP", "")
```
ChrisBoehmCA: It kind of looks like you accidentally reverted the Jenkinsfile back to the old one, right @zFernand0 ?

ChrisBoehmCA: Fixed

packages/main.ts
```diff
@@ -10,6 +10,11 @@
 *
 */
 
+import { PerfTiming } from "@zowe/perf-timing";
+
+const timingApi = PerfTiming.api;
+
+timingApi.mark("PRE-INIT");
```
ChrisBoehmCA: Not a big deal, but I wonder if we could  change the name as "PRE-INIT" vs. "BEFORE_INIT" is a little confusing to me. Maybe "PRE_IMPORT_IMPERATIVE" and "POST_IMPORT_IMPERATIVE" or something like that?

Also even if we don't change the name, I think you have a hyphen here but an underscore in the call to .measure below so the measurement might not work without an update

PeteSwauger: I am okay with any change.  The only thing I wanted was all uppercase so they stood out.

zFernand0: I'm so sorry I missed this 😢

PR #246, opened 2019-02-27
223 add uss delete

Colin-Stone:


Colin-Stone:
Based upon the equivalent function for Datasets. Implemented API and CLI and tests for unit, integration(CLI only) and system.

packages/zosfiles/__tests__/__system__/cli/delete/uss/uss.files.delete.directory.system.test.ts
```diff
@@ -0,0 +1,152 @@
...
+            testName: "zos_delete_uss_directory"
+        });
+
+        systemProps = new TestProperties(TEST_ENVIRONMENT.systemTestProperties);
+        defaultSystem = systemProps.getDefaultSystem();
+
+        REAL_SESSION = TestEnvironment.createZosmfSession(TEST_ENVIRONMENT);
+        user = defaultSystem.zosmf.user.trim();
+        ussname = `${defaultSystem.zosmf.user.trim()}.aTestUssDirectory`;
+        ussname = ussname.replace(/\./g, "");
```
ChrisBoehmCA: I didn't exactly follow why you replace `.` after adding one in the line above. Probably not a problem but wanted to understand the rationale

packages/zosfiles/__tests__/__system__/cli/delete/uss/uss.files.delete.directory.system.test.ts
```diff
@@ -0,0 +1,152 @@
+/*
```
ChrisBoehmCA: These two test files deviate slightly from the pattern we used for the other commands. Other test file names begin with `cli.` and the rest of the filename mimics the command structure e.g. `cli.files.delete.uss.directory.system.test.ts`. Something similar to that would be good. Good job on the tests though, they all passed for me and look good.

Colin-Stone: Ah yes. That's makes sense.

packages/zosfiles/__tests__/__system__/cli/delete/uss/uss.files.delete.directory.system.test.ts
Colin-Stone: Hmm.. Neither do I. I reckon I've copied it from the dataset delete along with the overall flow and didn't ask myself why. My bad.

PR #257, opened 2019-03-02
Archived Delete API

melvl01:
Based on delete command, created delete from archive API. Added unit tests and system tests.


packages/workflows/src/api/ArchivedDelete.ts
```diff
@@ -0,0 +1,40 @@
...
+export class ArchivedDeleteWorkflow {
+
+    /**
+     * Delete a workflow from an archived
+     * @param {AbstractSession} session                     - z/OSMF connection info
+     * @param {string} workflowKey                          - Unique identifier of the workflow instance.
+     * @param {string} zOSMFVersion                         - Identifies the version of the zOSMF workflow service.
+     * @returns {string}
+     */
+ public static async archivedDeleteWorkflow(session: AbstractSession, workflowKey: string,
```
uzuko01: can we have 4 spaces of indentation ? : )

packages/workflows/src/api/ArchivedDelete.ts
```diff
@@ -0,0 +1,40 @@
...
+import { WorkflowConstants, nozOSMFVersion, noWorkflowKey } from "./WorkflowConstants";
+import { WorkflowValidator } from "./WorkflowValidator";
+
+/**
+ * Class to handle deletion of workflow from an archived z/OSMF workflow from a z/OS system.
+ */
+export class ArchivedDeleteWorkflow {
+
+    /**
+     * Delete a workflow from an archived
```
uzuko01: Delete archived workflow ?

packages/workflows/src/api/ArchivedDelete.ts
```diff
@@ -0,0 +1,40 @@
...
+*/
+
+
+import { AbstractSession } from "@brightside/imperative";
+import { ZosmfRestClient } from "../../../rest";
+import { WorkflowConstants, nozOSMFVersion, noWorkflowKey } from "./WorkflowConstants";
+import { WorkflowValidator } from "./WorkflowValidator";
+
+/**
+ * Class to handle deletion of workflow from an archived z/OSMF workflow from a z/OS system.
```
uzuko01: Class to  handle deletion of an archived z/OSMF workflow from a z/OS system?

packages/workflows/src/api/ArchivedDelete.ts
```diff
@@ -0,0 +1,40 @@
...
+     */
+ public static async archivedDeleteWorkflow(session: AbstractSession, workflowKey: string,
+                                            zOSMFVersion = WorkflowConstants.ZOSMF_VERSION){
+    WorkflowValidator.validateSession(session);
+    WorkflowValidator.validateNotEmptyString(zOSMFVersion, nozOSMFVersion.message);
+    WorkflowValidator.validateNotEmptyString(workflowKey, noWorkflowKey.message);
+    let resourcesQuery: string = `${WorkflowConstants.RESOURCE}/${zOSMFVersion}/`;
+    resourcesQuery += `${WorkflowConstants.ARCH_WORKFLOW_RESOURCE}/${workflowKey}`;
+
+    return ZosmfRestClient.deleteExpectString(session, resourcesQuery, []);
```
uzuko01: maybe null instead of [] ?

packages/workflows/src/api/ArchivedDelete.ts
melvl01: done

packages/workflows/src/api/ArchivedDelete.ts
melvl01: ok

packages/workflows/src/api/ArchivedDelete.ts
melvl01: changed

packages/workflows/src/api/ArchivedDelete.ts
melvl01: changed

zFernand0:
I'm curious about the rationale of having the "action" after the potential CLI "object".
For example, `zowe console issue command` has a corresponding API class called `IssueCommand` in the [IssueCommand.ts](https://github.com/zowe/zowe-cli/blob/master/packages/zosconsole/src/api/IssueCommand.ts) file

@melvl01, Should this be changed?

zFernand0:
Also, Should this be associated with the already existing Delete APIs of the Workflow package/group?
I don't think we need a separate class if there is already a Delete class containing said workflow APIs.

melvl01:
Thanks zFernand0, Will discuss both comments with team...

zFernand0:
> Thanks zFernand0, Will discuss both comments with team...
@melvl01, No need to change how it was done.
FYI, if we do change it to use that `Delete.workflow` or `Delete.arvhivedWorkflow` model, then we will definitely have another breaking change, and this feature may bring further complications when porting it to the `lts-incremental` supported branch

PR #258, opened 2019-03-04
Workflows/fix/remove DS check

### LINKED ISSUE #234: Validation of data set names / uss files in create workflow commands produces malformatted error message
The error message contains [object Object] which at least should be fixed

```
$ zowe wf create wfds "namehere" --ds "**BADDATASET" --owner myid --sn system
Command Error:
Register workflow: Error: Expect Error: [object Object]Wrong format of USS path or DSNAME supplied.
```

I know in previous iterations of the CLI, we've done this kind of validation of parameters. However maintaining a regex for data set names can be error prone, especially since international characters like `£` can be used in some regions for data sets. It's easy to miss a valid data set name format. For these reasons, in other commands in core we've decided to not validate data set names or USS files. I would consider removing the validation and letting z/OSMF return an error message about the bad path/ DS name

@tucker01 
@muzikovam

abastyr:
- Removes the DS name check from the *Create workflow API*
- Removes failing unit tests
- Fixes system tests to be working properly
- Removes the error message constant
- Removes the DS validation from the `WorkflowValidator`

*Note* There still might be a problem in the CLI or CLI tests

Resolves #234


ChrisBoehmCA:
Thanks for the changes! Can you explain what you mean by this:
`Note There still might be a problem in the CLI or CLI tests`

abastyr:
Sure thing, Chris. I have made changes in the API and also modified the unit tests and system tests for the API. As for the CLI, unit tests and integration tests I have run are passing. However, in system tests, the tests are failing because they refer to the test file workflow which doesn't exist.

```typescript
const workflowDs = join(__dirname, "../../testfiles/demods.xml");
```

This file doesn't exist at all. I could have tried to recreate the workflow, but I am not sure what it did or if the workflow output is required for the tests.

ChrisBoehmCA:
Thanks. We must have deleted that file mistakenly.

abastyr:
Bump

zoweshare:
Testing for Zowe SHARE session.

ChrisBoehmCA:
Sorry guys. Could you resolve the merge conflict on demods.xml?

abastyr:
> Sorry guys. Could you resolve the merge conflict on demods.xml?

Done :)

Issue #253, opened 2019-02-28
Create USS List files function in List.ts


Colin-Stone:
As with https://github.com/zowe/zowe-cli/issues/223 a list function files by path name would support VSCode extension requirements and be of value to organizations using USS files.

PR #245, opened 2019-02-27
Fix/204

Alexandru-Dumitru:
- Fix for https://github.com/zowe/zowe-cli/issues/204 by introducing `--max-concurrent-requests` option on the `files upload dtu` command. 

Refactored `dirToUSSDir` API and CLI to:
- accept pool of options
- create empty directories but skip processing them if empty
- added progress bar  
- added option  `--max-concurrent-requests`

Updated snapshots
Updated command used in test shell script (it was wrongly using file-to-uss)
Updated related tests

PR #261, opened 2019-03-07
253 uss list

Colin-Stone:
This function provides a USS API and CLI functionality to list the files and directories in a USS path. Interfaces directly with the zosmf function described: https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.izua700/IZUHPINFO_API_GetListUNIXfiles.htm
CLI implemented to exploit the response format functionality --rft --rfh --rff

Addresses https://github.com/zowe/zowe-cli/issues/253


packages/zosfiles/src/api/methods/list/List.ts
```diff
@@ -118,6 +119,48 @@ export class List {
...
+     * @throws {ImperativeError} data set name must be set
+     * @throws {Error} When the {@link ZosmfRestClient} throws an error
+     */
+    public static async fileList(session: AbstractSession, path: string, options: IUSSListOptions = {}): Promise<IZosFilesResponse> {
+        ImperativeExpect.toNotBeNullOrUndefined(path, ZosFilesMessages.missingUSSFileName.message);
+        ImperativeExpect.toNotBeEqual(path, "", ZosFilesMessages.missingUSSFileName.message);
+
+        try {
+            const endpoint = posix.join(ZosFilesConstants.RESOURCE,
+                `${ZosFilesConstants.RES_USS_FILES}?${ZosFilesConstants.RES_PATH}=${path}`);
```
js665999:
```suggestion
                `${ZosFilesConstants.RES_USS_FILES}?${ZosFilesConstants.RES_PATH}=${encodeURIComponent(path)}`);
```
same issue as https://github.com/zowe/zowe-cli/pull/245 found by @ChrisBoehmCA 
encode USS path when appending it to the resource so that paths with special characters can still be uploaded/ downloaded

Colin-Stone:
Thanks Jason, Chris. Chris good idea regarding the examples. I'll add the rff one in as I can see that it's probably the more applicable to most users as opposed to getting the complete details back. Should I be adding something for --rft and rfh ?. May be overkill if we have too many examples?

ChrisBoehmCA:
Thanks. I see you added a string for the example, but I don't think you exposed it on the definition yet.  Maybe add one more example for just --rfh so that users can see how to view what fields they are looking at in the table.

packages/zosfiles/src/cli/-strings-/en.ts
```diff
@@ -313,6 +313,18 @@ export default {
...
+            },
+            USS_FILE: {
+                SUMMARY: "List USS files",
+                DESCRIPTION: "List USS files and directories in a UNIX file path",
+                POSITIONALS: {
+                    PATH: "The directory containing the files and directories to be listed"
+                },
+                EXAMPLES: {
+                    EX1: `Show the files and directories in path '/u/ibmuser'`,
+                    EX2: "Show the files and directories in path '/u/ibmuser displaying only the file or directory name:",
```
ChrisBoehmCA: Tiny nitpick... I think this shouldn't have a `:` at the end as one is added automatically when the help is generated

ChrisBoehmCA: I edited it for you but you might have to update the snapshots again

PR #275, opened 2019-03-12
Enhance delete active wf cli

uzust01:
Enhance delete active-workflow action to support deletion using workflow name.


uzust01:
> Hi,
> 
> Looks good to me, but seems like we should add a system test for the new option
> 
> Also, this PR and previous changes will need to be ported to the lts-incremental branch

:D I forgot to push the last commit.

PR #276, opened 2019-03-12
Enhance list active wf details cli

uzust01:
Enhance list active-workflow-details to accept workflow-name as well as workflow-key


packages/workflows/src/cli/list/activeWorkflowDetails/ActiveWorkflowDetails.definition.ts
```diff
@@ -38,6 +38,10 @@ export const ActiveWorkflowDetails: ICommandDefinition = {
         {
             description: "To list the details of an active workflow with key \"7c62c790-0340-86b2-61ce618d8f8c\" including its steps and variables",
             options: "--workflow-key \"7c62c790-0340-86b2-61ce618d8f8c\" --list-steps --list-variables"
+        },
```
ChrisBoehmCA: Thanks for adding a new example. However I think you might have meant to put --workflow-name here?

uzust01: yes, I meant that, thanks 🤖

PR #282, opened 2019-03-13
Port workflow directory from master

uzust01:
Add current Zowe-master-workflows code to lts-incremental.
Includes: #249 #254 #257 #258 #262 #263


ChrisBoehmCA:
Thanks very much for porting the recent workflow changes. It looks like a snapshot needs to be updated

packages/workflows/src/cli/create/dataset/Dataset.definition.ts
```diff
@@ -23,7 +23,7 @@ import { join } from "path";
 export const DataSet: ICommandDefinition = {
     name: "workflow-from-data-set",
     aliases: ["wfds"],
-    description: "Create a workflow instance in z/OSMF using a Data set",
+    description: "Create a z/OMF workflow on a z/OS system using a Data set",
```
zFernand0: shouldn't it be `z/OSMF`?

packages/workflows/src/api/WorkflowValidator.ts
```diff
@@ -68,42 +68,6 @@ export class WorkflowValidator {
...
 
-    /**
-     * Validate supplied parameter
-     * @static
-     * @param {string} path - string to check if contains USS path or DSNAME
-     * @param {string} errorMsg - message to show in case validation fails
-     * @memberof WorkflowValidator
-     * TODO maybe also validation if the file or DSNAME exists on Mainframe
-     */
-    public static validatePath(path: string, errorMsg: string) {
```
zFernand0: I get it that people often do not use the CLI APIs that we provide, but we should be more careful when removing functionality from the project.

zFernand0:
I marked this as a "minor" enhancement for now.
If someone else has a concern that this is potentially a breaking PR, then changes will be requested.

packages/workflows/src/cli/create/dataset/Dataset.definition.ts
uzust01: Yes, it should be, thanks 🤖

ChrisBoehmCA:
I think it's not breaking. The removed API was an internal input validation API, so I think it's okay

zFernand0:
> I think it's not breaking. The removed API was an internal input validation API, so I think it's okay

Thank you for pointing out that the validation APIs were never exported.
https://github.com/muzikovam/zowe-cli/blob/c09a691fc7ec37339f3c5a6b63473a8fcafc007b/packages/workflows/index.ts#L22-L29

packages/workflows/src/api/WorkflowValidator.ts
zFernand0: Validation APIs were never exported

PR #284, opened 2019-03-13
Resolve SHARE issue

MikeBauerCA:
Signed-off-by: Bauer, Michael <Michael.Bauer2@ca.com>


ChrisBoehmCA:
Did you know there is a `summary` field that is displayed in the list of groups? This way we could keep the longer description on `zowe provisioning -h`  but have a shorter description on `zowe -h`. But this is also a demonstration so no big deal, haha

MikeBauerCA:
@ChrisBoehmCA thanks for the info. It motivated me to adjust the top-level help so that each zowe group fit on a single line when there was available space.

packages/provisioning/src/cli/Provisioning.definition.ts
```diff
@@ -20,6 +20,7 @@ const definition: ICommandDefinition = {
     name: "provisioning",
     aliases: ["pv"],
     type: "group",
+    summary: "Perform z/OSMF provisioning tasks",
```
zFernand0: We need to implement these changes in LTS incremental.
@ChrisBoehmCA, any thoughts?

ChrisBoehmCA: I can open a new PR for LTS incremental.

zFernand0: Thank you :)

PR #294, opened 2019-03-18
lts-incremental update 

Alexandru-Dumitru:
Update based on https://github.com/zowe/zowe-cli/pull/245


docs/CLIReadme.md
```diff
@@ -2408,6 +2408,127 @@ delimiters are added between records.
 
 * `          $  zowe zos-files upload file-to-uss "file.txt" "/a/ibmuser/my_text.txt"`
 
+<<<<<<< HEAD
+=======
```
zFernand0: Is this the result of a merge conflict?
Since Tech Writers can generate this file, maybe we can remove this file from source to avoid confusion.
@BrandonJenkins14, @JamesBauman, What do you think?

BrandonJenkins14: Sounds OK to me, I don't think we link to it from anywhere.

Alexandru-Dumitru: Hi @zFernand0, yes, it was from an outstanding merge conflict. Don't know how it passed. I've fixed it now. Thanks for bringing it up.

PR #295, opened 2019-03-18
fix system test for files upload dtu api

Alexandru-Dumitru:
Was calling different API. Now it is fixed.

Signed-off-by: Alexandru-Dumitru <alexandru.dumitru@broadcom.com>


ChrisBoehmCA:
Hi, 

I ran this test : 

    const localDirWithSpaces = `${__dirname}/testfiles/space dir`;

It doesn't seem like the "space dir" directory is checked into the project. Maybe we forgot to add this?

Alexandru-Dumitru:
> Hi,
> 
> I ran this test :
> 
> ```
>     const localDirWithSpaces = `${__dirname}/testfiles/space dir`;
> ...
>  it("should upload local directory (with space in name) to USS", async () => {
> ```
> It doesn't seem like the "space dir" directory is checked into the project. Maybe we forgot to add this?

Hi @ChrisBoehmCA . You are right. I was not aware that git doesn't include empty directories by default, and failed to notice this absence. Corrected now.

PR #299, opened 2019-03-25
Retrieve mstr definition action cli

uzust01:
Retrieve details for workflow file definition which can either be a full USS path or a data set. 

Limitations: --list-steps command only returns the highest steps level, no sub steps are displayed.


ChrisBoehmCA:
Hi please port this to lts-incremental as well

Issue #214, opened 2019-02-07
Enhancements to issuing shell command


ChrisPark89:
## Description
Issue a z/OS UNIX command and echo the output.

## Motivation
Users often have a need to run individual commands on z/OS USS, for example as part of deploying an application. They could of course have a separate SSH session, but this is not as convenient or easy to script as part of a workflow. Our primary use-case is running 'npm install' on the target system as part of deploying a Node.js application on z/OS. 

## Design
Use ssh connection or submit BPXBATCH jcl through z/OSMF.
There are two options for how the command would look like. 

### 1.  Default to SSH connection and provide a --jcl flag to submit via JCL

```zowe zos-uss|uss issue command "npm install"  [ --cwd /mypath/myapp ] [ --jcl ]```

**Command Options**
--cwd : current working directory where the command will be executed.
--jcl : execute command by submitting JCL instead of SSH.

**Examples**
`zowe zos-uss|uss issue command "npm install"`
`zowe zos-uss|uss issue command "npm install" --jcl`
`zowe zos-uss|uss issue command "npm install" --cwd "/home/user/myapp"`

### 2. Have two separate commands for SSH and JCL

```zowe zos-uss|uss issue jcl "npm install"  [ --cwd /mypath/myapp ]```
```zowe zos-uss|uss issue ssh "npm install"  [ --cwd /mypath/myapp ]```

**Command Options**
--cwd : current working directory where the command will be executed.

**Examples**
`zowe zos-uss|uss issue jcl "npm install"`
`zowe zos-uss|uss issue ssh "npm install"`
`zowe zos-uss|uss issue ssh "npm install" --cwd "/home/user/myapp"`

### SSH
Uses an interactive SSH connection to execute the command. The connection is closed after the command is executed.

- Zowe ssh profile
Uses a new Zowe profile for ssh to contain the host, port number, username, password, and privateKey.
- SSH connection
Users can use a public key and a private key pair instead of a username and a password to connect to the ssh server
  - username / password
  - Private/Public key pair 
  - The public key has to be stored on the server side. (USSHOME/.ssh/authorized_keys)
  - The user can specify the path of the secret key in the zowe ssh profile.
- A SSH server must be running on the host.
- Output from the command is displayed line by line, which would be useful for commands that have intermediate progress messages or long output, for example "npm install".
- Users can terminate the execution of the command part way through using a key sequence to send  SIGINT signal to the server (e.g. CTRL+C in UNIX)
[...]

PR #308, opened 2019-04-03
zos-uss issue ssh

matthewpwilson:
Introduce the capability to issue ssh commands on the host as described in #214 

In the end we didn't think it was necessary to provide a 'fallback' implementation as discussed in #214.
We've not yet implemented mirroring the return code from the remote command, we can do this in a subsequent PR. 

Note to run the system tests we've added you'll need to update custom_properties.yaml to add the new ssh properties as per default_properties.yaml.


matthewpwilson:
Doc will be updated by this PR https://github.com/zowe/docs-site/pull/440

matthewpwilson:
Thanks for the speedy feedback Chris.
Re: 2, I've done as suggested and most of the system test cases will now work without a privateKey. However, the 'bad passphrase' test does require a valid private key in order to pass.  We could remove this test if you would prefer to have all the tests run without needing a key.

ChrisBoehmCA:
Can the bad passphrase test be updated to not need a valid private key? If not, maybe we could remove that test.

ChrisBoehmCA:
Could you also add more log messages to Shell.ts?

I'm getting a strange behavior when I use this command against one of our internal mainframe systems. The command exits with exit code 0, but no output to stderr or stdout. It works on multiple other systems I've tried though. Any clue what could be happening? SSH is available on the system with the issue, as I'm able to use the `ssh` command line program to connect to it.  Also I think you should reject the promise on line 120 of Shell.ts instead of throwing the error, but changing that locally didn't seem to help the issue I'm seeing.

ChrisPark89:
>     1. In your handler, you write directly to the process' standard out. This breaks --rfj if it's specified by the user. In order to fix this you can write using `commandParameters.response.console.log(data)` which will buffer the output and include it in the JSON response.

Is there another function that does not automatically append a new line character for each data passed to the function? The problem of this is that some stdout messages doesn't use a new line character on purpose (e.g progress bar).

ChrisPark89:
> Could you also add more log messages to Shell.ts?
> 
> I'm getting a strange behavior when I use this command against one of our internal mainframe systems. The command exits with exit code 0, but no output to stderr or stdout. It works on multiple other systems I've tried though. Any clue what could be happening? SSH is available on the system with the issue, as I'm able to use the ssh command line program to connect to it. Also I think you should reject the promise on line 120 of Shell.ts instead of throwing the error, but changing that locally didn't seem to help the issue I'm seeing.

I am suspecting the stdout clean-up logic. I added this because it becomes too verbose otherwise. I might have missed some corner cases. I just added a log message to see the original data stream. For the each stream, the log starts with '[Received data begin]' and ends with '[Received data end]'.

ChrisPark89:
The problem was I checked `@@START OF COMMAND@@` as a flag, but to distinguish with the first and the second flag, I made it check new line character at the beginning e.g. `\n@@START OF COMMAND@@` making it bash shell specific. I added a logic to check with `$` as well as `>` for Bourne shell.

Could you check if that works now?

ChrisBoehmCA:
Oh, or you could force a PS1 by doing `export PS1=$` and then it would always be the same, but that command has to succeed/ finish before it would take effect. But maybe that helps

ChrisPark89:
> Hmm... you can echo PS1 is $PS1 and that will print it out. Can you possibly just use the last appearance of @@START OF COMMAND@@ rather than relying on the PS1 though? If not maybe echoing out the PS1 would work

Because the data it gets is stream, it is hard to tell which one is the last. 

> Oh, or you could force a PS1 by doing export PS1=$ and then it would always be the same, but that command has to succeed/ finish before it would take effect. But maybe that helps

This is a brilliant idea! It seems to work with that. Could you check it on your systems too?

ChrisBoehmCA:
Thanks. That fixed the output on the system that had the problem before. Sorry for so much back and forth, but is there any way you can make the timeout configurable in the SSH profile or default to zero? I'm getting "Timed out while waiting for handshake" with one system

ChrisBoehmCA:
I'm also getting a few test failures on our internal test system: 

```

  ● zowe uss issue ssh without running bash scripts › should resolve --cwd option

    expect(received).toMatch(expected)

    Expected pattern: /\/u\/users\/username\/\s/
    Received string:  "
    $ /u/users/username
    $ 
    "

      85 |         checkResponse(response);
      86 |         // match only "/"+cwd with no following alpha-numeric character e.g. "/cwd   "
    > 87 |         expect(response.stdout.toString()).toMatch(new RegExp("\\" + cwd + "\\s"));
         |                                            ^
      88 |     });
      89 | 
      90 |     it("should get directory invalid --cwd option", async () => {

      at Object.<anonymous> (packages/zosuss/__tests__/__system__/cli/issue/cli.issue.ssh.system.test.ts:87:44)
      at fulfilled (packages/zosuss/__tests__/__system__/cli/issue/cli.issue.ssh.system.test.ts:14:58)

  ● Use a test directory to do stuff in that creates files › should write a long directory

    expect(received).toContain(expected) // indexOf

    Expected substring: "/u/users/username/usstest"
    Received string:    "
    $ $ 
    "

      117 |             const response = await runCliScript(__dirname + "/__scripts__/issue_ssh_no_cwd.sh", TEST_ENVIRONMENT, [commandName]);
      118 |             checkResponse(response);
    > 119 |             expect(response.stdout.toString()).toContain(directory + "usstest");
          |                                                ^
      120 |         });
      121 |         afterAll(async () => {
      122 |             // delete the test directory

      at Object.<anonymous> (packages/zosuss/__tests__/__system__/cli/issue/cli.issue.ssh.system.test.ts:119:48)
      at fulfilled (packages/zosuss/__tests__/__system__/cli/issue/cli.issue.ssh.system.test.ts:14:58)

  ● zowe uss issue ssh running bash scripts › script asks for input

    expect(received).toContain(expected) // indexOf

    Expected substring: "Welcome Tester!"
    Received string:    "
    $ read: askForName.sh 3: FSUM7357 no active co-process
    Welcome !
    $ 
    "

      262 | 
      263 |         checkResponse(response);
    > 264 |         expect(response.stdout.toString()).toContain("Welcome Tester!");
          |                                            ^
      265 |     });
      266 | 
      267 |     it("script kills itself", async () => {

      at Object.<anonymous> (packages/zosuss/__tests__/__system__/cli/issue/cli.issue.ssh.system.test.ts:264:44)
      at fulfilled (packages/zosuss/__tests__/__system__/cli/issue/cli.issue.ssh.system.test.ts:14:58)

```

And also the passphrase test is failing for me but I'm thinking we could remove that

ChrisPark89:
> Thanks. That fixed the output on the system that had the problem before. Sorry for so much back and forth, but is there any way you can make the timeout configurable in the SSH profile or default to zero? I'm getting "Timed out while waiting for handshake" with one system

Don't worry at all! It's good to have feedback and it really helps. I just added `handshakeTimeout` option in ssh profile. If it isn't specified, it is set to 0.

ChrisBoehmCA:
A few more things I noticed: 
- We usually require unit tests of handlers in addition to APIs. See other files ending in .handler.unit.test.ts for examples
- We usually have integration tests for each command for testing syntax problems and creating help snapshots. You can test omitting the command here and make sure it produces a syntax error. See other files ending in `.integration.test.ts` for examples

Could you add these types of tests please? Let me know any guidance you need.

ChrisPark89:
>   ● zowe uss issue ssh running bash scripts › script asks for input
> 
>     expect(received).toContain(expected) // indexOf
> 
>     Expected substring: "Welcome Tester!"
>     Received string:    "
>     $ read: askForName.sh 3: FSUM7357 no active co-process
>     Welcome !
>     $ 
>     "
> 
>       262 | 
>       263 |         checkResponse(response);
>     > 264 |         expect(response.stdout.toString()).toContain("Welcome Tester!");
>           |                                            ^
>       265 |     });
>       266 | 
>       267 |     it("script kills itself", async () => {
> 
>       at Object.<anonymous> (packages/zosuss/__tests__/__system__/cli/issue/cli.issue.ssh.system.test.ts:264:44)
>       at fulfilled (packages/zosuss/__tests__/__system__/cli/issue/cli.issue.ssh.system.test.ts:14:58)
> 
> 

For this test, it seems like the script was run with sh, instead of bash. Does your system not have bash?

ChrisBoehmCA:
Not all of our USS systems have bash, no

ChrisPark89:
I removed the "script asks for input" test because 
A. It was using a builtin bash command that doesn't work with sh.
B. We are not expecting users to use ssh command interactively, so the test isn't really relevant to our purpose.

Also, I fixed the other tests, so hopefully it passes.

> A few more things I noticed:
> 
>     We usually require unit tests of handlers in addition to APIs. See other files ending in .handler.unit.test.ts for examples
>     We usually have integration tests for each command for testing syntax problems and creating help snapshots. You can test omitting the command here and make sure it produces a syntax error. See other files ending in .integration.test.ts for examples
> 
> Could you add these types of tests please? Let me know any guidance you need.

I will have a look into this as well as the passphrase test.

ChrisPark89:
I've added the uni tests for the handler and integration tests.
Also, I decided to remove the passphrase test because it's hard to test it without having the private key installed in the server side.

ChrisBoehmCA:
It seems like you just merged in changes related to the upload dir-to-uss command. Was that intentional? I'll try re-running your system tests soon

ChrisPark89:
I didn't meant to take the changes for upload dir-to-uss command. I just merged the master to this branch because the build was failing and thought there was a conflict between the master and this branch.

ChrisBoehmCA:
Okay. Is there any way for you to remove the USS directory related changes?  We would probably like for those changes to be in a separate pull request especially if it's not ready for review yet.

ChrisBoehmCA:
Sorry for the confusion about the build failing. There is a vulnerability that we've discovered that we are waiting for a patch to be delivered for. So until then the build will fail on the "check for vulnerabilities" step.

ChrisPark89:
Sorry it was my mistake! I should have checked what I was doing more carefully. I think it's roll-backed to the one before merging now.

ChrisBoehmCA:
I still seem to have the invalid passphrase test, but all the other tests are passing for me

ChrisPark89:
I might have omitted the change! I updated now.

ChrisBoehmCA:
The command works well for me and the system tests are passing now. Good job. Thanks for enduring all the feedback : D

ChrisBoehmCA:
Sorry for the inconvenience by the way, but would you be willing to also port these changes in a new PR to the lts-incremental branch? This would make the command available to users of Zowe that install with the @lts-incremental tag which is documented on zowe.org . Master is our forward development version that allows breaking changes -- lts-incremental doesn't. But this can be ported since it is an enhancement with no breaking changes. 

The easiest way would probably be to copy the whole zos-uss directory, tests and all, into a new branch based on `lts-incremental`. Since you have so many commits, it would be hard to cherry pick them.

It would require you to change any instances of `@zowe/imperative` to  `@brightside/imperative` in your new branch.  If you're not able to, I can try to port it soon. Thanks

ChrisBoehmCA:
Ahhh! I'm really sorry, I tried to make a simple change to the pipeline and I somehow restored the in-progress USS stuff. I guess I somehow had the changes from before your force push. Sorry to be a bother, but any idea how to get rid of the USS stuff for now in this PR?

ChrisPark89:
Hmm. I don't really understand what you meant there. Do you think it's going to be easier, if we close this P&R and make another one, with all the commits based on `lts-incremental` branch rather than the `master`?

ChrisBoehmCA:
No sorry, it's a little confusing.  We need a pull request for both master and lts-incremental, as they are two divergent versions of the same product. Master has all the latest code including breaking changes. Lts-incremental can't receive breaking changes. Most people install using lts-incremental, but the next lts-incremental release will be cut from master. So all non-breaking changes have to be ported to both places

matthewpwilson:
> Ahhh! I'm really sorry, I tried to make a simple change to the pipeline and I somehow restored the in-progress USS stuff. I guess I somehow had the changes from before your force push. Sorry to be a bother, but any idea how to get rid of the USS stuff for now in this PR?

I had a local of the branch which I *think* was in the right state, so I've force-pushed that, so hopefully we're back.

ChrisBoehmCA:
Thanks. Sorry about that

ChrisBoehmCA:
Sorry for the delay guys. We want to merge soon, but we're having issues with the artifactory we deploy to and install from, so we can't really merge anything until that's resolved.  We'll get it in ASAP. Thanks for your patience

ChrisBoehmCA:
Looks like there is a failing unit test: 

Snapshot name: `zos-uss issue group definition should not have changed 1`
- Snapshot
+ Received
@@ -12,11 +12,11 @@
Object {
"description": "Issue a simple command, giving the working directory",
"options": ""npm install express" --cwd /u/cicprov/mnt/CICPY01I/bundles/myapp ",
},
],
- "handler": "/Users/jeonghyukpark/Documents/dev/cics/zowe-cli/packages/zosuss/src/cli/issue/ssh/Ssh.handler",
+ "handler": "/home/jenkins/workspace/brightside_PR-308-RTHJW7GWFKRUAWAYDOA2M62VHCCJSM6JZB7VMRHSPRYMOC2PQCEQ/packages/zosuss/src/cli/issue/ssh/Ssh.handler",
"name": "ssh",
"options": Array [
Object {
"description": "Working directory in which to execute the command",
"name": "cwd",


You have to delete the handler before storing it in a snapshot so it doesn't have a system specific path in the snapshot. I would fix it but I don't want to accidentally push the USS changes in again

ChrisPark89:
I don't know if that was right way to get rid of handler. but looking at the error message, I expect it to fix the problem!

ChrisBoehmCA:
Thanks, I think that's only one part of getting the test to work though.

Try `delete definition.handler` from your test before calling `toMatchSnapshot()`. Otherwise when the test runs, it will fail

ChrisPark89:
I had to delete the `definition.children` instead because handler is in the children object and I couldn't find a way to just delete the handler in the children.

ChrisBoehmCA:
Okay. could you make the same change in the lts-incremental PR please? thank you

ChrisPark89:
Done!

ChrisBoehmCA:
Thank you

PR #322, opened 2019-04-09
Updated description of overwrite (create workflow CLI) - master branch

muzikovam:
Updated snapshots
Variable input file can have two / in automation script (same as workflow definition file), so handled that in Create workflow API


muzikovam:
And now I have no idea what is wrong :disappointed:

Issue #320, opened 2019-04-08
Zowe Webinar (do not assign) - Help text URL link broken from VS Code


solsu01:
zowe --h displays the help text for the CLI.

In this display, there's a line that says "for additional zowe CLI documentation, visit <URL>."

![image](https://user-images.githubusercontent.com/35504644/55801578-5e9d1500-5aa4-11e9-8b15-a7d0c3691474.png)

Clicking on this link from VS-code gives me a 404 error on the browser. My guess is that this is because of the period at the end of the line which is being interpreted by the VS Code terminal as part of the URL.

![image](https://user-images.githubusercontent.com/35504644/55801591-6361c900-5aa4-11e9-8051-100cb061484d.png)

MikeBauerCA:
It is a valid issue and will be resolved by the PR I opened during the webinar (https://github.com/zowe/zowe-cli/pull/323).

PR #323, opened 2019-04-09
Resolve issue #320

MikeBauerCA:
Signed-off-by: Bauer, Michael <Michael.Bauer2@ca.com>


zFernand0:
> Looks like it was a hiccup with the npm audit step. Fixed by replaying the build

This seems to be happening quite often lately. I did a bit of research and could find much more than just the general suggestions (clear the cache, update the package.json, update npm).
We will have to come up with a plan to prevent builds from failing because of this.

Issue #207, opened 2019-02-06
Enhancements to dir-to-uss for encoding, file tagging and ignoring files


matthewpwilson:
## Description
This is a proposal for enhancements to `dir-to-uss` (introduced by PR #193) for file encoding, tagging and ignoring to be configured using a metadata file. Also provide an option to delete the destination dir. 

## Motivation
When uploading files and directories to USS, it's necessary to have control over which files are uploaded in binary mode, and which are converted from ASCII to EBCDIC.
`zowe files upload dir-to-uss` provides the `--binary-files` and `-ascii-files` options, however these require each file to be named explicitly, which is tedious and time-consuming.

In addition, many USS utilities can take advantage of file tagging to autoconvert data (e.g. from ISO8859-1 to IBM-1047).  The file upload commands should set tags appropriately (equivalent to `chtag -tc UTF-8 filename`.) 

Also, it can be useful to ignore sets of files to avoid transferring them entirely. 

These issues are particularly relevant when upload applications from a workstation to USS. For example, a Node.js application may contain a large number of files, which are a mixture of binary and text content. It will also contain a `node_modules` directory which should not be transferred as it may contain platform specific code and will be created when npm install is run on z/OS.   Also, if the application was previously deployed to the same location, it's desirable to delete the old directory first to avoid merging. 

## Implementation
A `.zosattributes` file in the source directory can be used to control file conversion and tagging. The `--binary-files` and `--ascii-files` options would override what's specified in file. Also these options should be extended to match wildcards. 

Here's an example of the file:
```
# pattern          local-encoding     remote-encoding
**                 ISO8859-1          ISO8859-1
*.profile          ISO8859-1          EBCDIC
**/*.png           binary             binary
node_modules    -                                         
```

The format is inspired by `.gitignore` and `.gitattributes`. Each line specifies a pattern to match a set of files, with the same [rules as .gitignore](https://git-scm.com/docs/gitignore).
Local encoding specifies the character set of the file found on the local workstation.
Remote encoding specifies the desired character set of the file on USS. It appears the z/OSMF Files REST API doesn't let us specify the target encoding, so we'll get whatever EBCDIC codepage the system is configured for.    Therefore the remote encoding must either match the local encoding, or be EBCDIC. 
If remote encoding is EBCDIC, the file is transferred in text mode and converted.

Otherwise the file is transferred in binary mode.
The remote file is tagged with the remote encoding or as binary.
[...]

PR #334, opened 2019-04-18
upload dir-to-uss with .zosattributes

matthewpwilson:
Add support for a .zosattributes metadata file as described in #207 

We weren't quite sure what to do with the description of the file format.  I've added it to the help for dir-to-uss, but it does make it quite long. Also it's not easy to format the example nicely.  https://github.com/zowe/imperative/pull/203 is needed in order to render the Markdown correctly. We have a few minor updates to words to come too but wanted to get this PR raised for review. 

The default behaviours when using a .zosattributes file turn out to be the opposite of the existing defaults of dir-to-uss i.e. dir-to-uss defaults to uploading in text mode (converting to EBCDIC), whereas with a .zosattributes the default is uploading in binary and tagging as ISO8859-1.   We felt this was a more sane default for application deployment use-cases but it's potentially confusing having different defaults.


ChrisBoehmCA:
Hi, thanks for the changes

>The default behaviours when using a .zosattributes file turn out to be the opposite of the existing defaults of dir-to-uss i.e. dir-to-uss defaults to uploading in text mode (converting to EBCDIC), whereas with a .zosattributes the default is uploading in binary and tagging as ISO8859-1. We felt this was a more sane default for application deployment use-cases but it's potentially confusing having different defaults.

This would constitute a breaking change then and would mean it couldn't be ported to our lts-incremental branch, which means most people won't be using it for several months (possibly until january 2020, not sure of the timeline of the next major release) unless they choose to install with the `@latest` or `@daily` tags

matthewpwilson:
> This would constitute a breaking change
OK, so seems best to keep the currently behaviour for dir-to-uss without .zosattributes and live with the fact that the default is difference once a .zosattributes exists

ChrisBoehmCA:
So to clarify are you going to change the implementation so that it's no longer a breaking change? Thanks

matthewpwilson:
I don't think it is a breaking change currently - I've not changed the existing behaviour or defaults when using dir-to-uss without a .zosattributes file. It's only when you use a .zosattributes file the default behaviour changes.  Is this acceptable?

ChrisBoehmCA:
Yeah, that makes sense. Thanks. Please resolve the merge conflicts in this PR and then I will try to review soon

matthewpwilson:
Should be ready for review now. I'm seeing one test case in `packages/zosfiles/__tests__/__system__/cli/upload/dtu/cli.dir.upload.dtu.system.test.ts` fail:
```
Upload directory to USS › Success scenarios › should give error when upload local directory to
 USS directory in default ascii if it contains also binary files

    expect(received).toContain(expected) // indexOf

    Expected substring: "Rest API failure with HTTP(S) status 500"
    Received string:    ""Upload directory to USS › Success scenarios › should give error when upload local directory to
 USS directory in default ascii if it contains also binary files

    expect(received).toContain(expected) // indexOf

    Expected substring: "Rest API failure with HTTP(S) status 500"
    Received string:    ""
```
but this fails for me at master as well so I don't believe I've regressed it.

ChrisBoehmCA:
Thanks. I'll open an issue for the failing test

matthewpwilson:
Thanks for sorting out those tests, I am indeed a Linux user so hadn't seen those issues. 
I'll port to lts-incremental in due course.

PR #348, opened 2019-04-26
List defined systems

uzust01:
Add _list systems_ to zosmf group:
- zowe zosmf list systems


packages/zosmf/__tests__/__system__/cli/list/systems/zosmf.list.systems.system.test.ts
```diff
@@ -0,0 +1,124 @@
...
+describe("zosmf list systems", () => {
+
+    // Create the unique test environment
+    beforeAll(async () => {
+        testEnvironment = await TestEnvironment.setUp({
+            tempProfileTypes: ["zosmf"],
+            testName: "zos_list_systems"
+        });
+
+        systemProps = new TestProperties(testEnvironment.systemTestProperties);
```
ChrisBoehmCA: Hi, 

You will have to update this slightly because of #332 

use host = testEnvironment.systemTestProperties.zosmf.host; instead of this  new TestProperties syntax

packages/zosmf/__tests__/__integration__/cli/list/zosmf.list.systems.integration.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import { runCliScript, stripNewLines } from "../../../../../../__tests__/__src__/TestUtils";
+
+let testEnvironment: ITestEnvironment;
+
+describe("zosmf list systems", () => {
+
+    // Create the unique test environment
+    beforeAll(async () => {
+        testEnvironment = await TestEnvironment.setUp({
+            testName: "zos_list_systems_integration"
```
ChrisBoehmCA: Please add skipProperties: true  here

ChrisBoehmCA:
Hi, 

Cool command! Worked for me 

I think having the message "Number of retreived system definitions:" is not consistent with other commands. Also there is a typo (retreived -> retrieved). @tucker01  and I think it should probably be removed and we should just display the results of the query like `list jobs` does.

PR #358, opened 2019-04-30
Added return code to ssh command

ChrisPark89:


ChrisBoehmCA:
Hi, I'm checking out these changes now. One thing I noticed though is that you have the ssh profile as required. Could you change this to optional in packages/zosuss/src/cli/issue/ssh/Ssh.definition.ts? That way people can use the connection options without a profile if they want to

ChrisBoehmCA:
Also I noticed that when my password expired (as it did just now) the ssh command seemed to hang forever. Any idea what could cause this?

ChrisPark89:
> Also I noticed that when my password expired (as it did just now) the ssh command seemed to hang forever. Any idea what could cause this?

I'm having a look at this

ChrisPark89:
I can't recreate the issue. For my case, when the password is expired, the ssh command gives this error: 
`Command Error:
All configured authentication methods failed`
Could it be the handshakeTimeout being set too long?

ChrisBoehmCA:
It could be -- I'm using the default setting so it doesn't time out.  If the timeout was changed it would probably fail.  Seems acceptable to have to adjust the timeout I guess to avoid the issue

ChrisPark89:
Thanks for the review!

zFernand0:
It seems there is a failing unit test.
Could you check it out?

ChrisPark89:
Sorry! I forgot to change the unit test accordingly. It should pass now.

PR #359, opened 2019-04-30
Added return code to ssh command (same as #358) 

ChrisPark89:
Same update as #358.


zFernand0:
Same as: https://github.com/zowe/zowe-cli/pull/358#issuecomment-488791802_

PR #363, opened 2019-05-01
Autocompletion

vvvlc:
Poor-mans bash autocompletion including instructions how to activate it.


zFernand0:
@vvvlc, commits need to be signed off.

Issue #33, opened 2018-10-09
Generalize provisioning automated tests to be system-independent


JTonda:
_From @ChrisBoehmCA on September 6, 2018 12:26_

Currently the provisioning tests only pass on a certain system that has been pre-configured with a provisioning workflow. It would be ideal to be able to run these tests on any system

Maybe we could store the workflow XML in the test resources directory and call the z/OSMF REST APIs to define the workflow, then delete it at the end of the tests so that it could be run on any system? 

_Copied from original issue: gizafoundation/brightside#233_

tucker01:
There is also an issue where if you run tests in parallel some will fail because the instances have already been provisioned. I think we need to clean up provisioned instances between runs.

AHumanFromCA:
@dogboydog and I looked into this a bit earlier today. It looks like there might be some zosmf apis that we could use to make all the provisioning tests self contained.

See:

[Zosmf Software Services Template APIs](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.izua700/izuprog_API_SoftwareCatalogServices.htm)

- [Create Template](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.izua700/izuprog_API_SoftwareTemplateCreateNew.htm?view=kc)
- [Delete Template](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.izua700/izuprog_API_SoftwareTemplateDelete.htm?view=kc)

nurra01:
@dogboydog  @AHumanFromCA 
I think for now it is impossible fully automate provisioning system tests. Even if it is possible to create a template

nurra01:
@ChrisBoehmCA @AHumanFromCA @tucker01  I researched this issue a bit and documented what are requirements. Provisioning system tests are slow now, and if we add more steps to it will be much slower.
[research.docx](https://github.com/zowe/zowe-cli/files/3072391/research.docx)
.

ChrisBoehmCA:
Thanks for documenting the requirements. Our goal is to 
a) get the tests passing on our current system we're assigned to. Right now they fail
b) make the tests as portable as possible so that when we do move to a new system, they can still pass

I understand the z/OSMF security is a complication but maybe we can get these rights added to our user IDs or at least a shared build user ID. The other tests such as console commands require a decent amount of security as well.

[...]

PR #365, opened 2019-05-02
Refactor prov tests

nurra01:
Refactor provisioning system tests to fix issue #33 
User will have to pass template-name to which has all required rights to provision, list, delete, etc.
As jest runs all system tests in once, if z/OSMF is slow or template is badly configured some tests can fail.


nurra01:
Ready to be merged if there are no suggestions

zFernand0:
@nurra01, please sign-off your commits next time

PR #369, opened 2019-05-07
Fix prov tests

nurra01:
The same pull request as #365 for lts-incremental, shouldn't have any breaking changes.


nurra01:
@zFernand0  it should be signed now

PR #382, opened 2019-05-14
fix broken link Readme > FAQ

JamesBauman:
Signed-off-by: bauja03 <james.bauman@broadcom.com>

Only one FAQ.. removed mini-toc, changed to bullet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,12 +134,10 @@ If the `custom_properties.yaml` file cannot be found or loaded, an error with re
 
 ## Frequently Asked Questions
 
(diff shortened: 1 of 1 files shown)


JamesBauman:
@jackjia-ibm 

Hi Jack - this PR is to fix a broken link in the readme file.  is there anything that i can do from my end to correct the commit/PR?   Or can only you solve the problem?

thx

ChrisBoehmCA:
I don't think there's anything to fix on here. Just the vulnerabilities check is failing which is failing for everyone right now until a fix is published for the vulnerabilities

ChrisBoehmCA:
No as far as I know you'd have to make a PR for each other branch. But personally I think not many people will be looking at the README in anything but master so maybe it's not a big deal if master is the most up to date one.

PR #399, opened 2019-05-23
Allow users to delete multiple workflows using workflow name with wildcard

uzuko01:
Allow users to delete multiple workflows using workflow name with wildcard


uzuko01:
@ChrisBoehmCA thanks for the review, I will remove the unnecessary snapshots and add system tests on the CLI level.

PR #465, opened 2019-06-26
Correct incremental readme - install from source

BrandonJenkins14:
Equivalent to PR #464 , but for incremental branch.


zFernand0:
@BrandonJenkins14,
Thanks for changing the target branch. The DCO failed because of that mismatch initially, but now the SonarCloud server seems to be queued up.
I'll go ahead and merge these PRs
Thanks for fixing the issue at hand.

PR #467, opened 2019-06-30
Allow JCL to be submitted from stdin

crshnburn:
While working on a project it became clear that being able to supply the JCL to submit on stdin would be useful as it would allow the Zowe-cli program to run as part of a unix pipeline.

The change adds a new `stdin` option which can be used in place of the existing data set or local file options.

Example of usage:
`mustache config.json template.jcl | zowe jobs sub stdin --vasc`

Signed-off-by: Andrew Smithson <smithson@uk.ibm.com>

### DIFF
--- a/packages/zosjobs/src/cli/submit/stdin/stdin.definition.ts
+++ b/packages/zosjobs/src/cli/submit/stdin/stdin.definition.ts
@@ -1,0 +1,45 @@
+/*
+* This program and the accompanying materials are made available under the terms of the
+* Eclipse Public License v2.0 which accompanies this distribution, and is available at
+* https://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import { ICommandDefinition, ICommandOptionDefinition } from "@zowe/imperative";
+
+export const StdinDefinition: ICommandDefinition = {
+    name: "stdin",
+    aliases: ["in"],
+    type: "command",
+    summary: "Submit a job read from standard in",
+    description: "Submit a job (JCL) passed to the command via the stdin stream. " +
+        "The command presents errors verbatim from the z/OSMF Jobs REST endpoints. " +
+        "For more information about z/OSMF Jobs API errors, see the z/OSMF Jobs API REST documentation.",
+    handler: __dirname + "/../Submit.shared.handler",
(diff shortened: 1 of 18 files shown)


zFernand0:
Hey @crshnburn, 
This looks great! Thanks for adding this functionality.
Would you mind updating that one snapshot with the new `stdin` option?

packages/zosjobs/src/cli/submit/stdin/stdin.definition.ts
```diff
@@ -0,0 +1,45 @@
...
+    type: "command",
+    summary: "Submit a job read from standard in",
+    description: "Submit a job (JCL) passed to the command via the stdin stream. " +
+        "The command presents errors verbatim from the z/OSMF Jobs REST endpoints. " +
+        "For more information about z/OSMF Jobs API errors, see the z/OSMF Jobs API REST documentation.",
+    handler: __dirname + "/../Submit.shared.handler",
+    options: ([
+        {
+            name: "view-all-spool-content", aliases: ["vasc"],
+            description: "View all spool content for specified job ID",
```
tucker01: This description doesn't look quite right here?

tucker01:
Also, would it be possible to add a CLI system test for this?

packages/zosjobs/src/cli/submit/stdin/stdin.definition.ts
crshnburn: I've updated the description, originally it had been copied from the local-file definition but this suffers from the same issues, so have updated them to match the dataset command.

crshnburn:
> Also, would it be possible to add a CLI system test for this?

Yep, I'll add a system test, plus unit test for the new definition.

tucker01:
Thanks for adding the tests.

tucker01:
Actually, unless I missed it, looks like a help test was added, but no system/CLI test script to submit a job from stdin. Mind adding one more? You can model it after the tests found here: https://github.com/zowe/zowe-cli/blob/master/packages/zosjobs/__tests__/__system__/cli/submit/local-file/cli.zos-jobs.submit.local-file.system.test.ts

crshnburn:
I've added some tests which mirror the local file tests but pipe the JCL into the Zowe CLI and use the stdin option.

tucker01:
Awesome. Thanks for contributing! Once, the merge with base branch build competes. I'll approve and merge.

PR #479, opened 2019-07-10
Testing Codecov

jellypuno:
Add the coverage.xml that needs to be uploaded


Jenkinsfile
```diff
@@ -141,6 +141,17 @@ node('ca-jenkins-agent') {
         junitOutput: INTEGRATION_JUNIT_OUTPUT
     )
 
+    //Upload Reports to Code Coverage
+    pipeline.createStage(
+        name: "Codecov",
+        stage: {
+            withCredentials([usernamePassword(credentialsId: 'CODECOV_ZOWE_CLI', usernameVariable: 'CODECOV_USERNAME', passwordVariable: 'CODECOV_CLI_TOKEN')]) {
+                sh "echo ${CODECOV_USERNAME} ${CODECOV_CLI_TOKEN}"
```
zFernand0: Please do not try to output secret keys to the job log.
Even though Jenkins does it's best to keep them hidden, Once you figure out how to "bypass" their masking, it will be kept in the logs unless you manually delete it.

README.md
```diff
@@ -1,5 +1,7 @@
 # Zowe CLI  <!-- omit in toc -->
 
+[![codecov](https://codecov.io/gh/zowe/zowe-cli/branch/master/graph/badge.svg)](https://codecov.io/gh/zowe/zowe-cli)
```
zFernand0: Awesome addition of the badge!

.codecov.yml
```diff
@@ -0,0 +1,8 @@
+coverage:
+  status:
+    patch:
+      default:
+        target: 60
+    project:
+      default:
+        target: 55
```
zFernand0: I'm sure we can set a higher target in the future :)

Jenkinsfile
jellypuno: thank you for reviewing :) I am debugging it that's why there's an echo. But I will certainly delete it.

PR #482, opened 2019-07-12
Export some CLI functionality to an API]

uzust01:
No changes in commands or options.


packages/workflows/__tests__/api/Properties.unit.test.ts
```diff
@@ -713,7 +869,21 @@ describe("Get workflow properties", () => {
...
+            try {
+                response = await PropertiesWorkflow.processStepSummaries(STEP_WITH_SUBSTEPS);
+                Imperative.console.info(`Response ${response}`);
+            } catch (thrownError) {
+                error = thrownError;
+                Imperative.console.info(`Error ${error}`);
+            }
 
+            expectZosmfResponseSucceeded(response, error);
+            expect(response).toEqual(FLATTEN_STEP);
```
tucker01: Totally up to you, but you could expect to match a snapshot. This way you wouldn't need to maintain the validation structure.

uzust01: Thanks @tucker01 , I have been having some issues with the build (I know why they are but I will not get into details here) so I decided to make it like this.

packages/workflows/src/api/Properties.ts
```diff
@@ -54,5 +56,53 @@ export class PropertiesWorkflow {
...
+            if(step.submitAs && step.submitAs.match(/.*JCL/)) {
+                if(step.jobInfo && step.jobInfo.jobstatus) {
+                    miscValue = step.jobInfo.jobstatus.jobid;
+                }
+            } else if(step.template) {
+                miscValue = "TSO";
+            } else if(step.isRestStep) {
+                miscValue = `HTTP ${step.actualStatusCode}`;
+            }
+            const stepSummary: IStepSummary = {
```
tucker01: Are you dropping any of the fields from `step` when creating `stepSummary`? You could use the spread operator here and append the `misc` property.

zFernand0:
@uzust01, @tucker01,
Should this be ported to master?

tucker01:
Yes.

PR #487, opened 2019-07-18
Include more options for dsntype

Colin-Stone:
Signed-off-by: Colin-Stone <30794003+Colin-Stone@users.noreply.github.com>

### DIFF
--- a/packages/zosfiles/__tests__/api/methods/create/Create.unit.test.ts
+++ b/packages/zosfiles/__tests__/api/methods/create/Create.unit.test.ts
@@ -14,7 +14,7 @@ import { Create, CreateDataSetTypeEnum, ZosFilesConstants, CreateDefaults, Invok
 import { ZosmfRestClient } from "../../../../../rest/";
 import { ZosFilesMessages } from "../../../../src/api/constants/ZosFiles.messages";
 
-describe("Create data set", () => {
+describe.only("Create data set", () => {
     const dummySession: any = {};
     const dataSetName = "testing";
     const dsOptions: any = {alcunit: "CYL"};
@@ -40,7 +40,7 @@ describe("Create data set", () => {
             expect(mySpy).toHaveBeenCalledWith(dummySession, endpoint, [], JSON.stringify({...CreateDefaults.DATA_SET.PARTITIONED, ...dsOptions}));
         });
 
-        it("should be able to create an extended partitioned data set (PDSE)", async () => {
+        it("should be able to create an extended partitioned data set (PDSE) - test with LIBRARY", async () => {
 
             dsOptions.dsntype = "LIBRARY";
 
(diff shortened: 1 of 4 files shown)


packages/zosfiles/__tests__/api/methods/create/Create.unit.test.ts
```diff
@@ -52,7 +52,42 @@ describe("Create data set", () => {
...
+
+        it("explicit testing of dsntype", async () => {
+            let success: boolean = false;
+            const dsntypes = ["BASIC", "EXTPREF", "EXTREQ", "HFS", "LARGE", "PDS", "LIBRARY", "PIPE"];
+            for (const type of dsntypes) {
+                dsOptions.dsntype = type;
+                try {
+                    await Create.dataSetValidateOptions(dsOptions);
+                } catch (err) {
+                    expect(success).toBe(true);
```
zFernand0: Could we maybe output more information to see which option failed?
As it is right now, people will just see a failure on this test without much details as to what went wrong with the test itself.

packages/zosfiles/__tests__/api/methods/create/Create.unit.test.ts
```diff
@@ -52,7 +52,42 @@ describe("Create data set", () => {
...
+                }
+            }
+            try {
+                dsOptions.dsntype = "PDSE";
+                await Create.dataSetValidateOptions(dsOptions);
+            } catch (err) {
+                success = true;
+            }
+            expect(success).toBe(true);
```
zFernand0: Same as last comment, we will just see "expected false to be true". I guess the line number will help us in this case.

packages/zosfiles/src/api/methods/create/Create.ts
```diff
@@ -189,15 +189,28 @@ export class Create {
...
+                            case "HFS":
+                                break;
+                            case "LARGE":
+                                break;
+                            case "PDS":
+                                break;
                             case "LIBRARY":
                                 break;
+                            case "PIPE":
+                                break;
```
kristinochka: Why does each of these values have a separate `case`? 
What about something like:
```
const type = tempOptions.dsntype.toUpperCase();
const availableTypes = ["BASIC", "EXTPREF", ...]
if (availableTypes.includes(type)) {
  return;
}
```

Colin-Stone: ECMAScript 2016 Array.prototype.includes() function doesn't appear to be available and it looks like CLI is at ES5. I am not aware of consequences of this decision so would need to defer to the rest of the squad

kristinochka: This can be done as an alternative in ES5
```
if (availableTypes.indexOf(type) !== -1) {
  return;
}
```

zFernand0:
Thanks for updating :smile:

hogstrom:
It appears this has been resolved.

Issue #325, opened 2019-04-09
`zowe files upload ftds` fails on `GET`.


dkelosky:
To write a file, it seems like [PUT](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.izua700/IZUHPINFO_API_PutWriteDataSet.htm) should be all that's necessary.   

However, I have several data sets (the latter two are archived):
- `user.WORK.JCL`
- `user.WORK.JCL.OLD2`
- `user.WORK.JCL.BACKUP`

If I download a file:
`zowe files download ds "user.WORK.JCL(SOMEMEM)"`

Then try to upload:
`zowe files upload ftds somemem "user.WORK.JCL(SOMEMEM)"`

I get a failure on `GET`:
```
$ zowe files upload ftds somemem "user.work.jcl(cashcreq)"
Command Error:
z/OSMF REST API Error:
Rest API failure with HTTP(S) status 500
rc:       4
reason:   3
details:
  - DMS2987 DATA SET CATALOGED TO CA DISK PSEUDO-VOLUME ARCIVE

  - DMS2987 CA DISK HAS ARCHIVED user.WORK.JCL.OLD2

  -
  - DMS2971 DO YOU WANT TO RESTORE THE DATA SET? (Y/N)
category: 2
message:  ServletDispatcher failed - received TSO Prompt when expecting TsoServletResponse


Error Details:
HTTP(S) error status "500" received.
Review request details (resource, base path, credentials, payload) and ensure correctness.

Host:      tso1
Port:      443
Base Path:
Resource:  /zosmf/restfiles/ds?dslevel=user.work.jcl
Request:   GET
Headers:   [{"X-IBM-Attributes":"base"},{"X-IBM-Max-Items":"0"},{"X-CSRF-ZOSMF-HEADER":true}]
Payload:   undefined
```

It seems like some upfront checking is occurring that does a `GET` but causes problems for additional LLQs which are archived.

Is there a way to bypass the `GET`?  Is this necessary?

ChrisBoehmCA:
There is currently no way to bypass the get. From looking at the code, it looks like this is done to determine whether the target data set is a PDS or PS. Looks like it's just to validate that you are not trying to upload multiple files to a non-pds data set. The check could potentially be removed.

dkelosky:
[...]

PR #489, opened 2019-07-22
Fix isse #325: `zowe files upload ftds` fails on `GET`

uzust01:
This seems to work for me, I made some manual tests (on top of the existing automated). 
Open for feedback :D 

https://github.com/zowe/zowe-cli/issues/325

### DIFF
--- a/packages/zosfiles/__tests__/__system__/api/methods/list/List.system.test.ts
+++ b/packages/zosfiles/__tests__/__system__/api/methods/list/List.system.test.ts
@@ -198,11 +198,12 @@ describe("List command group", () => {
                 expect(response.apiResponse.items[0].dsname).toEqual(dsname);
             });
 
-            it("should list a data set with attributes", async () => {
+            it("should list a data set with attributes and start options", async () => {
                 let error;
                 let response: IZosFilesResponse;
                 const option: IListOptions = {
-                    attributes: true
+                    attributes: true,
+                    start: dsname
                 };
 
                 try {
--- a/packages/zosfiles/__tests__/api/methods/list/List.unit.test.ts
+++ b/packages/zosfiles/__tests__/api/methods/list/List.unit.test.ts
@@ -315,7 +315,7 @@ describe("z/OS Files - List", () => {
                 items: ["test"]
             };
             const endpoint = posix.join(ZosFilesConstants.RESOURCE,
-                `${ZosFilesConstants.RES_DS_FILES}?${ZosFilesConstants.RES_DS_LEVEL}=${dsname}[&start=${dsname}]`);
+                `${ZosFilesConstants.RES_DS_FILES}?${ZosFilesConstants.RES_DS_LEVEL}=${dsname}&start=${dsname}`);
 
             expectJsonSpy.mockResolvedValue(testApiResponse);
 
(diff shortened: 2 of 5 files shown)


tucker01:
While I agree with some of the comments in the issue (about refactoring), this does seem to work as a simple fix. 

However, my concern is that this change relies on the collating sequence that IBM has defined for the return list (e.g. DATA.SET will be index 0 if DATA.SET.OTHER exists). Although, I suppose it wouldn't make much sense if you requested to list DATA.SET with max items 1 and z/OSMF returned DATA.SET.OTHER. It also depends on IBM not attempting to gather attributes on items that were filtered out by setting `X-IBM-Max-Items: 1`.

IBM does provide a `start` query param on ds list, that looks like it could be used to ensure you're getting exactly the data-set you want if you used `dsname=DATA.SET&start=DATA.SET` with the `X-IBM-Max-Items: 1`.

What do you think?

uzust01:
> While I agree with some of the comments in the issue (about refactoring), this does seem to work as a simple fix.
> 
> However, my concern is that this change relies on the collating sequence that IBM has defined for the return list (e.g. DATA.SET will be index 0 if DATA.SET.OTHER exists). Although, I suppose it wouldn't make much sense if you requested to list DATA.SET with max items 1 and z/OSMF returned DATA.SET.OTHER. It also depends on IBM not attempting to gather attributes on items that were filtered out by setting `X-IBM-Max-Items: 1`.
> 
> IBM does provide a `start` query param on ds list, that looks like it could be used to ensure you're getting exactly the data-set you want if you used `dsname=DATA.SET&start=DATA.SET` with the `X-IBM-Max-Items: 1`.
> 
> What do you t

uzust01:
> While I agree with some of the comments in the issue (about refactoring), this does seem to work as a simple fix.
> 
> However, my concern is that this change relies on the collating sequence that IBM has defined for the return list (e.g. DATA.SET will be index 0 if DATA.SET.OTHER exists). Although, I suppose it wouldn't make much sense if you requested to list DATA.SET with max items 1 and z/OSMF returned DATA.SET.OTHER. It also depends on IBM not attempting to gather attributes on items that were filtered out by setting `X-IBM-Max-Items: 1`.
> 
> IBM does provide a `start` query param on ds list, that looks like it could be used to ensure you're getting exactly the data-set you want if you used `dsname=DATA.SET&start=DATA.SET` with the `X-IBM-Max-Items: 1`.
> 
> What do you think?

Idk what I did there with the close and reopen :D 

I was thinking about this, too, and I dropped it.  In `dsname=DATA.SET&start=DATA.SET` DATA.SET will always be the same and regardless of start= being there or not the list will always start at DATA.SET.

uzust01:
Added <start> to the interface for the options parameter of List.dataSet API (and condition to handle it)
This option is not yet exposed to any CLI. It is there to make sure that the now limited to 1 data set list (used in used in Upload.pathToDataSet) starts with the correct data set. 

These changes do not fix the bug described in #325 completely. If the initial data set does not exist and the subsequent data set is archived the same error will occur. 

Currently (to my knowledge) there is no way to determine if a data set is archived without using any "X-IBM-Attributes" and using them will generate a TSO recall prompt asking if you want to restore the data-set.

tucker01:
Thanks @uzust01!

Eventually, we can refactor the `pathToDataSet` API into multiple explicit APIs to avoid the `GET` request. However, these changes should avoid the error as initially reported. If we decide to change the API and interfaces that would be a major breaking change given that they are exported/public, so I'm in favor of this simpler fix for now. 

Plus, these changes add some additional functionality to our existing list APIs, which is great. I have helpers in various projects that determine if a data set exists, and using the max items 1 with the start query param (without requesting attributes) will help me there. 

I'll open a separate issue to track the remaining problem and suggest a rework of the existing APIs when we have more time to devote to it.

tucker01:
Also, if you get a chance, please port to master.

zFernand0:
I like how this was resolved.
Very clever way to fix it! 😄

Issue #429, opened 2019-05-30
Add `zowe zos-files list fs` command


tjohnsonBCM:
Implement command for listing mounted z/OS file systems as documented [here](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.izua700/IZUHPINFO_API_GetListUnixFilesystems.htm#IZUHPINFO_API_GetListUnixFilesystems).

Usage: `zowe zos-files list zfs [options]`

Options:
```
--path | -p (string)
  Specifies the path where the file system is mounted.
  This option and --fsname are mutually exclusive.
--fsname | -f (string)
  Specifies the name of the mounted file system.
  This option and --path are mutually exclusive.
--max-length | --max (number)
  Specifies the maximum number of items to return.
  Default value: 1000
```

Examples:
```
To list all mounted file systems:
  $ zowe zos-files list zfs

To list file systems mounted to a specific path:
  $ zowe zos-files list zfs -p /a/ibmuser

To list file systems mounted with a specific name:
  $ zowe zos-files list zfs -f MY.ZFS
```

tucker01:
This was merged to master, but I'm going to add the tag to port to lts-incremental.

tjohnsonBCM:
Before porting to lts-incremental, this command should be changed to `zowe zos-files list fs` (dropping the `z`), since it is not specific to ZFS and can list other types of Unix file systems.

The reasoning behind this change is that according to [the IBM docs for z/OSMF](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.izua700/IZUHPINFO_API_RESTFILES.htm), the API endpoint for this command is `/zosmf/restfiles/mfs`. For commands that are specific to ZFS (`zos-files create/delete zfs`), the endpoint is `/zosmf/restfiles/mfs/zfs`.

tucker01:
Also, looks like we need an API system test as well.

cycaoecw:
@tucker01 should I change it to fs?

tucker01:
It's our fault having the original issue `list zfs`... we didn't make the connection until we went to create other filesystem commands (mount/unmount/etc.) that you can do more than zfs. 

If you would like to make the changes that would be excellent. It's up to you.

tjohnsonBCM:
Along with changing `zfs` to `fs`, `file-system` should be added as a long form of the command `fs`.

[...]

PR #496, opened 2019-07-29
PR for Add `zowe zos-files list zfs` command #429

cycaoecw:
Re-create PR for #429


cycaoecw:
@tucker01 Hi, I open another PR for it, would you have a look?

tucker01:
Hi @cycaoecw, thanks! Was out for a bit, but I'll make sure to review this in the next day or so...

packages/zosfiles/src/cli/list/zfs/Zfs.handler.ts
```diff
@@ -0,0 +1,54 @@
...
+        else
+        {
+            response = await List.zfs(session, {
+                path: null,
+                fsname: commandParameters.arguments.fsname,
+                maxLength: commandParameters.arguments.maxLength
+            });
+        }
+
+        if (commandParameters.arguments.attributes && response.apiResponse.items.length > 0) {
```
tucker01: I know some other commands have an `--attributes` argument, but it doesn't look like this is present on your command definition? Can this code be removed?

packages/zosfiles/src/cli/list/zfs/Zfs.handler.ts
```diff
@@ -0,0 +1,54 @@
...
+            commandParameters.response.console.log(TextUtils.prettyJson(response.apiResponse.items));
+        } else {
+            commandParameters.response.data.setObj(response);
```
tucker01: Can we move this line right after the API calls? This way it'll always get set?

cycaoecw:
Hi @tucker01 thanks for your suggest! I have do some update, but not sure why codecov/patch is not pass. For if you supply neither fsname or path what is the behavior? It will list all mount point.

packages/zosfiles/src/cli/list/zfs/Zfs.handler.ts
```diff
@@ -0,0 +1,55 @@
...
+            response = await List.zfs(session, {
+                path: null,
+                fsname: commandParameters.arguments.fsname,
+                maxLength: commandParameters.arguments.maxLength
+            });
+        }
+
+        commandParameters.response.data.setObj(response);
+
+        if (response.apiResponse.items.length > 0) {
```
tucker01: This should probably be removed also... just outputting the table (via `commandParameters.response.format.output`) is sufficient for the list commands.

tucker01:
I must have overlooked this, but it looks like there are no unit tests for `Zfs.handler.ts`? Probably why the codecov failed.

cycaoecw:
@tucker01 yes, you are right, seems it really missing. let me check why

cycaoecw:
@tucker01 I added back the unit test, but seems it still can not pass the codecov

tucker01:
It looks like the unit tests for `Zfs.handler.ts` are incorrect? It looks like the file committed as `Zfs.handler.unit.test.ts` is a copy of the contents from `List.unit.test.ts`. The codecov report is showing the `Zfs.handler.ts` file to only be 10% tested. 

If you write some unit tests for `Zfs.handler.ts` I think we'd be good to go... you can model your handler unit tests after https://github.com/zowe/zowe-cli/blob/master/packages/zosfiles/__tests__/cli/list/am/AllMembers.handler.unit.test.ts. 

Sorry for the confusion. Let me know if you have more questions.

cycaoecw:
Hi @tucker01 sorry that I am not aware the unit test files had re-structured. For the ZfsHandler unit test, I backup List.unit.test before and just using the backup as Zfs.handler.unit.test.ts last night. Let me update it and let you know.

tucker01:
Sounds good. Thank you!

cycaoecw:
Hi @tucker01 , sorry that take some times on it. I think should be ok now.

tucker01:
No problem at all, there is no rush to finish anything 😄 . I'll take a look today.

PR #501, opened 2019-08-06
Add script to automate building web help

tjohnsonBCM:
Also add logo to display above web help command tree


packages/imperative.ts
```diff
@@ -30,6 +30,7 @@ const config: IImperativeConfig = {
     rootCommandDescription: Constants.DESCRIPTION,
     defaultHome: Constants.HOME_DIR,
     envVariablePrefix: Constants.ENV_PREFIX,
+    webHelpLogoImgPath: require("find-up").sync("web-help/logo.png", {cwd: process.mainModule.filename}),
```
tucker01: I think in order for the logo to be included in the package, it needs to be listed under `files` in `package.json`? 

Also, every require/dependency added to initialization of the CLI decreases response time. Is `find-up` absolutely necessary here?

tjohnsonBCM: Oops, I did forget to list the file in `package.json`. Thanks for noticing that, will fix next week.

I believe the location of the image file could vary based on whether the CLI repo has been cloned from GitHub or installed as a package from NPM. To eliminate the extra dependency, I suppose it would be possible to detect the environment the CLI is running in to determine where to look for the file, like Imperative does [here](https://github.com/zowe/imperative/pull/262/files#diff-f8e90a58045f1997cb2d7de60bcdf18bR124).

tucker01: Sounds good thanks.

tjohnsonBCM: @tucker01 I removed usage of the `find-up` package and replaced what it was doing with a separate function (`tryFindWebHelpLogoImgPath`). Adding this function above the definition of `IImperativeConfig` feels wrong to me, but not sure where else to put it since I don't see a module for utility functions. I'm wondering if there's a better way to guide Imperative where to look for the logo image?

web-help/build-web-help.ts
```diff
@@ -0,0 +1,62 @@
...
+    // Exclude undesired command groups
+    const cmdDefinitions: any = imperativeModule.Imperative.fullCommandTree;
+    cmdDefinitions.children = cmdDefinitions.children
+        .filter((group: any) => (config.excludeGroups || []).indexOf(group.name) === -1);
+
+    // Build command help pages
+    const helpGenerator = new WebHelpGenerator(cmdDefinitions, myConfig, outDir);
+    helpGenerator.sanitizeHomeDir = true;
+    helpGenerator.buildHelp(new CommandResponse({ silent: false }));
+})();
```
tucker01: Since this is a build script, maybe we should handle any errors thrown and exit with a non-zero code (you can add a `catch` to your `async` IIFE). I think at this point in time (at least with node v10)... unhanded promise rejections don't result in a non-zero exit code... so the build wouldn't fail if your script encountered an issue.

PeteSwauger:
From my email: 

1. Is there any documentation on running build:webHelp?  I figured out I needed to copy example.config.yaml to config.yaml after getting the error message below.

> @zowe/cli@5.14.1-alpha.201908051621 build:webHelp C:\Users\ws617385\GitHub\zowe\zowe-cli
> ts-node ./web-help/build-web-help.ts

(node:7036) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods inst
ead.
(node:7036) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open 'C:\Users\ws617385\GitHub\zowe\zowe-cli\web-help/config.yaml'

2. Could you add a message indicating where the web help files are located after execution of build:webhelp?

tucker01:
Actually, for that matter, can we just hard-code the defaults we'll always use when generating the help via this script? Is there a need to customize the `config.yaml` each run?

tjohnsonBCM:
> Actually, for that matter, can we just hard-code the defaults we'll always use when generating the help via this script? Is there a need to customize the `config.yaml` each run?

That would be simpler, reason I created a separate example config file was so the `config.yaml` customized by the user wouldn't get checked in to Git. But I could instead commit the default version of the file and still list it in `.gitignore` so that user customizations won't get checked in.

PR #508, opened 2019-08-15
Fixes to build web help script

tjohnsonBCM:
Remove `config.yaml` file and always build help for CLI in repo.


tucker01:
Based on IRL conversation, we're going to simplify this a bit to work only against the code in the repo.

Issue #432, opened 2019-05-30
Add `zowe zos-files unmount fs` command


tjohnsonBCM:
Implement command for unmounting Unix file systems as documented [here](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.izua700/IZUHPINFO_API_UnmountUnixFile.htm#UnMountUnix).

Usage: `zowe zos-files unmount fs <fileSystemName>`

Options: none

Examples:
```
To unmount a mounted file system:
  $ zowe zos-files unmount fs MY.FS
```

Tasks:
- [X] Add command
  - [X] CLI
  - [X] API
  - [X] Test manually
- [x] Finalize summary and description for command
- [x] Add automated tests
  - [x] API unit tests
  - [x] API system tests
  - [x] CLI system tests
  - [x] CLI integration tests
  - [x] CLI handler unit tests

Issue #431, opened 2019-05-30
Add `zowe zos-files mount fs` command


tjohnsonBCM:
Implement command for mounting Unix file systems as documented [here](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.izua700/IZUHPINFO_API_MountUnixFile.htm#MountUnix).

Usage: `zowe zos-files mount fs <fileSystemName> [options]`

Options:
```
--mount-point | --mp (string)
  Specify the directory to use as a mount point.
--fs-type | --ft (string)
  Specify the file system type to mount.
  Default value: ZFS
--mode | -m (string)
  Specify the mode for mounting the file system (rdonly - read only, rdwr - read write).
  Default value: rdonly
```

Examples:
```
To mount a z/OS file system using default options:
  $ zowe zos-files mount fs MY.FS --mp /a/ibmuser/mountdir

To mount a hierarchical file system with write access:
  $ zowe zos-files mount fs MY.FS --mp /a/ibmuser/mountdir --ft HFS -m rdwr
```

Tasks:
- [X] Add command
  - [X] CLI
  - [X] API
  - [X] Test manually
- [x] Finalize summary and description for command
- [x] Add automated tests
  - [x] API unit tests
  - [x] API system tests
  - [x] CLI system tests
  - [x] CLI integration tests
  - [x] CLI handler unit tests

PR #513, opened 2019-08-22
Add zos files mount command

tjohnsonBCM:
* Resolves issue #431 by adding `mount fs` command
* Resolves issue #432 by adding `unmount fs` command


tucker01:
Hey, it's definitely not a problem here, but for your future PRs, we generally try to keep the changes as small as possible... In this case, we would open a PR per command (and each PR would address a single issue) 😄

packages/zosfiles/__tests__/__system__/api/methods/mount/Mount.system.test.ts
```diff
@@ -0,0 +1,173 @@
...
+        try{
+            response = await List.zfs(REAL_SESSION, {fsname});
+            Imperative.console.info("Response: " + inspect(response));
+        } catch (e) {
+            error = e;
+            Imperative.console.info("Error: " + inspect(error));
+        }
+
+        expect(error).toBeUndefined();
+        expect(response).toBeTruthy();
```
tucker01: Just curious, why `toBeTruthy` vs `toBeDefined`?

packages/zosfiles/__tests__/__system__/cli/mount/fs/cli.files.mount.fs.system.test.ts
```diff
@@ -0,0 +1,169 @@
...
+        });
+
+        defaultSystem = TEST_ENVIRONMENT.systemTestProperties;
+        volume = defaultSystem.datasets.vol;
+
+        REAL_SESSION = TestEnvironment.createZosmfSession(TEST_ENVIRONMENT);
+        fsname = getUniqueDatasetName(defaultSystem.zosmf.user);
+
+        const dirname = getUniqueDatasetName(defaultSystem.zosmf.user).split(".")[1];
+        mountPoint = "//tmp/" + dirname;
```
tucker01: should the extra slash be here?

packages/zosfiles/__tests__/__system__/cli/mount/fs/cli.files.mount.fs.system.test.ts
```diff
@@ -0,0 +1,169 @@
...
+        const sshCommand = "mkdir " + mountPoint;
+
+        const response = runCliScript(__dirname + "/__scripts__/command/command_setup.sh",
```
tucker01: It's totally fine how you have it, but in the `before` or `after` blocks, you can use the APIs instead of running a command.

packages/zosfiles/__tests__/__system__/cli/mount/fs/cli.files.mount.fs.system.test.ts
```diff
@@ -0,0 +1,169 @@
...
+            });
+
+            defaultSys = TEST_ENVIRONMENT_NO_PROF.systemTestProperties;
+        });
+
+        afterEach(async () => {
+            await TestEnvironment.cleanUp(TEST_ENVIRONMENT_NO_PROF);
+        });
+
+        it("should mount a file system", () => {
```
tucker01: Same comment here as the other test. They should be self contained and not depend on one another.

packages/zosfiles/__tests__/__system__/cli/mount/fs/cli.files.mount.fs.system.test.ts
```diff
@@ -0,0 +1,169 @@
...
+                    defaultSys.zosmf.port,
+                    defaultSys.zosmf.user,
+                    defaultSys.zosmf.pass]);
+            expect(response.stderr.toString()).toBeTruthy();
+            expect(response.status).toBe(1);
+        });
+    });
+
+    describe("With profiles", () => {
+        it("should mount a file system", () => {
```
tucker01: Same comment here as the other test. They should be self contained and not depend on one another.

packages/zosfiles/src/api/methods/mount/Mount.ts
```diff
@@ -0,0 +1,122 @@
...
+     */
+    public static async fs(
+        session: AbstractSession,
+        fileSystemName: string,
+        mountPoint: string,
+        options?: Partial<IMountFsOptions>)
+        : Promise<IZosFilesResponse> {
+        // We require the file system name and mount point
+        ImperativeExpect.toNotBeNullOrUndefined(fileSystemName, ZosFilesMessages.missingFileSystemName.message);
+        ImperativeExpect.toNotBeEqual(fileSystemName, "", ZosFilesMessages.missingFileSystemName.message);
```
tucker01: This is probably fine, but I think there is something along the lines of `toNotBeBlank`... it might also trim the value and check (to account for all spaces passed)?

packages/zosfiles/src/cli/mount/fs/fs.options.ts
```diff
@@ -0,0 +1,46 @@
...
+        description: strings.FSTYPE,
+        type: "string",
+        defaultValue: "ZFS"
+    },
+
+    /**
+     * The read/write mode for the file system
+     * @type {ICommandOptionDefinition}
+     */
+    mode: {
```
tucker01: In many cases, we don't perform validation of parameters for fear of falsely restricting something that is valid... however, in this case, the doc is pretty clear on the valid values for mode... and you have checks in your APIs to validate the values. I would add the available values on the definition.

packages/zosfiles/__tests__/api/methods/mount/Mount.unit.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    it("should succeed with correct parameters", async () => {
+        (ZosmfRestClient as any).putExpectString = jest.fn(() => {
+            // Do nothing
+        });
+        const options: IMountFsOptions = {
+            "fs-type":"ZFS",
+            "mode":"rdonly"
+        };
+        await Mount.fs(dummySession, fileSystemName, mountPoint, options);
```
tucker01: Can you add an expect here to ensure that the rest client is called one time?

packages/zosfiles/__tests__/api/methods/unmount/Unmount.unit.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    const dummySession: any = {};
+
+    describe("fs", () => {
+        const fileSystemName = "TEST.ZFS";
+
+        it("should succeed with correct parameters", async () => {
+            (ZosmfRestClient as any).putExpectString = jest.fn(() => {
+                // Do nothing
+            });
+            await Unmount.fs(dummySession, fileSystemName);
```
tucker01: Same comment as the mount unit tests... can we expect that the rest client is called?

packages/zosfiles/__tests__/cli/mount/fs/fs.handler.unit.test.ts
```diff
@@ -0,0 +1,180 @@
...
+            expect(profFunc).toHaveBeenCalledWith("zosmf", false);
+            expect(Mount.fs).toHaveBeenCalledTimes(1);
+            expect(Mount.fs).toHaveBeenCalledWith(fakeSession, fileSystemName, mountPoint, {});
+            expect(jsonObj).toMatchSnapshot();
+            expect(apiMessage).toMatchSnapshot();
+            expect(logMessage).toMatchSnapshot();
+        });
+    });
+
+    it("should raise an error", async () => {
```
tucker01: Can you make this title a bit more specific? Maybe include what error?

packages/zosfiles/__tests__/cli/unmount/fs/fs.definition.unit.test.ts
```diff
@@ -0,0 +1,36 @@
...
+        expect(definition.children).toBeUndefined();
+
+        // Should require a zosmf profile
+        expect(definition.profile.optional).toEqual(["zosmf"]);
+
+        // Should only contain one positional
+        expect(definition.positionals.length).toEqual(1);
+
+        // The positional should be required
+        expect(definition.positionals[0].required).toBeTruthy();
```
tucker01: This may be a nitpick, but the value should be exactly `true`. I'm sure we have this style expect everything in the code though.

packages/zosfiles/src/api/methods/mount/doc/IMountFsOptions.ts
```diff
@@ -0,0 +1,30 @@
...
+     * The file system type to mount
+     * @type {string}
+     */
+    "fs-type"?: string;
+
+    /**
+     * The mode for mounting the file system
+     * @type {string}
+     */
+    mode?: string;
```
tucker01: Since you have restricted this value to the supported values in the doc... how about we create a type for it? Helps out the programmer.

packages/zosfiles/__tests__/__system__/cli/mount/fs/cli.files.mount.fs.system.test.ts
tjohnsonBCM: Yes, it needs to be since the system tests run in Git Bash. Without the double slashes in front, `/tmp` is automatically expanded into the full path of the temp folder on the machine where tests are being run (e.g., `C:/Users/<user>/AppData/Local/Temp`).

packages/zosfiles/__tests__/__system__/api/methods/mount/Mount.system.test.ts
awharnBCM: We were going off of some of the other tests for ours, and found quite a few tests that used `expect(response).toBeTruthy`.

packages/zosfiles/__tests__/cli/mount/fs/fs.handler.unit.test.ts
tjohnsonBCM: Not sure if we can be more specific about the type of error, its just a fake one that we mocked out in the handler.

packages/zosfiles/__tests__/__system__/cli/mount/fs/cli.files.mount.fs.system.test.ts
tucker01: I know there is that oddity using gitbash, but what about on mac or linux? Would this cause a problem? The build system I believe uses Ubuntu containers.

packages/zosfiles/__tests__/cli/mount/fs/fs.handler.unit.test.ts
tucker01: It's not a big deal... sorry, I should have been more clear... What exactly is this test validating? The describe/its are supposed to read like a sentence: "Mount file system handler should raise an error" ?? Why? Does it always raise an error?

 The other test you have is more clear "Mount file system handler should mount a FS if requested".

packages/zosfiles/__tests__/__system__/cli/mount/fs/cli.files.mount.fs.system.test.ts
tjohnsonBCM: Have tested it in Ubuntu VM, and `//tmp` is treated the same as `/tmp`. I assume Mac behaves the same but would need to verify that. Since this is a system test, does it still run in a pipeline?

tucker01: OK great thanks.

packages/zosfiles/__tests__/cli/mount/fs/fs.handler.unit.test.ts
tjohnsonBCM: This test was copied over from other unit test handler suites that had the same one. Renamed it to "should fail to mount a FS when there is an error" which is hopefully a bit clearer. I believe the intent is to verify that if the `Mount.fs` API function fails, the handler function for the `mount` command should reflect that by returning an error.

tucker01: Your explanation of the intention of the test is what I was originally looking for... I know that we're not consistent everywhere. I think some of the tests have verbiage more like... "should not transform an error thrown from the rest client" or something along those lines.

packages/zosfiles/__tests__/__integration__/cli/mount/fs/__snapshots__/cli.files.mount.fs.integration.test.ts.snap
```diff
@@ -0,0 +1,127 @@
...
+   mountPoint		 (string)
+
+      The directory to use as a mount point.
+
+ OPTIONS
+ -------
+
+   --fs-type  | --ft (string)
+
+      Specify the file system type to mount (HFS, ZFS, NFS, TFS).
```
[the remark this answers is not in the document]
tjohnsonBCM: Updated to match IBM text

packages/zosfiles/src/api/methods/mount/doc/IMountFsOptions.ts
tjohnsonBCM: Added type declaration in IMountFsMode.ts:
```typescript
export type IMountFsMode =
    "rdonly" |
    "rdwr";
```

packages/zosfiles/src/api/methods/mount/Mount.ts
tjohnsonBCM: Replaced these two functions with single call to `ImperativeExpect.toBeDefinedAndNonBlank`

packages/zosfiles/src/cli/mount/fs/fs.options.ts
tjohnsonBCM: Added `allowedValues` to command option definition, it is also now restricted by the type `MountFsMode`

PR #519, opened 2019-09-04
Add Zowe logo to web help (lts-incremental)

tjohnsonBCM:


web-help/build-web-help.ts
```diff
@@ -0,0 +1,59 @@
+import * as fs from "fs";
+import * as path from "path";
+import { CommandResponse, Imperative, ImperativeConfig, WebHelpGenerator } from "@brightside/imperative";
+
+/**
+ * List of command groups to exclude from generated docs.
+ * For example, if you have the CICS plugin globally installed but don't want
+ * it included in the docs, add "cics" to this list.
+ */
+const excludeGroups: string[] = [];
```
tucker01: Is this a common occurrence? (or are there situations on the build machines where we'd want to exclude groups?).

tjohnsonBCM: This may not be a common occurrence, it was an option originally requested by @PeteSwauger. At the time we still needed to build web help manually because it was not yet in a pipeline. Given the way this is defined as a value at the top of the TypeScript file, I wasn't really expecting it to be used by build machines, but rather as a convenience for human users.

tucker01: OK Thanks, maybe I should have lead with the suggestion, which is to make it a script parameter, so we can adjust it when running the npm script, but if it's not going to be used much, then it's fine as is.

gejohnston:
The changes seem fine to me as well.

tjohnsonBCM:
Considering removal of `build-web-help.ts` script based on feedback from @tucker01

PR #527, opened 2019-09-09
Wrap "List.dataSet" (GET API call) in try-catch block to bypass CA Disk archived ds error

MatthewB19:
This is kind of a hack, but it seems that wrapping the GET API call in a try-catch block might temporarily solve the issue without losing functionality

Could you please take a look at the commit and see if it would be appropriate to narrow the catch block down to a specific rc & reason code?

### DIFF
--- a/packages/zosfiles/src/api/methods/upload/Upload.ts
+++ b/packages/zosfiles/src/api/methods/upload/Upload.ts
@@ -336,30 +336,35 @@ export class Upload {
 
         // Retrieve the information on the input data set name to determine if it is a
         // sequential data set or PDS.
-        const listResponse = await List.dataSet(session, dataSetName, {attributes: true, maxLength: 1, start: dataSetName});
-        if (listResponse.apiResponse != null && listResponse.apiResponse.returnedRows != null && listResponse.apiResponse.items != null) {
-            // Look for the index of the data set in the response from the List API
-            const dsnameIndex = listResponse.apiResponse.returnedRows === 0 ? -1 :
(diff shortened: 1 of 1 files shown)


tjohnsonBCM:
Thanks for the contribution and for looking into this issue! As you mentioned, wrapping the API in a try-catch block is not ideal. I believe the appropriate way to fix this would be to pass the HTTP header `X-IBM-Migrated-Recall` set to "wait". The HTTP headers for the z/OSMF list dataset API are documented [here](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.2.0/com.ibm.zos.v2r2.izua700/IZUHPINFO_API_GetListDataSetMembers.htm).

The "files upload" command group has a "--migrated-recall" CLI option that already does this, so it is a good example. This CLI option could probably be copied over to "files list" to add the same functionality.
https://github.com/zowe/zowe-cli/blob/d850655e9b51a280572aa589c0f2357425476d40/packages/zosfiles/src/cli/upload/Upload.options.ts#L47-L57

https://github.com/zowe/zowe-cli/blob/d850655e9b51a280572aa589c0f2357425476d40/packages/zosfiles/src/api/methods/upload/Upload.ts#L174-L190

tjohnsonBCM:
Consulted with a teammate about this, CA Disk may not support the `X-IBM-Migrated-Recall` value. Going to look into this further and post an update when I find out more.

MatthewB19:
@tjohnsonBCM Thanks!

When searching for z/OSMF REST API on Google, IBM only has the 2.1.0 version indexed in the search results.
So I wasn't even aware that there was a 2.2.0 version with more options for the request header

MatthewB19:
Closing this PR

PR #528, opened 2019-09-09
Fixed typo in ZosmfHeaders + added recall option to List

### LINKED ISSUE #493: If CA DISK is installed, the `zowe files upload ftds` may fail with an "archive" error if the data-set does not exist
When issuing `zowe files upload ftds` if the data set specified does NOT exist (e.g. DATA.SET), but a data set that matches its "pattern/level" exists and is archived by CA DISK (e.g. DATA.SET.ARCHIVE) the command will fail with the CA DISK prompts to un-archive. 

Fixing this issue should be a rework of the `pathToDataSet` API to remove the necessity for the `GET` request. We can allow z/OSMF to fail if the data-set upload conditions are not correct, instead of checking for the data-set organization in our code.

However, keep in mind that on z/OS it is possible to overlay the directory information in a PDS if, for example, you allocate and write to it as a physical sequential data-set. I'm not sure if z/OSMF would allow this, but it's something to keep in mind. 

This issue is an extension of #325

MatthewB19:
Still doesn't fix #493, but might be addressed in the future on the CA Disk-side

### DIFF
--- a/packages/rest/src/ZosmfHeaders.ts
+++ b/packages/rest/src/ZosmfHeaders.ts
@@ -153,8 +153,8 @@ export class ZosmfHeaders {
      * @static
      * @memberof ZosmfHeaders
      */
-    public static readonly X_IBM_MIGRATED_RECALL_WAIT: IHeaderContent = {"X-IBM-Migrated-Reacall": "wait"};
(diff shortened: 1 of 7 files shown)


tucker01:
Hey @MatthewB19, thanks for fixing these. Give me a day or so and I'll review! Thanks!

MatthewB19:
@tucker01 👍

tucker01:
Thanks @MatthewB19, busy week for me, so it'll be a moment before I can re-review 😄

Issue #531, opened 2019-09-27
How can I specify several variables in the command (zowe zos-workflows create  --VS?


George0013:
How can I specify several variables in the command (zowe zos-workflows create workflow-from-uss-file "testworkflow" "SYS2" --vs destination = CTA1 --vs firstvar=aaaaaa secondvar=bbbbb .....

uzust01:
Hello, @George0013
Try this:
zowe zos-workflows create workflow-from-uss-file --uss-file "a/user/ussFile" --system-name "SYS1" --owner "userID" --variables "var1=value1,var2=value2,var3=value3"

George0013:
Hello, @uzust01,
I'm afraid it doesn't work, it only works with one variable.

zowe zos-workflows cre wfuf "testworkflow" --uf "/SYS2/workflow/workflow_ConnectDirect.xml" --sn "SYS2" --ow "U529279" --ov --vs "destino=CTA1,indsn_TC0Z=u529279.temp.wlm,outdsn_CTA1=u529279.temp.wlmj"
Command Error:
Creating z/OSMF workflow with uss file: /SYS2/workflow/workflow_ConnectDirect.xml failed. More details:
Error: z/OSMF REST API Error:
Rest API failure with HTTP(S) status 400
messageText:            IZUWF5003E: The specified parameter name or parameter value "indsn_TC0Z:u529279.temp.wlm" is not valid.
validSingletonInstance: true
messageID:              IZUWF5003E

Anyway, I can always do it with the "variable in file" that works well.

tucker01:
Should we mark this with the bug label?

uzust01:
Thanks for pointing this out, @George0013 
@tucker01 , I will take a look

uzust01:
I did a little research and it seems like we are building the variables object in a way defined by IBM documentation. This looks like a Rest API validation error from z/OSMF itself. The only description of the error I found was this: 
[IZUWF5003E](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.2.0/com.ibm.zosmfworkflows.messages.help.doc/IZUWF5003E.html)

George0013:
Hi,
It seems to be correct, the variables are: destination, indsn_TC0Z and outdsn_CTA1j
They are the same that I have put with the option "variable in file".

---------------------------
[...]

PR #532, opened 2019-09-30
Update create wf example

uzust01:
Clarify --variables usage. Related to #531


packages/workflows/src/cli/create/Create.common.options.ts
```diff
@@ -97,7 +97,8 @@ export const CreateCommonOptions: { [key: string]: ICommandOptionDefinition } =
         name: "variables",
         aliases: ["vs"],
         description: "Includes a list of variables for the workflow. " +
-        "The variables that you specify here take precedence over the variables that are specified in the workflow variable input file.",
+        "The variables that you specify here take precedence over the variables that are specified in the workflow variable input file." +
```
tucker01: Looks like there is a missing space here: `file.Make`. Also, missing period at the end of the sentence.

PR #540, opened 2019-10-07
add --wait-for-output | --wfo  and --wait-for-active | --wfa option to zos-jobs submit commands 

ChrisBoehmCA:
* Try with `zowe zos-jobs submit data-set my.data.set --wait-for-output` and  `zowe zos-jobs submit data-set my.data.set --wait-for-active`
* if --vasc is not specified, the command will return the IJob object so that the return code can be easily checked without issuing `zowe zos-jobs view job-status-by-jobid <jobid>`

PR #546, opened 2019-10-16
--wait-for-active and --wait-for-output (incremental)

ChrisBoehmCA:
#540 for lts-incremental


tucker01:
I think you mentioned this, but this PR also ports submit from stdin right?

PR #550, opened 2019-10-23
Fix vulnerabilities

zFernand0:
Signed-off-by: zFernand0 <fernando.rijocedeno@broadcom.com>


zFernand0:
This PR should allow for #546 to be published.
So, the next version will be `2.32.0`

PR #560, opened 2019-10-29
fixed uncaught exception handling (lts-incremental)

### LINKED ISSUE #554: SSH Handler throws an uncaught exception when a connection is refused
```
Error: connect ECONNREFUSED <hostname>:22
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1106:14)
Command Error:
connect ECONNREFUSED <hostname>:22
Imperative encountered an unexpected exception
Command entered: "zowe uss issue ssh ls -la ..."
Use "zowe uss issue --help" to view groups, commands, and options.
Error Details:
connect ECONNREFUSED <hostname>:22
```

The connection is refused and an uncaught exception is generated (perhaps a promise rejection, etc. is not being handled?). 

@ChrisPark89 @chrisparkibm Any ideas here?

ChrisPark89:
resolve #554 
- added handling for uncaught exceptions.
- added some error handling messages.

### DIFF
--- a/packages/zosuss/__tests__/__system__/api/Shell.system.test.ts
+++ b/packages/zosuss/__tests__/__system__/api/Shell.system.test.ts
@@ -13,6 +13,7 @@ import { Shell, SshSession } from "../../../index";
 import { ITestEnvironment } from "../../../../../__tests__/__src__/environment/doc/response/ITestEnvironment";
 import { TestEnvironment } from "../../../../../__tests__/__src__/environment/TestEnvironment";
 import { ITestPropertiesSchema } from "../../../../../__tests__/__src__/properties/ITestPropertiesSchema";
+import { ZosUssMessages } from "../../../src/api/constants/ZosUss.messages";
 
 let TEST_ENVIRONMENT: ITestEnvironment;
 let SSH_SESSION: SshSession;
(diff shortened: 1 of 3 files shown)


zFernand0:
@ChrisPark89, could you add a quik test for this?

ChrisPark89:
Yes, I will. Sorry for the late reply. I wasn't getting the notification for some reasons :(

tucker01:
> Yes, I will. Sorry for the late reply. I wasn't getting the notification for some reasons :(

No problem at all. Take your time.

ChrisPark89:
I've added the tests

packages/zosuss/__tests__/__system__/api/Shell.system.test.ts
```diff
@@ -99,4 +100,44 @@ describe("zowe uss issue ssh api call test", () => {
...
+        });
+
+        try {
+            const command = "uname";
+            let stdoutData = "";
+            await Shell.executeSsh(invalidSshSession, command, (data: string) => {
+                stdoutData += data;
+            });
+        } catch (error) {
+            expect(error.toString()).toContain(ZosUssMessages.unexpected.message);
```
tucker01: Normally for this kind of test, we define an `err` var above the try/catch that we then assign in the catch block to the error thrown, then after the try/catch, we expect `err` to be defined (and any other fields/properties). This way, if no error is thrown we'll fail the test.

packages/zosuss/__tests__/__system__/api/Shell.system.test.ts
```diff
@@ -99,4 +100,44 @@ describe("zowe uss issue ssh api call test", () => {
...
+        });
+
+        try {
+            const command = "uname";
+            let stdoutData = "";
+            await Shell.executeSsh(invalidSshSession, command, (data: string) => {
+                stdoutData += data;
+            });
+        } catch (error) {
+            expect(error.toString()).toContain(ZosUssMessages.connectionRefused.message);
```
tucker01: Same comment here as above.

ChrisPark89: done!

packages/zosuss/__tests__/__system__/api/Shell.system.test.ts
ChrisPark89: I've made the change

tucker01: I'm sorry, I should have been more clear and linked an example, what you have is nearly there... the expects should be outside of the catch block:

https://github.com/zowe/zowe-cli/blob/de5afcebddb973d46da4041452202eae866430a4/packages/zosfiles/__tests__/api/methods/download/Download.unit.test.ts#L59

This way, if the test does not throw an error, it'll fail.

ChrisPark89:
> I'm Sorry 🤕, should have been more clear... checkout my comment. One small tweak.

My bad! I didn't realise it even by looking at other codes 😞 I fixed it

PR #562, opened 2019-10-29
Clarify workflow system in default props

awharn:
Update the value for the workflows system in default properties in an attempt to make it easier to populate.

### DIFF
--- a/__tests__/__resources__/properties/default_properties.yaml
+++ b/__tests__/__resources__/properties/default_properties.yaml
@@ -2,43 +2,43 @@
 #-------              ZOWE CLI EXAMPLE PROPERTIES FILE                --------#
 #  This file is used to parameterize runs of system tests with the relevant   #
 #  connection details for your z/OS system.                                   #
-#-----------------------------------------------------------------------------#
-#  find and replace the following:                                            #
-#                                                                             #
-#  my-user-name      - This is your TSO user ID                               #
-#  my-password       - This is your TSO password                              #
-#  my-jes-account    - This is your JES Account number for batch jobs         #
-#  my-uss-dir        - A directory on USS in which files and                  #
-#                      directories can be created                             #
-#  sysaff-system     - System Affinity for some batch jobs, use the system    #
-#                      that you are running the tests on                      #
-#  ds-iefbr14-pds    - PDS qualified by an IEFBR14 member JCL to submit       #
-#  ds-iefbr14-ps     - PS with an IEFBR14 JCL to submit                       #
-#  zosmf-host        - The host name for the LPAR with z/OSMF                 #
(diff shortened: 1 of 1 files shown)


__tests__/__resources__/properties/default_properties.yaml
```diff
@@ -97,7 +97,7 @@ provisioning:
 workflows:
   # The system where you want to run your workflow
   # (must exist and be active within your zOSMF)
-  system: system
+  system: workflow-system
```
zFernand0: I don't see a problem with this.
🤔 How about we set some sort of standard?
Each value will have a "full path" of its property
For example: `zosjobs-iefbr14Member` will be the value of `zosjobs.iefbr14Member`

zFernand0: That way, they will always be unique and traceable (hopefully making it easier for automation purposes

zFernand0:
@tucker01 @dkelosky, any thoughts?

__tests__/__resources__/properties/default_properties.yaml
tucker01: Sounds reasonable. Are you suggesting this change be made in this PR?

tucker01:
Also, make sure you document in the comment header

__tests__/__resources__/properties/default_properties.yaml
zFernand0: I guess 😝

__tests__/__resources__/properties/default_properties.yaml
```diff
@@ -2,43 +2,43 @@
...
+#----- please also update the ITestPropertiesSchema interface so that           -----#
+#----- developers can use auto-complete when using the properties in test code  -----#
 
 #-------------------------------------------------------------------------#
 #  Connection details/credentials for z/OSMF                              #
 #-------------------------------------------------------------------------#
 zosmf:
   # your z/OSMF User ID
-  user: &zosmfuser my-user-name
+  user: &zosmfuser zosmf-user
```
zFernand0: Thanks for updating.
Question, should the `&reference` be removed since they are not used in the yaml file?

tjohnsonBCM: I think they exist for convenience so that the same username and password can be used in the SSH credentials below.

zFernand0: So, should we use it (i.e. `*zosmfuser`) and remove the `ssh-user` (and password too)?

tucker01:
Where are we at with this PR?

zFernand0:
So, can we remove the `&zosmfuser` variable if it won't be used for the `ssh.user`?
Same goes for password.

Issue #563, opened 2019-10-29
When creating a data set from the CLI, the volser option isn't applied


CForrest97:
When you call the cmd `zowe files create data-set-partitioned DSNAME --volume-serial VOLSER`, the volser option isn't applied. This is because the function `generateZosmfOptions` expects the command argument to be called `volumeSerial`, whereas createOptions defines it as `volser`.
This will apply to the creation of all types of data sets.

There aren't any unit tests which confirm if the volser option is correctly passed on.

CForrest97:
From the commands below you can see vol `P2P210` exists however when I specify this volume in the create command it gets put into vol `P2P101`
```
craigs-mbp% zowe files ls ds CRAIGF.a.b5 -a  
- 
  dsname: CRAIGF.A.B5
  blksz:  6160
  catnm:  ICFCAT.SYSPLEX2.CATALOGB
  cdate:  2019/10/30
  dev:    3390
  dsorg:  PS
  edate:  ***None***
  extx:   1
  lrecl:  80
  migr:   NO
  mvol:   N
  ovf:    NO
  rdate:  ***None***
  recfm:  FB
  sizex:  15
  spacu:  CYLINDERS
  used:   0
  vol:    P2P210
  vols:   P2P210
craigs-mbp% zowe files create ps CRAIGF.a.b3 --volume-serial P2P210
Data set created successfully.
craigs-mbp% zowe files ls ds CRAIGF.a.b3 -a
-
  dsname: CRAIGF.A.B3
  blksz:  6160
  catnm:  ICFCAT.SYSPLEX2.CATALOGB
  cdate:  2019/10/30
  dev:    3390
  dsorg:  PS
  edate:  ***None***
  extx:   1
  lrecl:  80
  migr:   NO
  mvol:   N
  ovf:    NO
  rdate:  ***None***
  recfm:  FB
  sizex:  15
  spacu:  CYLINDERS
  used:   0
  vol:    P2P101
  vols:   P2P101
```

tucker01:
This might be due to SMS and how your volumes are managed? In house here, if you specify a certain volser, you'll get one in that SMS class (not necessarily the exact one).

PR #564, opened 2019-10-29
fix volser for create data sets

CForrest97:
Fix for issue #563


tucker01:
Hi @CForrest97, I'm a little confused here... I see that the command argument is defined as such:
```
    volser: {
        name: "volume-serial",
        aliases: ["vs"],
        description: strings.VOLSER,
        type: "string"
    },
```
This would be referenced in code as `arguments.volumeSerial` not as `arguments.volser`. Am I missing something?

CForrest97:
Hi @tucker01, I see what you mean now. I'll look into this some more as I'm still having problems creating data sets in specific volumes

PR #568, opened 2019-11-01
#557 migrating dataset api

### LINKED ISSUE #558: Create an API which migrates a data set
Scenario 1: API created for migrating data set
Given: A data set that isn't migrated
When: The user migrates data set
Then: They will be able to put the data set into a migration state

Pranay154:
Resolves #558

### DIFF
--- a/packages/zosfiles/__tests__/__system__/api/methods/hMigrate/HMigrate.system.test.ts
+++ b/packages/zosfiles/__tests__/__system__/api/methods/hMigrate/HMigrate.system.test.ts
@@ -9,13 +9,14 @@
 *
 */
 
-import { Create, Delete, CreateDataSetTypeEnum, HMigrate, ZosFilesMessages, Get } from "../../../../..";
+import { Create, Delete, CreateDataSetTypeEnum, HMigrate, ZosFilesMessages } from "../../../../..";
 import { Imperative, Session } from "@zowe/imperative";
 import { inspect } from "util";
 import { ITestEnvironment } from "../../../../../../../__tests__/__src__/environment/doc/response/ITestEnvironment";
 import { TestEnvironment } from "../../../../../../../__tests__/__src__/environment/TestEnvironment";
 import { ITestPropertiesSchema } from "../../../../../../../__tests__/__src__/properties/ITestPropertiesSchema";
-import { List, IZosFilesResponse, IListOptions } from "../../../../../src/api";
+import { List, IListOptions } from "../../../../../src/api";
+import { IMigrateOptions } from "../../../../../src/api/methods/hMigrate/doc/IMigrateOptions";
 
 let REAL_SESSION: Session;
 let testEnvironment: ITestEnvironment;
@@ -23,7 +24,7 @@ let defaultSystem: ITestPropertiesSchema;
 let dataSet1: string;
 let dataSet2: string;
 
-const options: IListOptions = { attributes: true };
+const listOptions: IListOptions = { attributes: true };
 
 describe("Migrate Dataset", () => {
     beforeAll(async () => {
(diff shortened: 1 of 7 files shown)


Colin-Stone:
If you could get the dataSet method in your HMigrate class to throw an error you should be able to achieve 100% coverage

packages/zosfiles/__tests__/__system__/api/methods/hMigrate/HMigrate.system.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    afterEach(async () => {
+        try {
+            await Delete.dataSet(REAL_SESSION, dataSet1);
+            await Delete.dataSet(REAL_SESSION, dataSet2);
+        } catch (err) {
+            Imperative.console.info(`Error: ${inspect(err)}`);
+        }
+    });
+    describe("Success Scenarios", () => {
+        it("Migrate sequential", async () => {
```
tucker01: `it` blocks normally contain a sentence starting with `should` (e.g. `it("should migrate a sequential dataset")`)

packages/zosfiles/src/api/methods/hMigrate/HMigrate.ts
```diff
@@ -0,0 +1,73 @@
...
+            if(!isNullOrUndefined(options.wait)) {
+                payload.wait = options.wait;
+            }
+
+            const headers: IHeaderContent[] = [
+              Headers.APPLICATION_JSON,
+              { "Content-Length": JSON.stringify(payload).length.toString() },
+            ];
+
+            await ZosmfRestClient.putExpectString(session, endpoint, headers, payload);
```
tucker01: Is there no response to the migrate? If there is, we should be returning this to the caller on the `apiResponse` field.

packages/zosfiles/src/api/methods/hMigrate/HMigrate.ts
```diff
@@ -0,0 +1,73 @@
...
+        ImperativeExpect.toNotBeEqual(dataSetName, "", ZosFilesMessages.missingDatasetName.message);
+
+        try {
+            const endpoint = posix.join(ZosFilesConstants.RESOURCE, ZosFilesConstants.RES_DS_FILES, dataSetName);
+
+            Logger.getAppLogger().debug(`Endpoint: ${endpoint}`);
+
+            const payload = { request: "hmigrate" } as any;
+
+            if(!isNullOrUndefined(options.wait)) {
```
tucker01: `isNullOrUndefined` is deprecated. Use `options.wait != null` for a not "is null or undefined".

packages/zosfiles/__tests__/__system__/api/methods/hMigrate/HMigrate.system.test.ts
Pranay154: Hey @tucker01, thanks for your comments, I have changed the `it` blocks so that they start with a `should`.

packages/zosfiles/src/api/methods/hMigrate/HMigrate.ts
Pranay154: Hey @tucker01, thanks for your comments, I have changed this

packages/zosfiles/src/api/methods/hMigrate/HMigrate.ts
Pranay154: Hey @tucker01, thanks for your comments, I have confirmed that there is no response to the migrate (it returns an empty object/string)

Pranay154:
Resolves #558

packages/zosfiles/src/api/methods/hMigrate/HMigrate.ts
tucker01: OK cool thank you.

PR #577, opened 2019-11-05
Add rename for data set api

### LINKED ISSUE #571: Add API functionality to rename a data set
#### User Story
**As a** CLI extender,  
**I want** to be able to call a rename API  
**so that** I can change the name of a data set.

#### Acceptance Criteria
**Scenario 1:** Rename sequential data set
*Given* A sequential data set with name `USER.BEFORE.SET`
*When* I call the rename API with `USER.BEFORE.SET` and `USER.AFTER.SET`
*Then* the data set will be renamed to `USER.AFTER.SET`
**Scenario 2:** Rename partitioned data set
*Given* A partitioned data set with name `USER.BEFORE.SET`
*When* I call the rename API with `USER.BEFORE.SET` and `USER.AFTER.SET`
*Then* the data set will be renamed to `USER.AFTER.SET`

See: https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.izua700/IZUHPINFO_API_PutDataSetMemberUtilities.htm

CForrest97:
Resolves #571

### DIFF
--- a/packages/zosfiles/src/api/methods/rename/Rename.ts
+++ b/packages/zosfiles/src/api/methods/rename/Rename.ts
@@ -1,0 +1,71 @@
+/*
+* This program and the accompanying materials are made available under the terms of the
+* Eclipse Public License v2.0 which accompanies this distribution, and is available at
+* https://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import { AbstractSession, ImperativeExpect, Logger, Headers } from "@zowe/imperative";
+import { posix } from "path";
+
+import { ZosmfRestClient, IHeaderContent } from "../../../../../rest";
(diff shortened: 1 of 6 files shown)


packages/zosfiles/src/api/methods/rename/Rename.ts
```diff
@@ -0,0 +1,71 @@
...
+            },
+        };
+
+        const reqHeaders: IHeaderContent[] = [
+            Headers.APPLICATION_JSON,
+            { [Headers.CONTENT_LENGTH]: JSON.stringify(payload).length.toString() },
+        ];
+
+        try {
+            await ZosmfRestClient.putExpectString(session, endpoint, reqHeaders, payload);
```
tucker01: Is there any payload response from the rename?

CForrest97: Hi @tucker01, no this API does not return a payload

PR #584, opened 2019-11-11
feat: add API of replace option for copy

liuxh0:
Add API functionality to specify replace option when copying dataset.

If implementation is okay, I can also add for enq option in the same way.

Related to #579

Issue #579, opened 2019-11-06
Add API functionality to specify all options when copying a data set or member


CForrest97:
### User Story
**As a** CLI extender,  
**I want** to be able to call an API function `copy` and add the options `volser, alias, enq, replace`
**so that** I can have full control over the copy API call

### Progress
- [ ] Volser
- [ ] Alias
- [ ] Enq
- [ ] Replace

Blocked by: #578 

See:
#521
https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.izua700/IZUHPINFO_API_PutDataSetMemberUtilities.htm

gejohnston:
This enhancement has had no activity for 12 months. The issue also has less than 5 up-votes by the community. No action on this enhancement is targeted for the next 2 calendar quarters. Therefore, this enhancement is being closed. If you feel that this enhancement should continue to be available for community up-votes, you may reopen this issue.

PR #608, opened 2019-11-25
feat: add API of enq option for copy

liuxh0:
Related to #579  and continuation of #584 .

### DIFF
--- a/packages/zosfiles/src/api/methods/copy/Copy.ts
+++ b/packages/zosfiles/src/api/methods/copy/Copy.ts
@@ -28,8 +28,8 @@ export class Copy {
      * Copy the contents of a dataset
      *
      * @param {AbstractSession}   session        - z/OSMF connection info
-     * @param {IDataSet}          fromDataSet    - The data set to copy from
      * @param {IDataSet}          toDataSet      - The data set to copy to
+     * @param {IDataSetOptions}   options        - Options
      *
      * @returns {Promise<IZosFilesResponse>} A response indicating the status of the copying
      *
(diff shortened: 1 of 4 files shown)


packages/zosfiles/src/api/methods/copy/Copy.ts
```diff
@@ -62,6 +62,10 @@ export class Copy {
             },
         };
 
+        if (options.enq !== undefined) {
+            payload.enq = options.enq;
+        }
```
zFernand0: Before we continue to add more, I'd suggest enhancing the ICopyDataSetOptions API and just use that object as the payload :)

liuxh0: Hi @zFernand0 , I have a new commit. Is that what you mean? I'm not sure if I get you correctly.

packages/zosfiles/src/api/methods/copy/Copy.ts
```diff
@@ -57,14 +56,12 @@ export class Copy {
...
+                dsn: options.fromDataSet.dataSetName,
+                member: options.fromDataSet.memberName,
             },
+            ...options
         };
-
-        if (options.replace !== undefined) {
-            payload.replace = options.replace;
-        }
+        delete payload.fromDataSet;
```
zFernand0: @liuxh0, this is kind of what I was looking for. Rather than deleting part of the object, maybe something more generic that can apply to all the [Utilities provided by ZOSMF](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.izua700/IZUHPINFO_API_PutDataSetMemberUtilities.htm) around this.
https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.izua700/IZUHPINFO_API_PutDataSetMemberUtilities.htm

For example, an interface that looks very much like the table on the link above : )
`IDataSet`
```
dsn: string,
member?: string,
volser?: string,
alias?: boolean
```

`IFileDescription`
```
filename: string,
type?: "binary|executable|text"
```

`IDataSetUtilities.ts`
```
request: string,
"from-dataset": IDataSet,
enq: string,
"from-file"?: IFileDescription,
replace?: boolean
```

zFernand0: Again, I believe this PR is good as it is for now :) 
We can definitely address this interface typing at a later time : )

liuxh0: @zFernand0 , thanks for your review. I'm happy to hear that this is what you want. 
I also think it is much better than deleting the object property. I was just not sure if I should change `IDataSet` interface since it also changes other parts. 
If it is okay, I would like to change it in another PR, to keep each PR's single responsibility.
And I also need your help to see if the system tests pass. Thanks!

zFernand0:
Bypassing audit
See https://github.com/zowe/imperative/issues/327

liuxh0:
@zFernand0 Many thanks! 👍🏼

PR #610, opened 2019-11-26
adding option to return Etag on Upload

Alexandru-Dumitru:
- update IUploadResult interface to optionally return the etag
- update IUploadOptions interface to request return of etag
- create IStringWithEtagResponse interface
    - needed to define return object from newly created REST action
- create new zosmfRestClient action
    - needed so that I don't introduce a breaking change
- update Upload method
    - `bufferToDataSet` can now return the etag inside `apiResponse`
    - `pathToDataSet` can now return the etag inside `apiResponse`
    for each member in the list
    - create new action `bufferToUssFileWithEtag`
        - it has an implicit `returnEtag = true`
        - couldn't update `bufferToUssFile` without introducing breaking change
    - `fileToUSSFile` can now return the etag inside `apiResponse`

Signed-off-by: Alexandru-Paul Dumitru <ad670553@broadcom.net>


zFernand0:
Bypassing known issue
https://github.com/zowe/imperative/issues/327

PR #614, opened 2019-11-28
Fixes #613 - broken links + typos in CLIReadme.md

### LINKED ISSUE #613: Incorrect links and spelling in CLIReadme.md
1. zos-files download/list data-set links to delete data-set command. This is because they are using the same # ref.
2. ssh-profile update has misspellings of "private" in the example

JDUNNIN:
Fixes #613

use unique names for data-set commands so link resolves correctly + fix typos in ssh-profile create example


JDUNNIN:
@zFernand0 Hi, I hadn't appreciated that, I'll take a look and update.

JDUNNIN:
@zFernand0 Hi, my latest commit changes how the anchor tags are generated to ensure uniqueness (on the assumption all zowe commands are unique) by creating an anchor that matches the command (with hyphens).

e.g. 
`zowe profiles create ssh-profile` becomes `#profiles-create-ssh-profile`
`zowe zos-jobs submit local-file` becomes `#zos-jobs-submit-local-file`

I then ran `npm run doc:generate` and `CLIReadme.md` was generated with the updated tags. It's also been updated significantly elsewhere, so I suspect this hasn't been run in a while. I initially made a mistake and had local plugins installed so they were added in commit 707524b... but removed in 2758daa.

Please review :)

gulp/DevelopmentTasks.ts
```diff
@@ -145,7 +145,14 @@ const doc: ITaskFunction = async () => {
             commandNameSummary += " (experimental)";
         }
 
-        const anchorTag = "module-" + definition.name;
+        // create anchor tag that matches full command
+        var anchorTag = "";
```
CForrest97: In js/ts, `let` and `const` are generally preferred to `var` as they have limited scope. Could you change it to `let` here instead

gulp/DevelopmentTasks.ts
```diff
@@ -158,7 +165,15 @@ const doc: ITaskFunction = async () => {
                 continue;
             }
             totalCommands++;
-            const childAnchorTag = "command-" + child.name.replace(/\s/g, "-");
+
+            // create anchor tag that matches full command
+            var childAnchorTag = ""
```
CForrest97: Could you use `let` here as well and a `;`

gulp/DevelopmentTasks.ts
```diff
@@ -145,7 +145,14 @@ const doc: ITaskFunction = async () => {
             commandNameSummary += " (experimental)";
         }
 
-        const anchorTag = "module-" + definition.name;
+        // create anchor tag that matches full command
+        var anchorTag = "";
+        if (oldCommandName !== "") {
+            anchorTag = oldCommandName + "-" + definition.name;
```
CForrest97: This is optional really but you could use template literals here
see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals
so it could be 
```
anchorTag = `${oldCommandName}-${definition.name}`;
```

gulp/DevelopmentTasks.ts
```diff
@@ -158,7 +165,15 @@ const doc: ITaskFunction = async () => {
...
             }
             totalCommands++;
-            const childAnchorTag = "command-" + child.name.replace(/\s/g, "-");
+
+            // create anchor tag that matches full command
+            var childAnchorTag = ""
+            if (oldCommandName !== "") {
+                childAnchorTag = oldCommandName + "-" + definition.name +"-" + child.name.replace(/\s/g, "-");
+            } else {
+                childAnchorTag = definition.name + "-" + child.name.replace(/\s/g, "-");
```
CForrest97: You could use template literals here too but not necessary

gulp/DevelopmentTasks.ts
JDUNNIN: Good to know. Done.

gulp/DevelopmentTasks.ts
JDUNNIN: Done.

gulp/DevelopmentTasks.ts
JDUNNIN: Changed to template literals.

gulp/DevelopmentTasks.ts
JDUNNIN: Changed to template literals.

zFernand0:
We need to include the `gulp/` scripts folder into the tslint config maybe?

The assumption that `commands are unique` is correct because there cannot be two major level groups with the same name (i.e. `plugins`, `config` ...). 

Implementation seems good to me and you are correct that this has not been updated in a little while.

@JDUNNIN, One comment, do you think this can be addressed...
`/Users/jackdunning/\.zowe/plugins/plugins\.json\`

JDUNNIN:
@zFernand0 I hadn't spotted my settings has been embedded into it. The doc build seems to pull in much more of my local environment than I would like, I'll work on removing it.

I don't know enough about tslint to comment really, but including running this script in a build makes sense to me.

JDUNNIN:
@zFernand0 I have made the changes directly to the generated `CLIReadme.md`. I could not find, even with global string search, where the sections which had my home directory in were being pulled from. 

May be worth raising another issue to find and update the source values to not use variables. Possibly another issue for running this doc build automatically as well.

Hopefully that's everything needed for the PR.

packages/imperative.ts
```diff
@@ -204,9 +204,9 @@ const config: IImperativeConfig = {
                     description: "Create a ssh profile called 'ssh222' to connect to z/OS SSH server at host 'zos123' and port 13022"
                 },
                 {
-                    options: "ssh333 --host sshhost --user ibmuser --priavetKey /path/to/privatekey --keyPassphrase privateKeyPassphrase",
+                    options: "ssh333 --host sshhost --user ibmuser --privateKey /path/to/privatekey --keyPassphrase privateKeyPassphrase",
                     description: "Create a ssh profile called 'ssh333' to connect to z/OS SSH server at host 'zos123' " +
-                                 "using a privatekey '/path/to/privatkey' and its decryption passphrase 'privateKeyPassphrase' " +
+                                 "using a privatekey '/path/to/privatekey' and its decryption passphrase 'privateKeyPassphrase' " +
```
zFernand0: Thanks for fixing :+1:

docs/CLIReadme.md
```diff
@@ -11,144 +11,154 @@ For Zowe CLI support, visit https://zowe.org
...
+	* [install](#plugins-install)
+	* [list](#plugins-list)
+	* [uninstall](#plugins-uninstall)
+	* [update](#plugins-update)
+	* [validate](#plugins-validate)
+* [profiles](#profiles)
+	* [create | cre](# profiles-create)
+		* [zosmf-profile](# profiles-create-zosmf-profile)
+		* [tso-profile](# profiles-create-tso-profile)
+		* [ssh-profile](# profiles-create-ssh-profile)
```
zFernand0: I believe the anchors won't work because of the space
image

gulp/DevelopmentTasks.ts
```diff
@@ -145,7 +145,14 @@ const doc: ITaskFunction = async () => {
             commandNameSummary += " (experimental)";
         }
 
-        const anchorTag = "module-" + definition.name;
+        // create anchor tag that matches full command
+        let anchorTag = "";
+        if (oldCommandName !== "") {
+            anchorTag = `${oldCommandName}-${definition.name}`;
```
zFernand0: I believe we need to trim the `oldCommandName` since a space is added below

JDUNNIN: Ah I missed this. Will trim the anchor tags as they are passed an an argument because the examples rely on the space to when inserting `zowe` before the command.

JDUNNIN: Resolved by 1d4ef12

PR #619, opened 2019-12-10
Refactor: IDataSet and ICopyDatasetOptions

liuxh0:
Hi @zFernand0 ,

As discussed in #608 , here is the PR in which I changed property names in IDataSet and ICopyDatasetOptions, and therefore simplified payload generation in `Copy.ts`.

I would add `volser` and `alias` options later when we start to implement them, but maybe not directly into `IDataSet`. Because `IDataSet` is also used by *rename*,  which don't have these two options. I would probably create a new interface which extends `IDataSet`. An alternative is to use an anonymous type that extends `IDataSet` in `ICopyDatasetOptions`, like:
```
"from-dataset": IDataSet & { volser?: string, alias?: boolean };
```

Does this sound good to you?


tucker01:
Hi @liuxh0, @zFernand0 and I discussed this and we should probably hold this PR until more breaking changes are necessary. We don't want to bump the major version for interface changes.

zFernand0:
If possible, could you change the target to `v7.0.0-next`.
This is a new branch where we will group all upcoming breaking changes.
Apologies for the inconveniences.

dkelosky:
hi @zFernand0 - what do you think about a generic `next` branch always ahead of master (`@latest`) and eventually cut over to it?  I'm asking because this is how we started with imperative (right or wrong) and it might be helpful to be consistent across these CLI repos.

liuxh0:
Hi @zFernand0 ,
No problem. Now I have changed the target to `v7.0.0-next`.

zFernand0:
@liuxh0, could you address the very few instances on cli handlers before merging?

liuxh0:
@zFernand0 , is this what you mean?

zFernand0:
Yeah, That's it.
One last thing... Lint errors.

If it helps, could you run `npm run build` and `npm run lint` locally when opening PRs?
Hopefully this should avoid these sort of things in the future : )

liuxh0:
@zFernand0 ,
I'm sorry, this is my mistake...

zFernand0:
No problem at all : )
Thanks for the contribution! 👍

zFernand0:
FYI, Going forward, We will follow Dan's suggestion about using the `next` branch for breaking changes that we want to introduce to the CLI and other projects as well. :)

PR #624, opened 2019-12-17
switch to zowe artifactory

jackjia-ibm:
Signed-off-by: Jack (T.) Jia <jack-tiefeng.jia@ibm.com>

Part of https://github.com/zowe/zowe-install-packaging/issues/875

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "web-help/logo.png"
   ],
   "publishConfig": {
-    "registry": "https://gizaartifactory.jfrog.io/gizaartifactory/api/npm/npm-local-release/"
+    "registry": "https://zowe.jfrog.io/zowe/api/npm/npm-local-release/"
   },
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
     "audit:public": "npm audit --registry https://registry.npmjs.org/"
   },
   "dependencies": {
-    "@brightside/imperative": "2.7.4",
+    "@brightside/imperative": "^2.7.4",
```
MarkAckert: Is there a reason the ^ was added to the version number here?

zFernand0: I believe it's ok to have it in case someone builds from source and wanted to receive features and fixes without having to reinstall imperative with the corresponding tag every time :) 
I could be wrong though

zFernand0: We can change it later if we should be fully locking things down :)

PR #625, opened 2019-12-17
switch to zowe artifactory

jackjia-ibm:
Signed-off-by: Jack (T.) Jia <jack-tiefeng.jia@ibm.com>

Part of https://github.com/zowe/zowe-install-packaging/issues/875


jackjia-ibm:
@zFernand0 I think we should also update package-lock.json because it's pointing to gizaartifactory, but seems like have to rebuild it. Do you think it's safe?

PR #633, opened 2020-01-10
Update Typescript and other package versions

tjohnsonBCM:
When verifying that the system tests still pass after making these changes, it was discovered there were a few that failed on Linux, which were fixed.


packages/zosfiles/__tests__/__system__/api/methods/download/Download.system.test.ts
```diff
@@ -116,6 +116,7 @@ describe("Download Data Set", () => {
                 // convert the data set name to use as a path/file
                 const regex = /\./gi;
                 file = dsname.replace(regex, "/") + ".txt";
+                file = file.toLowerCase();
```
zFernand0: nice catch 👍

packages/zosfiles/__tests__/__system__/api/methods/hMigrate/HMigrate.system.test.ts
```diff
@@ -79,7 +79,7 @@ describe("Migrate Dataset", () => {
             expect(listResponse.apiResponse.items[0].migr).toBe("YES");
             expect(migrateResponse.commandResponse).toContain(ZosFilesMessages.datasetMigratedSuccessfully.message);
         });
-        it("should migrate a partioned dataset", async () => {
+        it("should migrate a partitioned dataset", async () => {
```
zFernand0: Good thing there are no snapshots 😛

packages/zosfiles/__tests__/__system__/api/methods/upload/Upload.system.test.ts
```diff
@@ -114,6 +114,11 @@ describe("Upload Data Set", () => {
             });
 
             it("should upload a file to a physical sequential data set using Windows relative path", async () => {
+                if (process.platform !== "win32") {
+                    // TODO These tests cannot run on Linux, how should we handle this?
+                    return;
+                }
```
zFernand0: There may be a few tricks to work around this.
 See https://github.com/facebook/jest/issues/3652#issuecomment-521359778

packages/zosfiles/__tests__/__system__/cli/download/ds/cli.files.download.ds.system.test.ts
```diff
@@ -179,9 +179,9 @@ describe("Download Data Set", () => {
             expect(response.stderr.toString()).toContain("dataSetName");
         });
 
-        it("should fail due to specified data set name does not existed", async () => {
+        it("should fail due to specified data set name does not exist", async () => {
             const shellScript = path.join(__dirname, "__scripts__", "command", "command_download_data_set.sh");
-            const response = runCliScript(shellScript, TEST_ENVIRONMENT, [dsname + ".dummy"]);
+            const response = runCliScript(shellScript, TEST_ENVIRONMENT, ["dummy." + dsname]);
```
zFernand0: Why the order change?
Are we concerned of some other test creating the dataset and not properly cleaning afterwards?

packages/zosjobs/__tests__/__system__/api/CancelJobs.system.test.ts
```diff
@@ -42,8 +42,9 @@ describe("CancelJobs System tests", () => {
         });
 
         const ACCOUNT = systemProps.tso.account;
+        const maxStepNum = 255;  // Use lots of steps to make the job stay in INPUT status longer
 
-        iefbr14JCL = JobTestsUtils.getIefbr14JCL(REAL_SESSION.ISession.user, ACCOUNT);
+        iefbr14JCL = JobTestsUtils.getIefbr14JCL(REAL_SESSION.ISession.user, ACCOUNT, maxStepNum);
```
zFernand0: ❤️  Nice 🤣

packages/zosfiles/__tests__/__system__/cli/download/ds/cli.files.download.ds.system.test.ts
tjohnsonBCM: Suppose the dsname is "ABC.DEF.GHI". Earlier tests in the suite would have created a file "abc/def/ghi" when downloading the dataset. Then when this test runs, the original behavior was that it would create a folder "abc/def/ghi" to contain the "dummy" file. This was a problem, because it's trying to create a folder with the same name as an existing file. Windows seems to allow this, but on Linux `mkdirSync` fails in this case.

An alternative fix would be to pick a new random dsname at the start of this test to avoid the filename conflict. Would that be a better workaround?

packages/zosfiles/__tests__/__system__/api/methods/upload/Upload.system.test.ts
tjohnsonBCM: Ah that is clever :)

packages/zosfiles/__tests__/__system__/cli/download/ds/cli.files.download.ds.system.test.ts
awharn: I think the changed dsname would be better. It makes the tests consistent.

PR #658, opened 2020-02-03
Encode the account so # is supported.

### LINKED ISSUE #485: Encoding: CLI Error for zos-tso command group when account contains '#'
When trying to issue a zos-tso command with an account that contains the character `#`, the CLI command fails and an opaque error message is returned.  When issuing the command with `%23` instead of `#` it succeeds.

CLI command and output:
```
 zowe --version
5.9.1

zowe zos-tso issue command "NETSTAT CONN (PORT <SOME_PORT>" --user <user> --pass <password> --host <LPAR_HOST> --port <zosmf_port> --account ACCT#
Command Error:
Expect Error: [object Object]No servlet key was supplied.
```

z/OSMF log:
```
Ýtx00000000000135D5:<user>@<LPAR> (POST) /zosmf/tsoApp/tso?acct=ACCT¨   
2019-07-16T19:04:28.681Z!001C0F34!com.ibm.zoszmf.util.tso.session.TsoController!
handleError(uiResponse, requestLocale, userName, tsoASID, ex)                   
SEVERE: userName: <USER>; tsoASID:0; Exception:IZUG1102E: The TSO/E address space was not started because the following TSO/E parameters or values are missing or have incorrect syntax: proc, chset, cpage, rows, cols. Specify any missing parameters or values, and correct any syntax errors.                              
```

Working command (output omitted):
`zowe zos-tso issue command "NETSTAT CONN (PORT <SOME_PORT>" --user <user> --pass <password> --host <LPAR_HOST> --port <zosmf_port> --account ACCT%23`

awharn:
Fixes #485 


Alexandru-Dumitru:
@awharn Thanks for addressing this issue.
One question though, should we cover this by a test? 
Also, I can see that almost the whole EBCDIC is allowed to be used in the account number 😄 
I am thinking if we should gracefully fail if one of the not allowed characters are used (this list is shorter). This means an enhancement to check for them, of course, and not be part of this PR 😋

awharn:
Yes, I am in the process of writing a test to capture this behavior. However, I am finding it difficult to find a system that actually allows us to use ACCT# as the account number using the command referenced in the issue.

packages/zostso/__tests__/__system__/cli/issue/cli.zos-tso.issue.cmd.system.test.ts
```diff
@@ -104,5 +104,31 @@ describe("zos-tso issue command", () => {
...
+            }
+
+            const response = runCliScript(__dirname + "/__scripts__/as/issue_command_fully_qualified.sh",
+                TEST_ENVIRONMENT_NO_PROF,
+                [
+                    SYSTEM_PROPS.zosmf.host,
+                    SYSTEM_PROPS.zosmf.port,
+                    SYSTEM_PROPS.zosmf.user,
+                    SYSTEM_PROPS.zosmf.pass,
+                    "ACCT#"
```
Alexandru-Dumitru: In current form the test fails if used against a system which doesn't allow `ACCT#` as account number.
I was thinking it would be better to read it from SYSTEM_PROPS as well, and implement it in the test definition, something like this:
```
        (isAccountWithHash ? it : it.skip)("should successfully issue command = \"time\" with an account with an # in it", async () => {
```
but I was not able to make is work, because we instantiate the test environment in the `beforeAll()`, and Jest seems to be building all the tests even before this step.
I tried with `isAccountWithHash = systemProps.tso.account.includes("#");` in both `beforeAll()` statements, but doesn't help .... debugger shows that the tests are instantiated before these functions are executed, so `isAccountWithHash` is undefined at this time.
I have also tried to have `systemProps.tso.account.includes("#");` as the direct condition when deciding to use `it` or `it.skip`, but it errors out.

Bottom line is that it would be nice to have it as a conditional test, in case someone actually uses it in the custom_properties.yaml, but we need some creative solution for that.

zFernand0: > `(isAccountWithHash ? it : it.skip)("should successfully issue command = \"time\" with an account with an # in it", async () => {`

Really good thinking

Alexandru-Dumitru: Update on this.

It works fine if we determine if the account has `#` in it before the parent `describe()` definition. This way, when the tests are initialized, the variable is already set. I did it by reading the `custom_properties.yaml` file directly. Small caveat is that when we have only numbers, it gets cast into `number` type, so the `.includes("#")` function doesn't exit. Had to cast to string to make it work.
The reading of the file was right after the imports, but I don't see why this test can't come from `testUtils` (makes it more elegant).

Biggest showstopper is that we create the test environment dynamically, so this means that this variable might be wrong in case we are not taking the properties from the yaml file.

Still thinking how to make this more foolproof 😄 .

zFernand0:
@Alexandru-Dumitru, Are you Ok with an approach like what @awharn implemented. loading the properties file before all jest hooks to determine if we really need to run any given test?

packages/zostso/__tests__/__system__/cli/issue/cli.zos-tso.issue.cmd.system.test.ts
```diff
@@ -24,6 +30,9 @@ let acc: string;
 
 describe("zos-tso issue command", () => {
 
+    const jsonObject = yaml.safeLoad(fs.readFileSync(propfile, "utf8"));
+    acc = jsonObject.tso.account;
```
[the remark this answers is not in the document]
zFernand0: Since running tests conditionally based on the system properties may be a more frequent practice in the future. I say we need to build some of this logic into a `test_utilities` class/group to make tests more maintainable in the future (if this is the path we want to move forward with)

packages/zostso/__tests__/__system__/cli/issue/cli.zos-tso.issue.cmd.system.test.ts
zFernand0: Sorry for taking too long to respond here.
After looking at this with a fresh pair of eyes (👀 ), it's seem to be more trouble that the benefit we get out of it (i.e. skipping one test).
However, I don't think it's a bad practice. Maybe we can revisit this in the future to use/update the testing utilities.

zFernand0: Silly me. https://github.com/zowe/zowe-cli/pull/658#discussion_r383248657
Sorry for repeating myself.

PR #660, opened 2020-02-04
605 disallow trailing comma

Pranay154:
Resolves #605, changed `"trailing-comma": false` to `"trailing-comma": [true, {"multiline": "never", "singleline": "never"}]` in `tslint.json` file and made appropriate changes in relevant files (good luck reviewing)

PR #662, opened 2020-02-05
#555 recall migrated dataset

### LINKED ISSUE #555: Create an API which recalls the migrated data set
**Scenario 1:** API created for recalled migrated data set
**Given:** A data set in migration state
**When:** The user recalls a migrated data set
**Then:** They will be able to edit the data set that was migrated back

Pranay154:
Resolves #555. Manually tested migrating datasets and recalling them with the different options.


Pranay154:
I believe the code coverage is low due to changes for a trailing comma #660, where for some reason certain lines are touched by this PR were not covered already by any unit test. Unit tests for hRecall all pass and have 100% coverage.

Pranay154:
> Looking good Pranay but the one system test fails i.e.
> ✕ should recall a sequential data set with wait = true (30917ms)
> I don't believe we can use this as a reliable test because system to system we don't know how long these will actually take to respond. Have you got this test to work on the Hursley systems ?. I would remove or comment out this test for now

Hey @Colin-Stone , thanks for taking the time to review, the system test for me fails on the Hursley systems as it times out waiting for a response, however upon testing manually with the wait = true option, I can confirm that the recall function works, so I think it would be best to comment out the test for now.

packages/zosfiles/src/api/methods/hRecall/HRecall.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * This class holds helper functions that are used to recall files through the
+ * z/OSMF APIs.
+ */
+export class HRecall {
+    /**
+     *
+     * @param {AbstractSession}       session      z/OSMF connection info
+     * @param {string}                dataSetName  The name of the data set to recall
+     *
```
CForrest97: Could you add a param for the options

packages/zosfiles/src/api/methods/hRecall/HRecall.ts
```diff
@@ -0,0 +1,76 @@
...
+
+            Logger.getAppLogger().debug(`Endpoint: ${endpoint}`);
+
+            const payload = { request: "hrecall" } as any;
+
+            if(options.wait != null) {
+                payload.wait = options.wait;
+            }
+
+            if(options.nowait != null) {
```
CForrest97: not required as spoken about

packages/zosfiles/__tests__/__system__/api/methods/hRecall/HRecall.system.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    });
+
+    afterAll(async () => {
+        await TestEnvironment.cleanUp(testEnvironment);
+    });
+
+    afterEach(async () => {
+        try {
+            await Delete.dataSet(REAL_SESSION, dataSet1);
+            await Delete.dataSet(REAL_SESSION, dataSet2);
```
CForrest97: if the first one of these fails (which is likely if your tests havent been executed before) then the second will not run. Also these can be executed at the same time. Could you change it to the following style
```
await Promise.all([
            Delete.dataSet(REAL_SESSION, beforeDataSetName),
            Delete.dataSet(REAL_SESSION, afterDataSetName)
        ].map((p) => p.catch((err) => err)));
```

PR #680, opened 2020-03-10
Set exit code to 1 when upload command fails

tjohnsonBCM:
This resolves an issue with the command `zowe files upload ftds`, where if the upload fails, the command reports an error, but the exit code is still 0.

This PR is filed against the `next` branch because it contains a breaking change. It removes the `apiResponse` property from the JSON output of the upload API when it fails.

This change to make the upload handler throw is consistent with the behavior of other commands. I tested commands in the categories `files create`, `files delete`, `files download`, `jobs view`, and `tso stop`. They all throw when a REST API error is encountered and do not return an `apiResponse`.


tjohnsonBCM:
Closing in favor of an alternate solution that doesn't require breaking changes

PR #685, opened 2020-03-20
refactor hsmCommands

### LINKED ISSUE #684: Base/common handler function for Hmigrate and Hrecall
There are duplicated lines of code for the implementation of the Hmigrate api code and Recall api code, #558 and #555 respectively. Will create a base handler function in the `ZosFilesUtils.ts` file.

CForrest97:
Resolves #684. Work completed by @Pranay154


packages/zosfiles/__tests__/__system__/api/methods/hMigrate/HMigrate.system.test.ts
```diff
@@ -57,6 +60,27 @@ describe("Migrate Dataset", () => {
...
+            //         Imperative.console.info(`Error: ${inspect(err)}`);
+            //     }
+
+            //     expect(error).toBeFalsy();
+
+            //     expect(migrateResponse).toBeTruthy();
+            //     expect(migrateResponse.success).toBe(true);
+            //     expect(listResponse.apiResponse.items[0].migr).toBe("YES");
+            //     expect(migrateResponse.commandResponse).toContain(ZosFilesMessages.datasetMigratedSuccessfully.message);
+            // });
```
zFernand0, line 83: Should we removed commented out tests?
Or is this something that could be reintroduced?

packages/zosfiles/__tests__/__system__/api/methods/hMigrate/HMigrate.system.test.ts
```diff
@@ -115,7 +139,7 @@ describe("Migrate Dataset", () => {
     describe("Failure Scenarios", () => {
         beforeEach(async () => {
             try {
-                await Create.dataSet(REAL_SESSION, CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL, dataSet1);
+                await Create.dataSet(REAL_SESSION, CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL, dataSet3);
```
zFernand0, line 142: Thanks for using the `...FAIL...` dataset name.
In the future we can use this patter (relying on specific values in the test-data) to experiment with more failure scenarios. 👍

packages/zosfiles/__tests__/__system__/cli/hMigrate/ds/cli.files.HMigrate.ds.system.test.ts
```diff
@@ -44,6 +45,7 @@ describe("Migrate Dataset", () => {
     user = defaultSystem.zosmf.user.trim().toUpperCase();
     dataSetName1 = `${user}.SDATAC.SET`;
     dataSetName2 = `${user}.PDATAC.SET`;
+    dataSetName3 = `${user}.FAIL.SET`;
```
zFernand0, line 48: I wonder why we didn't make the names the same 😋 
JK, no need to change the LLQ to MIGR 😸

packages/zosfiles/src/api/utils/ZosFilesUtils.ts
```diff
@@ -177,5 +183,50 @@ export class ZosFilesUtils {
...
+     *  - If true then the function waits for completion of the request. If false (default) the request is queued.
+     * @param {IRecallOptions | IMigrateOptions} options
+     */
+    public static async dfsmsHsmCommand(
+        session: AbstractSession,
+        dataSetName: string,
+        returnMessage: string,
+        hsmCommand: any,
+        options: Partial<IRecallOptions> | Partial<IMigrateOptions> = {}
+    ): Promise<IZosFilesResponse> {
```
zFernand0, line 201: Should we create an interface for this function in case we need to expand the number of parameters in the future?

Pranay154: We could create an interface, I know that there is still hDelete hsm command that can be added in, in which case creating an interface would make sense

packages/zosfiles/__tests__/__system__/api/methods/hMigrate/HMigrate.system.test.ts
Pranay154: Hey @zFernand0 , thanks for taking the time to review this, these tests are commented out as they fail on the system due to timeout (similar to how it does for the cli system tests) hence I decided to comment it out, as in the future it could potentially be useful.

packages/zosfiles/__tests__/__system__/cli/hMigrate/ds/cli.files.HMigrate.ds.system.test.ts
Pranay154: I suppose the `.MIGR.` would make a lot more sense, same for recall, `.REC.`

Pranay154: @zFernand0 changed this in the wrong PR 😅 changed it in #695.

packages/zosfiles/src/api/utils/ZosFilesUtils.ts
zFernand0: Let's try to get an interface when hDelete comes along! (if we still think it's necessary)

PR #686, opened 2020-03-24
Add pipeline changelog logic

awharn:
Add logic to the pipeline to check for a Changelog update.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,8 +22,6 @@ def PRODUCT_NAME = "Zowe CLI"
 node('ca-jenkins-agent') {
     // Initialize the pipeline
     def pipeline = new NodeJSPipeline(this)
-    def changelog_updated
-    changelog_updated = false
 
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -183,12 +183,48 @@ node('ca-jenkins-agent') {
...
     // 30 minutes before an auto timeout approve.
     pipeline.deploy(
         versionArguments: [timeout: [time: 30, unit: 'MINUTES']]
     )
 
+    pipeline.createStage(
+        name: "Update Changelog Version",
+        stage: {
+            def firstTen = sh(returnStdout: true, script: "head -10 CHANGELOG.md")
+            if (firstTen.contains("Recent Changes")) {
```
tjohnsonBCM: Should this check be done earlier in the "Changelog Verification" stage? If there are any conditions that could result in an error, it would seem ideal to check for them before the deploy stage runs.

Issue #631, opened 2020-01-08
Downloading all members of the dataset changes their names in lowercase


zimlu02:
The following command to download all members of the specified dataset renames all files in lowercase:

zowe files dl am "mydsn.files.asm" -d myfiles -e asm

This definitely matters in case of assembler files that use references to macros in uppercase format only.

To download an HLASM file and all its macros we need to preserve their exact names.
Parameter to specify the case of downloaded files would be a great help.

MikeBauerCA:
I agree with preserving the exact names.

Ideally, I think this would be the default behavior. However, changing the default behavior would be a breaking change that might impact many scripts.

Perhaps a flag to change the behavior to preserve the uppercase format would be a good first step. 

I recognize the same behavior when downloading a single member so this flag should be added to the `zowe files download ds` command as well.

zFernand0:
I may not understand the full implications of this, however, I believe that if you were to use the CLI to upload them back, they will all be uploaded in uppercase. Someone can correct me if I'm wrong though.

I'm not saying we couldn't improve, I'm just trying to provide some piece of mind : )

zimlu02:
That's correct - from that perspective, it would rather make more sense to keep it uppercase all the time, wouldn't it? As Mike mentioned, keep it as a default. 
Anyway - for the use case I've mentioned initially, we need all locally downloaded files to stay uppercase. We're talking about hundreds of files referenced in the case-sensitive code.

liuxh0:
Hi all,
I'm trying to follow this issue. As I understand, we want an additional option for download to keep folders and files in original case (which is always uppercase), right?

zimlu02:
Correct, this issue is about being able to download folders and files in the **original case**.
(whether it is always uppercase or not is actually not relevant)

liuxh0:
Then I will try to create a PR asap :-)

PR #687, opened 2020-03-24
feat: add option to preserve original case

liuxh0:
See issue #631 

Add option preserve-original-letter-case for download, so that
automatically generated directories and files use the original letter
case (uppercase) when downloading data sets and members of a data set.


liuxh0:
Since I don't have access to test systems, I can't run system tests.

And I noticed `defaultValue: false` has no effect. I think this is a issue of [zowe/imperative](https://github.com/zowe/imperative).

liuxh0:
Update: defaultValue issue fix is merged in zowe/imperative#372

packages/zosfiles/__tests__/__system__/api/methods/download/Download.system.test.ts
```diff
@@ -281,6 +320,45 @@ describe("Download Data Set", () => {
...
+                }
+                expect(error).toBeFalsy();
+                expect(response).toBeTruthy();
+                expect(response.success).toBeTruthy();
+                expect(response.commandResponse).toContain(
+                    ZosFilesMessages.datasetDownloadedSuccessfully.message.substring(0, "Data set downloaded successfully".length + 1));
+
+                // convert the data set name to use as a path/file
+                const regex = /\./gi;
+                file = dsname.replace(regex, "/") + "/MEMBER.txt";
```
zFernand0, line 348: Since the only difference between the tests added seems to be this line, Should we try to combine the two?
I'm thinking purely about maintenance overhead of having duplication of tests in case we had to go back and refactor something.

packages/zosfiles/__tests__/__system__/api/methods/download/Download.system.test.ts
```diff
@@ -281,6 +320,45 @@ describe("Download Data Set", () => {
...
 
+            it("should download a data set and create folders and file in original letter case", async () => {
+                let error;
+                let response: IZosFilesResponse;
+
+                // TODO - convert to UPLOAD APIs when available
+                // upload data to the newly created data set
+                const data: string = "abcdefghijklmnopqrstuvwxyz";
+                const endpoint: string = ZosFilesConstants.RESOURCE + ZosFilesConstants.RES_DS_FILES + "/" + dsname + "(member)";
+                const rc = await ZosmfRestClient.putExpectString(REAL_SESSION, endpoint, [], data);
```
zFernand0, line 331: Also, we should try our best to implement some of the TODO items, or at least not add more technical debt.
If we feel it's Ok to use the rest client instead of the upload APIs, it's fine. we just need to remove the TODO in that case.

packages/zosfiles/__tests__/__system__/api/methods/download/Download.system.test.ts
```diff
@@ -123,6 +123,45 @@ describe("Download Data Set", () => {
                 expect(fileContents).toEqual(data);
             });
 
+            it("should download a data set and create folders and file in original letter case", async () => {
```
zFernand0, line 126: If we are still keeping two tests, at least we need to make the names different so that reports (or possible snapshots) don't present unexpected behavior.

packages/zosfiles/__tests__/__system__/api/methods/download/Download.system.test.ts
liuxh0: The one tests on Download.dataSet() while the other tests on Download.allMembers(). Since both parts have changes, I think it makes sense to separate into two tests.

packages/zosfiles/__tests__/__system__/api/methods/download/Download.system.test.ts
zFernand0: You are right, I should've expanded the file a bit more! 😋 
Thanks!

liuxh0:
@tjohnsonBCM No problem. I was also busy recently. Just updated the changelog.

PR #692, opened 2020-03-28
fix: false positives regarding allowableValues

liuxh0:
The values in allowableValues are regular expressions. Current settings
using plain string will match input containing the string, which is not
desired. This fix changes allowableValues to regex that exactly matches
the string.

See zowe/imperative#326


tjohnsonBCM:
Thanks for identifying and fixing this issue 😄 

Replacing "allowableValue" with "^allowableValue$" requires that users understand regexes when they read the command line help:
`Allowed values: ^rdonly$, ^rdwr$`

A potential fix would be to strip off the leading character `^` and trailing character `$` when printing the allowable values. Not sure if this is ideal to be modifying regexes before displaying them to the user.

Alternatively, the allowableValue string could be implemented as a wildcard instead of a regex so that adding `^` and `$` wouldn't be necessary. However, that would be a breaking change.

Going to consult with the team and get back to you with a consensus.

tjohnsonBCM:
> Replacing "allowableValue" with "^allowableValue$" requires that users understand regexes when they read the command line help:
> `Allowed values: ^rdonly$, ^rdwr$`
> 
> A potential fix would be to strip off the leading character `^` and trailing character `$` when printing the allowable values.

Could you please make this change, so that the allowed values contain the `^` and `$` characters, but are displayed without them in command line help?

liuxh0:
Hi @tjohnsonBCM 

I agree with you that such display is not very intuitive for users. 

Either way there are changes in imperative required, breakingly or unbreakingly (but not 100% correctly).

In my opinion it makes sense to **always** match the whole string. That is to say ^ and $ are added implicitly. Because I can't think of meaningful use cases where I need user input to only contain a part of what is required. 

For example:
- Regex `Public` will become implicitly regex `^Public$` which only matches string `Public`.
- Regex `.*\.txt` will become implicitly regex `^.*\.txt$` which will matches string ends with `.txt`.

tjohnsonBCM:
> In my opinion it makes sense to **always** match the whole string. That is to say ^ and $ are added implicitly.

Excellent idea, feel free to go ahead with that approach and please disregard my earlier comment.

liuxh0:
Cool. Then I will open a PR in imperative repo. 

Even though in most cases there's no need to change usage, it is still a breaking change for very rare cases. 

In the meantime can somebody have a look at my other PR #687? Thanks!

zFernand0:
Thanks for the discussion and collaboration.
Since we have updated the version of imperative (4.5.6), I believe the only thing we need to do is update the Imperative version. :smile:

Is it ok if we close this PR? (@liuxh0 and @tjohnsonBCM)

liuxh0:
@zFernand0 
Of course! Since it is fixed in imperative, there's nothing to do in this repo.
So I'll close this PR. Thanks for helping!

PR #706, opened 2020-04-22
Handle duplicate JCL step names when downloading spool output

### LINKED ISSUE #163: `zowe jobs submit ds xxx --directory` Doesn't Get All Output
To reproduce, use the same step / dd name twice in a job:
```
//*
//* Assemble TEMPLATE
//*
//         IF (RC = 0) THEN
//ASSEMBLE EXEC PGM=ASMA90
//ASMAOPT  DD  *
ADATA
RENT
MACHINE(ZSERIES-5)
LIST(133)
//SYSADATA DD  DISP=SHR,DSN=KELDA16.METALC.ADATA(TEMPLATE)
//SYSLIB   DD  DISP=SHR,DSN=KELDA16.METALC.ASMMAC
//         DD  DISP=SHR,DSN=SYS1.MACLIB
//         DD  DISP=SHR,DSN=SYS1.MODGEN
//         DD  DISP=SHR,DSN=ASMA.SASMMAC2
//         DD  DISP=SHR,DSN=CBC.SCCNSAM
//         DD  DISP=SHR,DSN=TCPIP.AEZAMAC1
//SYSPRINT DD  SYSOUT=*                                      
//SYSIN    DD  DISP=SHR,DSN=KELDA16.METALC.ASMPGM(TEMPLATE)
//SYSLIN   DD  DISP=SHR,DSN=KELDA16.METALC.OBJLIB(TEMPLATE)
//         ENDIF
//*
//* Assemble WTOUTILS
//*
//         IF (RC = 0) THEN
//ASSEMBLE EXEC PGM=ASMA90
//ASMAOPT  DD  *
ADATA
RENT
MACHINE(ZSERIES-5)
LIST(133)
//SYSADATA DD  DISP=SHR,DSN=KELDA16.METALC.ADATA(WTOUTILS)
//SYSLIB   DD  DISP=SHR,DSN=KELDA16.METALC.ASMMAC
//         DD  DISP=SHR,DSN=SYS1.MACLIB
//         DD  DISP=SHR,DSN=SYS1.MODGEN
//         DD  DISP=SHR,DSN=ASMA.SASMMAC2
//         DD  DISP=SHR,DSN=CBC.SCCNSAM
//         DD  DISP=SHR,DSN=TCPIP.AEZAMAC1
//SYSPRINT DD  SYSOUT=*                                      
//SYSIN    DD  DISP=SHR,DSN=KELDA16.METALC.ASMPGM(WTOUTILS)
//SYSLIN   DD  DISP=SHR,DSN=KELDA16.METALC.OBJLIB(WTOUTILS)
//         ENDIF
```

In the output, you only the last ASSEMBLE SYSPRINT.

Generally, I'd prefer that the STEP and DD name are not transformed, but perhaps in this case when they're the exact same and will overwrite each other - some unique identifier should be appended,

tjohnsonBCM:
Fixes #163 by ensuring that files downloaded from spool output have unique names.

This PR adds a number after DD names, only when needed to ensure they are unique.

For example, if a job has two steps both named `STEP1` which both contain DDs named `SYSPRINT` and `SYSUT2`, the following files will be downloaded. Previously the files output by the 2nd step would have overwritten the files from the 1st step.
image

Ideally we could distinguish between identically named steps and download their files to separate folders. However, I couldn't find a way to distinguish between steps with the same name (there is no concept of a step ID).


packages/zosjobs/src/api/DownloadJobs.ts
```diff
@@ -68,12 +68,23 @@ export class DownloadJobs {
...
 
         const jobFiles = await GetJobs.getSpoolFiles(session, parms.jobname, parms.jobid);
+        // Maintain a mapping of step names and associated DD names to detect duplicate step names
+        const usedStepNames: {[key: string]: string[]} = {};
         for (const file of jobFiles) {
+            let uniqueDDName = file.ddname;
+            if (usedStepNames[file.stepname] != null) {
+                let index = 1;
+                while (usedStepNames[file.stepname].indexOf(uniqueDDName) !== -1) {
+                    uniqueDDName = `${file.ddname}(${index})`;
```
tjohnsonBCM, line 78: Not sure what the best naming convention is to make filenames unique. I followed the Windows convention of appending `(1)`, `(2)`, etc., but perhaps a different convention like appending `_2`, `_3`, etc. would be better.

dkelosky: Perhaps pass it an as a part and set your default?  API users could adjust it when programmatically using it.  Someday if someone cared you could surface it as an option on the CLI?

tjohnsonBCM: Sounds good but I'm not clear how it would be passed in. Perhaps the suffix could be specified with a wildcard, like `(*)` or `_*`, where `*` would be replaced by a number?

dkelosky: oh yeah good point haha - maybe you should leave it as you have it and leave that stuff for another day

awharn: I would suggest against implementing with parenthesis just for the sake of problems that arise on *nix systems. Granted, they can be easily mitigated if you are certain you want to implement in this way.

```
[andrew@test-VM GitHub]$ touch "file.txt"
[andrew@test-VM GitHub]$ mv file.txt file(1).txt
bash: syntax error near unexpected token `('
[andrew@test-VM GitHub]$ mv file.txt "file(1).txt"
[andrew@test-VM GitHub]$ 
```

awharn: Nevermind, I just came to the realization that modern web browsers and file explorers on linux do take the approach of appending `(copy)` or `(1)` to duplicate files.

PR #707, opened 2020-04-23
Make remaining z/OSMF profile options optional

awharn:
Make the remaining required options `user`, `password`, and `host` optional when creating z/OSMF profiles. This change allows an end user to decide what profile information they store on disk.


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Make username, password, and host optional on z/OSMF profiles and update profile creation doc to reflect the change.
+- Create z/OSMF profile creation system tests.
```
tjohnsonBCM, line 8: Would suggest removing the second line here since it's not a user-facing change, and to be consistent with our practice that if a PR only changes tests, it doesn't warrant a version bump or changelog entry.

__tests__/__system__/__scripts__/imperative_zosmf_create_profile.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+profile=$1
```
tjohnsonBCM, line 2: This variable is unused, since `$*` is used below

__tests__/__system__/__scripts__/imperative_zosmf_delete_profile.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+profile=$1
```
tjohnsonBCM, line 2: This variable is unused, since `$*` is used below

gejohnston:
Just let me know if there is any other administrative action that you need me to take.

PR #717, opened 2020-05-12
Delete migrated data sets api

### LINKED ISSUE #715: Delete Migrated data sets API
Scenario 1: API created for deleting migrated data sets
Given: A data set that is migrated
When: The user wants to delete a migrated data set
Then: They will be able to delete the data set in a migration state

Pranay154:
Resolves #715 

Adds API to delete migrated data sets using the `hdelete` mainframe command. Other minor changes include updated variable naming (to make it more clear) and updates to comments (update doc links etc)


packages/zosfiles/src/api/methods/hDelete/doc/IDeleteOptions.ts
```diff
@@ -0,0 +1,24 @@
...
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import { IMigrateOptions } from "../../hMigrate";
+
+/**
+ * This interface defines the options that can be sent into the delete data set function.
+ */
+export interface IDeleteOptions extends IMigrateOptions {
```
Pranay154, line 17: Ive extended the previous interface `IMigrateOptions`, I feel there could be a better implementation, just not sure what so I am open to suggestions

packages/zosfiles/src/api/methods/hRecall/doc/IRecallOptions.ts
```diff
@@ -9,14 +9,12 @@
...
  */
-export interface IRecallOptions {
 
-  /**
-   * If true then the function waits for completion of the request.
-   * If false the request is queued.
-   */
-  wait?: boolean;
+// tslint:disable-next-line: no-empty-interface
+export interface IRecallOptions extends IMigrateOptions{
```
Pranay154, line 19: Refer to comment on extending `IMigrateOptions`, is there a better way to implement this?

packages/zosfiles/src/api/methods/hDelete/doc/IDeleteOptions.ts
zFernand0: I personally don't dislike reusing interfaces whenever possible. 😋 

how does `IDatasetUtils` sound? and the only required field is the `request` property? (and wait and purge are optional) 😋

packages/zosfiles/src/api/utils/ZosFilesUtils.ts
```diff
@@ -197,7 +196,7 @@ export class ZosFilesUtils {
         dataSetName: string,
         returnMessage: string,
         hsmCommand: any,
```
zFernand0, line 198: I believe `hsmCommand` will also be of type IZosFilesUtils?
Maybe I'm confused.

packages/zosfiles/src/api/methods/hDelete/doc/IDeleteOptions.ts
Pranay154: Hey @zFernand0, thanks for taking the time to review! Wasn't sure on if reusing interfaces was a good implementation, just needed to make sure. For the idea of `IZosFilesUtils`, just to make sure we're on the same page, it would just be a base "options" handler? so all the `IMigrateOptions`,`IRecallOptions`,`IDeleteOptions` "options" would all be inside this `IZosFilesUtils`(meaning API's would use this to append the options to the `payload`). As for the `request` property, could you explain which property you are referring to?

packages/zosfiles/src/api/utils/ZosFilesUtils.ts
Pranay154: @zFernand0 `hsmCommand` here refers to the request type, so for example the hDelete API would have the request type `{ request: "hdelete" }`  https://github.com/zowe/zowe-cli/blob/84280b7837ea76af65cb02f24cdef0c5728f428a/packages/zosfiles/src/api/methods/hDelete/HDelete.ts#L45

Whereas the hRecall API would send the request `{ request: "hrecall" }` to the base handler `dfsmsHsmCommand` https://github.com/zowe/zowe-cli/blob/84280b7837ea76af65cb02f24cdef0c5728f428a/packages/zosfiles/src/api/methods/hRecall/HRecall.ts#L45

packages/zosfiles/src/api/methods/hRecall/doc/IRecallOptions.ts
Pranay154: @zFernand0 The `//tslint:disable-next-line: no-empty-interface` is what made me think that there could be a better implementation as effectively this is just another file when all it does is use the same options as `IMigrateOptions`. The `IZosFilesUtils` you suggested would be good to look into, has something like a `IZosfilesUtils` been done before?

zFernand0:
Will re-review right after merging #728, as there may be some cool practices and maybe a conflict here.

zFernand0:
Thanks for addressing merge conflicts!
If nobody has any comments, I think this is good to be merged today! 😋

Pranay154:
Thanks @zFernand0, just looking through your comment for the `IZosFilesUtils`. I certainly like the idea of having a common interface for options would be good to have, I'll raise a separate issue for this. Regarding the `request` property, It would be interesting to have it pass it altogether to the payload. Its currently done seperatly :

https://github.com/zowe/zowe-cli/blob/de874fcec6c37b12968afbacfa9e906a3059f5e9/packages/zosfiles/src/api/methods/hDelete/HDelete.ts#L45

where the `request` is passed to the `ZosFileUtils` and then the options are appended to the `payload`:

https://github.com/zowe/zowe-cli/blob/de874fcec6c37b12968afbacfa9e906a3059f5e9/packages/zosfiles/src/api/utils/ZosFilesUtils.ts#L239

https://github.com/zowe/zowe-cli/blob/de874fcec6c37b12968afbacfa9e906a3059f5e9/packages/zosfiles/src/api/utils/ZosFilesUtils.ts#L242

So again I would be more than happy to create an issue to address these changes as I agree its quite a big change to the way the `payload` works. Thanks for reviewing and responding to my queries! 😄

zFernand0:
Since it may require a breaking change (or maybe not and I'm just worrying too much about it), we could add some sort of label `next` or `v7.x` to the issue? 😋

Pranay154:
> Since it may require a breaking change (or maybe not and I'm just worrying too much about it), we could add some sort of label `next` or `v7.x` to the issue? 😋

@zFernand0 Sounds good to me!

Pranay154:
Once this is merged in, ill create an issue for it  😄

PR #719, opened 2020-05-13
Initial token support

gejohnston:
- Converted all core Zowe commands to request that the Imperative infrasturcture prompt for missing credentials, and enforce a 30 second timeout on the response.
    - This was done by having the commands use the new Imperative CredsForSessCfg.addCredsOrPrompt() function.
- Added a placeholder for the 'zowe auth login apiml' command.
    - Once a real login command is implemented, the Zowe CLI commands will be able to authenticate with APIML tokens instead of only with user name and password.
- This PR cannot be merged into master (it will not even build) until the Imperative PR 399, on which it depends, is merged into master and published to public NPM.


packages/zosmf/src/api/ZosmfSession.ts
```diff
@@ -144,21 +193,33 @@ export class ZosmfSession {
...
             port: args.port,
-            user: args.user,
-            password: args.password,
-            base64EncodedAuth: args.auth,
             rejectUnauthorized: args.rejectUnauthorized,
             basePath: args.basePath
-        });
+        };
+
+        this.log.debug("Using basic authentication");
```
awharn, line 210: Would we want this in an if statement so we are not adding both the "Using basic authentication" and "Using token authentication" messages to the debug log?

packages/zosmf/src/api/ZosmfSession.ts
```diff
@@ -117,25 +151,40 @@ export class ZosmfSession {
...
     public static createBasicZosmfSession(profile: IProfile): Session {
         this.log.debug("Creating a z/OSMF session from the profile named %s", profile.name);
         return new Session({
-            type: "basic",
+            type: SessConstants.AUTH_TYPE_BASIC,
             hostname: profile.host,
             port: profile.port,
             user: profile.user,
             password: profile.password,
-            base64EncodedAuth: profile.auth,
```
awharn, line 138: Do we want to remove base64EncodedAuth yet?

packages/zosmf/src/api/ZosmfSession.ts
```diff
@@ -144,21 +193,33 @@ export class ZosmfSession {
...
         this.log.debug("Creating a z/OSMF session from arguments");
-        return new Session({
-            type: "basic",
+
+        const sessionConfig: ISession = {
             hostname: args.host,
             port: args.port,
-            user: args.user,
-            password: args.password,
-            base64EncodedAuth: args.auth,
```
awharn, line 158: See above comment.

packages/zosmf/src/api/ZosmfSession.ts
gejohnston: I added an if test.

packages/zosmf/src/api/ZosmfSession.ts
gejohnston: @dkelosky - I did not intentionally remove base64EncodedAuth. So I assume that it was removed in the POC. If we have not removed it as a possible profile option, I think that it should remain in the configuration of a session. The new imperative/..../CredsForSessCfg module was derived from this ZosmfSession logic, so base64EncodedAuth should also be added there. Unless you identify an objection, I will make both changes tomorrow.

dkelosky: That makes sense to me.  I think I cleaned up other things I did that were off in this PR.  Since you will make changes, I won't push anything else.

Maybe as a future exercise we can document authentication precedence: user & password, base64, ltpa, jwt, and client certs in the future

gejohnston: Upon further investigation, I found that even in master, there is no functionality to supply a base64 auth option in the profile or on the command line. That functionality must have been removed some time ago. Removing these references now would merely be further cleanup.
 
Note: the ISession property base64EncodedAuth remains so that we do not break any part of the Imperative API, even though it is unlikely that a anyone is using that property in a session.

PR #720, opened 2020-05-18
Add base profile

tjohnsonBCM:
Add `baseProfile` schema to Imperative config. This adds a `--base-profile` option to any existing command that uses service profiles.

Most of the changes in this PR are snapshot updates. This PR is temporarily set up to use the `add-base-profile` branch of Imperative so that it will build.


packages/Constants.ts
```diff
@@ -90,4 +92,63 @@ For ${Constants.DISPLAY_NAME} support, visit ${Constants.SUPPORT_LINK}
...
+     * Option used in profile creation and commands for rejectUnauthorized setting for connecting to z/OSMF
+     */
+    public static BASE_OPTION_REJECT_UNAUTHORIZED: ICommandOptionDefinition = {
+        name: "reject-unauthorized",
+        aliases: ["ru"],
+        description: "Reject self-signed certificates.",
+        type: "boolean",
+        defaultValue: true,
+        group: Constants.BASE_CONNECTION_OPTION_GROUP
+    };
```
gejohnston, line 153: Disclaimer: Not something to address now:

The non-SSL request of UPS makes me wonder if an option of --protocol [http | https] might be a candidate for base profile. Such an option exists on the JCLCheck plugin. It was quite helpful to be able to quickly switch the CLI when the server was reconfigured.

packages/imperative.ts
```diff
@@ -38,6 +38,76 @@ const config: IImperativeConfig = {
...
+                    optionDefinition: Constants.BASE_OPTION_REJECT_UNAUTHORIZED
+                },
+                tokenType: {
+                    type: "string"
+                },
+                tokenValue: {
+                    type: "string",
+                    secure: true
+                },
+                serviceIds: {
```
gejohnston, line 77: How we could or should use serviceIds still feels like a subject that our group needs to clarify.
Again, nothing to change at this time.

packages/Constants.ts
zFernand0: I agree that being able to influence `--protocol` will prove useful for some users!
Let's try to get it included
@tjohnsonBCM 👉 https://github.com/zowe/zowe-cli/issues/723
:yum: :wink:

PR #722, opened 2020-05-19
Update the Token Support to use the APIML endpoint

awharn:
Use the APIML endpoint
Modify the login function to allow for different request types and endpoints
Default the zowe auth login apiml command to use the APIML token
Add a 'show token' option to the zowe auth login apiml command

### DIFF
--- a/packages/auth/src/api/Login.ts
+++ b/packages/auth/src/api/Login.ts
@@ -9,9 +9,8 @@
 *
 */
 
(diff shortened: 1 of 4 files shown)


awharn:
> hey @awharn - what would you think about putting the `Login.ts` into a `packages/apiml`?
> 
> I'm wondering about the possibility of extending this group for more `apiml` api and cli possibilities and a possible future `zowe auth login` to something else in core...

We could do that, but I am wondering why we wouldn't want a common login function for the `zowe auth login` command. I suppose we could have specific function for anything that does token authentication if we want it.

Issue #632, opened 2020-01-09
Add Support for Code Page Setting Through Zowe CLI


dkelosky:
We've had several requests to support alternate code pages through Zowe CLI and should support code page for Zowe CLI file operations via [PH15263](https://www-01.ibm.com/support/docview.wss?rs=63&uid=isg1PH15263).

F-LEMANER:
Hello,
Is codepage setting available now ?
My config : z/OSMF 2.3 ZOWE 1.9
Thanks and regards

dkelosky:
Hi @F-LEMANER  - not yet.  Are you also using Zowe CLI on your workstation or only using the Zowe "Desktop" (browser-based Zowe UI)?

F-LEMANER:
Hello, I'm not using CLI because I'm not a developper. As zOS system engineer I have installed ZOWE and just use JES and MVS explorers to validate the installation. But our developpers want to use CLI.

Do you have an planned date for support of codepage ?
Regards

Marty850:
I'm also very interested in this feature for CLI! ZOWE file upload is useless for us because of that issue and it would be such a great possibility to do automatic deployment.

I'm checking the satus of this issue every week ;-)

dkelosky:
Sorry for the delay - we will be working on this in the coming weeks.

FALLAI-Denis:
Hi
Please, see comment in [#260](https://github.com/zowe/vscode-extension-for-zowe/issues/260#issuecomment-634991101).

dkelosky:
@venkatzhub do you have any insight to add here?

dkelosky:
@Marty850 the code is completed and should be available on the next published release - `6.14.0`

PR #728, opened 2020-05-28
API/CLI Data Set Code Page Support

dkelosky:
This PR adds code page support to data set & USS file `Download`, `Upload`, and `Get` API methods and download / upload commands to address #632 :

```
$ zowe files download ds "ibmuser.work.jcl(asmtest1)" --encoding=285
Data set downloaded successfully.
Destination: ibmuser/work/jcl/asmtest1.txt

```
And:
```
import { ISession, Session, Logger, LoggingConfigurer } from "@zowe/imperative";
import { Get, Download } from "./zosfiles";

Logger.initLogger(LoggingConfigurer.configureLogger('lib', {name: 'test'}));

const conn: ISession = {
    hostname: "somehost.net",
    port: 443,
    rejectUnauthorized: false,
    user: "ibmuser",
    password: "password",
    type: "basic",
};
const session = new Session(conn);

(async () => {
    const resp = (await Get.dataSet(session, "IBMUSER.WORK.JCL(ASMTEST1)", {
        encoding: 285
    })).toString();
    console.log(resp);

    await Download.dataSet(session, "IBMUSER.WORK.JCL(ASMTEST1)", {encoding: 285});

    await Download.allMembers(session, "IBMUSER.PUBLIC.REGSERV.LNKINC", {encoding: 285});
    
})();
```


packages/zosfiles/src/api/methods/get/Get.ts
```diff
@@ -46,6 +46,14 @@ export class Get {
             reqHeaders = [ZosmfHeaders.X_IBM_BINARY];
         }
 
+        if (options.encoding) {
```
awharn, line 49: Is this a different behavior when compared to Upload.ts? Imagine binary and encoding are both specified. In Upload.ts, the text;fileEncoding header is never added because it is in an else block. Here, the encoding overrides the binary header.

I would suggest making the if an else if to make it consistent.

packages/zosfiles/src/api/methods/download/Download.ts
```diff
@@ -84,8 +84,15 @@ export class Download {
                 reqHeaders = [ZosmfHeaders.X_IBM_BINARY];
             }
 
-            // Get contents of the data set
+            if (options.encoding) {
```
awharn, line 87: Same comment as Get.ts

dkelosky: good point - i did this one first and forget when i got to Upload

packages/zosfiles/src/api/methods/download/Download.ts
```diff
@@ -80,12 +80,23 @@ export class Download {
...
             let reqHeaders: IHeaderContent[] = [];
+
             if (options.binary) {
                 reqHeaders = [ZosmfHeaders.X_IBM_BINARY];
+
+            } else if (options.encoding) {
+
+                const keys: string[] = Object.keys(ZosmfHeaders.X_IBM_TEXT);
+                const value = ZosmfHeaders.X_IBM_TEXT[keys[0]] + ZosmfHeaders.X_IBM_TEXT_ENCODING + options.encoding;
+                const header = ZosmfHeaders.X_IBM_TEXT;
```
dkelosky, line 91: This causes some tests to fail... updating in an upcoming commit

PR #732, opened 2020-06-02
Add Encoding Setting to z/OSMF Profile

dkelosky:
Encoding will be an option when creating the profile:
image

### DIFF
--- a/packages/imperative.ts
+++ b/packages/imperative.ts
@@ -71,6 +71,16 @@ const config: IImperativeConfig = {
                     basePath: {
                         type: "string",
                         optionDefinition: ZosmfSession.ZOSMF_OPTION_BASE_PATH
+                    },
+                    encoding: {
+                        type: "number",
+                        optionDefinition: {
+                            name: "encoding",
+                            aliases: ["ec"],
+                            description: "The encoding for download and upload of z/OS data set and USS files." +
+                                " The default encoding if not specified is 1047.",
+                            type: "number"
+                        }
                     }
                 },
                 required: []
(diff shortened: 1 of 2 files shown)


packages/imperative.ts
```diff
@@ -71,6 +71,16 @@ const config: IImperativeConfig = {
                     basePath: {
                         type: "string",
                         optionDefinition: ZosmfSession.ZOSMF_OPTION_BASE_PATH
+                    },
+                    encoding: {
+                        type: "number",
```
gejohnston, line 76: Just a question: Will the encoding always be a number? Will we never never need something like IBM1234?

dkelosky: I worried about this as well.  Although they document that the default is `IBM-1047`, when you provide a string value on the header you received errors:
```
{
    "category": 16,
    "rc": 121,
    "reason": -1037303780,
    "message": "iconv_open() failed.",
    "stack": "  1 3BB1C0D0+0000036C Translator::initialize()\n  2 3BB19738+000002E8 ServletOutputStream::write(const char*,unsigned int)\n  3 3BB0DFC8+0000057A DatasetReader::copyTextTo(ServletOutputStream&,unsigned int)\n  4 3BB06660+00001664 GetDatasetServlet::doGet()\n  5 3BB22A88+000000A0 TsoServlet::service()\n  6 3BB53BC0+00000190 TsoServlet::run()\n  7 3BB53520+00000396 main\n  8 0DA958D8+0000127C CEEVROND\n  9 3B2EF8E8+000000B4 EDCZHINV\n 10 0D931F68+000001C6 CEEBBEXT\n",
    "details": [
        "EDC5121I Invalid argument. (errno2=0xC22C001C)"
    ]
}
```

From what I read, `code page` comes from the literal page numbers in a manual and other API methods used numbers.  I hope I'm not wrong on this since the zowe CLI command type is also a number 😟

gejohnston: Thanks for the info.

PR #737, opened 2020-06-12
Base profiles and APIML authentication

tjohnsonBCM:
Add base profiles, and auth command group for connecting to APIML.

Also use Imperative's token handling capabilities and prompting for missing connection options.

Each round of changes were previously reviewed as PRs from other feature branches into this tokens branch.


TODO.txt
```diff
@@ -0,0 +1,5 @@
+right now zosmf is an optional profile, everything using apiml would need be added as an optional profile
+session objects would need to use apiml token if it is there
+this is a major hack for now
+insert basepath automatically
+depends on same profile named entities
```
awharn, line 5: Do we want to keep the TODO file?

__tests__/__resources__/properties/default_properties.yaml
```diff
@@ -49,6 +49,20 @@ zosmf:
...
+  # your Base User ID
+  user: base-user
+  # your Base password
+  pass: base-password
+  # host for your base / API ML instance
+  host: base-host
+  # port for your base / API ML instance
+  port: base-port
+  # false to allow self signed certificates when connecting
+  rejectUnauthorized: false
```
zFernand0, line 64: Contributors of the CLI will be required to add these properties to their custom_properties.yaml file if they run integration and/or systems tests if they want to try base profile stuff in their tests.

TODO.txt
gejohnston: This was left over the the proof of concept. I removed it.

__tests__/__resources__/properties/default_properties.yaml
awharn: That is correct. I don't think you were looking for a comment here, but you are getting one. 😄

gejohnston:
Are we waiting for anything else, or should this be merged to master now?

zFernand0:
We should be good to merge! :yum:

PR #759, opened 2020-07-07
Add fail fast option to downloads

awharn:
Adds the --fail-fast option to `zowe zos-files download all-members` to keep downloading members even if some fail.

Failed member names are output in the error message and in additionalDetails when --rfj is specified. Errors are preserved, mapped, and output in the error in case members fail to download for different reasons.


packages/zosfiles/src/api/methods/download/Download.ts
```diff
@@ -238,6 +249,16 @@ export class Download {
                 await asyncPool(maxConcurrentRequests, memberList, createDownloadPromise);
             }
 
+            // Handle failed downloads if no errors were thrown yet
+            if (downloadErrors.length > 0) {
+                throw new ImperativeError({
+                    msg: "Failed to download the following members: \n" + failedMembers.join("\n") + "\n\n" +
```
tjohnsonBCM, line 255: Should the message "Failed to download the following members:" be defined as a constant in `api/constants/ZosFiles.messages.ts`? Just a question, it doesn't need to be

awharn: We can do that. Would make it easier to keep the test updated if we decide to change it.

Issue #754, opened 2020-06-26
Zowe SDK - Implement allocation of dataset 'like' feature


zFernand0:
Enhance `zosmf/restfiles` API to leverage the new `like` field provided in the request body.
Doc: https://www.ibm.com/support/knowledgecenter/SSLTBW_2.4.0/com.ibm.zos.v2r4.izua700/IZUHPINFO_API_CreateDataSet.htm
APAR: [PH22030](https://www.ibm.com/support/pages/apar/PH22030)

PR #766, opened 2020-07-20
Implementation of new "like" parameter for data set creation

katelynienaber:
Solves linked issue: [#754](https://github.com/zowe/zowe-cli/issues/754)

The changes in this PR will allow calling of zowe.Create.dataSet with the new "like" REST parameter implemented in [zOSMF v2.4](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.4.0/com.ibm.zos.v2r4.izua700/IZUHPINFO_API_CreateDataSet.htm).

I am not absolutely certain this doesn't belong in Imperative. Please let me know if it should be moved.


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## `6.18.1`
```
zFernand0, line 5: Please leave this as `Recent changes`
That will avoid having to merge PRs in a certain order.
Also, this may be an enhancement since it's adding new functionality to already existing APIs

packages/zosfiles/src/api/methods/create/Create.ts
```diff
@@ -70,6 +74,10 @@ export class Create {
             case CreateDataSetTypeEnum.DATA_SET_CLASSIC:
                 tempOptions = {...CreateDefaults.DATA_SET.CLASSIC, ...tempOptions};
                 break;
+            // tslint:disable-next-line:no-magic-numbers
```
zFernand0, line 77: Please do not use magic-numbers.
If this needs to be a constant, then add it to the corresponding constants file and document it.

CHANGELOG.md
katelynienaber: Fixed in recent push :)

packages/zosfiles/src/api/methods/create/Create.ts
katelynienaber: Fixed in recent push :)

katelynienaber:
> Left a few comments.
> Also, please don't source test files:
> ******/zosfile/download/a1595240/a775092.txt

Fixed! Also changed ID. Will check to see if I left it anywhere...

packages/zosfiles/__tests__/__system__/api/methods/invoke/DefineVSAM.ams.temp
```diff
@@ -0,0 +1 @@
+DEFINE CLUSTER (NAME () CYLINDERS (5 2 ) VOLUMES())
```
tjohnsonBCM, line 1: Please remove file left behind from running system tests

packages/zosfiles/__tests__/__system__/api/methods/upload/testfiles/upload.txt
```diff
@@ -1 +1 @@
-abcdefghijklmnopqrstuvwxyz
\ No newline at end of file
+abcdefghijklmnopqrstuvwxyz
```
tjohnsonBCM, line 1: Please remove file left behind from running system tests

packages/zosfiles/__tests__/api/methods/create/Create.unit.test.ts
```diff
@@ -139,6 +140,24 @@ describe("Create data set", () => {
...
+                alcunit: "CYL",
+                primary: 20,
+                recfm: "FB",
+                blksize: 6160,
+                lrecl: 80,
+                showAttributes: true
+            };
+            const custOptions2 = { like: dataSetName, showAttributes: true };
+
+            await Create.dataSet(dummySession, CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL, dataSetName, custOptions1);
```
tjohnsonBCM, line 155: The first `Create.dataset` call with `custOptions1` is unnecessary here. Since this is a unit test rather than a system test, the second `Create.dataset` call with `custOptions2` will succeed even if the dataset `dataSetName` hasn't been created yet.

packages/zosfiles/__tests__/api/methods/create/Create.unit.test.ts
```diff
@@ -139,6 +140,24 @@ describe("Create data set", () => {
             );
         });
 
+        it("should be able to allocate like from a sequential data set", async () => {
```
tjohnsonBCM, line 143: Please add an API system test in `packages/zosfiles/__tests__/__system__/api/methods/create` modeled after this test.

katelynienaber:
> Thanks for adding this parameter! Left a few comments related to tests.
> 
> Also I'm having a problem when calling the create dataset API like this:
> `await Create.dataSet(session, CreateDataSetTypeEnum.DATA_SET_LIKE, "DSNAME.NEW", { like: "DSNAME.OLD" });`
> 
> An error 500 is returned:
> 
> ```
> RestClientError: z/OSMF REST API Error:
> Rest API failure with HTTP(S) status 500
> category: 1
> rc:       4
> reason:   13
> message:  primary
> ```
> 
> I think the problem may be with the payload being sent:
> 
> ```
> Resource:  /zosmf/restfiles/ds/DSNAME.NEW
> Request:   POST
> Headers:   [{"X-CSRF-ZOSMF-HEADER":true}]
> Payload:   '{"like":"DSNAME.OLD","secondary":1}'
> ```
> 
> Seems like the `secondary` property is getting set in the payload but shouldn't be. Since `secondary` is set, z/OSMF expects that `primary` should also be set.

I have gotten to the easier comments, now I will take a look at this one...

katelynienaber:
Okay I think I have also gotten to the accidental setting of the secondary parameter...

packages/zosfiles/src/api/methods/create/doc/ICreateDataSetOptions.ts
```diff
@@ -123,4 +123,9 @@ export interface ICreateDataSetOptions {
      * @type {string}
      */
     size?: string;
+    /**
+     * The name of another data set from which to copy attributes
+     * @type {string}
+     */
+    like?: string;
```
tjohnsonBCM, line 130: When "like" is used on `ICreateDataSetOptions`, no other properties are needed. Should we make every property on the interface optional? Currently "primary" and "lrecl" are required.

This may be a question for @zFernand0 🙂 If we change required values to be optional, we may need to add validation logic where those values were being used before.

zFernand0: Sounds to me that we may need a separate `object` i.e. `zowe files create data-set-like` since we usually aim to avoid confusion with what options are conditionally required and/or mutually exclusive.

I do agree that if we decide to make them all optional, additional validation (at the API level) will be required in order to guarantee the same behavior.

katelynienaber: I like the first idea :) But @tjohnsonBCM I will wait for your input before implementing anything...

tjohnsonBCM: @katelynienaber I agree with the suggestion to define a new interface (perhaps `ICreateDataSetLikeOptions`?), to minimize confusion about which options on the interface are required.

If the "allocate like" functionality is implemented as a CLI command in the future, it would make sense for it to be a new command (`zowe files create data-set-like` as Fernando mentioned). This is consistent with creating a new interface.

katelynienaber: @tjohnsonBCM @zFernand0 I added a new interface and a separate function in the Create class, please check if this meets your specifications ^^

tjohnsonBCM: Should "like" be removed from `ICreateDataSetOptions` now?

packages/zosfiles/src/api/methods/create/doc/ICreateDataSetLikeOptions.ts
```diff
@@ -0,0 +1,32 @@
...
+     */
+    like: string;
+    /**
+     * The indicator that we need to show the attributes
+     *
+     * DO NOT SEND THIS TO ZOSMF
+     *
+     * @type {boolean}
+     */
+    showAttributes?: boolean;
```
tjohnsonBCM, line 31: The "showAttributes" property doesn't do much when "like" is used, since it only returns a single line in the command response:
`like: DATASET.OLD`

Should we call the z/OSMF API to list all the attributes of the new data set created, so they can be included in the command response?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Added new command "dataSetLike(session, dataSetName, options)" to `Create` class, for use when creating a dataset with parameters like another data set
```
tjohnsonBCM, line 7: Please change "new command" to "new method", to avoid confusion with CLI commands.

packages/zosfiles/__tests__/api/methods/create/Create.unit.test.ts
tjohnsonBCM: Please add an API system test. I think there was one previously, but it seems to have gone missing.

CHANGELOG.md
katelynienaber: Done ^^

packages/zosfiles/__tests__/api/methods/create/Create.unit.test.ts
katelynienaber: Added

packages/zosfiles/src/api/methods/create/doc/ICreateDataSetOptions.ts
katelynienaber: Removed!

packages/zosfiles/src/api/methods/create/doc/ICreateDataSetLikeOptions.ts
katelynienaber: Yes, I have added this. It is not so much work and it would be nice to have the attributes back :)

packages/zosfiles/src/api/methods/create/Create.ts
```diff
@@ -134,6 +136,34 @@ export class Create {
...
+        ImperativeExpect.toNotBeNullOrUndefined(dataSetName, ZosFilesMessages.missingDatasetName.message);
+
+        try {
+            const endpoint: string = ZosFilesConstants.RESOURCE + ZosFilesConstants.RES_DS_FILES + "/" + dataSetName;
+
+            const data = await ZosmfRestClient.postExpectString(session, endpoint, [], JSON.stringify(tempOptions));
+
+            let response = "";
+            if (!isNullOrUndefined(tempOptions.showAttributes)) {
+                delete tempOptions.showAttributes;
```
tjohnsonBCM, line 152: Deleting `showAttributes` property from `tempOptions` object should be done sooner, before calling `ZosmfRestClient`.

packages/zosfiles/src/api/methods/create/Create.ts
```diff
@@ -134,6 +136,34 @@ export class Create {
...
+                if (tempOptions.showAttributes) {
```
tjohnsonBCM, line 153: This condition should be checking `options` rather than `tempOptions`. `tempOptions.showAttributes` property is no longer defined after being deleted above.

packages/zosfiles/src/api/methods/create/Create.ts
```diff
@@ -134,6 +136,34 @@ export class Create {
...
+                    response = TextUtils.prettyJson(List.dataSet(session, dataSetName, { attributes: true }));
```
tjohnsonBCM, line 154: `List.dataSet` returns a Promise so must be awaited. Please also limit how much of the response is printed to `List.dataSet(...).apiResponse.items[0]`, so that it only displays the dataset attributes rather than an entire API response.

packages/zosfiles/__tests__/__system__/api/methods/create/Create.system.test.ts
```diff
@@ -106,6 +106,76 @@ describe("Create data set", () => {
...
+            error = err;
+            Imperative.console.info("Error: " + inspect(error));
+        }
+
+        expect(error).toBeFalsy();
+        expect(response).toBeTruthy();
+
+        expect(response.success).toBe(true);
+        expect(response.commandResponse).toContain(ZosFilesMessages.dataSetCreatedSuccessfully.message);
+        expect(response.commandResponse).toContain(options);
```
tjohnsonBCM, line 175: This `expect` fails, because `TextUtils.prettyJson` generates YAML rather than JSON output, and the attributes returned by `List.dataSet` aren't exactly identical to the options provided to `Create.dataSet`.

packages/zosfiles/src/api/methods/create/Create.ts
katelynienaber: Hello again @tjohnsonBCM :)

I am not able to return the same dataset attributes that are returned from function `Create.dataSet()`.

The dataset attributes returned by the API are [these attributes](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.2.0/com.ibm.zos.v2r2.izua700/IZUHPINFO_API_RESTFILES_JSON_Documents.htm).

But the attributes returned by the `Create.dataSet()` function are [these attributes](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.4.0/com.ibm.zos.v2r4.izua700/IZUHPINFO_API_CreateDataSet.htm).

`Create.dataSet()` just returns the attributes it has been passed as YAML, if `{ attributes = true }`. Since `Create.dataSetLike()` doesn't receive any attributes (it is just receiving the name of the dataset to copy attributes from) I am not sure how to retrieve the attributes available at the second link.

I'd like to do this so that the return value from `Create.dataSetLike()` is exactly the same as that from `Create.dataSet()`.

For now, `Create.dataSetLike()` is returning a selection of relevant attributes listed at the first link.

@tjohnsonBCM if you know how to use the API to get the attributes listed at the second link, please let me know! I looked at the API endpoints [here](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.2.0/com.ibm.zos.v2r2.izua700/IZUHPINFO_API_RESTFILES.htm) but couldn't find anything that would work.

tjohnsonBCM: @katelynienaber You're right, I don't think it's possible to use the API to get the attributes listed at the second link.

Discussed this at Zowe CLI Squad meeting. Either of the following approaches are fine:
* Keep the current behavior in this PR of `Create.dataSetLike` returning the list of attributes provided by z/OSMF list API.

  It's OK that this differs from the list of attributes returned by `Create.dataSet`. If we wanted to make all attribute lists consistent, perhaps `Create.dataSet` output should be changed. But that would be a breaking change that can be considered in the future.

  If this approach is taken, we need to be careful about calling multiple z/OSMF APIs in one API method. In the case where the z/OSMF create API succeeds and the z/OSMF list API fails, we should make it clear to the user that the dataset was created successfully. Most API methods call only one z/OSMF API so they don't have to handle this scenario.

  To address this, it would be good to add a unit test that validates if the create API call succeeds and the list API call throws an error, the command should still output a success message.
* Remove the `showAttributes` option from "Create.dataSetLike". It's a nice to have, but if it's too much effort to add now and handle the above scenario, it can be done in the future per community request.

katelynienaber: Hello, thanks for your review!

I will probably remove the show attributes parameter, but I am asking our team first.

Maybe they can think of a use case for returning them.

packages/zosfiles/__tests__/api/methods/create/Create.unit.test.ts
```diff
@@ -13,6 +13,7 @@ import { ImperativeError, TextUtils } from "@zowe/imperative";
 import { Create, CreateDataSetTypeEnum, ZosFilesConstants, CreateDefaults, Invoke, ICreateVsamOptions } from "../../../../";
 import { ZosmfRestClient } from "../../../../../rest/";
 import { ZosFilesMessages } from "../../../../src/api/constants/ZosFiles.messages";
+import { Get, List } from "../../../../src/api";
```
tjohnsonBCM, line 16: Please remove unused imports

test.ts
```diff
@@ -0,0 +1,55 @@
+import { Create } from "./packages/zosfiles/src/api/methods/create/Create";
```
tjohnsonBCM, line 1: Please remove this test file

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to the Zowe CLI package will be documented in this file.
 
 ## Recent Changes
 
+- Added new method "dataSetLike(session, dataSetName, options)" to `Create` class, for use when creating a dataset with parameters like another data set
 - Fix compilation error
```
tjohnsonBCM, line 8: Please merge master into your branch and resolve conflicts. The line "Fix compilation error" is no longer needed here since it was fixed in 6.19.1.

packages/zosfiles/src/api/methods/create/Create.ts
```diff
@@ -134,6 +134,25 @@ export class Create {
...
 
+    public static async dataSetLike(session: AbstractSession, dataSetName: string, likeDataSetName: string): Promise<IZosFilesResponse> {
+        // Required
+        ImperativeExpect.toNotBeNullOrUndefined(dataSetName, ZosFilesMessages.missingDatasetName.message);
+        ImperativeExpect.toNotBeNullOrUndefined(likeDataSetName, ZosFilesMessages.missingDatasetLikeName.message);
+
+        try {
+            const endpoint: string = ZosFilesConstants.RESOURCE + ZosFilesConstants.RES_DS_FILES + "/" + dataSetName;
+
+            const data = await ZosmfRestClient.postExpectString(session, endpoint, [], { like: likeDataSetName });
```
zFernand0, line 145: Could you stringify the `{like: likeDataSetName}` object before passing it to the expectString function?
This will prevent issues caused by what we believe is an old issue, (or maybe it was introduced not long ago?)

packages/zosfiles/__tests__/__system__/api/methods/create/Create.system.test.ts
```diff
@@ -106,6 +106,68 @@ describe("Create data set", () => {
...
+
+        try {
+            response = await Create.dataSetLike(REAL_SESSION, dsnameLike, dsname);
+            Imperative.console.info("Response: " + inspect(response));
+        } catch (err) {
+            error = err;
+            Imperative.console.info("Error: " + inspect(error));
+        }
+
+        expect(response).toBeTruthy();
```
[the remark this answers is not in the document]
zFernand0: I meant for both to exists in this order
`expect(error).toBeFalse();` (this one guarantees that any errors are properly displayed instead of `undefined expected to be truthy`)
`expect(response).toBeTruthy();` (This guarantees that the response.commandResponse expect does not fail with `cannot read property commandResponse of undefined`)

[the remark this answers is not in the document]
katelynienaber: Sorry was being lazy to check how it was working.

katelynienaber:
woohoo!

Issue #751, opened 2020-06-25
Zowe SDKs - API usage examples


MikeBauerCA:
Each package (e.g zosjobs) should contain overview info and documentation on how to leverage its APIs in a README

BrandonJenkins14:
Was addressed in PR #776

PR #768, opened 2020-07-20
Describe Node APIs in Readme

BrandonJenkins14:
Addresses #751 when combined with the `add-api-example` branch. 

**Question:* Do we need instruction about how to import the CLI in the main readme? Or is that covered in every package readme? Anything else I didn't consider?

**Added:** 
- Small blurb at beginning of readme. 
- Section about what the APIs are for and accessing the APIs.
- List of links to readme examples. 

**Misc:** 
- Linked to more complete info about profiles in Create a Profile section.
- Random minor grammar changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+```
+## Recent Changes
+
+- Added documentation about leveraging the Zowe Node APIs to the Readme.
+```
+
 ## `6.18.0`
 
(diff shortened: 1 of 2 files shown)


awharn:
As for your question, I believe the import appears in every one of the readmes you linked. It may be good to show how to import the overall CLI (if you want to use most, if not all, of the API packages built into it) in the overall readme, and then reference that the individual readmes provide examples on how to import the individual packages (which will be true when we revisit this once the SDKs are released on NPM).

PR #777, opened 2020-07-29
Exit with status code 1 in case of severe Imperative.init() errors (fix #748)

### LINKED ISSUE #748: Zowe CLI exits with status code 0 if Imperative.init() fails
https://github.com/zowe/zowe-cli/blob/master/packages/main.ts#L45

The failure of Imperative.init() implies no work can be completed and the program may be unusable. The CLI should exit with a non zero exit code to indicate the error

ChrisBoehmCA:
Tried to figure out a `.sh` test, but it was a challenge since I couldn't figure out how to mock the require of `@zowe/imperative` in a child process. So if it's okay I just did a typescript-level test with jest mocking. 

Fix #748


zFernand0:
Some status-checks seem to be missing for forks?
@awharn @tjohnsonBCM  @dkelosky, thoughts?

awharn:
> Some status-checks seem to be missing for forks?
> @awharn @tjohnsonBCM @dkelosky, thoughts?

I brought this up in the Zowe Explorer repo, but not here. If we have GitHub actions run on pull requests and on push, then pull requests from the branches will run twice and leave no runners left for the repository for about half an hour, possibly longer. So either we are going to need to only run the actions on pull requests, or we are going to need to skip the pull request actions for a PR whose origin is from the same repository as the destination (but this will result in about 12 aborted tests).

zFernand0:
is there a way to delete status-checks? or prevent them from showing up within GH Actions?

dkelosky:
@awharn - I dont doubt youre right, but can you link some of these capacity details?  It seems like some hugely popular repos use github actions, and I'm wondering how they function with these kinds of limitations

dkelosky:
we also saw they may not all run twice if complete on a push before opening the PR

awharn:
@dkelosky  Here are the details. We fall under the 'free' tier - https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits

awharn:
Keep in mind each of our builds generates 9 tests using a matrix of OSes (linux, windows, macos) and node versions (8, 10, 12). We also have the changelog check, audit check, and linting, all of which only take a few seconds.

dkelosky:
I’m too dense to understand. Where does thirty minutes come into play? Are the 9 tests in the matrix each 30 minutes long?

awharn:
Each of them are roughly 20-30 minutes long per test, yes.

dkelosky:
OK, so to your point 9 x 2 = 18 which is almost all of the available test runners.  But as a runner frees up, a next test should run, so hopefully it'll even out well.  

The jest repo only runs on pull requests but they take on 12 runs in their matrix.  If we observe a lot of issues with simultaneous PRs, we can look at ways to streamline a bit.   

In the meantime, I'll copy / paste in stuff for running on PRs if you all think that's a good first step.

awharn:
Sounds good. Another thing to consider is that one macOS build will need to wait until one of the other five complete, since GitHub limit each repository to 5 simultaneous macOS builds.

Issue #764, opened 2020-07-17
Add "--encoding" option for uploading dir to PDS


tjohnsonBCM:
As a CLI user that frequently uploads a local directory to a PDS,
I want an "--encoding" option on the command `zos-files upload dir-to-pds`,
so that I don't need to upload each member individually using the command `zos-files upload file-to-data-set` (which already has an "--encoding" option).

PR #793, opened 2020-08-19
Add dir to pds encoding

mehmetalpsumer:
This should add encoding option for "dir to pds" as requested on #764 .


packages/zosfiles/src/cli/upload/dtp/DirToPds.handler.ts
```diff
@@ -29,12 +29,18 @@ export default class DirToPdsHandler extends ZosFilesBaseHandler {
             stageName: TaskStage.IN_PROGRESS
         };
         commandParameters.response.progress.startBar({task: status});
-        const response = await Upload.dirToPds(session, commandParameters.arguments.inputdir, commandParameters.arguments.dataSetName, {
-            volume: commandParameters.arguments.volumeSerial,
-            binary: commandParameters.arguments.binary,
-            task: status,
-            responseTimeout: commandParameters.arguments.responseTimeout
-        });
+        const response = await Upload.dirToPds(
```
zFernand0, line 32: I don't mean to be annoying, but could you maybe add a test to the dirToPds API related to the encoding option?
I understand that these lines are already covered by other tests, hence why overall coverage is not going down as much.

mehmetalpsumer: As you mention encoding tests were covered in other parts, but I created a unit test for dirToPds API encoding from what I understood. I am not sure whether this test is redundant or serves the purpose. Any comments? @zFernand0

zFernand0: Thank you very much for adding it!
Now, In case the option name changes in the future (or is removed), it will be more than just snapshots that fail, (which people sometimes update with the `jest -u` command) 😋 
It'd be nice to capture the behavior in a system test as well, but it will end up failing depending on the system configuration.

PR #799, opened 2020-08-26
784 build lerna packages

### LINKED ISSUE #784: Build independent packages in monorepo
Prerequisites
* #782 so that single command can be run to build all packages (`lerna run build`)
* #783 to restructure source folders

Tasks
* Add package.json file for each subfolder in `packages`. Here is an example of what package.json should look like for `zosfiles` package: https://github.com/zowe/zowe-cli/blob/test-lerna2/packages/zosfiles/package.json
  * Package name can be found in [this list](https://github.com/zowe/zowe-cli/issues/750)
  * Package version should be same as in top-level package.json
  * Build, watch, and clean scripts can be copied from example package.json above
  * Include dependencies/dev dependencies on a per-package basis. For example, every package needs `@zowe/imperative` as a dependency, and `typescript` as a dev dependency.
* Add tsconfig.json file for each subfolder in `packages`. It doesn't need many settings if it extends the top-level tsconfig.json:
```json
{
  "extends": "../../tsconfig.json",
  "compilerOptions": {
    "rootDir": "src",
    "outDir": "lib"
  }
}
```
* Imports will be broken everywhere after completing first half of #783. Fix them until `lerna run watch` successfully compiles all packages.
  * Inside `packages/cli`, any imports from `../api` should be replaced with imports from the corresponding SDK package (e.g., import from `../api/SubmitJobs` should be changed to import from `@zowe/zos-jobs-for-zowe-sdk`)
    * List any SDK packages required as dependencies in `packages/cli/package.json`
    * Run `lerna bootstrap` to cross-link these dependencies within the repository

AndrewTwydell:
Resolves #784.

Opening as a draft PR to see the output of the build pipeline. Running `lerna run build` still produces errors that we are unsure how to fix, but we are in the process of looking into them.  

image

We believe the errors starting with `Cannot write file...` are to do with the tsconfig.json files in the packages, but we are unsure how to go about fixing those.

The other type of error is to do with files being outside the `src` folder in the packages although, in most cases, they are.


tjohnsonBCM:
@AndrewTwydell I encountered similar issues when testing Lerna in the `test-lerna2` branch.

For the 1st error "Cannot write file X because it would overwrite input file": I'm not sure what causes this, but a workaround is to define "rimraf" as a dev dependency in each SDK package.json and include the NPM script `"clean": "rimraf lib"` (example [here](https://github.com/zowe/zowe-cli/blob/test-lerna2/packages/zosfiles/package.json#L28)). Then to build all SDKs from the top level, you can run `lerna run --parallel clean && lerna run build`.

For the 2nd error "File X is not under 'rootDir'": This can be resolved by setting `"rootDir": "src"` in each SDK tsconfig.json, which should work fine in the monorepo structure. Here's a sample tsconfig.json for an SDK package like "zosfiles": https://github.com/zowe/zowe-cli/blob/test-lerna2/packages/zosfiles/tsconfig.json

Pranay154:
@zFernand0 Massive thanks for helping out with the merge conflicts! Changes look good to me 😋

packages/zosfiles/src/methods/create/index.ts
```diff
@@ -10,7 +10,9 @@
 */
 
 export * from "./doc/ICreateDataSetOptions";
+export * from "./doc/ICreateUssOption";
```
tjohnsonBCM, line 13: One of the benefits of splitting API and CLI code into separate packages is that it forces us to make sure we aren't missing any exports 👍

tjohnsonBCM (approved):
Thanks for doing this! Hopefully much of the painful part of the monorepo migration is done now 🙂 

I removed `dependencies` from the top-level package.json to ensure that SDK packages still build, which they do 👍

PR #809, opened 2020-09-02
Update readme for SDK

BrandonJenkins14:
@tjohnsonBCM @zFernand0 Wanted to adjust this readme section to prepare for publishing the SDK. 

Changed references to "Zowe Node Client SDK". Linked to Zowe Docs - Using SDKs section, (which will tell them how to download it from Zowe.org). 

For the last sentence "alternatively, you can import zowe cli into your project" is there any first step or a command that the developer needs to know in order to do that?


awharn:
For the last sentence, they would just change the import statement they would use in their code; instead of `import { stuff } from @zowe/zosfiles-node-sdk` or similar (I cannot remember the package names for the life of me) in the code, it would be replaced with `import { stuff } from @zowe/cli`, which is how most, if not all projects currently import the CLI APIs.

tjohnsonBCM (approved):
LGTM

I'm assuming that after the SDK packages are published, we''ll provide more details about how to import them in the Using SDK section.

awharn (approved):
Looks good to me! 👍

zFernand0 (approved):

PR #825, opened 2020-09-23
Daemon Mode PoC

dkelosky:
Planned updates to allow running Zowe CLI as a persistent process (~daemon) for faster response time and a new native client is introduced to communicate with the persistent process.  Implementation, behavior, command names are all subject to change.  

To test:
- Obtain `zowex.exe` binary for your platform and add to `PATH`
- Run `zowe` as a background task with a hidden `--daemon` option:
  - Windows `start zowe --daemon`
  - Linux `zowe --daemon &`
  - pm2 `pm2 start <global-npm-location>/node_modules/@zowe/cli/lib/main.js --name zowe-daemon -- --daemon`
    - `npm list -g | head -1` to see `<global-npm-location>`

Daemon port is configurable via `ZOWE_DAEMON` - the default port is `4000`

Then run commands using `zowex` with the same parameters that you normally use e.g. `zowex jobs list jobs --help`.  

**Known Issues**
* `prompt*` does not work
* `login` does not work


package.json
```diff
@@ -70,6 +73,7 @@
     "get-stdin": "7.0.0",
     "js-yaml": "3.13.1",
     "minimatch": "3.0.4",
+    "pm2": "^4.4.1",
```
dkelosky, line 76: Is this license a problem?  AGPLv3?

dkelosky: removed as a dependency

zFernand0:
LGTM!

awharn (dismissed):
Aside from the things already marked TODO, looks good to me! Can't wait for this to become a feature. 😄

packages/zosfiles/src/api/methods/download/Download.ts
```diff
@@ -87,10 +87,14 @@ export class Download {
             // Note that the "extension" options do not affect the destination if the "file" options were provided
             const destination = (() => {
 
-                const cwd = options.cwd;
+                let lcd = options.lcd;
+
+                if (lcd && lcd[lcd.length - 1] !== '/') {
+                    lcd += '/';
+                }
```
zFernand0, line 94: Could we use `path` functions here?
What do you think about `let lcd = options.lcd ? path.normalize(path.join(options.lcd, "/")); : null` ?

packages/zosjobs/src/api/DownloadJobs.ts
```diff
@@ -107,11 +107,11 @@ export class DownloadJobs {
             parms.outDir = DownloadJobs.DEFAULT_JOBS_OUTPUT_DIR;
         }
 
-        if (parms.cwd) {
-            if (parms.cwd[parms.cwd.length - 1] !== '/') {
-                parms.cwd += '/';
+        if (parms.lcd) {
+            if (parms.lcd[parms.lcd.length - 1] !== '/') {
+                parms.lcd += '/';
```
zFernand0, line 112: similar as before (`path` functions) maybe?

packages/Processor.ts
```diff
@@ -0,0 +1,194 @@
...
+
+import { Imperative } from "@zowe/imperative";
+import * as net from "net";
+import { DaemonClient } from "./DaemonClient";
+
+// TODO(Kelosky): handle prompting cases from login command
+// TODO(Kelosky): prompt* broken - hangs, must restart daemon
+// TODO(Kelosky): colors do not come through on some terminals (must be started via `node lib/main --daemon` to see colors)
+// TODO(Kelosky): stderr
+// TODO(Kelosky): plugins install
```
zFernand0, line 21: 🤔 hmmm

packages/Processor.ts
```diff
@@ -0,0 +1,194 @@
...
+     */
+    private static readonly DAEMON_KEY = "--daemon";
+
+    /**
+     * Undocumented paramter for launching in server mode with a port
+     * @private
+     * @static
+     * @memberof Processor
+     */
+    private static readonly DAEMON_PORT_KEY = "--daemon=";
```
zFernand0, line 52: if this is intended to configure the port in which the daemon instance is listening to, should it be renamed to something like `--daemon-port=`

zFernand0 (commented):
Left a few comments/questions.
Looks great so far (and awesome error handling all around) 😋 
Still looking at `zowe2/src/main.rs` though.

packages/Processor.ts
dkelosky: what should i do here?

dkelosky:
` zowex files upload ftds` errors with `--rfj`, with progress bar, and with `zowe`

package.json
```diff
@@ -65,7 +65,7 @@
     "audit:public": "npm audit --registry https://registry.npmjs.org/"
   },
   "dependencies": {
-    "@zowe/imperative": "5.0.0-next.202009221839",
+    "@zowe/imperative": "file:..\\imperative",
```
t1m0thyj, line 68: Should the "@zowe/imperative" version be "zowe/imperative#next", so the CI build can find it on GitHub?

zowex/src/main.rs
```diff
@@ -0,0 +1,283 @@
...
+    if headers.contains_key(X_ZOWE_DAEMON_VERSION) {
+        let &version = headers.get(X_ZOWE_DAEMON_VERSION).unwrap();
+        if version != 1i32 {
+            headers.clear();
+        }
+    } else {
+        headers.clear();
+    }
+
+    headers
```
awharn, line 170: I'm not very familiar with RUST - what is the purpose of this line?

awharn (commented):
I have one question in addition to Fernando's comments. 😃

dkelosky:
yes its a return, rust is dumb 😄 - is there an extra comment from Fernando that I'm missing?  

ill try to continue to refine / shorten the rust piece (it's supposed to be thin so that i can be re-written if needed) in future PRs

awharn:
> yes its a return, rust is dumb 😄 - is there an extra comment from Fernando that I'm missing?
> 
> ill try to continue to refine / shorten the rust piece (it's supposed to be thin so that i can be re-written if needed) in future PRs

Okay, I was just confused, because the next function specified that it was a return. Fernando's comments were hidden, but on the DaemonClient unit test, some tests have snapshots and some have other expects. Though, on second thought, I think it makes sense.

awharn (approved):

dkelosky:
ok ill merge soon and see if the binary is built - i expect ill need subsequent PRs to fix my automation.  thank you for looking

PR #836, opened 2020-10-06
Improve error handling for starting TSO address spaces

### LINKED ISSUE #28: tso issue command fails with cryptic error
_From @tucker01 on September 18, 2018 13:3_

If an invalid value is specified on your tso profile (for example `columns: 9999`) the start address space command returns an error from zosmf:

```
Command Error:
TSO address space failed to start.
Error Details:
IZUG1115E: The request could not be completed. The common event adapter (CEA) component returned reason code 0x1013, reason "For more information, look up the reason code in the table of CEA reason codes included in IBM z/OS Management Facility Configuration Guide ." , and diagnostic codes 0x20, 0x0, 0x0, 0x0.
```

However, `tso issue cmd` returns a cryptic error:

```
Command Error:
Expect Error: [object Object]No servlet key was supplied.
```

We should surface the same details (assuming its the same error) if possible.

_Copied from original issue: gizafoundation/brightside#249_

ghost:
Fixes #28 and addresses some other minor issues that were discovered:
* Fixed the API for starting TSO address spaces to always fail in the case that a servlet key is missing from the z/OSMF response. Previously the API would appear to succeed if both the "servletKey" and "msgData" fields were missing from the response.
* Updated text of several messages that showed "[object Object]" instead of "Imperative API Error".
* Renamed `IZosfmMessages` to `IZosmfMessages`. Old interface name was preserved to avoid a breaking change, but has been marked as deprecated.


zFernand0 (approved):
LGTM!
Thanks for deprecating the misspelled functions/interfaces. : )

packages/zostso/src/TsoResponseService.ts
```diff
@@ -65,12 +65,12 @@ export class TsoResponseService {
             messages: isNullOrUndefined(collectedResponses)? "" : collectedResponses.messages
         };
 
-        if (isNullOrUndefined(zosmfResponse.servletKey) && zosmfResponse.msgData) {
+        if (!isNullOrUndefined(zosmfResponse.servletKey)) {
```
awharn, line 68: Since we are already changing things around a bit, would it be appropriate to replace the deprecated isNullOrUndefined here?

awharn (commented):
Looks good to me, just have one question.

awharn (approved):
Looks good to me.

PR #840, opened 2020-10-20
help with tests

zFernand0:
Signed-off-by: zFernand0 <fernando.rijocedeno@broadcom.com>

### DIFF
--- a/__mocks__/os.ts
+++ b/__mocks__/os.ts
@@ -1,0 +1,31 @@
+/*
+* This program and the accompanying materials are made available under the terms of the
+* Eclipse Public License v2.0 which accompanies this distribution, and is available at
+* https://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+const os = jest.genMockFromModule("os") as any;
+const realOs = require.requireActual("os");
+
+let osProperties = Object.create(null);
+function __setValues(properties: { [key: string]: string }) {
+    osProperties = Object.create(null);
+    // tslint:disable-next-line forin
+    for (const prop in properties) {
+        osProperties[prop] = properties[prop];
+    }
+}
+
+function homedir() {
+    return osProperties.homedir;
+}
+
+os.__setValues = __setValues;
+os.homedir = homedir;
+os.release = realOs.release;
+
+module.exports = os;
--- a/packages/core/__tests__/utils/__integration__/ProfileUtils.integration.test.ts
+++ b/packages/core/__tests__/utils/__integration__/ProfileUtils.integration.test.ts
@@ -9,12 +9,15 @@
 *
 */
 
+
 import * as imperative from "@zowe/imperative";
 import * as profileUtils from "../../../src/utils/ProfileUtils";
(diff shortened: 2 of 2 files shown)


awharn:
It looks good to me. Thanks for getting that test suite working: I was completely stumped trying to figure that out.  😄

zFernand0:
Just opened the PR to maybe spark some discussion around updating the TEST_ENV class.
@awharn, @tjohnsonBCM, @ChrisBoehmCA, @tucker01, @dkelosky, thoughts?

awharn:
I think that a TEST_ENV option to put everything inside of a .zowe directory would be nice to have. It's where we expect to find everything related to the CLI and profiles, at least for the time being. As we move forward with these SDKs, I can see more situations arising where people are looking in that directory or providing functions that look in that directory.

ChrisBoehmCA:
I think it's OK to even automatically append /.zowe (if it's not already there and use that as a`workingDir` in TestEnvironment if it doesn't break anything else.  Just curious why  `.zowe/` needed to be there for this test? Because .zowe is not created if a ZOWE_CLI_HOME environment variable is specified, so if not having .zowe breaks the test, it might break the thing it's testing when a user doesn't manually include .zowe in ZOWE_CLI_HOME

zFernand0:
> I think it's OK to even automatically append /.zowe (if it's not already there and use that as a`workingDir` in TestEnvironment if it doesn't break anything else. Just curious why `.zowe/` needed to be there for this test? Because .zowe is not created if a ZOWE_CLI_HOME environment variable is specified, so if not having .zowe breaks the test, it might break the thing it's testing when a user doesn't manually include .zowe in ZOWE_CLI_HOME

Adding the `/.zowe` automatically will not cause issues for any other tests (I'll confirm that later today), but that only solves half of the problem since the `getZoweDir` function appends `.zowe` to the `os.homedir`.

**Pros & Cons:** Having an option will at least clear the testName a little bit, but if people need to back out (like in this test), it may not be intuitive as to why we are doing `os.homedir() + "/.."`

**Context:** This particular tests has to do with getting default profiles, so we tried to simulate a real scenario as much as possible 😋

tjohnsonBCM:
> Because .zowe is not created if a ZOWE_CLI_HOME environment variable is specified, so if not having .zowe breaks the test, it might break the thing it's testing when a user doesn't manually include .zowe in ZOWE_CLI_HOME

The thing being tested here is the value returned by the getter `ImperativeConfig.instance.cliHome`. If the ZOWE_CLI_HOME env var is specified, it will return that path without appending ".zowe" to the end. If the env var is *not* specified, then ".zowe" will always be appended to the end.

In this test, ZOWE_CLI_HOME was not defined, hence why ".zowe" was required at the end of the path. Perhaps a simpler way to fix this test would have been to define `process.env.ZOWE_CLI_HOME`, but we didn't think of that 🙂

ChrisBoehmCA:
OK thanks I see. Wasn’t sure if setting the environment variable would invalidate the test or not. Just an idea

zFernand0:
🤯 Thanks for the suggestion @ChrisBoehmCA and @tjohnsonBCM 😋 
https://github.com/zowe/zowe-cli/pull/841/commits/ddca0ae549c60182512fb2f6299636f3a144db54

PR #847, opened 2020-10-29
Revert #845

awharn:
In order to have the .tgz and hosted package on NPM behave in the same fashion, revert #845 to restore peer dependencies. The problem that #845 tried to solve was actually a problem in the bundling pipeline.


t1m0thyj (approved):
LGTM 🙂

zFernand0:
Thank you! : )

PR #848, opened 2020-10-29
Add task to build typedoc for Zowe Node.js SDK

t1m0thyj:
Exclude CLI package from API docs for Zowe Node.js SDK. Also extract SDK version from lerna.json, since it can no longer be found in top-level package.json.


zFernand0 (approved):
LGTM

PR #859, opened 2020-11-11
Use new Imperative without dependency vulnerability

awharn:


t1m0thyj (approved):
LGTM! Thanks for fixing this vuln 🙂

Only failed status checks I see are due to changelogs which can probably be ignored

PR #862, opened 2020-11-20
[json-config-next] Define values for Imperative to generate config template

t1m0thyj:
This PR enhances the config templates generated by https://github.com/zowe/imperative/pull/466
- Set `includeInTemplate` to true for profile properties that should be included in the template
- Define `templateProfileName` to change the name of the root profile that wraps zosmf, tso, & ssh to `lpar1`


gejohnston (approved):
Looks good.

awharn (approved):

Issue #861, opened 2020-11-20
Long response times using Zowe CLI


Rune-Christensen:
Hi,
We are experiencing very long response times using the Zowe CLI.
 
We want to fetch a list of all the members in our main COBOL source library.
This library contains approx 280k members.
We tried this using the zOSMF rest interface: 
```
$ time curl -s --insecure -X GET -b "C:\TEMP\.sysvcookies" -H 'X-IBM-Max-Items: 300000' "https://ourzOS/zosmf/restfiles/ds/mainSourceLibrary/member" > members-rest.txt

real    0m2,126s
user    0m0,000s
sys     0m0,000s
```
And then we tried it using the Zowe CLI:
```
$ time zowe zos-files list all-members mainSourceLibrary > members-cli.txt

real    3m10,039s
user    0m0,045s
sys     0m0,167s
```

When running the Zowe CLI command, we see that after a few seconds the output file contains data, and then grows in size until all members are written.
We therefore suspect that it is the processing of the output from zOSMF, that takes a long time.

Is it possible to reduce the time it takes to proces the output?
This is currently a major roadblock for our adoption of using the VSCode Zowe plugin.

Regards
Rune Christensen

t1m0thyj:
Thanks for reporting this! I've been able to reproduce the long response time when running `zowe zos-files list all-members` on a PDS with 300k members.

The performance should be improved once https://github.com/zowe/zowe-cli/pull/868 is merged. On my machine it reduced the response time from >2 min to 15 sec. It may be even less depending on how fast your z/OSMF instance is (since the Curl command took 10 sec for me).

t1m0thyj:
@Rune-Christensen Could you please try v6.24.6 of Zowe CLI which is available on [the NPM registry](https://www.npmjs.com/package/@zowe/cli)? It includes the fix which should improve response time.

Rune-Christensen:
@t1m0thyj I can confirm that response times have improved a lot :-D
Thank you

PR #868, opened 2020-12-01
Improve performance of zosfiles list handlers

t1m0thyj:
This PR addresses the performance issue reported in #861.

When logging a long list to the console, it is much more efficient to print all lines at once rather than one-by-one.

This reduces the time of the command `zowe zos-files list all-members` for a PDS with 300k members, from 2m 10s to about 15s.

### DIFF
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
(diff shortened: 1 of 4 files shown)


zFernand0 (approved):
Very Smart way to avoid all the buffering done by CommandResponse APIs 👍

awharn (approved):
Looks great to me! Nice fix. 👍

PR #882, opened 2020-12-22
Rust CLI Publish CI:  Upload Rust Code Only

dkelosky:
Rust CLI publish fails because operations are done from the wrong directory - this PR should fix that.  

### DIFF
--- a/.github/workflows/zowe-cli.yml
+++ b/.github/workflows/zowe-cli.yml
@@ -5,9 +5,15 @@ name: Zowe CLI CI
 
 on:
   push:
-    paths-ignore: 'zowex/**'
+    paths-ignore:
+    - 'zowex/**'
+    - '.github/workflows/rust-cli-publish.yml'
+    - '.github/workflows/rust-cli.yml'
   pull_request:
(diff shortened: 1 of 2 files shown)


awharn (approved):

.github/workflows/zowe-cli.yml
```diff
@@ -5,9 +5,15 @@ name: Zowe CLI CI
 
 on:
   push:
-    paths-ignore: 'zowex/**'
+    paths-ignore:
+    - 'zowex/**'
+    - '.github/workflows/rust-cli-publish.yml'
+    - '.github/workflows/rust-cli.yml'
```
t1m0thyj, line 11: Nothing important, but curious why not `- '.github/workflows/rust-cli*.yml'` ?

dkelosky: clever idea - i can switch over in the next one if that's ok.  im expecting a few more PRs like this to get the binary publishing right

t1m0thyj (approved):
LGTM ༼ つ ◕_◕ ༽つ

PR #899, opened 2021-01-09
Like param issue pull request

### LINKED ISSUE #771: LIKE parameter for zowe zos-files create data-set-*
**Is your feature request related to a problem or limitation? Please describe.**
LIKE parameter missing from zowe zos-files create data-set-*

**Describe the experience you'd like**
I would like to have --LIKE parameter for create data-set functions 

**Additional context**

Nike682631:
Fix #771


packages/cli/src/zosfiles/-strings-/en.ts
```diff
@@ -134,7 +140,7 @@ export default {
                     CREATE_DIRECTORY: `Create a USS directory named "testDir" `,
                     SPECIFY_MODE: `Create a USS directory named "testDir" with mode "rwxrwxrwx" `
                 }
-            }
+            }
```
t1m0thyj, line 143: Please remove trailing spaces

packages/cli/src/zosfiles/create/Create.options.ts
```diff
@@ -51,6 +51,16 @@ export const ZosFilesCreateExtraOptions: { [key: string]: ICommandOptionDefiniti
         aliases: ["sz"],
         description: strings.SIZE,
         type: "string"
+    },
+     /**
+     * The like value of data set
+     * @type {ICommandOptionDefinition}
+     */
```
t1m0thyj, line 58: Please add one more space in front of these lines to make linter happy 😋

packages/cli/src/zosfiles/create/like/like.definition.ts
```diff
@@ -0,0 +1,65 @@
...
+        {...ZosFilesCreateOptions.recfm, defaultValue: CreateDefaults.DATA_SET.CLASSIC.recfm},
+        {...ZosFilesCreateOptions.blksize, defaultValue: CreateDefaults.DATA_SET.CLASSIC.blksize},
+        {...ZosFilesCreateOptions.lrecl, defaultValue: CreateDefaults.DATA_SET.CLASSIC.lrecl},
+        ZosFilesCreateOptions.storclass,
+        ZosFilesCreateOptions.mgntclass,
+        ZosFilesCreateOptions.dataclass,
+        ZosFilesCreateOptions.unit,
+        ZosFilesCreateOptions.dsntype,
+        ZosFilesCreateExtraOptions.showAttributes,
+        ZosFilesCreateExtraOptions.like
```
t1m0thyj, line 56: Original comment from @awharn:
> For the initial implementation of this command, I believe the only option should be `ZosFilesCreateExtraOptions.like`.

Please remove lines 42-55. The `CreateDefaults` import at top of file should also be unnecessary after making this change and can be removed.

t1m0thyj (changes requested):
Thanks for addressing comments on the first PR 👍 I've left a few more - after addressing them hopefully the build will pass 🙂 

Please remember to add some unit, integration, and system tests (found in `packages/cli/__tests__/zosfiles`).

packages/cli/__tests__/zosfiles/__system__/create/ds/__scripts__/create_ds_help.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+echo "================Z/OS FILES CREATE DATA-SET HELP==============="
+zowe files create data-set my.data.copy --like my.data.orig --lrecl 1024 --help
```
t1m0thyj, line 5: Please remove the arguments "my.data.copy --like my.data.orig --lrecl 1024". Only the "--help" argument should be passed for this test.

packages/cli/__tests__/zosfiles/__unit__/create/like/__snapshots__/ds.definition.unit.test.ts.snap
```diff
@@ -0,0 +1,14 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
t1m0thyj, line 1: Please rename this folder from "create/like" to "create/ds" to make it consistent with the command name. Sorry for not noticing this earlier.

packages/cli/__tests__/zosfiles/__unit__/create/like/ds.definition.unit.test.ts
```diff
@@ -0,0 +1,36 @@
...
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import { ICommandDefinition } from "@zowe/imperative";
+
+describe("zos-files create  command definition", () => {
+    it("should not have changed", () => {
+        const definition: ICommandDefinition = require("../../../../../src/zosfiles/create/classicPds/ds.definition").DsDefinition;
```
t1m0thyj, line 16: Please update the directory in path for your new handler: "../../../../../src/zosfiles/create/ds/ds.definition" (after renaming "create/like" to "create/ds" as mentioned below).

packages/cli/__tests__/zosfiles/__unit__/create/like/ds.handler.unit.test.ts
```diff
@@ -0,0 +1,100 @@
...
+*/
+
+import { Create, CreateDataSetTypeEnum } from "@zowe/zos-files-for-zowe-sdk";
+import { UNIT_TEST_ZOSMF_PROF_OPTS } from "../../../../../../../__tests__/__src__/mocks/ZosmfProfileMock";
+
+describe("Create data set handler", () => {
+    describe("process method", () => {
+        it("should create a  data set if requested", async () => {
+            // Require the handler and create a new instance
+            const handlerReq = require("../../../../../src/zosfiles/create/classicPds/ds.handler");
```
t1m0thyj, line 19: Please update the directory in path for your new handler: "../../../../../src/zosfiles/create/ds/ds.handler" (after renaming "create/like" to "create/ds" as mentioned below).

packages/cli/__tests__/zosfiles/__unit__/create/like/ds.handler.unit.test.ts
```diff
@@ -0,0 +1,100 @@
...
+                    }
+                } as any);
+            } catch (e) {
+                error = e;
+            }
+
+            expect(error).toBeUndefined();
+            expect(profFunc).toHaveBeenCalledWith("zosmf", false);
+            expect(Create.dataSet).toHaveBeenCalledTimes(1);
+            expect(Create.dataSet).toHaveBeenCalledWith(fakeSession, CreateDataSetTypeEnum.DATA_SET_CLASSIC, dataSetName, {});
```
t1m0thyj, line 94: DATA_SET_CLASSIC should be DATA_SET_LIKE here

packages/cli/src/zosfiles/create/like/ds.handler.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { AbstractSession, IHandlerParameters } from "@zowe/imperative";
+import { Create, CreateDataSetTypeEnum, IZosFilesResponse } from "@zowe/zos-files-for-zowe-sdk";
+import { ZosFilesBaseHandler } from "../../ZosFilesBase.handler";
+
+/**
+ * Handler to like a data set
+ */
+
+export default class BinaryPDSHandler extends ZosFilesBaseHandler {
```
t1m0thyj, line 20: Please rename to applicable name like "DataSetHandler"

t1m0thyj (changes requested):
Thanks for adding tests! I've left a few comments that should help to get them passing 🙂

packages/cli/__tests__/zosfiles/__unit__/create/like/ds.handler.unit.test.ts
Nike682631: DATA_SET_LIKE does not exist in createdatasettypeenum

t1m0thyj: Sorry about that 😢 In this case you'll want to test that `Create.dataSetLike` was called rather than `Create.dataSet`. Please also update the expected parameters to match `Create.dataSetLike(session: AbstractSession, dataSetName: string, likeDataSetName: string)`.

packages/cli/__tests__/zosfiles/__unit__/create/like/ds.definition.unit.test.ts
t1m0thyj: Test here still failing because "classicPds" in the path needs to be replaced with "ds".

t1m0thyj (changes requested):
Thanks for addressing comments 👍 

Integration tests appear to be failing because the new "create ds" command isn't recognized by Zowe CLI. Please register the new command in the "files create" group by adding `DsDefinition` to the list in this file: https://github.com/zowe/zowe-cli/blob/master/packages/cli/src/zosfiles/create/Create.definition.ts#L37-L46

packages/cli/__tests__/zosfiles/__unit__/create/ds/ds.handler.unit.test.ts
```diff
@@ -0,0 +1,102 @@
...
+                        get: profFunc
+                    }
+                } as any);
+            } catch (e) {
+                error = e;
+            }
+
+            expect(error).toBeUndefined();
+            expect(profFunc).toHaveBeenCalledWith("zosmf", false);
+            expect(Create.dataSet).toHaveBeenCalledTimes(1);
```
t1m0thyj, line 95: Please change `Create.dataSet` to `Create.dataSetLike` as done on the line below. `Create.dataSetLike` is the only SDK method being called by the "create ds" command and mocked for this test.

packages/cli/src/zosfiles/create/ds/ds.definition.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Create dataSet command definition containing its description, examples and/or options
+ * @type {ICommandDefinition}
+ */
+export const DsDefinition: ICommandDefinition = {
+    name: "data-set",
+    aliases: ["ds"],
+    description: strings.ACTIONS.DATA_SET_LIKE.DESCRIPTION,
+    type: "command",
+    handler: __dirname + "/like.handler",
```
t1m0thyj, line 28: Please update from "like.handler" to "ds.handler" since the file was renamed.

packages/cli/src/zosfiles/create/ds/ds.handler.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Handler to like a data set
+ */
+
+export default class DataSetHandler extends ZosFilesBaseHandler {
+    public async processWithSession(commandParameters: IHandlerParameters, session: AbstractSession): Promise<IZosFilesResponse> {
+        return Create.dataSetLike(
+            session,
+            commandParameters.arguments.dataSetName,
+            commandParameters.arguments.likeDataSetName
```
t1m0thyj, line 25: Please change to `commandParameters.arguments.like`. There originally was a positional called "likeDataSetName", but it was changed into the "--like" option which can be accessed with `arguments.like`.

packages/cli/__tests__/zosfiles/__system__/create/ds/__scripts__/command/command_create_ds_like.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+echo "================Z/OS FILES CREATE DS==============="
+zowe files create data-set my.data.copy --like my.data.orig
```
t1m0thyj, line 5: Please parameterize this line so the system test can pass values into the script. Something like this should work:
`zowe files create data-set "$1.test.data.set.like" --like $2`

For a similar example, see `packages/cli/__tests__/zosfiles/__system__/create/binaryPds/__scripts__/command/command_create_binary_pds.sh`

packages/cli/__tests__/zosfiles/__system__/create/ds/__snapshots__/cli.files.create.dataset.classic.system.test.ts.snap
```diff
@@ -0,0 +1,29 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
t1m0thyj, line 1: Please delete this file since there isn't a system test anymore that matches its name.

packages/cli/__tests__/zosfiles/__system__/create/ds/cli.files.create.ds.system.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        afterEach(async () => {
+            // use DELETE APIs
+            if (dsnameSuffix !== "") {
+                await delay(delTime);
+                const response = await Delete.dataSet(REAL_SESSION, dsname + "." + dsnameSuffix);
+            }
+        });
+
+        it("should create a data set", () => {
+            dsnameSuffix = "";
```
t1m0thyj, line 65: Please define a suffix here (for example `dsnameSuffix = "like"`), so that the condition on line 58 will evaluate to true and the data set will be cleaned up after test finishes running.

packages/cli/__tests__/zosfiles/__system__/create/ds/cli.files.create.ds.system.test.ts
```diff
@@ -0,0 +1,75 @@
...
+            const response = runCliScript(__dirname + "/__scripts__/command/command_create_ds_like.sh",
+                TEST_ENVIRONMENT, [user]);
```
t1m0thyj, line 67: In addition to `user`, please supply a second parameter with the name of a dataset to "allocate like" from. For example, you could use `defaultSystem.zosjobs.iefbr14PSDataSet` which is a data set already defined to exist in the system test properties.

t1m0thyj (changes requested):
Tests are getting close to being finished and passing 🙂 

Please address the comments, and then run tests with `-u` flag to update snapshots:
```shell
npm run test:unit -- -u
npx jest cli.files.create.dataset.integration.test.ts -u
npx jest cli.files.create.ds.system.test.ts -u
```

If you don't have access to a mainframe and are unable to run the last command to update system test snapshots, that's ok - just let me know and I can run the system test for you.

Nike682631:
> Tests are getting close to being finished and passing 
> 
> Please address the comments, and then run tests with `-u` flag to update snapshots:
> 
> ```shell
> npm run test:unit -- -u
> npx jest cli.files.create.dataset.integration.test.ts -u
> npx jest cli.files.create.ds.system.test.ts -u
> ```
> 
> If you don't have access to a mainframe and are unable to run the last command to update system test snapshots, that's ok - just let me know and I can run the system test for you.

yes i don't have access of mainframe, please do that part for me.

packages/cli/__tests__/zosfiles/__integration__/create/ds/__scripts__/command/command_create_ds_fail_missing_like.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+echo "================Z/OS FILES CREATE DS==============="
+zowe files create data-set my.data.copy
```
t1m0thyj, line 5: Please add host, user, and password arguments to this command, so that it skips prompting for those values and fails due to missing like data set name as desired. For example:
`zowe files create data-set my.data.copy --host fakehost --user fakeuser --pw fakepass`

packages/cli/__tests__/zosfiles/__unit__/create/ds/ds.handler.unit.test.ts
```diff
@@ -0,0 +1,102 @@
...
+            });
+
+            try {
+                // Invoke the handler with a full set of mocked arguments and response functions
+                await handler.process({
+                    arguments: {
+                        $0: "abstract",
+                        _: ["abstract"],
+                        dataSetName,
+                        likeDataSetName,
```
t1m0thyj, line 59: This is shorthand for `likeDataSetName: likeDataSetName`. Since the command was reworked to have a "--like" argument instead of a "likeDataSetName" positional, it needs to be called "like" instead. Please change to `like: likeDataSetName`

gejohnston:
There is no description of the expected behavior of this PR and no reference to the issue driving this PR.
After searching it looks like issue #771 is the relevant issue.

After reading code and the issue, I have one remaining question: What is the behavior when no --like option is supplied?

    zowe files create new.data.set

Nike682631:
> There is no description of the expected behavior of this PR and no reference to the issue driving this PR.
> After searching it looks like issue #771 is the relevant issue.
> 
> After reading code and the issue, I have one remaining question: What is the behavior when no --like option is supplied?
> 
> ```
> zowe files create new.data.set
> ```

@gejohnston It should give out an error i believe

packages/cli/src/zosfiles/create/ds/ds.definition.ts
```diff
@@ -0,0 +1,65 @@
...
+    ],
+    options: [
+        ZosFilesCreateExtraOptions.size,
+        ZosFilesCreateOptions.volser,
+        {...ZosFilesCreateOptions.primary, defaultValue: CreateDefaults.DATA_SET.C.primary},
+        ZosFilesCreateOptions.secondary,
+        {...ZosFilesCreateOptions.dirblk, defaultValue: CreateDefaults.DATA_SET.C.dirblk},
+        {...ZosFilesCreateOptions.recfm, defaultValue: CreateDefaults.DATA_SET.C.recfm},
+        {...ZosFilesCreateOptions.blksize, defaultValue: CreateDefaults.DATA_SET.C.blksize},
+        {...ZosFilesCreateOptions.lrecl, defaultValue: CreateDefaults.DATA_SET.C.lrecl},
```
t1m0thyj, line 49: Please remove the default values for these options - the "create ds" command shouldn't have any as mentioned in #771:
> Now, for this zowe files create data-set command, I'm thinking all of the options should be left without defaults. The expectation being that users will either leverage the --like option and specify optional overrides or not leverage the --like option and fully specify the attributes. If they didn't want to fully specify the attributes they could always opt to use one of the template type creation commands.

packages/zosfiles/src/methods/create/Create.ts
```diff
@@ -134,7 +134,10 @@ export class Create {
         }
     }
 
-    public static async dataSetLike(session: AbstractSession, dataSetName: string, likeDataSetName: string): Promise<IZosFilesResponse> {
+    public static async dataSetLike(session: AbstractSession,
+                                    dataSetName: string,
+                                    likeDataSetName: string,
+                                    options?: Partial<ICreateDataSetOptions>): Promise<IZosFilesResponse> {
```
t1m0thyj, line 140: I like this approach to add optional argument so there isn't a breaking change 👍

packages/zosfiles/src/methods/create/Create.ts
```diff
@@ -134,7 +134,10 @@ export class Create {
...
         // Required
         ImperativeExpect.toNotBeNullOrUndefined(dataSetName, ZosFilesMessages.missingDatasetName.message);
         ImperativeExpect.toNotBeNullOrUndefined(likeDataSetName, ZosFilesMessages.missingDatasetLikeName.message);
```
t1m0thyj, line 143: Please validate the options before calling the ZosmfRestClient. Example of doing this in `Create.dataSet`:
https://github.com/zowe/zowe-cli/blob/6b993c90093e84ab3af7a0e281fda7d8b86fdff7/packages/zosfiles/src/methods/create/Create.ts#L123-L125

t1m0thyj, line 143: Only the "like" option is being passed to the REST client: `JSON.stringify({ like: likeDataSetName })`
Other options in the `options` object must also be passed in. Please modify to stringify those options as well. Something like this should work: `{ like: likeDataSetName, ...(options || {}) }`

t1m0thyj (changes requested):
Thanks for adding the requested CLI options 🙂 Left a few comments.

To fix outdated snapshots in tests, please run these commands:
Unit tests: `npm run test:unit -- -u`
Integration tests: `npx jest <testFilename> -u`

packages/zosfiles/src/methods/create/Create.ts
```diff
@@ -142,6 +142,13 @@ export class Create {
...
 
+        // Removes undefined properties
+        let tempOptions = !isNullOrUndefined(options) ? JSON.parse(JSON.stringify({ like: likeDataSetName, ...(options || {}) })) : {};
+        Create.dataSetValidateOptions(tempOptions);
+        const endpoint: string = ZosFilesConstants.RESOURCE + ZosFilesConstants.RES_DS_FILES + "/" + dataSetName;
+
+        const data = await ZosmfRestClient.postExpectString(session, endpoint, [], JSON.stringify(tempOptions));
```
t1m0thyj, line 150: The `Create.datasetLike` API appears to be calling the z/OSMF endpoint twice. Please delete lines 153-155, and move lines 148-150 inside the `try` block. This should result in the desired behavior to call the endpoint only once with `tempOptions`.

packages/cli/src/zosfiles/create/ds/ds.definition.ts
```diff
@@ -58,7 +58,8 @@ export const DsDefinition: ICommandDefinition = {
     examples: [
         {
             description: strings.ACTIONS.DATA_SET_LIKE.EXAMPLES.EX1,
-            options: "NEW.DATASET --like EXISTING.DATASET"
+            options: {"NEW.DATASET --like EXISTING.DATASET",
+                      "NEW.DATASET --like EXISTING.DATASET --lrecl 1024"
```
t1m0thyj, line 62: The `options` property doesn't support an array with multiple values. Please add the second example as a new `{description: string, options: string}` object in the `examples` array.

t1m0thyj (changes requested):
Thanks for making the requested updates 🙂 

Before I can try running the `create ds` command on the mainframe, there are a few more issues to address.

packages/cli/src/zosfiles/create/ds/ds.definition.ts
t1m0thyj: After making this change, please update unit and integration test snapshots to reflect the updated command help. 🙂

packages/cli/src/zosfiles/-strings-/en.ts
```diff
@@ -40,6 +40,12 @@ export default {
                     EX1: "Create an empty C code PDS with default parameters"
                 }
             },
+            DATA_SET_LIKE: {
+                DESCRIPTION: "Create data sets based on the properties of an existing data set",
+                EXAMPLES: {
+                    EX1: "Create an dataset with default parameters and like flag",
+                    EX2: "Create an dataset with default parameters and like flag and lrect flag"                }
```
t1m0thyj, line 47: Please change `lrect` to `lrecl`, and update unit and integration test snapshots after making this change.

packages/zosfiles/src/methods/create/Create.ts
```diff
@@ -134,15 +134,22 @@ export class Create {
...
+                                    dataSetName: string,
+                                    likeDataSetName: string,
+                                    options?: Partial<ICreateDataSetOptions>): Promise<IZosFilesResponse> {
         // Required
         ImperativeExpect.toNotBeNullOrUndefined(dataSetName, ZosFilesMessages.missingDatasetName.message);
         ImperativeExpect.toNotBeNullOrUndefined(likeDataSetName, ZosFilesMessages.missingDatasetLikeName.message);
 
+        // Removes undefined properties
+        const tempOptions = !isNullOrUndefined(options) ? JSON.parse(JSON.stringify({ like: likeDataSetName, ...(options || {}) })) : {};
+        Create.dataSetValidateOptions(tempOptions);
```
t1m0thyj, line 147: The `Create.dataSetValidateOptions` API doesn't recognize "like" as a valid option. Please add a case for "like" (e.g., `case "like":`) to the `switch` statement below:
https://github.com/zowe/zowe-cli/blob/7c16e6587628ce86f26ecc60bd64356df4ac4a02/packages/zosfiles/src/methods/create/Create.ts#L296-L301

t1m0thyj (approved):
Looks good to me, just needs 1 more approval and a vulnerability fix which we can handle before merging 🙂

Nike682631:
> Looks good to me, just needs 1 more approval and a vulnerability fix which we can handle before merging slightly_smiling_face

finally :D

zFernand0 (approved):
LGTM! 😋

awharn (approved):

zFernand0 (approved):
Thanks for the contribution!
Good to go!

PR #902, opened 2021-01-11
Job statement

abastyr:
Added a possibility to use a custom job statement during the workflow creation.


t1m0thyj (commented):
LGTM, just noticed a typo that I commented about.

@zFernand0 Could you take a look at this PR? I don't have experience with workflows to able to test the updated command.

packages/workflows/src/Create.ts
```diff
@@ -72,6 +72,7 @@ export class CreateWorkflow{
     public static createWorkflow(session: AbstractSession, WorkflowName: string, WorkflowDefinitionFile: string,
                                  systemName: string, Owner: string, VariableInputFile?: string, Variables?: string,
                                  AssignToOwner?: boolean, AccessType?: accessT, DeleteCompletedJobs?: boolean,
+                                 JobStatement?: Array<string>,
                                  zOSMFVersion = WorkflowConstants.ZOSMF_VERSION
                                     // add job statement, account info, comments and resolveGlobalConflictByUsing,
                                     ): Promise<ICreatedWorkflow> {
```
zFernand0, line 78: I'm afraid this may be a breaking change : (
Someone may already be expecting zosmfVersion to be the 11th parameter on the `createWorkflow`  API.

packages/workflows/__tests__/__unit__/Create.unit.test.ts
```diff
@@ -592,7 +598,7 @@ describe("Create workflow from local file", () => {
             let response: any;
             try {
                 response = await CreateWorkflow.createWorkflowLocal(PRETEND_SESSION, wfName, wfDefinitionFile, systemName, wfOwner,  null,
-                    null, null, null, null, false, null, "");
+                    null, null, null, null, false, null, null, "");
```
zFernand0, line 601: Should we create an interface that substitutes all 10+ parameters going into the `createWorkflow` API?
Also, we should probably deprecate it instead of breaking it right away. 😋

zFernand0 (commented):
Hey @abastyr,
Left a few comments. Wondering what your thoughts are on this. : )

zFernand0 (changes requested):
See previous comments

packages/workflows/__tests__/__unit__/Create.unit.test.ts
abastyr: Oh, that would be so much better :D.

packages/workflows/src/Create.ts
abastyr: Hard to say. Technically I could keep the zosmfVersion as the 10th parameter, but then it would be better to switch it from WorkflowConstants.ZOSMF_VERSION to Null and just replace it in the code with the constant, else I would have to import the WorkflowConstants everywhere where the job statement was added.

I did it this way because that parameter is barely used in the code and unfortunately Typescript doesn't allow to skip random parameters, only the last ones.

abastyr:
> LGTM, thanks for contributing to CLI 🙂 Just noticed a typo that I commented about.
> 
> @zFernand0 Could you take a look at this PR? I don't have experience with workflows to be able to test the updated command.

Fixed :)

packages/workflows/src/doc/ICreateWorkflow.ts
```diff
@@ -91,4 +91,6 @@ export interface ICreateWorkflow {
     // accountInfo?: string;
     // jobStatement?: string;
     // comments?: string;
+
+    jobStatement?: string;
```
zFernand0, line 95: Please add JsDoc comments :yum:

packages/workflows/src/Create.ts
zFernand0: Hey @abastyr,
Gotcha. I definitely see the benefit in having the version as the last parameter since it can be skipped (and a value will still be available for it.

However, say that someone is calling this function like `createWorkflow(session, "name", "defFile","SYS","myID", null, null, null, null, false, "1.0")`

Once they update their CLI version to `^6.x.x`, they may start getting compilation errors since they didn't provide an array in the place they should have. And even if they don't get it at compile/transpile -time, `"1.0".join("\n")` should complain since strings do not know about `.join`.

We can still do the breaking change, but this PR will have to go out as part of the `next` release.

@dkelosky, thoughts?

zFernand0 (changes requested):
left a few comments.
Should we move this to `next` instead of `master`?

packages/workflows/src/Create.ts
dkelosky: I think I like breaking the API for `next` and passing an interface of options going forward; otherwise, that's a lot of `null`s to keep straight 😄

packages/workflows/src/Create.ts
```diff
@@ -96,31 +99,86 @@ export class CreateWorkflow{
...
             data.accessType = "Public";
         }
-        if (isNullOrUndefined(DeleteCompletedJobs)){
+        if (isNullOrUndefined(DeleteCompletedJobs)) {
             data.deleteCompletedJobs = false;
         }
 
         const resourcesQuery: string = `${WorkflowConstants.RESOURCE}/${zOSMFVersion}/${WorkflowConstants.WORKFLOW_RESOURCE}`;
 
         return ZosmfRestClient.postExpectJSON<ICreatedWorkflow>(session, resourcesQuery, [Headers.APPLICATION_JSON], data);
```
zFernand0, line 123: Should the `createWorkflow` and `createWorkflowLocal` create a data object and call their respective `version2` functions instead of revalidating everything?

packages/workflows/src/doc/IParamsCreateWorkflow.ts
```diff
@@ -0,0 +1,107 @@
+import { AbstractSession } from "@zowe/imperative";
+import { accessT } from "./ICreateWorkflow";
```
zFernand0, line 2: Maybe I got this one wrong, but since `IParamsCreateWrofklow` and IParamsCreateWorkflowLocal` only differ in `keepFiles` and `customDir`, is it worth trying to reduce the duplication by extending `IParamsCreateWorkflow`  (Or creating a common interface) ?

packages/cli/src/workflows/create/Create.common.handler.ts
```diff
@@ -75,18 +76,19 @@ export default class CreateCommonHandler extends ZosmfBaseHandler {
...
-                        this.arguments.dataSet,
-                        this.arguments.systemName,
-                        this.arguments.owner,
-                        this.arguments.variablesInputFile,
-                        this.arguments.variables,
-                        this.arguments.assignToOwner,
-                        this.arguments.accessType,
-                        this.arguments.deleteCompleted
-                    );
+                    resp = await CreateWorkflow.createWorkflow2({
```
zFernand0, line 79: Even though the common handler is calling a different function (which usually translates into different/breaking behavior), I believe this is an exception as no existing end-user automation should be negatively impacted.

zFernand0 (commented):
Thanks for addressing the initial concern.
Left a few comments/suggestions.

Now that we have can have two versions of the same function in `main`, should `createWorkflow` be "broken" in the `next`?

packages/cli/src/workflows/create/Create.common.handler.ts
dkelosky: @zFernand0 - perhaps we could have a guideline or contribution recommendation in the future on `v2` API naming standards.  In other cases, it seems like we introduced new method names instead of a incremental suffix.  No action needed, just observations.

abastyr:
Now it says that Java used for the SonarQube scanner is too old.

t1m0thyj:
Marking this PR as draft temporarily - the CLI squad is working with UX to determine the best way to implement the functionality added in this PR 🙂

zFernand0:
@MikeBauerCA @dkelosky Should we get UX involved so we can determine what changes (if any) are needed for this PR?

zFernand0:
Hey @jelaplan

We wanted to get your input on this contribution with the intention to achieve a better user experience.

The way this contribution works now is that users can add any statements that they want to use as prefix for the JCL contained in the created workflow.

I believe a common use case is to overwrite the default job card specified by z/OSMF to customize job parameters.

Given that different terminals handle special characters in a different way, we may want to avoid complex input directly through the CLI
At the bottom you will find an example in various terminals that may help illustrate the differences.

An idea is to read the Job statements from a local-file, stdin, a USS path or even a dataset.
However this also presents a small problem as there will be too many different combinations.
For example: 
`create workflow-from [ds, local-file, uss] --job-statements-from [CLI, sdtin, local-file, uss, ds]`
(total: 15 different combinations ; `ds:CLI`, `ds:local-file`, `...`, `local-file:CLI`, `local-file:uss`, `...`)

Any thoughts / recommendations?

---

"Side-by-side" comparison at the bottom.

In Windows Command Prompt: <code>zowe zos-workflows create workflow-from-local-file "USERID.TESTWF01" --local-file "test-workflow.xml" --owner "USERID" --system-name "SYS1" --workflow-job-statement "//USERIDx JOB (ACCT#)," "//    CLASS=A,MSGCLASS=A,REGION=0M,NOTIFY=USERID" "//* complex line: \\" ` ' @ % $ . / ? ) ( ] [ } { \\\\" --overwrite</code>

In Ubuntu on WSL2: <code>zowe zos-workflows create workflow-from-local-file "USERID.TESTWF01" --local-file "test-workflow.xml" --owner "USERID" --system-name "SYS1" --workflow-job-statement "//USERIDx JOB (ACCT#)," "//    CLASS=A,MSGCLASS=A,REGION=0M,NOTIFY=USERID" "//* complex line: \\" \\\` ' @ % $ . / ? ) ( ] [ } { \\\\" --overwrite</code>
Notice the escaped backtick <code>\\\`</code>

In Windows PowerShell: <code>zowe zos-workflows create workflow-from-local-file "USERID.TESTWF01" --local-file "test-workflow.xml" --owner "USERID" --system-name "SYS1" --workflow-job-statement "//USERIDx JOB (ACCT#)," "//    CLASS=A,MSGCLASS=A,REGION=0M,NOTIFY=USERID" "//* complex line: \\`" `` ' @ % $ . / ? ) ( ] [ } { \\\\" --overwrite</code>
Notice the different escaping characters

In Git Bash: <code>zowe zos-workflows create workflow-from-local-file "USERID.TESTWF01" --local-file "test-workflow.xml" --owner "USERID" --system-name "SYS1" --workflow-job-statement "///USERIDx JOB (ACCT#)," "///    CLASS=A,MSGCLASS=A,REGION=0M,NOTIFY=USERID" "//* complex line: \\" \\\` ' @ % $ . / ? ) ( ] [ } { \\\\" --overwrite</code>
Notice the triple slash for non-commented statements

Seeing them "side-by-side" may help.
(Order: Windows CMD, Ubuntu on WSL2, PowerShell and Git Bash)
```"//USERIDx JOB ..." "//* complex line: \" ` ' @ % $ . / ? ) ( ] [ } { \\"```
```"//USERIDx JOB ..." "//* complex line: \" \` ' @ % $ . / ? ) ( ] [ } { \\"```
```"//USERIDx JOB ..." "//* complex line: \`" `` ' @ % $ . / ? ) ( ] [ } { \\"```
```"///USERIDx JOB ..." "//* complex line: \" \` ' @ % $ . / ? ) ( ] [ } { \\"```

zFernand0:
Rebased with the `zowe-v1-lts` branch since this is the V1-LTS implementation of the job statement enhancement.
However, there are still merge conflicts.

Closing since this technically an enhancement and Zowe V1 is in maintenance mode.

PR #907, opened 2021-01-14
Update Early Access - Using Global Profile Configuration.md

MichaelABroadcom:
Updated with comments from Mike B and Timothy J.

### DIFF
--- a/docs/Early Access - Using Global Profile Configuration.md
+++ b/docs/Early Access - Using Global Profile Configuration.md
@@ -93,7 +93,7 @@ To begin, define a connection to z/OSMF and initialize your configuration files.
 
    The CLI provides a series of prompts.
 
-2. Respond to the prompts to enter a username, and password for a mainframe service such as z/OSMF. The `--global` option ensures that your credentials are stored securely on your computer by default.
+2. Respond to the prompts to enter a username and password for a mainframe service such as z/OSMF. The `--global` option ensures that your credentials are stored securely on your computer by default.
 
    After you respond to the prompts, the following file is added to your local `.zowe` directory:
 
@@ -227,9 +227,9 @@ The CLI checks for option values in the following order. If not found, the next 
 4. Base type profiles
 5. If no value is found, the default value for the option is used.
 
-If you have two services that share the same username and password on the mainframe, you can define your username and password just once in the base array, leaving those fields blank in each service definition.
+The user name and password fields are not supplied in the service definitions.
 
-In the following example, the username and password fields for ZOSMF1 and ZOSMF2 are empty to allow them to inherit values from the base array:
+In the following example, the username and password fields for ZOSMF1 and ZOSMF2 are user name and password fields are not supplied in the service definitions to allow them to inherit values from the base array:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


docs/Early Access - Using Global Profile Configuration.md
```diff
@@ -84,11 +93,11 @@ To begin, define a connection to z/OSMF and initialize your configuration files.
 
    The CLI provides a series of prompts.
 
-2. Respond to the prompts to enter a service name, username, and password for a mainframe service such as z/OSMF. The `--global` option ensures that your credentials are stored securely on your computer by default.
+2. Respond to the prompts to enter a username, and password for a mainframe service such as z/OSMF. The `--global` option ensures that your credentials are stored securely on your computer by default.
```
gejohnston, line 96: Is it really the --global flag which ensures that credentials are stored securely?

gejohnston:
This document refers to the 'config set secure' command. I think that was changed to just 'config set' ?

docs/Early Access - Using Global Profile Configuration.md
```diff
@@ -222,18 +223,14 @@ Zowe CLI uses a "command option order of precedence" that lets your service defi
 The CLI checks for option values in the following order. If not found, the next location is checked:
 1. Options you define explicitly on the command-line
 2. Environment variables
-3. Service array definitions
-4. Base array definitions
+3. Service type profiles
+4. Base type profiles
 5. If no value is found, the default value for the option is used.
 
 If you have two services that share the same username and password on the mainframe, you can define your username and password just once in the base array, leaving those fields blank in each service definition.
```
gejohnston, line 230: In this sentence and the next I suggest that we replace the references to 'supplying a blank or empty user name or password', with a statement that 'the user name and password fields are not supplied' in each service definition.

docs/Early Access - Using Global Profile Configuration.md
MichaelABroadcom: @t1m0thyj I'll need your input on this.

t1m0thyj: The `--global` option stores the JSON configuration file in your CLI home directory (default location `~/.zowe`). If omitted, the JSON file would get created in the current working directory. The --global flag is not related to secure credential storage.

docs/Early Access - Using Global Profile Configuration.md
```diff
@@ -57,18 +60,24 @@ To get started, install the Zowe CLI `@next` version from the online registry. Y
...
    Zowe CLI and optional plug-ins are installed!
 
-5. If you previously had an instance of Zowe CLI installed, your old configuration files are no longer used in this version. Delete the following files from your local `.zowe/` directory:
-   - `.zowe/settings/imperative.json`
-   - `.zowe/profiles`
+5. Open a command-line window and issue the following command:
 
-   **Important!** Prior to deleting the contents of the `/profiles` directory, take note of any mainframe service details that you need (host, port, etc...). You might want to save the entire `/profiles` directory to another location on your computer so that you can reference or restore the profiles later.
+   ```
+   zowe scs revert --force
```
t1m0thyj, line 66: Perhaps we should:
1. inform users that this will remove old secure credentials stored in their operating system's credential vault (the desired behavior)
1. warn users that this also reverts all securely stored credentials to being stored in plain text (potentially undesirable side effect and could be bad if they neglect to delete `~/.zowe/profiles`)

PR #909, opened 2021-01-20
498 http support

### LINKED ISSUE #498: The zowe CLI does not honor the environment variables http_proxy nor https:proxy
The http_proxy and https_proxy environment variables are the standard way to configure a global network proxy in Linux style systems. It looks the zowe command line does not use those variables when trying to establish connection to the zosmf or api mediation layer server.

Alternatively, an option to set the https proxy in the command line and/or the profile would be nice to have.

Rationale: in our environment the contractor developers are not allowed to directly access the mainframes from their VPN and have to use a proxy to reach them, so they can't use the CLI (they can, however, use the Zowe Desktop, and can invoke the APIs using curl or other tools that honor the proxy environment variables).

AndrewTwydell:
Resolves #498 

This PR implements a `--protocol` or `-o` flag to specify either `http` or `https` as a protocol. The default is still `https`.

Imperative already supports a session including a protocol option and defaults it to https if not supplied. This PR sets the protocol parameter when the ISession is created.


zFernand0 (approved):
LGTM!
I know this probably doesn't concern this PR, but should we try to expose the protocol in the Imperative AbstractRestClient Error messages (now that it's available)?
See https://github.com/zowe/imperative/issues/539 for updates on the AbstractRestClient error details.

packages/zosmf/src/ZosmfSession.ts
```diff
@@ -115,6 +115,17 @@ export class ZosmfSession {
...
+    /**
+     * Option used to specify HTTP or HTTPS Protocol
+     */
+    public static ZOSMF_OPTION_PROTOCOL: ICommandOptionDefinition = {
+        name: "protocol",
+        aliases: ["o"],
+        description: "The protocol used (HTTP or HTTPS)",
+        type: "string",
+        group: ZosmfSession.ZOSMF_CONNECTION_OPTION_GROUP
+    };
```
t1m0thyj, line 127: Should we define "http" and "https" as the only allowed values for the --protocol command option?

zFernand0: I don't see a problem in doing so. 😋

t1m0thyj (approved):
LGTM 🙂 

Forgot when I reviewed earlier that CICS also has `defaultValue: "https"` defined on the command option definition for protocol 😢 

Perhaps we should do the same, but I'm ok with this being merged as is - the `defaultValue` property doesn't seem important since the behavior behind the scenes is the same (HTTPS is the default for ZosmfSession).

t1m0thyj (approved):
LGTM, thanks for adding `defaultValue` and good catch on the potential breaking change 😛

packages/zosmf/src/ZosmfSession.ts
```diff
@@ -156,7 +171,8 @@ export class ZosmfSession {
             user: profile.user,
             password: profile.password,
             rejectUnauthorized: profile.rejectUnauthorized,
-            basePath: profile.basePath
+            basePath: profile.basePath,
+            protocol: profile.protocol ? profile.protocol.toLowerCase() : 'https'
```
zFernand0, line 175: Thanks for preventing any strange behavior with the old method 😋

zFernand0 (approved):
LGTM!

t1m0thyj (approved):

PR #807, opened 2020-08-31
Add contributor guidelines for SDK development

tjohnsonBCM:
Resolves #753.

If you've already read the first draft of these guidelines, the Build Automation section is the only new part to review 🙂

PR #925, opened 2021-01-26
Create Early Access - Using Global Profile Configuration.md

dkelosky:
Signed-off-by: Michael Aimino <michael.aimino@broadcom.com>

### DIFF
--- a/docs/Early Access - Using Global Profile Configuration.md
+++ b/docs/Early Access - Using Global Profile Configuration.md
@@ -212,7 +212,7 @@ When you first run the `zowe config init --global` command, the `profiles.base.p
 
 Issue the `zowe config secure` command to re-prompt for all secure fields when you want to update them (for example, when you want to change your username and password).
 
-To secure a specific field, use the command `zowe config set secure --<option-name>`. For example, you can issue `zowe config set secure --host`. If you issue the command for an option that is already secure, the CLI re-prompts you to enter a new option value.
+To secure a specific field, use the command `zowe config set --secure <property-path>`. For example, you can issue `zowe config set --secure profiles.base.properties.password`. If you issue the command for an option that is already secure, the CLI re-prompts you to enter a new option value.
 
 Alternatively, you can use an editor to define options to the secure array in `zowe.config.json` manually. Any option that you define to there becomes secure/prompted-for.
 
--- a/docs/Early Access - Using Global Profile Configuration.md
+++ b/docs/Early Access - Using Global Profile Configuration.md
@@ -93,13 +93,15 @@ To begin, define a connection to z/OSMF and initialize your configuration files.
 
    The CLI provides a series of prompts.
 
-2. Respond to the prompts to enter a username and password for a mainframe service such as z/OSMF. The `--global` option ensures that your credentials are stored securely on your computer by default.
+2. Respond to the prompts to enter a username and password for a mainframe service such as z/OSMF. The `config init` command ensures that your credentials are stored securely on your computer by default.
 
    After you respond to the prompts, the following file is added to your local `.zowe` directory:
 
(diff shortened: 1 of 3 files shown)


docs/Early Access - Using Global Profile Configuration.md
```diff
@@ -0,0 +1,559 @@
...
+
+1. Issue the following command:
+
+   ```
+   zowe config init --global
+   ```
+
+   The CLI provides a series of prompts.
+
+2. Respond to the prompts to enter a username and password for a mainframe service such as z/OSMF. The `--global` option ensures that your credentials are stored securely on your computer by default.
```
gejohnston, line 96: I think that we should replace "'--global' option" with "config init command"

docs/Early Access - Using Global Profile Configuration.md
```diff
@@ -0,0 +1,559 @@
...
+
+   After you respond to the prompts, the following file is added to your local `.zowe` directory:
+
+   - `zowe.config.json` - A global configuration file. This is the primary location where your MF service connection details such as host and port are defined. Use this configuration file for the following procedures.
+
+3. Issue a Zowe CLI command to test that you can access z/OSMF. For example, list all data sets under your user ID:
```
gejohnston, line 102: I think that after the "config init" command and before the "zowe zos-files list" command, we need another step, telling the user that (s)he must edit values in the zowe.config.jcon file. You have a section later about editing the config file, but I think a user needs step here telling him/her to edit the config file before running a command. It could be something like:

Edit the zowe.config.json file to change the name for your LPAR, to supply a value for host name, and to change your port number if needed.

gejohnston (changes requested):
I suggested a couple of changes. Let's see if the other reviewer agree.

docs/Early Access - Using Global Profile Configuration.md
t1m0thyj: Agreed, from comment in #807:
> The --global flag is not related to secure credential storage.

MichaelABroadcom: Updated

docs/Early Access - Using Global Profile Configuration.md
MichaelABroadcom: Added a Step

Edit the zowe.config.json file to change the name of the LPAR, to supply a value for the host name, and to change the port number if needed.

gejohnston (approved):
Looks good. Thanks.

docs/Early Access - Using Global Profile Configuration.md
```diff
@@ -0,0 +1,561 @@
...
+```
+
+You can continue to add more LPARs, and more services within each LPAR. After you make changes, save the file and issue a Zowe CLI command to the service to verify connection.
+## Managing credential security
+
+When you first run the `zowe config init --global` command, the `profiles.base.properties.user` and `profiles.base.properties.password` fields are defined to the "secure" array in your configuration file. This ensures that username and password are stored securely on your computer.
+
+Issue the `zowe config secure` command to re-prompt for all secure fields when you want to update them (for example, when you want to change your username and password).
+
+To secure a specific field, use the command `zowe config set secure --<option-name>`. For example, you can issue `zowe config set secure --host`. If you issue the command for an option that is already secure, the CLI re-prompts you to enter a new option value.
```
t1m0thyj, line 215: The command for securing a specific field is `zowe config set --secure <property-path>`.

An example of this command would be `zowe config set --secure profiles.base.properties.password`.

t1m0thyj (approved):
LGTM 🙂

t1m0thyj:
Ignoring unit test failures on "next" branch, they can be addressed in another PR as this one only contains doc changes.

PR #928, opened 2021-01-27
Zowe daemon prompt

dkelosky:


t1m0thyj (approved):
LGTM. I see you are artificially inflating your # of lines contributed by changing v1 package-lock JSONs to v2 every time you update them with NPM 7 😁

dkelosky:
I did it as a draft because I was unsure which branch I should target.  

I really don't understand `package-lock.json`, and didn't realize it raised line counts 😞.  Do you think it would be best to downgrade my `npm` version?

PR #931, opened 2021-01-29
Moved the job card changes to the `next`

abastyr:
Job statement related changes from the `master` branch


abastyr:
@zFernand0 Hey, could you please check that error? The `checkTestsCompile` works on my machine, and it seems this error wasn't caused by my changes.

zFernand0:
@abastyr, seems like `checkTestsCompile` worked on permutations of the (OS, Node_version) matrix. 😋 
However, Jenkins seems to be complaining about the conflicts.

To avoid having to deal with all of these, I'll recommend waiting for #902 to be merged into `main` and then we can update next with the newest features (including the `jobStatement`)

How does that sound? 😄

abastyr:
> @abastyr, seems like `checkTestsCompile` worked on permutations of the (OS, Node_version) matrix. 😋
> However, Jenkins seems to be complaining about the conflicts.
> 
> To avoid having to deal with all of these, I'll recommend waiting for #902 to be merged into `main` and then we can update next with the newest features (including the `jobStatement`)
> 
> How does that sound? 😄

OK, sounds good to me :)

packages/workflows/src/api/doc/IParamsCreateWorkflow.ts
```diff
@@ -0,0 +1,118 @@
...
+     */
+    session: AbstractSession;
+
+    /**
+     * Workflow title.
+     *
+     * @type {string}
+     * @memberof IParamsCreateWorkflow
+     */
+    WorkflowName: string;
```
awharn, line 37: I noticed that some of these parameters on the interface are in UpperCamelCase, and some are in lowerCamelCase. I think we would want to be consistent in which one we are using, and most other interfaces seem to use lowerCamelCase.

packages/workflows/src/api/doc/IParamsCreateWorkflowLocal.ts
```diff
@@ -0,0 +1,133 @@
...
+     */
+    session: AbstractSession;
+
+    /**
+     * Workflow title.
+     *
+     * @type {string}
+     * @memberof IParamsCreateWorkflowLocal
+     */
+    WorkflowName: string;
```
awharn, line 36: Same comment as IParamsCreateWorkflow

awharn (changes requested):

zFernand0:
This PR is waiting for #902

zFernand0:
Reasons for closing this PR:
- As it stands, there are 2 years worth of changes that need to be integrated with this enhancement. 
  - Not impossible but tedious.
- With the maturity of the `zowe.config.json` files, it's likely for this feature to have more adoption by storing the Job Statement as an object as a profile property.
  - [Dan's blog](https://medium.com/zowe/tip-how-to-store-zowe-cli-arguments-in-team-config-b96c7d366edc) may help when/if we decide to support these complex structures.
  - Example: 
    ```jsonc
      //...
      "jobStatement": [
          "//USERIDx JOB (ACCT#),",
          "//    CLASS=A,MSGCLASS=A,REGION=0M,NOTIFY=USERID"
      ],
    ```
- There was a possible breaking change in the initial implementation of #902 and worked around in this one.
  - We can discuss if it still makes sense to introduce this enhancement alongside the desired changes to the `createWorkflow` APIs for Zowe V3.

PR #966, opened 2021-03-02
 made changes to definition files for zowe ssh commands - pt2; see issue 603

Jazzy-Jeff:


packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## `6.26.1`
```
zFernand0, line 5: I believe we don't want to lock the changelog down to a specific version (to prevent having to merge PRs in a specific order).

Could we change this to `## Recent Changes` for now?

packages/cli/src/zosuss/ZosUss.definition.ts
```diff
@@ -14,8 +14,8 @@ import { IssueCommand } from "./issue/Issue.definition";
 import { SshSession } from "@zowe/zos-uss-for-zowe-sdk";
 
 export const definition: ICommandDefinition = {
-    name: "zos-uss",
-    aliases: ["uss"],
+    name: "zos-ssh",
+    aliases: ["ssh", "zowe-uss", "uss"],
```
zFernand0, line 18: Changes look good to me.
There might be a few snapshots that need updated.
Feel free to run `npm run test:unit -- -u` in order to update any snapshots 😋

zFernand0 (changes requested):
LGTM!
Left a few comments, and also, there might still be an unsigned commit in the history.
Please follow the instructions in the Details of the DCO step 
https://github.com/zowe/zowe-cli/pull/966/checks?check_run_id=2014965988

packages/cli/src/zosuss/ZosUss.definition.ts
awharn: The alias 'zowe-uss' needs to be 'zos-uss' or else this change will be breaking.

awharn (changes requested):

zFernand0:
Thanks again for the contribution!

zFernand0 (approved):

Issue #947, opened 2021-02-11
Create.DataSet requires a DS type to be passed


katelynienaber:
Right now, if I call `zowe.Create.dataSet` I have to pass in `cmdType`, which is of type `CreateDataSetTypeEnum`. Meaning I must pass in one of the Zowe default data set types in order to call this command (_not_ dsntype, this is a CLI specific type).

I would like to be able to create a data set by just passing in parameters and a name, and not have to use the Zowe types.

-------------------------------------------------------------------------------------------------------------------------------------------

**Use case:** I want to let users make their own data set templates in Zowe Explorer. They should be able to create data sets with a minimum of 5 fields specified (alcunit, dsorg, primary space, recfm, lrecl). All of the standard data set types made available by the CLI contain 6 or 7 fields. So if the user leaves 1 or 2 fields blank, and I choose an arbitrary type for them, I will be allocating a data set with more parameters than they specified.

Ex. the user allocates a data set from this template:
```
{
            alcunit: "CYL",
            dsorg: "PO",
            primary: 10,
            recfm: "U",
            lrecl: 27998
}
```

I select for them the data set type PDS (`CreateDataSetTypeEnum` value 3), which has these default values:
```
{
            alcunit: "CYL",
            dsorg: "PO",
            primary: 1,
            dirblk: 5,
            recfm: "FB",
            blksize: 6160,
            lrecl: 80
}
```

So even if I overwrite the defaults, their data set will have the following unwanted parameters applied as well:
```
{
            dirblk: 5,
            blksize: 6160,
}
```

Instead, I should just be able to pass in the user's template, and if allocation fails, I receive an error back from MF.

zFernand0:
This enhancement has had no activity for 12 months. The issue also has less than 5 up-votes by the community. No action on this enhancement is targeted for the next 2 calendar quarters. Therefore, this enhancement is being closed. If you feel that this enhancement should continue to be available for community up-votes, you may reopen this issue

PR #973, opened 2021-03-08
Blank data set template

katelynienaber:
Adding a standard DS template which contains no parameters. For use with DS templates in Zowe Explorer.

Linked CLI issue (can be closed when this is merged): https://github.com/zowe/zowe-cli/issues/947


packages/cli/src/zosfiles/create/blankPds/BlankPDS.definition.ts
```diff
@@ -0,0 +1,62 @@
...
+// Does not use the import in anticipation of some internationalization work to be done later.
+const strings = (require("../../-strings-/en").default as typeof i18nTypings).CREATE;
+
+/**
+ * Create Blank PDS command definition containing its description, examples and/or options
+ * @type {ICommandDefinition}
+ */
+export const BlankPDSDefinition: ICommandDefinition = {
+    name: "data-set-blank",
+    aliases: ["bin"],
```
zFernand0, line 25: I'm afraid `bin` may be a conflicting alias.

packages/cli/src/zosfiles/create/blankPds/BlankPDS.definition.ts
```diff
@@ -0,0 +1,62 @@
...
+
+// Does not use the import in anticipation of some internationalization work to be done later.
+const strings = (require("../../-strings-/en").default as typeof i18nTypings).CREATE;
+
+/**
+ * Create Blank PDS command definition containing its description, examples and/or options
+ * @type {ICommandDefinition}
+ */
+export const BlankPDSDefinition: ICommandDefinition = {
+    name: "data-set-blank",
```
zFernand0, line 24: Have you tried the `zowe files create ds` command?
I believe it may allow for the same use case.

zFernand0 (commented):
Template addition looks OK to me.
Left a few comments 😋

packages/cli/src/zosfiles/create/blankPds/BlankPDS.definition.ts
katelynienaber: Can I ask where I can find more info about `zowe files create` command?

I see docs for `zowe zos-files create` command but it takes a template argument: https://docs.zowe.org/stable/web_help/index.html?p=zowe_zos-files_create

packages/cli/__tests__/zosfiles/__integration__/create/ds/__scripts__/create_ds_help.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 echo "================Z/OS FILES CREATE DATA-SET HELP==============="
-zowe files create data-set --help
+zowe files create --help
```
zFernand0, line 5: Wondering if this was intended.
Since this script belongs to the `.../create/ds` folder, shouldn't it display data-set help?

packages/cli/src/zosfiles/-strings-/en.ts
```diff
@@ -58,6 +58,12 @@ export default {
                     EX1: "Create an empty physical sequential data set with default parameters"
                 }
             },
+            DATA_SET_BLANK: {
```
zFernand0, line 61: I believe this was used in the `create data-set-blank` command definition, but seems like it no longer exists.

packages/zosfiles/src/methods/create/Create.ts
```diff
@@ -70,6 +70,9 @@ export class Create {
             case CreateDataSetTypeEnum.DATA_SET_CLASSIC:
                 tempOptions = { ...CreateDefaults.DATA_SET.CLASSIC, ...tempOptions };
                 break;
+            case CreateDataSetTypeEnum.DATA_SET_BLANK:
+                tempOptions = { ...CreateDefaults.DATA_SET.BLANK, ...tempOptions };
```
zFernand0, line 74: Should we add a unit test to `packages/zosfiles/__tests__/__unit__/methods/create/Create.unit.test.ts`?

packages/core/src/rest/ZosmfRestClient.ts
```diff
@@ -20,6 +20,7 @@ import { ZosmfHeaders } from "./ZosmfHeaders";
  * @extends {RestClient}
  */
 export class ZosmfRestClient extends RestClient {
+    mDecode: boolean;
```
zFernand0, line 23: I'm a little confused as to why this is needed in this PR.
@t1m0thyj, was this updated in master?

zFernand0 (changes requested):
Hey @katelynienaber, 
Is this change expected to be consumed by the Zowe Explorer this PI?

packages/core/src/rest/ZosmfRestClient.ts
t1m0thyj: Yes, `mDecode` should already be defined on master branch:
image

I encountered a similar error this morning where CLI failed to build because of complaints about `mDecode`, deleting all package-lock.json files and reinstalling fixed it:
```bash
cd zowe-cli
rm -f package-lock.json
rm -f packages/**/package-lock.json
npm install
```

packages/zosfiles/src/methods/create/Create.ts
katelynienaber: I added one...please lmk if it is enough. Since the blank template doesn't contain any properties, we don't really need to test overriding of template properties.

packages/cli/__tests__/zosfiles/__integration__/create/ds/__scripts__/create_ds_help.sh
katelynienaber: I was getting an error on the integration tests...I can try to remove it again and see if they pass now

zFernand0 (approved):
LGTM

katelynienaber:
> Hey @katelynienaber,
> Is this change expected to be consumed by the Zowe Explorer this PI?

Imho this won't be consumed until next PI. But if we get lucky and things go fast, maybe it would P:

To implement templates I need to use the API for `next`. Because I just want to build the templates system once, and have the templates stored in the config file. I would wait until the basic framework for consuming the API is built up in ZE before complicating things with templates...so...

Template prerequisites:
- The CLI API needs to be working
- We need to consume the API in Zowe Explorer

packages/core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe core SDK package will be documented in this file.
 
+## Recent Changes
```
t1m0thyj, line 5: Nothing has changed in the "core" package, could you please remove the added lines in this file?

packages/core/src/rest/ZosmfRestClient.ts
```diff
@@ -20,7 +20,6 @@ import { ZosmfHeaders } from "./ZosmfHeaders";
  * @extends {RestClient}
  */
 export class ZosmfRestClient extends RestClient {
-
     /**
```
t1m0thyj, line 23: Could you please add the blank line back above? Sorry for being picky 😢 If Lerna sees any Git changes in the "core" package, it will require a changelog update and publish a new version of the core SDK package.

t1m0thyj (changes requested):

t1m0thyj (approved):
LGTM 🙂

PR #975, opened 2021-03-10
Update next with v6.27.0-v6.28.0

awharn:
Also bumps the version of Imperative next being used.


zFernand0 (approved):
LGTM!

package.json
```diff
@@ -14,6 +14,7 @@
     "lint": "npm run lint:packages && npm run lint:tests",
     "lint:packages": "lerna run --parallel lint",
     "lint:tests": "tslint --format stylish -c ./tslint-tests.json \"/__tests__/**/*.ts\"",
+    "syncpack:imperative": "syncpack fix-mismatches --dev --prod --filter @zowe/imperative",
```
t1m0thyj, line 17: Thanks for adding 🙂

packages/core/src/rest/ZosmfRestClient.ts
```diff
@@ -21,6 +21,13 @@ import { ZosmfHeaders } from "./ZosmfHeaders";
  */
 export class ZosmfRestClient extends RestClient {
 
+    /**
+     * Construct zosmf rest client to use content decoding
+     * @param mSession The session object to use
+     */
+    constructor(mSession: AbstractSession){
+        super(mSession);
+    }
```
t1m0thyj, line 30: This constructor (and the AbstractSession import) can be removed since it is now identical to the `RestClient` constructor.

zFernand0 (approved):
LGTM!

PR #985, opened 2021-04-08
[next] Create test utils package

t1m0thyj:
This PR creates a new package @zowe/cli-test-utils that exports the TestEnvironment class. In the past, any CLI plug-in with system tests had to copy the TestEnvironment class for their own use which has resulted in code duplication and no easy way to update the TestEnvironment across all plug-ins.

The TestEnvironment class has been updated to use team config profiles. The new default behavior of `TestEnvironment.setUp` is to create team config profiles instead of old school ones. To still use old school profiles, set the environment variable `ZOWE_CLI_TEST_OLD_PROFILES=1` when running system tests.

Creation of profiles has been generalized in this PR - instead of separate methods for `createZosmfProfile`, `createSshProfile`, `createTsoProfile`, etc., a single method is used to create a profile of type X. This imposes a limitation on the structure of "custom_properties.yaml" - it must now match the CLI profile structure. The following are no longer valid:
```
zosmf:
  pass: XXX
cmci:
  host: example.com
```
and must be replaced with:
```
zosmf:
  password: XXX
cics:
  host: example.com
```

This package will only be published with the "next" tag on NPM, to avoid needless work to backport it to "zowe-v1-lts".

PR #988, opened 2021-04-12
Publish cli-test-utils package

t1m0thyj:
This package was supposed to get published when #985 was merged but didn't 😢 

The Lerna config needed to be updated so it no longer ignores all changes inside the `__tests__` directory.

Additionally, I had to update some comments in the cli-test-utils package so that Lerna sees something changed in this PR 😛


zFernand0 (approved):
LGTM!

PR #993, opened 2021-04-12
Add jcl symbolic substitution

gejohnston:
Enables the jobs submit command to accept JCL symbol substitution.

- Adds IBM headers accepted by the zosmf REST API to specify symbol names and values.

- Adds the option --jcl-symbols to the jobs submit command to enable users to specify JCL symbol names and values. Prime examples are:

    - Symbol value with no spaces:

        --jcl-symbols SYMNAME=VALUE

    - Multiple symbols specified:

        --jcl-symbols "SYM1=VAL1 SYM2=VAL2 SYM3=VAL3"

    - Symbol values with embedded spaces and an embedded single quote:

        --jcl-symbols "SYM1='O''Hare airport is busy' SYM2=QUOTESNOTREQUIRED"

- Description in our help:
    ```
      --jcl-symbols  | --js (string)

      A string of JCL symbols to use for substitution. For symbol values with no
      spaces: "symbol1=value1 symbol2=value2 ...". When a value contains spaces,
      enclose the value in single quotes: "symbol1='value 1 with spaces'
      symbol2='value 2 with spaces' ...". To embed a single quote in a value, use two
      single quotes: "NAME=O''Brian".
    ```


eej_prior_commits.txt
```diff
@@ -0,0 +1,26 @@
+The following commits were made pursuant to the Developer Certificate of Origin, even though a Signed-off-by: was not included in the commit message.
```
t1m0thyj, line 1: Could this file be placed inside a "dco_signoffs" folder?

Sorry for being picky 😢 I'd prefer the number of files at the top level to be as few as possible and have noticed that other Zowe repos use a subfolder for their DCO signoffs ([docs site](https://github.com/zowe/docs-site/tree/master/dco_signoffs), [Zowe Explorer](https://github.com/zowe/vscode-extension-for-zowe/tree/master/dco_signoffs)).

packages/cli/__tests__/zosjobs/__integration__/submit/data-set/__snapshots__/cli.zos-jobs.submit.data-set.integration.test.ts.snap
```diff
@@ -61,6 +61,14 @@ exports[`zos-jobs submit data-set command should display the help 1`] = `
 
       A file extension to save the job output with. Default is '.txt'.
 
+   --jcl-symbols  | --js (string)
```
t1m0thyj, line 64: Integration test snapshots only changed for `jobs submit data-set` and `jobs submit stdin`. Should a test be added that checks the CLI help text of the `jobs submit local-file` command? I don't think we necessarily need to add such a test in this PR, just bringing it up since I noticed this.

packages/cli/src/zosjobs/submit/data-set/DataSet.definition.ts
```diff
@@ -67,6 +67,15 @@ export const DataSetDefinition: ICommandDefinition = {
             description: "A file extension to save the job output with. Default is '.txt'.",
             implies: ["directory"],
             type: "string"
+        },
+        {
+            name: "jcl-symbols", aliases: ["js"],
+            description: "A string of JCL symbols to use for substitution. " +
```
t1m0thyj, line 73: It would be ideal to have a file for the "zosjobs" package where shared string constants are defined, rather than needing to duplicate them in each command definition. The "zosfiles" package (`packages/cli/src/zosfiles/-strings-/en.ts`) has such a file, but since "zosjobs" doesn't, this is just a remark, not a request for changes 🙂

packages/zosjobs/src/SubmitJobs.ts
```diff
@@ -119,6 +124,12 @@ export class SubmitJobs {
             // default to fixed format records
             headers.push(ZosmfHeaders.X_IBM_INTRDR_RECFM_F);
         }
+        if (parms.jclSymbols) {
+            const extraHeaders = this.getSubstitutionHeaders(parms.jclSymbols);
+            for (const header of extraHeaders) {
+                headers.push(header);
+            }
```
t1m0thyj, line 131: This could be replaced with one line: `headers.push(...extraHeaders);`

No need to change, just pointing out an ES6 trick 😛

packages/zosjobs/src/SubmitJobs.ts
```diff
@@ -289,6 +300,172 @@ export class SubmitJobs {
...
+        const equals = "=";
+        const maxSymLen = 8;
+        let symStartInx = 0;
+
+        moreSymLoop:
+        while(symStartInx < symbols.length) {
+            // skip all blanks at the start of a sym def
+            while (symbols[symStartInx] === blank) {
+                if (++symStartInx >= symbols.length) {
+                    break moreSymLoop;
```
t1m0thyj, line 321: This is cool, I didn't know you could assign names to loops to break out of an outer loop 🚀

t1m0thyj (approved):
LGTM! Left a few comments. The implementation for the JCL symbol parser and accompanying seems quite rigorous 🙂

eej_prior_commits.txt
gejohnston: I completely agree. I did not like it when I did it, but I obediently followed the instructions that I found.
However, this is our repo, so we should be able to do what makes sense to us.
I will move the file into a new dco_signoffs folder.

packages/cli/src/zosjobs/submit/data-set/DataSet.definition.ts
gejohnston: I thought about that when adding the --jcl-symbols definition. Because that was inconsistent with all other definitions in this component, I keep the duplicated text in-line.

On this same subject, because the --jcl-symbols description contains a few lines, it tripped a SonarCube detection for duplicated lines of code.

Do you think it would be more (or less) confusing to move just the definition of --jcl-symbols to a common file?
I feel like reworking all of the job submit flags would be beyond the scope of this PR.

packages/zosjobs/src/SubmitJobs.ts
gejohnston: Yeah, and SonarCloud hates it and wants me to remove the label. I am going to ignore SonarCloud.

packages/cli/src/zosjobs/submit/data-set/DataSet.definition.ts
t1m0thyj: > Do you think it would be more (or less) confusing to move just the definition of --jcl-symbols to a common file?
I feel like reworking all of the job submit flags would be beyond the scope of this PR.

I agree that reworking all flags seems outside of scope. I'll leave it up to you whether or not to move the definition of `--jcl-symbols` to a separate file. On one hand it would be a bit odd to have it by itself in a separate file. On the other hand, I don't think separating it would cause confusion, and if we want to have a common file we need to start somewhere. 🙂

packages/zosjobs/src/SubmitJobs.ts
awharn: Neat, didn't know that it'd work that way with the spread operator.

awharn (approved):
Looks good to me, no additional comments from what Timothy's.

packages/zosjobs/src/SubmitJobs.ts
gejohnston: This was easy enough to change while doing other things.

packages/cli/src/zosjobs/submit/data-set/DataSet.definition.ts
gejohnston: Moving just --jcl-symbols into strings would not provide a very good framework for future storage of other zos-job strings. Doing all of zos-jobs was beyond the scope of this PR. Moving just the SUBMIT definitions seemed like the right compromise.

packages/cli/__tests__/zosjobs/__integration__/submit/data-set/__snapshots__/cli.zos-jobs.submit.data-set.integration.test.ts.snap
gejohnston: I added a help test for jobs submit local-file.

PR #996, opened 2021-04-14
Add commit hook to require signoff

t1m0thyj:
This PR adds the `husky` package as a dev dependency, which enables Git hooks during the `npm install` process.

The `commit-msg` hook will validate every Git commit message to check whether it is signed off. If not, the commit will fail:
```
$ git commit -m "test"
Commit message must be signed off with your user name and email.
To sign off your commit, add the -s flag to the git commit command.
husky - commit-msg hook exited with code 1 (error)
```
image


gejohnston:
I assume that Husky will overwrite some of the files in .git/hooks/*. If someone is using a tool like DCO, it may overwrite some of the same files. If every file created by an existing tool is not completely replaced by Husky's overwrites, the user may need to manually correct some things. 

This is NOT an objection to this PR. It is just something for people to keep in mind, during a one-time transition.

By the way, a commit.template can protect you from forgetting the signature in the first place. Since Husky creates an error for a missing signature (as opposed to DCO, which automatically appends the signature when needed), I will likely rely on a commit.template more than ever.
```
    git config --list | grep template
    commit.template=C:\Users\ej608771\git_commit_template_signature.txt
```
Include your signature in your template, and you get a pre-formatted commit message. In the Git Extension GUI that I use,  my prompt for a commit message looks like this:

image

gejohnston (approved):
I like the fact that right out-of-the-box, a freshly cloned and installed repo will prevent you from creating any unsigned commits.

zFernand0 (approved):

.husky/commit-msg
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
```
dkelosky, line 1: this is cool - perhaps an implication is that if you install `git` without updating `PATH` you might get errors running shell commands from the hook.

we may already have these linux command line tools as a requirement for our repo

.husky/commit-msg
```diff
@@ -0,0 +1,8 @@
...
+. "$(dirname "$0")/_/husky.sh"
+
+if ! (cat "$1" | grep -q "Signed-off-by:"); then
+  >&2 echo "Commit message must be signed off with your user name and email."
+  >&2 echo "To sign off your commit, add the -S flag to the git commit command."
```
dkelosky, line 6: also, should this be a lower case `s`?  I get an error running with capital `S` but I usually sign off with lower case:
```
C:\dev\asm\make>git commit -S -m "test"
gpg: skipped "Dan Kelosky <dkelosky@gmail.com>": No secret key
gpg: signing failed: No secret key
error: gpg failed to sign the data
fatal: failed to write commit object
```

Versus:
```
C:\dev\asm\make>git commit -s -m "test" 
[master ad25c86] test
 1 file changed, 12 insertions(+), 5 deletions(-)
```

.husky/commit-msg
t1m0thyj: Husky claims to work on Windows as long as Git is installed: https://typicode.github.io/husky/#/?id=does-it-work-on-windows

Since Git hooks written in Bash are a feature built into Git itself, and Husky merely automates creating/managing these hooks, I don't think it's necessary for Bash to be on the PATH as long as it's present in the Git installation. I should probably test this in a Windows Sandbox to be sure 😛

.husky/commit-msg
t1m0thyj: Good catch! I assumed the options were case insensitive, but they definitely aren't based on your example.

.husky/commit-msg
dkelosky: cool beans 😎

dkelosky (approved):

gejohnston:
Do we plan to add this to other Zowe repos?

t1m0thyj:
> Do we plan to add this to other Zowe repos?

We can discuss this with the team at standup. I'd like to see it added to other Zowe repos eventually...for the CLI squad and also other squads if they're interested 🙂 But since this is new, I think it would be good to test it just in one repo first (maybe two if we do it for Imperative).

awharn (approved):
Neat! 👍

.husky/commit-msg
dkelosky: thank you 😄

t1m0thyj:
> I assume that Husky will overwrite some of the files in .git/hooks/*.

@gejohnston I also assumed this at first, but actually it doesn't. The contents of "zowe-cli/.git/hooks/*" haven't changed on my machine.

All the hooks managed by Husky are stored in the ".husky" directory. Git finds the hooks here because Husky modifies the Git config at the repo level: `git config set core.hooksPath .husky`. I'm not sure how this setting affects the behavior of the original hooks directory.

PR #998, opened 2021-04-16
added signed commit section to CONTRIBUTING.md

Jazzy-Jeff:
Signed-off-by: Jeffrey Willoughby <Jeffrey.David.Willoughby@ibm.com>

PR #1001, opened 2021-04-20
Update next with changes from master

t1m0thyj:
Port #996 and #998 to next branch


gejohnston (approved):
Looks as good as when we merged it into master.

PR #1017, opened 2021-05-19
Always send the like parameter regardless of the optional options

tanto259:
Signed-off-by: Hartanto Ario Widjaya <tanto259@users.noreply.github.com>

Regardless of what the user passed as the `options`, the like parameter must always be included for allocate like.

My understanding is if options is either null or undefined, it will be caught in `(options || {})`, thus removing the need of `isNullOrUndefined`.

The optional `options` parameter should still be there to allow user to override the obtained parameter from `likeDataSetName`.

### DIFF
--- a/packages/zosfiles/CHANGELOG.md
+++ b/packages/zosfiles/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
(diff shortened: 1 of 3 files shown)


t1m0thyj (approved):
LGTM, thanks for fixing this!

Looks like this problem would have been caught sooner if this unit test verified that the correct options were passed on the z/OSMF REST request:
https://github.com/zowe/zowe-cli/blob/cfda70a99169fac15d259f98518f4b64a2044f33/packages/zosfiles/__tests__/__unit__/methods/create/Create.unit.test.ts#L167-L171

Feel free to update this unit test in your PR 🙂 Or if you prefer, we can create a separate issue for improving the unit test.

tanto259:
Unit test updated :)

PR #1022, opened 2021-06-04
Rm deprecated next

gejohnston:
Remove items that have been deprecated in v1-lts for at least a year.

- Remove the use of Imperative's deprecated IImperativeErrorParms.suppressReport from automated tests
- Replace the use of Imperative's deprecated AbstractRestClient.performRest() with request() in automated tests.
- Removed the following functions from Zowe-CLI
  - ZosmfSession.createBasicZosmfSession
  - ZosmfSession.createBasicZosmfSessionFromArguments
  - SshSession.createBasicSshSession
  - SshSession.createBasicSshSessionFromArguments
- Utilized a semver expression that will only retrieve 5.0.0-next.XXX versions of Imperative.


packages/core/package.json
```diff
@@ -58,6 +58,6 @@
     "typescript": "^3.8.0"
   },
   "peerDependencies": {
-    "@zowe/imperative": "^5.0.0-next"
+    "@zowe/imperative": ">=5.0.0-next.202106021800 <=5.0.0"
```
t1m0thyj, line 61: Sorry if this is a dumb question, I probably don't understand semver well 😢

What are the reasons to use this format:
`>=5.0.0-next.202106021800 <=5.0.0`
instead of `^5.0.0-next`

If we want to change the version to a range rather than an exact version, then it seems odd to me to still reference an exact version.

I suppose we need to make sure the version of Imperative is compatible with Zowe CLI, but if we have to update the version range manually in the future, that seems like more work IMO than letting Jenkins automatically bump the version for us.

t1m0thyj (commented):
If we want custom version ranges for the @zowe/imperative package, it must be removed from the Jenkinsfile:
https://github.com/zowe/zowe-cli/blob/next/Jenkinsfile#L37

Otherwise when this PR is merged, Jenkins will attempt to update Imperative to the latest `@next` release and overwrite the version number in package.json.

packages/core/package.json
gejohnston: "^5.0.0-next"
    will match every prerelase plus every future 5.X.X GA release (5.1.0, 5.2.0, ..., 5.9.7, ...).

">=5.0.0-next.202106021800 <=5.0.0"
    will match the 5.0.0-next prereleases (starting with June 2) and the only 5.0.0 release.
    
Since we need to use a fixed version at GA, I thought this would help force us to change to a fixed version when 5.0.0 goes GA. I could remove the "=" from "<=5.0.0" to limit the dependency to only the pre-release versions.

I used a starting pre-relase version because I thought we would want this version of CLI to have at least the secure array update (or later pre-release of imperative) 

According to normal string comparison, it is a little odd that 5.0.0 is greater than 5.0.0-next.202106021800, but that is how semver works (at least according to the [npm semver calculator](https://semver.npmjs.com/))

Regarding the comment about Jenkinsfile:
I will remove that entry from the Jenkinsfile, if we decide to proceed with a semver range.
If the semver range does not provide enough value to justify the variety of changes, I can certainly go back to always updating fixed versions in the CLI-next dependencies of imperative whenever imperative changes in next.

t1m0thyj (approved):
LGTM!

PR #1039, opened 2021-06-21
Create SDK functions for automatic API ML configuration

### LINKED ISSUE #920: Support Automatic Configuration for API ML users - Creation / testing of SDK function
In support of #892, implement design developed in #919. 

The ultimate goal is to improve the getting started experience for API ML adopters. Users of the API ML will be able to leverage a new command to generate a config (either project or global) containing the needed information (base-path for each service, is SSO supported) from API ML endpoints. If only one service of a specific type is available, it should be set as the default. If multiple services of a specific type exist, then we select the first returned service of that type as the default, output the list of default services and inform how to change the defaults (edit `zowe.config.json`). The user would need to manually choose the default service for that type by editing the json config. 

For this issue, an SDK function will be developed that can perform such an operation. The CLI command that invokes this function will be developed in a later issue.

Desire to have a method that just returns an object, and a separate method to write to disk.

t1m0thyj:
Resolves #920 by creating a Services class in core SDK and implementing/testing 3 APIs:
* `getPluginApimlConfigs` - API to fetch plugin API ML configs (thanks to @gejohnston)
* `getServicesByConfig` - API to call API ML REST endpoint
* `convertApimlProfileInfoToProfileConfig` - API to convert JSON response into Zowe profiles (thanks to @zFernand0)

This PR does not yet expose these APIs in a CLI command - that will be done separately in a future PR.


gejohnston (dismissed):
Looks good.
The parts that I wrote seem particularly clear to me  :-)

packages/core/src/apiml/Services.ts
```diff
@@ -0,0 +1,244 @@
...
+            }
+        }
+
+        return profInfos;
+    }
+
+    /**
+     * Converts apiml profile information to team config profile objects to be stored in a zowe.config.json file
+     * @param profileInfoList List of apiml profiles
+     * @returns List of config profile objects
```
zFernand0, line 162: Any thoughts on adding an `@example`?

packages/core/src/apiml/Services.ts
```diff
@@ -0,0 +1,244 @@
...
+     * Calls the services endpoint of the APIML gateway to obtain a list of
+     * services that support Single Sign-On. This list is compared against a
+     * list of APIML service attributes defined in CLI plug-in configs. When a
+     * service's API ID is present in both lists, a profile info object is
+     * generated to store CLI profile info for connecting to that service.
+     * @static
+     * @param session Session with APIML connection info
+     * @param configs APIML service attributes defined by CLI plug-ins
+     * @returns List of objects containing profile info for APIML services
+     * @memberof Services
```
zFernand0, line 73: Should we document the errors that this function can throw (e.g. `@throw [XYZ error] when Session is undefined`)?
The `[XYZ error]` portion may not be necessary since they will all be `ImperativeError`s

packages/core/src/apiml/Services.ts
```diff
@@ -0,0 +1,244 @@
...
+ * @class Services
+ */
+export class Services {
+    /**
+     * Forms a list of APIML service attributes needed to query APIML for every
+     * REST service for every loaded command group. This information can later
+     * be used to create every connection profile required for every loaded
+     * command group.
+     *
+     * @returns The APIML service attributes needed to query APIML.
```
zFernand0, line 33: Should we document the errors that this function can throw (e.g. `@throw [XYZ error] when loadedConfig is not defined. Call Imperative.init()`)?
The `[XYZ error]` portion may not be necessary since they will all be `ImperativeError`s

packages/core/src/apiml/Services.ts
```diff
@@ -0,0 +1,244 @@
...
+        return profInfos;
+    }
+
+    /**
+     * Converts apiml profile information to team config profile objects to be stored in a zowe.config.json file
+     * @param profileInfoList List of apiml profiles
+     * @returns List of config profile objects
+     * @memberof Services
+     */
+    public static convertApimlProfileInfoToProfileConfig(profileInfoList: IApimlProfileInfo[]): IConfig {
```
zFernand0, line 165: This function does not throw any errors, but should we at least default the `profileInfoList` to an empty array `[]`?
Or rather do `profileInfoList?.forEach(...)` in case someone provides `null|undefined`

zFernand0: Maybe a unit tests should be added for such scenario?

packages/core/src/apiml/Services.ts
```diff
@@ -0,0 +1,244 @@
...
+        profileInfoList.forEach((profileInfo: IApimlProfileInfo) => {
+
+            profileInfo.pluginConfigs.forEach((pluginInfo: IApimlSvcAttrsLoaded) => {
+                configPlugins.add(pluginInfo.pluginName);
+            });
+
+            if (!configDefaults[profileInfo.profType]) {
+                configDefaults[profileInfo.profType] = profileInfo.profName;
+            }
+            if (profileInfo.conflictTypes.includes("profType") && profileInfo.profName !== configDefaults[profileInfo.profType]) {
```
zFernand0, line 189: This may be a wrong assumption but this `if (profileInfo.conflictTypes.includes("profType") && profileInfo.profName !== configDefaults[profileInfo.profType]))` could be replaced with an `else` and combined with the previous clause.

I do believe this is more explicit.

My only concern is about something like this:
```Javascript
const testCase: IApimlProfileInfo[] = [
  {
    profName: "test2.1",
    profType: "type2",
    basePaths: [
        "test2.1/v1"
    ],
    pluginConfigs: new Set(),
    conflictTypes: []
  },
  {
    profName: "test2.2",
    profType: "type2",
    basePaths: [
        "test2.2/v1"
    ],
    pluginConfigs: new Set(),
    conflictTypes: []
  }
];
```

Clearly two profiles exists of the same type but no profType conflict was detected.

Should an error be thrown in this scenario?
And a test added too 😋 ?

zFernand0 (commented):
Overall this looks great!
Left a few questions (even on my own code) 😋

packages/core/src/apiml/Services.ts
t1m0thyj: Not really, I haven't seen it used much but it could be a nice thing to add 🙂

zFernand0: Added example

packages/core/src/apiml/Services.ts
zFernand0: Went with `profileInfoList?.forEach(...)`
Added a test for this

packages/core/src/apiml/Services.ts
zFernand0: Removed `profType` check for a simplified conflict check.
Conflicts are now detected even if there is no conflictTypes provided.

zFernand0:
Based on discussions, a basepath conflict (named `<insert-name-here>` 😋 ) will be created with a structure similar to:
```json
{
  "plugin-name-1": ["gateway-url-1","gateway-url-2"],
  "plugin-name-2": ["gateway-url-2","gateway-url-3"],
  "plugin-name-3": ["gateway-url-3"]
}
```

zFernand0:
FYI, 
e06850c is a little broken with half of the `basePathConflicts` implementation

zFernand0 (approved):
Changes before merging from master looked good to me!
Thanks for updating the branch
Also for addressing the code-review comments

gejohnston:
There is already a later version of imperative published (5.0.0-next.202106221817) than the version used in this PR.

When we update a plugin's version of imperative to 5.0.0-next.202106221817, and if that plugin uses @zowe/cli-test-utils, which in turn depends on imperative, then an 'npm install' of the plugin may receive a version conflict error like this:

npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! Could not resolve dependency:
npm ERR!peer @zowe/imperative@"^5.0.0" from @zowe/cli-test-utils@7.0.0-next.202104132114

Our dependencies require a lot of synchronization. Maybe we need to revisit the use of semver expressions in more situations? If we stick with a specific imperative version, we should probably use the same version everywhere.

t1m0thyj:
Merging, will be interesting to see if the Imperative version gets auto-bumped

Issue #627, opened 2019-12-31
zowe cli failed to run workflow with step calling restapi to execute console command, while the same workflow runs well manually or though workflow apis


rjmaomao:
I have a workflow contains step to call a restapi to execute console commands. 

Created the workflow by command.

_zowe zos-workflows create wfuf "oper" --uss-file "/u/renjing/workflow/oper.xml" --system-name "J80" --variables-input-file /u/renjing/workflow/syslist.txt --owner "renjing" --assign-to-owner --overwrite --zosmf-p zosmfplx1_ 

Run with command, 
_zowe zos-workflows start workflow-full --wn "oper" --wait --zosmf-p zosmfplx1_

Failed with following message.
_renjing@renjings-mbp:~/Documents/Work/Zowe/zWIN$ zowe zos-workflows start workflow-full --wn "oper" --wait --zosmf-p zosmfplx1
Command Error:
Workflow failed or was cancelled or there is manual step.
Error Details:
{"owner":"renjing","accountInfo":null,"access":"Public","workflowName":"oper","automationStatus":{"messageText":"IZUWF8029E: The status code value of \"400\" received from the RESTful API call \"https://zpetplx1zosmf.pok.stglabs.ibm.com:34211/zosmf/restconsoles/consoles/wasadmcn\" does not match the expected status code value of \"200\" . The workflow step \"D_XCF\" failed.","startedTime":1577762153624,"currentStepNumber":"2.1","currentStepName":"D_XCF","stoppedTime":1577762153748,"messageID":"IZUWF8029E","startUser":"renjing","currentStepTitle":"call D XCF command"},"percentComplete":7,"workflowDescription":"This workflow contains different steps to perform daily operator check.","jobStatement":null,"deleteCompletedJobs":false,"globalVariableGroup":null,"containsParallelSteps":false,"workflowDefinitionFileMD5Value":"cb28e624c6d23b868498705bd2e94474","isCallable":null,"system":"UTCPLXJ8.J80","vendor":"zPET","scope":"none","workflowKey":"1d1864f4-d2a8-4b61-a87f-506d50a1911a","statusName":"in-progress","workflowVersion":"1.0","category":"general","workflowID":"Daily operator workflow"}_

Go to zOSMF UI and  saw following message in failure step's Status tab.

_{
   "reason": "The console name WASADMCN already been used by user RENJING, please use other name or connect other user to delete the console.",
   "reason-code": 19,
   "return-code": 1
}_

[...]

PR #1051, opened 2021-07-09
Adds the `config auto-init` CLI command

awharn:
- Adds the `config auto-init` command
- Adds tests surrounding the new command and its handler

⚠️ Do not merge until the Imperative version is bumped following the merge of [#627](https://github.com/zowe/imperative/pull/627) on zowe/imperative ⚠️


packages/cli/src/Constants.ts
```diff
@@ -93,6 +93,87 @@ For ${Constants.DISPLAY_NAME} support, visit ${Constants.SUPPORT_LINK}
      */
     public static readonly HOME_DIR = "~/.zowe";
 
+    public static readonly AUTO_INIT_OPTION_GROUP = "APIML Connection Options";
+
+    /**
+     * Option used in profile creation and commands for hostname
+     */
+    public static AUTO_INIT_OPTION_HOST: ICommandOptionDefinition = {
```
zFernand0, line 101: There seems to be a similar set of options (`BASE_OPTION_HOST`), should they be shared and the group changed on the handler definition?
Not sure if we will ever change these options description or aliases, but if we did... we will have to change them in more than one place.

zFernand0 (approved):
LGTM

t1m0thyj (approved):
LGTM!

gejohnston (approved):

package.json
```diff
@@ -42,6 +42,7 @@
     "@types/fs-extra": "^8.0.1",
     "@types/jest": "^22.2.3",
     "@types/node": "^12.12.24",
+    "@types/uuid": "^8.3.0",
```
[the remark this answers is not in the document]
awharn: I don't recall what the exact error was, but my system had complained that it was missing typings on uuid, and adding this fixed it. I can remove it if you like.

packages/cli/src/Constants.ts
awharn: There are some differences, mainly the group and descriptions. I'm not sure if I follow what you mean about changing the group on the handler definition. Would it be okay if we use the spread operator on the existing options when defining the new ones, and just override the group and description?

zFernand0: Sorry, handler and definition are two separate things. 
I meant the command definition since these options eventually make their way into a `.definition.ts` file.
As you said, the group and description could be replaced at that time : )
IDK if this needs done as part of this PR though. We can worry about that at a later time 😋

packages/cli/src/Constants.ts
```diff
@@ -93,166 +93,151 @@ For ${Constants.DISPLAY_NAME} support, visit ${Constants.SUPPORT_LINK}
...
-    public static BASE_OPTION_PORT: ICommandOptionDefinition = {
-        name: "port",
-        aliases: ["P"],
-        description: "Port number of service on the mainframe.",
-        type: "number",
-        group: Constants.BASE_CONNECTION_OPTION_GROUP
+    public static AUTO_INIT_OPTION_PORT: ICommandOptionDefinition = {
+        ...Constants.BASE_OPTION_PORT,
+        description: "Port number of API Mediation Layer on the mainframe.",
+        group: Constants.AUTO_INIT_OPTION_GROUP
```
zFernand0, line 195: Nice!

package.json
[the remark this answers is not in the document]
t1m0thyj: Since it's a dev dependency, this shouldn't increase the size of our package or affect our vulnerability scans, so I don't think it's a big deal 😛

t1m0thyj (dismissed):

zFernand0 (approved):
LGTM

PR #1066, opened 2021-07-21
Make zowex Linux binary compatible with more distros

t1m0thyj:
Python/PIP/PyPI provides ManyLinux Docker images which can be used to build binaries that support as many Linux distros as possible.

Building the zowex Linux binary in a ManyLinux Docker container (as suggested [here](https://kobzol.github.io/rust/ci/2021/05/07/building-rust-binaries-in-ci-that-work-with-older-glibc.html)) relaxes the GLIBC version requirement, which makes it compatible with CentOS/RHEL 7.

I've tested the binary generated by the GitHub workflow to verify that it works on both Ubuntu 20.04 and RHEL 7: https://github.com/zowe/zowe-cli/actions/runs/1052534508

When this PR is merged, it will publish a patch release (v0.2.1) of "zowex", and we can mention in the release notes that you don't need to update if running on a system where the "zowex" binary already worked 🙂

### DIFF
--- a/.github/workflows/rust-cli-publish.yml
+++ b/.github/workflows/rust-cli-publish.yml
@@ -118,6 +118,13 @@ jobs:
 
     steps:
 
+    - name: Install Rust toolchain
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+        override: true
+
     - uses: actions/download-artifact@v2
       with:
         name: repo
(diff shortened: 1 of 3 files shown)


.github/workflows/rust-cli-publish.yml
```diff
@@ -94,6 +94,61 @@ jobs:
         asset_content_type: application/octet-stream
 
 
+  build-ubuntu:
```
dkelosky, line 97: trivial detail - perhaps this could be `build-linux` or `build-many-linux`

dkelosky (dismissed):
another quirk is that when we get `0.2.1` published, we should file a false positive here for `zowex`: https://symsubmit.symantec.com/false_negative?lang=en

gejohnston (dismissed):

Issue #935, opened 2021-02-01
Improve user friendlyness error for listing members in a data set 


Joe-Winchester:
**Problem**

When using `zowe files list all-members <PDS-NAME>` if the PDS is invalid the error message includes information about 404 codes and rc: 8, which can make it look as though there is a runtime CLI problem rather than just a user error.

```
Joes-MBP-3:~ Joe$ zowe files list all-members "WINCJ.JCL"
Command Error:
z/OSMF REST API Error:
Rest API failure with HTTP(S) status 404
category: 4
rc:       8
reason:   0
message:  LMINIT error
details: 
  - ISRZ002 Data set not cataloged - 'WINCJ.JCL' was not found in catalog.


Error Details:
HTTP(S) error status "404" received.
Review request details (resource, base path, credentials, payload) and ensure correctness.

Host:      winmvs3b.hursley.ibm.com
Port:      32070
Base Path: 
Resource:  /zosmf/restfiles/ds/WINCJ.JCL/member
Request:   GET
Headers:   [{"X-IBM-Max-Items":"0"},{"X-CSRF-ZOSMF-HEADER":true}]
Payload:   undefined
``
 
**Solution**

echo a more user friendly error, such as `Partitioned Data Set {n)` not found.

There are other examples of where a CLI command with an invalid path or input PDS emits the raw https://. response.  If one exists that parses the response into a more friendly error, this could perhaps become a good example for others to base and form some good first issue cli items.  It's possible there is already such an example of how to do this so apologies if so.

**Comment by MikeBauerCA** [comment]
@dkelosky do you think we should open a z/OSMF RFE to improve this error message?

**Comment by dkelosky** [comment]
Potentially, as I read this though, I think we have a good message already.  The problem is that it’s buried by a bunch of other messages.  

Are you interested in only seeing the “details” @Joe-Winchester ?

**Comment by Joe-Winchester** [comment]
I agree there is a good message there, but it's wrapped in a quite a bit of chaff.  By comparison with other CLIs when a mistake is made there is usually a one-liner, e.g 
```
>chmod asd
chmod: asd: No such file or directory
>
```
In the example of `zowe files list all-members "WINCJ.JCL" the current echo to the user has "Command  Error" and "Error Details" that get colored red in my mac terminal and look like a low level crash.  Including the "HTTP(S) error status "404" received" also feels like low level debug info that a user can't really do a lot about.

  What if you just echo'd the details:

```
[...]

Issue #870, opened 2020-12-01
Enhancement: Display for a Given Command What Values Are Being Used


dkelosky:
I have two profiles:

tso1:
```
host: tso1
port: 443
rejectUnauthorized: false
```

ca11:
```
host: ca11
port: 1443
rejectUnauthorized: false
```

I have no base profile.  `zowe jobs list jobs --zosmf-p ca11` fails while `zowe jobs list jobs --zosmf-p tso1` succeeds.  In my windows credential manager I have:
![image](https://user-images.githubusercontent.com/14940561/100781239-c8380880-33d8-11eb-8edd-ced4751a9bd5.png)

I'm guessing that the `tso1` profile "works" because my credentials are being loaded be residually being in the credential manager?  

It would be nice to know all values that are being used for a given command to help troubleshoot what is going on in this case.  Perhaps a `--dry-run` flag that would show me the final order of precedence in:
env vars -> command args -> profiles -> base profile what values are actually being used for a command.  I could also run a command to see what values are currently set in my credential manager.

plavjanik:
Hello, the `--dry-run` flag would be very useful for other purposes as well. We are creating a tool that generates Zowe CLI commands and we would like to validate during our tests that the generated CLI options are valid.

PR #1075, opened 2021-07-23
Improve user friendliness of error for listing all members in a data set

namanpatel1112:
Resolves https://github.com/zowe/zowe-cli/issues/935


packages/zosfiles/src/methods/list/List.ts
```diff
@@ -76,7 +77,18 @@ export class List {
...
+            // Throw detailed error message with REST debugging info if that option is included
+            if (options.debugResponse) {
+                throw error;
+            } else {
+                // Isolate the main error message to display
+                const splitMessage: string[] = error.message.split("details:");
+
+                // Throw short error message
+                process.stdout.write("\n--> Use option `--debug-response` or `--dr` with the command to see REST API level debugging info.\n\n");
+                throw Error(splitMessage[splitMessage.length - 1].replace(/\s+$/, ""));
```
zFernand0, line 90: I believe that we want to avoid writing directly to process.stdout (or stderr) directly.
We should always aim to use the `params.response.console` functions to do so.
However, there is no such concept as a CommandResponse object here since this is an SDK method.

Given that we are throwing an error, maybe we could add the short error message in the `ImperativeError.additionalDetails` property (or some other `ImperativeError.shortDescription` optional property) and let the receiving handler react to such error and print the short message if available.

kudos to @t1m0thyj for helping come up with idea 😋

dkelosky:
Can we see some samples of the more friendly messages pasted here?

namanpatel1112:
**Without using the --dr or --debug-response option:**
image
In the above, if we can remove the 'Unexpected Command Error' and 'Stack' trace messages and just show the 'Message' part, it would be even better. I am not sure how to prevent it from showing the unexpected command error and the stack trace. Let me know if there is a way.


**Using the --dr or --debug-response option, you can see all the REST API info:**
image

zFernand0:
Should we reserve `--dr` for dry-run? #870

namanpatel1112:
Oh Okay, I wasn't aware --dr is being used for dry-run. So, do you think we can use --message-Response or --mr as Joe Winchester suggested in the issue related to this PR or any other name?

zFernand0:
> Oh Okay, I wasn't aware --dr is being used for dry-run. So, do you think we can use --message-Response or --mr as Joe Winchester suggested in the issue related to this PR or any other name?

`--dry-run` has not been implemented as a global option yet. But we may want to use something else for this option.
Unless we think we should use full option names for global ones (i.e. no aliases).
Thank you for considering using other option names like `--message-response | --mr`. 😋

Joe-Winchester:
@namanpatel1112 - what's the best way I can test this PR locally (i.e. get it installed on my mac ? 

Looking at the screen shots if there is a basic 
```
zowe files list all-members "HLQ.NOTHERE"
```
Does this then write the message 
```
--> Use option `--debug-response` ... 
```
if the option hasn't been added ?  

I think as a user the response to 
```
zowe files list ds "HLQ.NOTHERE.*
```
would be
```
ISRZ002 Data set not cataloged - 'HLQ.NOTHERE' was not found in catalog
```
and if a user did 
```
zowe files list ds "HLQ.NOTHERE.* -h
```
then list that the option about getting more detailed information, i.e. make that only echo'd if the user asks for it

For other commands that fall into the same category when I use them to access files that aren't there and get a big gnarly error messages that your more verbose response would benefit from are:
```
zowe files list uss "/nothere"
zowe files upload file-to-uss "file.txt" "/nothere/file.txt"
zowe files download uss "/nothere"
zowe files download ds "HLQ.PDS(NOTHERE)
zowe files upload dir-to-pds "localpath" "HLQ.PDS(NOTHERE)"
```
Generally I find it's the file APIs that are the culprits.  There are probably others than the list shown above.

namanpatel1112:
@Joe-Winchester Yes, you can install it on your local machine. But, hold off on that for now. I think your point about only showing the message '--> Use option `--debug-response...` when the user adds a help option is good. @zFernand0 What do you think? 

If I do this, then we don't have to go into ImperativeError.additionalDetails or create a ImperativeError.shortDescription property, which keeps it much more simple.

Joe, you can test it on your mac after I make the recommended changes.

zFernand0:
> showing the message "--> Use option --debug-response..." when the user adds a help option is good.> 
> If I do this, then we don't have to go into ImperativeError.additionalDetails or create a ImperativeError.shortDescription property, which keeps it much more simple.

It definitely sounds simpler.
Will this message (`Use option --mr`) show up on all errors or just the errors on the commands where the  `--mr` option is supported?

side-comment:
- Should we have a bot or some sort of automation that does the `npm update` for us periodically?

namanpatel1112:
@zFernand0 I think I might have misunderstood you about using ImperativeError.additional details property. I thought I would have to make changes in the imperative framework itself, but I just had to actually throw an imperative error with the error message in the additionalDetails property. So, I implemented it as you recommended.

**Now, without the --message-response | --mr option:**
image

**With the --message-response | --mr option:**
image


The message (Use option --mr) doesn't show up with the error output itself to keep the error clutter free as Joe recommended. Instead I added as an option in the --help doc. So, users can see it under OPTIONS when they check the help doc (zowe files list am "naman.*" -h). So, it would only show up for commands that have that option.
image

I am just working on fixing the issue of codecov check failing, and then, I can do the same thing for other zosfiles commands:
  zowe files list uss "/nothere"
  zowe files upload file-to-uss "file.txt" "/nothere/file.txt"
  zowe files download uss "/nothere"
  zowe files download ds "HLQ.PDS(NOTHERE)
  zowe files upload dir-to-pds "localpath" "HLQ.PDS(NOTHERE)"

I think we should have a way to do `npm update` automatically periodically.

zFernand0:
I'm a little worried about changing the default behavior of error messages.
Not sure if someone is depending on the REST output today to determine and react to errors in an automated pipeline.
Do we know if the `--response-format-json` output of the error is affected by us overwriting the error details?

namanpatel1112:
That's true, I agree with that. Someone depending on the REST output would have to add the --mr flag to ensure there are no breaks in their automated pipeline. It does affect the --response-format-json output. I need to do it differently.

zFernand0:
Reasons for closing this PR:
- It is a little bit behind the `master` branch.
- The squad has come to an agreement on how to move forward with improving error friendliness
  - See: https://github.com/zowe/zowe-cli/issues/935#issuecomment-1276179160
- This approach requires us to add our interpretation in every API where there is an error thrown.

Issue #495, opened 2019-07-23
Expose new option added to List.dataSet API


uzust01:
New option _start_ was added to the List.dataSet options interface - IListOption. 
It is only used when Upload.pathToDataSet uses List.dataSet.

It should me made available for the list DataSetHandler.

PR #1093, opened 2021-07-31
New `start` option for list dataset API

namanpatel1112:
Resolves https://github.com/zowe/zowe-cli/issues/495


packages/cli/src/zosfiles/-strings-/en.ts
```diff
@@ -468,7 +468,8 @@ export default {
             PATH: "Specifies the path where the file system is mounted." +
                 " This option and --fsname are mutually exclusive.",
             FSNAME: "Specifies the name of the mounted file system." +
-                " This option and --path are mutually exclusive."
+                " This option and --path are mutually exclusive.",
+            START: " An optional search parameter that specifies the first data set name to return in the response document."
```
t1m0thyj, line 472: Please remove the leading whitespace from this string 🙂

namanpatel1112: Alright, I made the change and ran the unit and integration tests successfully. I haven't made any other change. I am not sure why the Jenkins check is failing now but before it wasn't failing. I cannot see the 'Details' since I don't have access to it. Any ideas? Thanks!

t1m0thyj: The error in Jenkins is:
`[packages/core] Changelog missing valid header. Please see CONTRIBUTING.md for changelog format.`

I don't think you should have to do anything though, since it doesn't look like you made any changes to the core SDK.

t1m0thyj:
Snapshots are failing in Jenkins build after updating Imperative. Merging now since those snapshots are fixed in #1110.

PR #1124, opened 2021-08-27
Supply string, not error object to 'new Error()'

gejohnston:
The following section of code caused a compile error in a fresh clone of zowe-cli (next branch).
```
} catch (error) {
    logger.error("Error reading test properties yaml configuration file. Tests cannot continue. " +
        "Additional details:" + error);
    throw new Error(error);
}
```

The error it produced was:
```
> tsc --pretty

src/environment/TestEnvironment.ts:150:29 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'string | undefined'.
    Type 'unknown' is not assignable to type 'string'.

150             throw new Error(error);
                                ~~~~~
```

I think you can re-throw an error object:

- throw error;

or you can throw a newly created error object passing a string to the Error constructor:

- throw new Error("some string");

However, the compiler does not seem to like it when you to pass an Error object to the constructor of the Error class.

The code above has been in place for at least 5 months. I do not know why it fails to compile now. Perhaps an update to typescript made the compiler a little more strict?

I used the following code to prevent the error.

```
} catch (error) {
    const errMsg: string = "Error reading test properties yaml configuration file. Tests cannot continue. " +
        "Additional details:" + error;
    logger.error(errMsg);
    throw new Error(errMsg);
}
```

### DIFF
--- a/__tests__/__packages__/cli-test-utils/src/environment/TestEnvironment.ts
+++ b/__tests__/__packages__/cli-test-utils/src/environment/TestEnvironment.ts
@@ -145,9 +145,10 @@ export class TestEnvironment {
             // injectCliProps(properties);
             // console.log(properties);
         } catch (error) {
-            logger.error("Error reading test properties yaml configuration file. Tests cannot continue. " +
-                "Additional details:" + error);
-            throw new Error(error);
+            const errMsg: string = "Error reading test properties yaml configuration file. Tests cannot continue. " +
+                "Additional details:" + error;
+            logger.error(errMsg);
+            throw new Error(errMsg);
         }
         logger.info("Loaded configuration properties file.");
 
--- a/__tests__/__packages__/cli-test-utils/src/environment/TestEnvironment.ts
+++ b/__tests__/__packages__/cli-test-utils/src/environment/TestEnvironment.ts
@@ -145,10 +145,9 @@ export class TestEnvironment {
             // injectCliProps(properties);
             // console.log(properties);
         } catch (error) {
-            const errMsg: string = "Error reading test properties yaml configuration file. Tests cannot continue. " +
-                "Additional details:" + error;
-            logger.error(errMsg);
-            throw new Error(errMsg);
+            logger.error("Error reading test properties yaml configuration file. Tests cannot continue. " +
+                "Additional details:" + error);
+            throw error;
         }
         logger.info("Loaded configuration properties file.");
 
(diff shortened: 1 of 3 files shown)


__tests__/__packages__/cli-test-utils/src/environment/TestEnvironment.ts
```diff
@@ -145,9 +145,10 @@ export class TestEnvironment {
             // injectCliProps(properties);
             // console.log(properties);
         } catch (error) {
-            logger.error("Error reading test properties yaml configuration file. Tests cannot continue. " +
-                "Additional details:" + error);
-            throw new Error(error);
+            const errMsg: string = "Error reading test properties yaml configuration file. Tests cannot continue. " +
+                "Additional details:" + error;
+            logger.error(errMsg);
+            throw new Error(errMsg);
```
t1m0thyj, line 151: Should we rethrow the original error? `throw error;`

My only concern with the current approach, is that when the `error` object is stringified, some traceback details may be lost.

package-lock.json
```diff
@@ -62,7 +62,7 @@
     },
     "__tests__/__packages__/cli-test-utils": {
       "name": "@zowe/cli-test-utils",
-      "version": "7.0.0-next.202108121907",
+      "version": "7.0.0-next.202108202027",
```
t1m0thyj, line 65: Should we update the changelog for the `@zowe/cli-test-utils` package?

__tests__/__packages__/cli-test-utils/src/environment/TestEnvironment.ts
gejohnston: I thought about re-throwing the original error. It would have been easier. I thought I would add the extra text information that was available, but I did not think about the traceback.

I will change to re-throw the error.

package-lock.json
gejohnston: Since the @next release should have had to compile successfully during its last PR and promotion into @next, it is not clear whether this change will be apparent to any consumer.

I can add a Change log entry to the effect of 'Fixed compile error'. What do you think?

t1m0thyj (approved):
LGTM!

package-lock.json
t1m0thyj: It may be apparent to consumers of the cli-test-utils package (CLI plug-in developers), if they experience the same breakage that we did 🙂

The current changelog update looks good to me, I think there may be a bug in the GitHub workflow

zFernand0:
Seems like the changelog check worked on Jenkins, but on on GH Actions?
@awharn any thoughts?

zFernand0 (approved):
PR should be good to be merged! 😋 
Thanks for doing it!!

Issue #1092, opened 2021-07-30
Expose modifyVersion option for deleteJobs command


namanpatel1112:
There is support for "X-IBM-Job-Modify-Version" header in the SDK, but we need to expose modifyVersion option for deleteJobs command in the CLI to allow the user to set the version of the X-IBM-Job-Modify-Version header to use (versions: "1.0" or "2.0").

Refer to:
https://github.com/zowe/zowe-cli/blob/f93260312a60ab9f77c476f152535125835fedfd/packages/core/src/rest/ZosmfHeaders.ts#L98
https://github.com/zowe/zowe-cli/blob/57e30025034a001bcba1eaad1650a0f8828def9c/packages/zosjobs/src/doc/input/IDeleteJobParms.ts#L42
https://github.com/zowe/zowe-cli/blob/57e30025034a001bcba1eaad1650a0f8828def9c/packages/zosjobs/src/DeleteJobs.ts#L49

PR #1134, opened 2021-09-09
added option to include modifyVersion in `zowe jobs delete job` command

namanpatel1112:
Signed-off-by: Naman Patel <namanpatel1112@gmail.com>


namanpatel1112:
Resolves https://github.com/zowe/zowe-cli/issues/1092

namanpatel1112:
@zFernand0 Are you able to chime in on why those checks could be failing. Especially the Audit (vulnerability) check. Thanks!

zFernand0 (commented):
LGTM!
Should the same option be applied to the cancel job command definition?
It'd be cool to do them both in the same version/PR 😋

namanpatel1112:
@zFernand0 I added the `modifyVersion` for the cancel jobs command. 😃

packages/zosjobs/src/doc/input/ICancelJobParms.ts
```diff
@@ -33,8 +33,8 @@ export interface ICancelJobParms {
 
     /**
      * version of the cancel request
-     * @type {string}
+     * @type {boolean}
      * @memberof ICancelJobParms
      */
-    version?: string;
+    version?: boolean;
```
zFernand0, line 39: Apologies for not noticing sooner, but isn't this considered a breaking change?

zFernand0 (changes requested):
Changing the input types of an API should probably be done only on the `@next` version of the CLI since it may not be backwards compatible with other LTS versions.

Is is possible, to keep the version property a string and just provide a different value in the handlers (cancel and delete) if/when the user specifies `--modifyVersion`?

packages/zosjobs/src/doc/input/ICancelJobParms.ts
t1m0thyj: I agree this seems like a breaking change. At least in the LTS branch, we should leave the version property as type "string" same as it was before.

packages/zosjobs/src/doc/input/IDeleteJobParms.ts
```diff
@@ -36,8 +36,8 @@ export interface IDeleteJobParms {
      * If omitted, functionality is the same as if specifying "1.0" - the deletion of the job is asynchronous
      * If "2.0" is specified, the cancel and purge is synchronous
      *
-     * @type {"1.0" | "2.0"}
+     * @type {boolean}
      * @memberof IDeleteJobParms
      */
-    modifyVersion?: "1.0" | "2.0";
+    modifyVersion?: boolean;
```
t1m0thyj, line 42: Same comment as above regarding a breaking change

packages/cli/src/zosjobs/delete/job/Job.definition.ts
```diff
@@ -30,6 +30,15 @@ export const JobDefinition: ICommandDefinition = {
...
         }
     ],
+    options: [
+        {
+            name: "modifyVersion",
+            description: "If you use this option, X-IBM-Job-Modify-Version will be set to \"2.0\" and delete job API will be synchronous. " +
+            "Otherwise, it will be asynchronous.",
+            type: "boolean",
+            required: false
+        }
```
t1m0thyj, line 40: Regarding the CLI command option `--modifyVersion`, would it make sense to define it as an integer with `allowableValues` of 1 and 2? I'm not sure how futureproof it is to make it a boolean, in case a version 3 is created in the future.

t1m0thyj (changes requested):

packages/cli/src/zosjobs/delete/job/Job.definition.ts
awharn: I'm in agreement with Timothy here. While I understand that the main difference right now is whether or not the operation is asynchronous, it is defined in [IBM documentation](https://www.ibm.com/docs/en/zos/2.4.0?topic=interface-cancel-job-purge-its-output) as a string (specifically, "1.0" and "2.0"). If those headers were "true" and "false", I would be more open to changing the type of this argument on the next branch.

Issue #659, opened 2020-02-04
Invalid example provided in workflow command


zFernand0:
https://github.com/zowe/zowe-cli/blob/b1df89258cb5113dc91743494b201faa7a51ebf7/packages/workflows/src/cli/list/retrieveWorkflowDefinition/RetrieveWorkflowDefinition.definition.ts#L50-L54

Also, why is the Class name way different than the file name?
https://github.com/zowe/zowe-cli/blob/b1df89258cb5113dc91743494b201faa7a51ebf7/packages/workflows/src/cli/list/retrieveWorkflowDefinition/RetrieveWorkflowDefinition.handler.ts#L23

PR #1147, opened 2021-09-16
Exposed workflow-name option for `zowe zos-workflows list definition-file-details` and fixed example

namanpatel1112:
Resolves https://github.com/zowe/zowe-cli/issues/659


namanpatel1112:
I exposed the `workflow-name` option for the `zowe zos-workflows list definition-file-details` command and fixed the doc example by adding in the missing `definitionFilePath` positional argument.

t1m0thyj (approved):
LGTM!

t1m0thyj (changes requested):
Noticed a problem with the `workflow-name` option not being used. It seems redundant with the positional `definition-file-path`. Will add more detailed info after discussing further with the team.

awharn (changes requested):
It appears that this particular command breaks form from the rest of the workflow commands in the way that we populate the name of the workflow that we are attempting to retrieve. While most of the other commands populate `workflowName` as an option, this command uses the `definitionFilePath` positional. I cannot locate any code in the handler that will attempt to do anything with the `workflowName` argument.

In the interest of preventing a breaking change, perhaps it would be best to remove the `workflowName` option from this command and update the examples accordingly. We may want to create a follow-up issue to address this inconsistency in `@next`, when a breaking change is allowed.

namanpatel1112:
@t1m0thyj @awharn I made the requested changes. I also made a new issue for adding `workflowName` as an optional argument and removing the `definitionFilePath` as a positional.

packages/cli/src/workflows/list/retrieveWorkflowDefinition/RetrieveWorkflowDefinition.definition.ts
```diff
@@ -46,11 +46,6 @@ export const RetrieveWorkflowDefinition: ICommandDefinition = {
             description: "To list the contents of a workflow definition stored in the UNIX file \"/user/dir/workflow.xml\"" +
             " including its steps and variables",
             options: "\"/user/dir/workflow.xml\" --list-steps --list-variables"
-        },
-        {
-            description: "To list the contents of a workflow definition stored in the z/OS data set \"USER.DATA.SET.XML\"" +
-            " including its steps and variables",
-            options: "--workflow-name \"testWorkflow\" --list-steps --list-variables"
         }
```
t1m0thyj, line 49: Do we want the 2nd example deleted, or the `options` list corrected to match what the description says? I'm fine either way 😛

t1m0thyj (approved):

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Exposed the --workflow-name option for the 'zowe zos-workflows list definition-file-details' command and fixed the help example. [#659](https://github.com/zowe/zowe-cli/issues/659)
```
t1m0thyj, line 7: Should this be updated to reflect the option is no longer being exposed?

t1m0thyj (approved):

awharn (approved):
Looks good to me! 👍

PR #1149, opened 2021-09-23
Add PEM Certificate Support

awharn:
Requires Imperative to be merged first
Requires Unit, Integration, and System tests


packages/core/src/rest/ZosmfRestClient.ts
```diff
@@ -101,6 +101,10 @@ export class ZosmfRestClient extends RestClient {
                 original.additionalDetails = "Token is not valid or expired.\n\n" +
                     "For CLI usage, see `zowe auth login apiml --help`";
             }
+            // TODO: Add PFX support in the future
+            if (this.session.ISession.type === SessConstants.AUTH_TYPE_CERT_PEM) {
+                original.additionalDetails = "The certificate is not valid or expired.\n\n";
```
t1m0thyj, line 106: Sorry for being nitpicky, but should the word `The` be removed to be consistent with the messages for expired password/token?

t1m0thyj (approved):

packages/core/src/rest/ZosmfRestClient.ts
awharn: Yes, it probably should be removed.

PR #1160, opened 2021-10-01
[next] Change default log level to WARN

t1m0thyj:
Also document development tips including:
* The newly added `NODE_ENV=development` setting
* VS Code launch configuration so I don't have to keep digging for it in my email 🙂 (thanks to @gejohnston for providing it)


zFernand0 (approved):
Thanks for updating the version of imperative.

Issue #1195, opened 2021-11-04
Daemon Mode - Erros from the daemon are not forwarded through STDERR


zFernand0:
This should be fixed by #1182

Issue #1192, opened 2021-11-02
Daemon Mode - Socket connection error after every command [Mac-only]


zFernand0:
Version 0.2.1 (`zowex`) throws the following error after every command
_thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 57, kind: NotConnected, message: "Socket is not connected" }', src/main.rs:56:42
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace_

Version 0.3.0 (`next-zowex-is-zowe`) throws a similar error after every command
_The daemon failed to run your command due to this error:
Socket is not connected (os error 57)_

zFernand0:
FYI, 
In the old communication protocol between the Daemon and the Client, the socket errors are the only thing going to `stderr`
https://github.com/zowe/zowe-cli/issues/1195

t1m0thyj:
This error (`Socket is not connected (os error 57)`) is the reason why integration tests fail in our macOS CI build.

Issue #1183, opened 2021-10-26
Daemon Mode - Full suite of rust executable tests


MikeBauerCA:
Full suite of rust executable tests using some form of test toggling.

Issue #1177, opened 2021-10-26
Daemon Mode tests for rust binary on Windows, Linux, Mac


MikeBauerCA:
Create a separate set of automated sanity tests for the rust binary on Windows, Linux, Mac.

zFernand0:
Here is a quick Q&A discussed with @t1m0thyj and @awharn 😋 

1. Do we have to build the linux binary in a container?
  a. Option A: Do a simple build from `ubuntu-latest` for testing purposes
  b. Option B: Create a `pre-build` job and upload the linux artifact. Then download it from the `runs-on: {{ matrix.os }}`
  
2. Should the testing matrix be a bit more complex and include npm versions (6.x and 8.x)?
  a. Probably not since that will make the build have twice as many runners for every push
  
3. Should we remove the npm@7 installation from the CI pipeline?
  a. Yes, at least replace it with NPM 8.x
  b. Another alternative is to use the version of npm that comes with NodeJS

Prereq:
- [ ] `zowe config daemon` #1176 (probably)

PR #1196, opened 2021-11-04
Use binaries for CI testing :yum:

zFernand0:
Things included here:
- Ability to run all integrations tests in Daemon mode
- Ability to run GitHub actions locally using [`nektos/act`](https://github.com/nektos/act)

Issues fixed:
- #1177
- #1183 

Known Test Failures:
- Integration: Checking for `stderr` will fail.
    #1195
- Integration: Mac testing will fail.
    #1192


zFernand0:
Here is an example of how the matrix will be handled by `nektos/act`
image

__tests__/__src__/environment/TestEnvironment.ts
```diff
@@ -40,7 +40,7 @@ export class TestEnvironment extends BaseTestEnvironment {
 
         // Ensure correct path separator for windows or linux like systems.
         const separator = process.platform === "win32" ? ";" : ":";
-        result.env.PATH = `${nodePath.resolve(__dirname, "../../__resources__/application_instances")}${separator}${process.env.PATH}`;
+        result.env.PATH = `${process.env.PATH}${separator}${nodePath.resolve(__dirname, "../../__resources__/application_instances")}`;
```
t1m0thyj, line 43: Would this change cause integration/system tests on our local machines to use a different Zowe CLI, and if so is this desirable?

For example, if `%APPDATA%\npm` is already on my PATH, won't the Zowe bin there take priority over the local CLI instance because of this change?

zFernand0: you are right, this change should not be needed for the purpose of running the Integration tests with the binaries : )
reverting

zFernand0:
The `zowe-cli.yml` workflow may need to be renamed in order for the `workflow_dispatch` (and default values for its inputs) to take effect

.github/workflows/zowe-cli.yml
```diff
@@ -30,45 +75,86 @@ jobs:
...
+      id: upload-binary
+      uses: actions/upload-artifact@v2
+      with:
+        name: zowe-${{ matrix.os }}.tgz
+        path: zowe.tgz
+
+    - name: Setup Binary in PATH
+      if: github.event.inputs.test-type == "binary"
+      id: setup-binary
+      run: tar -xvzf zowe.tgz -C ./node_modules/.bin --overwrite
```
gejohnston, line 136: Should ./node_modules/.bin also be the location that the "zowe config daemon" command puts the zowe executable at a customer site?

gejohnston (approved):
Just a couple of questions to complete my understanding. I am NOT requesting any changes.

.github/README.md
```diff
@@ -0,0 +1,21 @@
...
+- Run it: `npm run test:act`
+- Clean it: `npm run test:act -- --clean`
+
+For more information, run: `npm run test:act -- --help`
+
+**Known Issues for `nektos/act@0.2.24`**
+
+`0.` The first time it will likely take ~3 minutes to run. Subsequent runs (with `--reuse`) should take less than 2 minutes. <br/>
+`1.` Artifact upload and download doesn't work on `nektos/act` (hence why we need to copy the zowe.tgz to a shared Docker volume). <br/>
+`2.` `nektos/act` only works with linux distros. <br/>
```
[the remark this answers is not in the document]
zFernand0: That's right, I will rephrase that statement 👍 
It should work fine on Windows, Linux, MacOS, and almost everywhere you can have Docker (e.g. Raspberry PI - armv7l) 😋

.github/workflows/zowe-cli.yml
zFernand0: I don't think `./node_modules/.bin` will work for end users since the `where/which zowe` location depends on their installation/configuration of node and npm, e.g. using nvm. 
For example:
- `/root/.nvm/versions/node/v16.12.0/bin/zowe`
- `C:\Users\USER\scoop\apps\nvm\current\nodejs\nodejs\zowe`
- `C:\Users\USER\AppData\Roaming\npm\zowe`

Maybe something along the lines of `"${process.env.ZOWE_CLI_HOME}/_bin"` or `"~/.zowe/_bin")`

gejohnston: Thanks. Makes sense. Many customers may not even have the privilege to write to those NodeJS locations when they run the "zowe config daemon" command.

zFernand0:
Even though `binary` is the default `test-type`, I believe that `if: github.event.inputs.test-type == 'binary'` will be skipped until master recognizes the workflow_dispatch event.
https://github.com/zowe/zowe-cli/blob/c746e43d0460e822badaf2b166891bc4cc61d4f3/.github/workflows/zowe-cli.yml#L21-L24

scripts/testCliWorkflow.js
```diff
@@ -0,0 +1,212 @@
...
+    console.log("New workflow saved:", genWfPath);
+  } else {
+    // Execute workflow locally
+    console.log("Executing new workflow...");
+    await _sleep();
+    cp.spawn("act", [`-r${verbose ? 'v' : ''}W`, genWfPath, "-e", epPath], { stdio: "inherit" });
+
+    console.log("Copying existing artifacts...");
+    await _sleep();
+    const copyPath = cp.execSync("docker inspect volume act-toolcache | jq '.[0].Mountpoint'", { encoding: "utf-8" }).trim().replaceAll('"', '');
```
zFernand0, line 174: `(－‸ლ)`
Thanks @t1m0thyj for pointing this out 😋

zFernand0:
Note: 
The following hack was necessary for the binary to be recognized.
```result.env.PATH = `${process.env.PATH}${separator}${nodePath.resolve(__dirname, "../../__resources__/application_instances")}`;```

Which means that the setup-binary step will need to be changed back to modifying the path instead of relying on the `./node_modules/.bin` location to be recognized by the integration tests
```yaml
  run: |
    mkdir -p $HOME/.zowe/_bin
    tar -xvzf zowe.tgz -C $HOME/.zowe/_bin
    echo "$HOME/.zowe/_bin" >> $GITHUB_PATH
```

scripts/testCliWorkflow.js
```diff
@@ -0,0 +1,221 @@
...
+      console.log("Copying existing artifacts...");
+      await _sleep();
+      const testPath = path.resolve(__dirname, "..", "__tests__", "__results__", "nektos_act");
+      _handle(() => {
+        const copyPath = JSON.parse(cp.execSync("docker inspect act-toolcache").toString().trim())[0].Mountpoint;
+        fs.mkdirSync(path.join(copyPath, "artifacts"), { recursive: true });
+        fs.mkdirSync(testPath, { recursive: true });
+        cp.execSync(`tar -zc${verbose ? 'v' : ''}f __act__artifacts.tgz -C ${copyPath} .`);
+        cp.execSync(`tar -zx${verbose ? 'v' : ''}f __act__artifacts.tgz -C ${testPath}`);
+        fs.rmSync("__act__artifacts.tgz", { force: true });
```
t1m0thyj, line 182: Would the following work instead?
`fs.unlinkSync("__act__artifacts.tgz");`

Then the note below about Node 14.14.0 may be unnecessary 🙂 

Sorry I didn't think about this when we discussed it before 😢

t1m0thyj (commented):
@zFernand0 Thanks for addressing all of my feedback related to the `test:act` script 🙂

Left a minor comment below, also I'm still having issues with using `nektos/act` locally 😢 Lots of integration tests fail with an error like the following:
```
Expected: ""
Received: "/mnt/c/Users/XXX/Projects/zowe-next/zowe-cli/packages/cli/__tests__/zostso/__integration__/issue/__scripts__/issue_help.sh: 2: set: Illegal option -·
"
```

Based on some Google searching, the error `set: Illegal option -·` seems to be caused by Bash scripts using Windows line endings (CRLF) instead of Unix ones (LF). I believe what is happening is that on my local machine, the Bash scripts are checked out with CRLF line endings, and then they're getting copied into the Ubuntu Docker container.

Curious what solution you would suggest here - should I change Git settings on my Windows machine to check out files with Unix line endings, or is there an easy way we can convert the line endings when files are copied into the Docker container?

t1m0thyj (approved):
LGTM! Moved my comment above to #1210 so that it can be addressed separately instead of holding up this PR 🙂

.github/workflows/zowe-cli.yml
zFernand0: Note: 
The location where we place the binary changed from `./node_modules/.bin` to `./__tests__/__resources__/application_instances` becuase of the `TestEnvironment.setUp` function. (see below)
https://github.com/zowe/zowe-cli/blob/06330fa6c50a907363717ee0412d044ab2587843/__tests__/__src__/environment/TestEnvironment.ts#L43

awharn (approved):

PR #1197, opened 2021-11-10
Improve Daemon Mode Initialization

awharn:
Improves Daemon startup and config error messaging


zFernand0 (approved):
LGTM! 😋

gejohnston (approved):
I really like 2 things:
1. You did not modify the existing socket communication to fix this problem.
2. We now have a property that we can check throughout our app to know if we are in daemon mode.

t1m0thyj (approved):
LGTM!

Issue #598, opened 2019-11-20
Expose Etag option via CLI 


Alexandru-Dumitru:
PR https://github.com/zowe/zowe-cli/pull/588 added option to request and pass an Etag while working with data sets and uss files. This feature is implemented only at the API level, since there was a need in https://github.com/zowe/vscode-extension-for-zowe/milestone/12.
We should look into exposing this feature via the CLI as well. 

Things to consider:
- how do we return the Etag on download for a single file
    - print to console? (needs scraping and storing)
    - assign some local environment variable which user can read afterwards?
    - combination of both?
- how do we return Etags when downloading a PDS (all members)
- passing Etag while uploading a single file should be pretty straightforward (optional parm, something like --etag _string_)
- how do we pass Etags (if we have them for each file) when uploading a directory to USS or PDS

Please, feel free to add your suggestions on this topic.

Thanks

dkelosky:
@Alexandru-Dumitru and @zFernand0,

Fernando had a nice API idea for RestClient, but if it's too difficult to implement, I was wondering if this would suffice for now for what you need: https://github.com/zowe/imperative/issues/318#issuecomment-551188955

PR #1208, opened 2021-11-15
etag and hidden alias

### LINKED ISSUE #1204: --show-attributes alias is --pa
`zowe files create pds --help`:
image

Should the alias be `--sa` ? On `zowe files list ds --help`, a similar option is shortened to `--attributes`.

dkelosky:
Fix #1204
Partially address #598 

The proposal is to add an `--etag` option on a `zowe files download ds` command.  ETag is only returned when `--rfj` is specified.  I did this because ["If this header is enabled for very large data sets, then performance is impacted since the data set must be read twice by the system."](https://www.ibm.com/docs/en/zos/2.3.0?topic=interface-retrieve-contents-zos-data-set-member).  

Here is a sample download response:
```
C:\dev\asm\source>zowe files download ds SOURE.DATA.SET(MEMBER) --extension .tst  --etag --rfj
{
  "success": true,
  "exitCode": 0,
  "message": "",
  "stdout": "Data set downloaded successfully.\nDestination: ...\n",
  "stderr": "",
  "data": {
    "success": true,
    "commandResponse": "Data set downloaded successfully.\nDestination: ...",
    "apiResponse": {
      "etag": "E765D80FF818BAA1772482569E6FF2DD"
    }
  }
}
```

Secondly, I added a `--match` keyword to the download command for which you can supply an ETag which presents a `If-None-Match` header.  

With both `--etag` and `--match`, I satisfy my own use case where I want to loop through and only download changed files from z/OS.   I'll leave this in draft for for feedback before proceeding to refine any tests.   

### DIFF
--- a/packages/cli/src/zosfiles/-strings-/en.ts
+++ b/packages/cli/src/zosfiles/-strings-/en.ts
@@ -347,6 +347,8 @@ export default {
                 "only when the data set is not cataloged on the system. A VOLSER is analogous to a drive name on a PC.",
             BINARY: "Download the file content in binary mode, which means that no data conversion is performed. The data transfer process " +
                 "returns each line as-is, without translation. No delimiters are added between records.",
+            ETAG: "Return the ETag token to determine if the data on z/OS host has changed.",
+            MATCH: "ETag token use for match checking before download begins.",
             ENCODING: "Download the file content with encoding mode, which means that data conversion is performed using the file encoding " +
                 "specified.",
             FAIL_FAST: "Set this option to false to continue downloading dataset members if one or more fail.",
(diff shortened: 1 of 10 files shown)


MikeBauerCA:
@dkelosky did you personally notice a significant performance improvement in your use case? Is there a reason to not always return `etag` if `--rfj` is specified? If we just returned `etag` then we could use the `--etag` flag to accept an input `etag` rather than `--match`.

MikeBauerCA:
Please disregard my comments above... I was reading about the `If-None-Match` header not the `X-IBM-Return-Etag` header which states `If this header is not present or set to something other than 'true', then the default is to only send an Etag in the response for data sets smaller than a system determined length, which is at least 8MB.` so I see now that you need that extra flag to guarantee you get the response you desire.

dkelosky:
I did a couple of tests on performance anyway...

- Downloading ~50 files with ~10 lines each takes around 27s.  Retrying with `etag` "saves" 2s. 
- Downloading 500 files with ~1000 lines each takes around 5m 36s.  Retrying with `etag` "saves" 1m.  

Perhaps I'm not going to get the savings I anticipated...

PR #1216, opened 2021-11-19
Update Daemon Bundling Script

awharn:
Include the Cargo.toml and Cargo.lock files so the dependencies that were used can be determined in other pipelines.
Archive the artifacts from the Daemon Bundling step in the Jenkins pipeline

### DIFF
--- a/jenkins/bundleDaemon.sh
+++ b/jenkins/bundleDaemon.sh
@@ -19,5 +19,8 @@ for platform in linux macos windows; do
     curl -fsLOJ https://github.com/zowe/zowe-cli/releases/download/native-v$daemonVersion/zowe-$platform.tgz
 done
 
+curl -fsLOJ https://raw.githubusercontent.com/zowe/zowe-cli/native-v$daemonVersion/zowex/Cargo.toml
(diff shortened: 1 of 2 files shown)


jenkins/bundleDaemon.sh
```diff
@@ -19,5 +19,8 @@ for platform in linux macos windows; do
     curl -fsLOJ https://github.com/zowe/zowe-cli/releases/download/native-v$daemonVersion/zowe-$platform.tgz
 done
 
+curl -fsLOJ https://raw.githubusercontent.com/zowe/zowe-cli/native-v$daemonVersion/zowex/Cargo.toml
+curl -fsLOJ https://raw.githubusercontent.com/zowe/zowe-cli/native-v$daemonVersion/zowex/Cargo.lock
```
t1m0thyj, line 23: In case we're concerned about rate-limiting for unauthenticated GitHub requests in the future, we could change the URLs to `https://$githubAuthHeader@raw.githubusercontent.com`, similar to line 8 of this script.

t1m0thyj (approved):
LGTM!

PR #1217, opened 2021-11-22
[next] Make credentials secure by default for v1 profiles

### LINKED ISSUE #1128: Zowe CLI vNext should be secure by default for v1 profiles
**Current Behavior**
1. Install `@zowe/cli@next` in a clean environment with no ~/.zowe folder
2. Create a profile with the `zowe profiles create` command
3. Notice that the credentials are stored in plain text

**Expected Behavior**
Secure credentials should be enabled in the scenario that:
* No team config files exist
* The Imperative setting CredentialManager=false does not exist

t1m0thyj:
Resolves #1128 by including zowe/imperative#690

### DIFF
--- a/.github/workflows/zowe-cli.yml
+++ b/.github/workflows/zowe-cli.yml
@@ -57,6 +57,10 @@ jobs:
       if: ${{ always() && steps.build.outcome == 'success' }}
       run: npm run test:unit >> file.txt
 
+    - name: Unlock Keyring
+      if: ${{ always() && steps.build.outcome == 'success' && matrix.os == 'ubuntu-latest' }}
+      uses: t1m0thyj/unlock-keyring@master
+
     - name: Integration Tests
       id: integration
       if: ${{ always() && steps.build.outcome == 'success' }}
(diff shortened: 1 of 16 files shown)


awharn (approved):

.github/workflows/zowe-cli.yml
```diff
@@ -57,6 +57,10 @@ jobs:
       if: ${{ always() && steps.build.outcome == 'success' }}
       run: npm run test:unit >> file.txt
 
+    - name: Unlock Keyring
+      if: ${{ always() && steps.build.outcome == 'success' && matrix.os == 'ubuntu-latest' }}
+      uses: t1m0thyj/unlock-keyring@master
```
zFernand0, line 62: @t1m0thyj 
Will this cause problems to the `nektos/act` scripts?
Maybe we can add an ID and skip/remove this step all together in the [testCliWorkflow.js](https://github.com/zowe/zowe-cli/blob/next/scripts/testCliWorkflow.js)

PR #1226, opened 2021-12-02
[next] Update SDK examples for vNext

### LINKED ISSUE #1141: Example in Zowe Node.js Client SDK documentation use deprecated code
For example, https://docs.zowe.org/stable/typedoc/modules/_zowe_zos_jobs_for_zowe_sdk.html has this code:

```
    ...
    const session: Session = ZosmfSession.createBasicZosmfSession(defaultZosmfProfile);
    ...
```

Which produces following warning:
```
The signature '(profile: IProfile): Session' of 'ZosmfSession.createBasicZosmfSession' is deprecated.ts(6387)
ZosmfSession.d.ts(48, 8): The declaration was marked as deprecated here.
```

What is the recommended alternative that is not deprecated?

t1m0thyj:
Fixes #1141 by updating SDK examples to use ProfileInfo API instead of the deprecated method `createBasicZosmfSession`.

There is one minor change required in Imperative for the SSH examples to work. The `ProfileInfo.initSessCfg` method needs to made public. If we're ok with this change, I can open a PR for it: https://github.com/zowe/imperative/compare/next...next-public-initsesscfg


packages/zostso/README.md
```diff
@@ -7,30 +7,18 @@ Contains APIs to interact with TSO on z/OS (using z/OSMF TSO REST endpoints).
...
-    let defaultTsoProfile: IProfile;
-    try {
-        defaultZosmfProfile = await getDefaultProfile("zosmf", true);
-        defaultTsoProfile = await getDefaultProfile("tso", false);
-    } catch (err) {
-        throw new ImperativeError({msg: "Failed to get a profile."});
-    }
-
-    const session: Session = ZosmfSession.createBasicZosmfSession(defaultZosmfProfile);
-    const accountNumber = defaultTsoProfile.account;
```
t1m0thyj, line 33: I stopped loading the TSO account number from a profile because the code to do is significantly more complex using the ProfileInfo API:
```typescript
const tsoProfAttrs = profInfo.getDefaultProfile("tso");
const tsoMergedArgs = profInfo.mergeArgsForProfile(tsoProfAttrs);
const accountNumber = tsoMergedArgs.knownArgs.find(arg => arg.argName === "account").argValue as string;
```

Curious what your thoughts are on this. Should I include the above code block in the example, or leave it out? Alternatively, is there a simpler way I've missed to load the account number?

zFernand0 (approved):
👍 🥳

packages/zostso/README.md
awharn: Maybe provide two examples? I can't really think of any other alternative.

gejohnston: Do we think that getting the TSO account is a common use-case for consumers of this API. If so, then we probably should show them how to accomplish it, even if it is involved. Maybe include a comment in the sample stating that getting the account is not a common practice?

gejohnston (dismissed):

gejohnston (approved):

zFernand0 (dismissed):

zFernand0 (approved):
LGTM!!!

awharn (approved):

PR #1228, opened 2021-12-03
[next] Get all tests passing with daemon mode on Ubuntu

t1m0thyj:
The current state of GitHub workflows is:
* Windows - _not_ using daemon binary 😢 so it passes
* Ubuntu, macOS - using daemon binary, lots of errors

This PR changes that status to:
* Ubuntu - using daemon binary, all tests pass 🚀 
* Windows, macOS - using daemon binary, still lots of errors (but all platform-specific ones)

Going forward, we can ensure that our tests pass in daemon mode 🙂

The following fixes were made:
* In daemon client, only check Node processes with at least 3 command line arguments (fixes Node 12 test failures).
* In daemon client, wrap empty string arguments in quotes (fixes a CLI integration test failure).
* In daemon client and server, pass current directory in the format `<currentDirectory>\r<zoweCommand>` where "\r" is a carriage return character. This moves the hardcoded `--dcd` argument down to a lower level, since the daemon client no longer knows about it, and the daemon server can now hide it from the Imperative command line (fixes several test failures).
* In test environment, put "daemon_instances" directory higher on PATH than "application_instances" so that daemon binary always has priority over NPM bin scripts on all platforms.

### DIFF
--- a/.github/workflows/zowe-cli.yml
+++ b/.github/workflows/zowe-cli.yml
@@ -80,7 +80,7 @@ jobs:
       if: github.event.inputs.test-type == 'binary' || github.event_name == 'push'
       run: |
         cargo build --verbose ${{ github.event.inputs.binary-type == 'release' && '--release' || '' }} --manifest-path=zowex/Cargo.toml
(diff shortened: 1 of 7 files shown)


zFernand0 (approved):
LGTM! 😋 
Thanks for adding the daemon_instances folder to the path 🚀

gejohnston (approved):

PR #1236, opened 2021-12-14
Update Imperative version and Changelog

### LINKED ISSUE #1106: zowe.config.json secure properties must not be logged and prompts should not show input
In vNext, users/admins can decide to store any value securely (not limited to user/password/tokens). Currently, when prompting for username/password, input is masked and these values are not logged or output to the terminal. I would expect this same behavior for all properties that are listed in the secure array.

zFernand0:
Upgrade Imperative so that secure prompts do not show input and zowe.config.json secure properties are not be logged.

### DIFF
--- a/__tests__/__packages__/cli-test-utils/package.json
+++ b/__tests__/__packages__/cli-test-utils/package.json
@@ -43,7 +43,7 @@
   },
   "devDependencies": {
     "@types/node": "^14.14.37",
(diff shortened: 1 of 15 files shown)


gejohnston (dismissed):

awharn (approved):

zFernand0:
Not sure why glob and rimraf are showing up in the package-lock still 😢

t1m0thyj (approved):

PR #1239, opened 2021-12-16
[next] Fix socket connection error on macOS in daemon mode

### LINKED ISSUE #1192: Daemon Mode - Socket connection error after every command [Mac-only]

### LINKED ISSUE #1237: Daemon fails with space in path on MacOs
From Joe Winchester:

When I had a space in the directory name containing the zowe executable it failed trying to do a chdir command so that may be because the directory name isn't escaped.

t1m0thyj:
Fixes #1192 by ignoring NotConnected errors when shutting down the daemon connection. These errors may happen on macOS when either half of the TCP stream is already closed. According to [Rust doc](https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.shutdown):
> Calling this function multiple times may result in different behavior, depending on the operating system. On Linux, the second call will return `Ok(())`, but on macOS, it will return `ErrorKind::NotConnected`. This may change in the future.

Adds a "Start Daemon" step before integration tests run on Windows in the GHA workflow. This step is necessary because Windows doesn't support forking processes. So when an `sh` script invoked by integration tests auto-starts the Zowe daemon, the script can't return until the daemon process has been terminated. Auto-starting the daemon as a background process before tests run works around this limitation.

Also removes the `bright` script from our application instances folder used in integration/system tests. Updating all the tests to invoke `zowe` means that they should run in daemon mode when the `zowe` daemon binary is at the top of the PATH.

Fixes #1237 (daemon fails with spaces in path) by wrapping `--dcd` path in quotes before passing it to `Imperative.parse`.

After merging this PR, integration tests should pass on all platforms using daemon mode 🎉 But system tests remain a WIP.


packages/zosmf/src/ZosmfSession.ts
```diff
@@ -82,11 +72,6 @@ export class ZosmfSession {
         group: ZosmfSession.ZOSMF_CONNECTION_OPTION_GROUP
     };
 
-    public static ZOSMF_OPTION_PASSWORD_PROFILE: ICommandOptionDefinition = {
```
zFernand0, line 85: should this breaking change be documented? 
(Same for the other porperties) 😋

zFernand0 (commented):
LGTM!
Thanks for updating all tests to use `zowe` 
Left one question

awharn (dismissed):
This looks good to me! Thanks for changing everything to use 'zowe'. 😄

awharn (approved):

zFernand0 (dismissed):
LGTM!

gejohnston (dismissed):

awharn (approved):

awharn (approved):

Issue #1179, opened 2021-10-26
Daemon mode - reliable data parsing


MikeBauerCA:
Pass commands & arguments in structures instead of text for more reliable data parsing. Confirm proper parsing on Windows (cmd, Powershell, gitBash) Linux, & Mac.

Spaces in parameters no longer require extra escape characters in daemon mode, but other special characters on the command line may not be correctly handled in daemon mode.

t1m0thyj:
We should also test how daemon mode handles large binary data (for example, piping a binary file larger than 1 MB).

Since stdin/stdout/stderr content is not being buffered into chunks, and in the case of stdout/stderr is being parsed as JSON, there could be some data loss or poor performance.

t1m0thyj:
Additionally, we should probably have better error handling, for example on the `JSON.parse` of data received from the daemon client 😛

MikeBauerCA:
Closing as resolved, if we feel additional work is needed then separate issues should be opened.

Issue #1241, opened 2021-12-17
Ctrl+C may kill daemon background process on Windows


t1m0thyj:
**Steps to Reproduce**
1. Ensure the daemon server is _not_ running
2. Run a command with the daemon client binary that prompts for input
    ```
    $ zowe jobs ls jobs
    Starting a background process to increase performance ...
    Enter the host name of your service (will be hidden):
    ```
3. Press Ctrl+C to terminate

**Actual Behavior**
Run `ps aux | grep zowe` and the daemon server is not running anymore.

**Expected Behavior**
The daemon server should still be running. Ctrl+C should only trigger an ECONNRESET without killing the background process. This is the behavior you see if you repeat the steps above but for `#1` ensure the daemon server _is_ running.

t1m0thyj:
This issue seems to be Windows-only. Since Windows doesn't fork processes, terminating the same Rust process that spawned the Node.js daemon server will terminate both the parent and child process simultaneously.

PR #1244, opened 2021-12-22
[next] 2-way JSON communication for daemon mode

### LINKED ISSUE #1215: ZOWE_OPT environment variables ignored in daemon mode
**Steps to reproduce**
1. Launch Zowe daemon in the background: `zowe --daemon &`
2. Run a command with the `zowex` binary that uses environment variables. For example:
    ```
    export ZOWE_OPT_USER=XXX
    export ZOWE_OPT_PASSWORD=XXX
    zowex jobs ls jobs
    ```
    OR in one line: `ZOWE_OPT_USER=XXX ZOWE_OPT_PASSWORD=XXX zowex  jobs ls jobs`
3. You are prompted to enter user & password even though they are defined in environment variables.

**Workarounds**
The environment variables must be defined before launching the `zowe --daemon` process, but this is not ideal for secure values like user & password that should persist for no longer than necessary.

### LINKED ISSUE #1240: ZOWE_CLI_HOME not used in daemon mode
When running zowe commands with the zowe.exe daemon running it does not use ZOWE_CLI_HOME to find the config files.
Running without the daemon uses ZOWE_CLI_HOME.

t1m0thyj:
ℹ️ For tests to pass, https://github.com/zowe/imperative/pull/714 must be published, and the Imperative version must be bumped in this PR.

Implement JSON communication for data sent from client-to-server, similar to what already exists for server-to-client.

This PR fixes #1215 ~~and may also fix #1240 (waiting on information from user who reported it to confirm)~~

## Old Implementation
* Sends commands in the following format containing cwd and args: `/home/dev/my_app\rzowe zosmf check status --host XXX`
* Sends prompt replies as a JSON object: `{ id: "daemon-client", reply: "XXX" }`
* Sends shutdown command as a string `--shutdown` (undocumented feature)
* Does not handle environment variables like `ZOWE_OPT_PASSWORD`
* Does not handle data piped to stdin (required by commands like `zowe files upload stds` and `zowe jobs submit stdin`)

## New Implementation
Sends the following JSON object for commands:
```
{
  argv: ["zosmf", "check", "status", "--host", "XXX"],
  cwd: "/home/dev",
  env: {
    ZOWE_OPT_PORT: "1443"
  },
  stdinLength: 0
}
```
* Argv is kept as an array in the Rust client rather than converting to a string, since yargs can parse either `string` or `string[]`
* Sends prompt replies as a JSON object: `{ stdin: "XXX" }`
* Sends shutdown command as a JSON object: `{ stdin: "\x03" }` (ASCII 3 = Ctrl+C)
* Includes environment variables prefixed with "ZOWE_" in the JSON object above, so env vars that define CLI options like `ZOWE_OPT_PASSWORD` will be loaded in daemon mode.
* Includes stdin data outside the JSON object so that binary data is supported. The `stdinLength` value is the number of stdin bytes, similar to the "Content-Length" header on REST requests. For example, to upload a file to a data set with the contents `Hello world!`:
```
{
  argv: ["zosmf", "files", "upload", "stds", "--host", "XXX"],
  cwd: "/home/dev",
  env: {
    ZOWE_OPT_PORT: "1443"
  },
  stdinLength: 12
}
Hello world!
```

**Limitations**
* Some env vars like `ZOWE_CLI_HOME` are only loaded when Imperative initializes, so they won't take effect if the background process has already been launched. If specified on the first Zowe command a user runs, they will be passed on to the daemon server that is auto-started.
* All the stdin data is sent in one chunk. We may want to investigate streaming it in the future, particularly for the use case of uploading large files. I have not yet tested what happens when the stdin data is very large.
* The way environment variables and stdin data are handled will need to be enhanced in the future when we support multitenancy. While worth noting, this is not a regression since the old implementation also didn't support multitenancy.


zowex/Cargo.toml
```diff
@@ -9,8 +9,9 @@ repository = "https://github.com/zowe/zowe-cli"
...
 
 [dependencies]
+atty = "0.2.14"
 pathsearch = "0.2.0"
 rpassword = "5.0.1"
 sysinfo = "0.21.1"
-serde_json  = "1.0.69"
-serde  = { version = "1.0.130", features = ["derive"]}
+serde_json = "1.0.69"
+serde = { version = "1.0.130", features = ["derive"]}
```
zFernand0, line 17: Before Version 0.5.0 gets included in the Zowe Release, the BoM may need updated.

packages/cli/src/DaemonClient.ts
```diff
@@ -67,48 +56,79 @@ export class DaemonClient {
...
      * Data event handler triggered for whenever data comes in on a connection
      * @private
      * @param {Buffer} data
      * @memberof DaemonClient
      */
     private data(data: Buffer) {
-
+        // Split JSON body and binary data from multipart response
         const stringData = data.toString();
+        const jsonEndIdx = stringData.indexOf("}\n");
```
zFernand0, line 112: What if `}\n` is part of the data coming back ?
Maybe reading a JSON file from USS ?

zowex/src/main.rs
```diff
@@ -129,59 +132,40 @@ fn main() -> std::io::Result<()> {
...
+    let response: DaemonResponse = DaemonResponse {
+        argv: Some(args.to_vec()),
+        cwd: Some(cwd.into_os_string().into_string().unwrap()),
+        env: Some(get_zowe_env()),
+        stdinLength: Some(stdin.len() as i32),
+        stdin: None,
+    };
+    let mut _resp = serde_json::to_string(&response)?;
+    if response.stdinLength.unwrap() > 0 {
+        _resp.push('\n');
```
zFernand0, line 156: Curious if this may need changed based on the previous comment.
Please disregard this comment if this is not where the `}\n` lookup is coming from.

zowex/src/main.rs
```diff
@@ -215,7 +199,7 @@ fn establish_connection(host: String, port: String) -> std::io::Result<TcpStream
                         "Command used to start the Zowe daemon was:\n    {}\nTerminating.",
                         cmd_to_show
                     );
-                    std::process::exit(EXIT_CODE_DEAMON_NOT_RUNNING_AFTER_START);
+                    std::process::exit(EXIT_CODE_DAEMON_NOT_RUNNING_AFTER_START);
```
zFernand0, line 202: 🤣

zowex/src/main.rs
```diff
@@ -570,6 +564,7 @@ fn start_daemon(njs_zowe_path: &str) -> String {
     let daemon_arg = "--daemon";
     match Command::new(njs_zowe_path.to_owned())
         .arg(daemon_arg)
+        .envs(get_zowe_env())
         .stdout(Stdio::null())
```
zFernand0, line 568: Just curious...
What happens if we do `.stdin(Stdio::null())` ?

zFernand0 (commented):
The implementation seems like a great improvement from before. 👍

zowex/Cargo.toml
gejohnston: When the next-daemon-enable branch is merged we will be at 0.6.0.

zowex/src/main.rs
```diff
@@ -129,59 +132,40 @@ fn main() -> std::io::Result<()> {
...
 
-/**
- * Convert a vector of command line arguments into a single string of arguments.
- * @param cmd_line_args
- *      The user-supplied command line arguments to the zowe command.
- *      Each argument is in its own vector element.
- * @returns
- *      A String containing all of the command line arguments.
- */
-fn arg_vec_to_string(arg_vec: Vec<String>) -> String {
```
gejohnston, line 140: The arg_vec_to_string function is now used in a second, unrelated place in the next-daemon-enable branch. So, it will be resurrected when that branch merges.

gejohnston (approved):

packages/cli/src/DaemonClient.ts
t1m0thyj: This method only parses data sent from client-to-server, which is limited to the properties listed above: argv, cwd, env, stdin, stdinLength.
As long as none of these values contain new-line characters, then `}\n` should never be present inside the JSON object.
Any data coming back to be written to stdout/stderr (like the contents of a USS file) are in the data that gets sent back from server-to-client which is not parsed here.
Perhaps I could change the separator to use a form-feed (`}\f`) rather than new-line if that seems more robust?

packages/cli/src/DaemonClient.ts
```diff
@@ -67,48 +56,79 @@ export class DaemonClient {
...
-                }
-            } catch {
-                // do nothing
+        if (jsonData.stdin != null) {
+            if (jsonData.stdin !== "\x03") {
+                // This data is related to a prompt reply so we ignore it
+                return;
+            } else if (this.mServer) {
+                // Ctrl+C was sent (ASCII 3) so we shutdown the server
+                this.shutdown();
```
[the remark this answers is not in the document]
t1m0thyj: When Ctrl+C is pressed while a prompt is active, it does not get sent as a Ctrl+C character to the daemon server.
Instead, the Rust process intercepts the Ctrl+C input and terminates.
The issue #1241 appears to be Windows-only and unrelated to this parsing logic.

zowex/src/main.rs
t1m0thyj: According to Rust doc for `Stdio::null`:
> This stream will be ignored. This is the equivalent of attaching the stream to `/dev/null`.

I believe we want to do this for stdout/stderr so they don't get written to the console twice, but for stdin it shouldn't be necessary.

packages/cli/src/DaemonClient.ts
zFernand0: It may be more consistent with the use of the form feed char to indicate termination in other areas of the daemon implementation 😋

zFernand0:
This PR may also fix https://github.com/zowe/zowe-cli/issues/1179

gejohnston (approved):

awharn (approved):

zFernand0 (approved):
LGTM!

gejohnston (approved):

Issue #1109, opened 2021-08-16
[next] zowe config auto-init command should respect rejectUnauthorized flag


awharn:
When you specify --ru on the `zowe config auto-init` command, the sessions all respect the rejectUnauthorized option, but the base profile does not possess the rejectUnauthorized property.

If --ru is specified, apply the option to the base profile properties if the base profile is created.

namanpatel1112:
Closing as associated PR is approved and merged.

PR #1258, opened 2022-01-10
Save rejectUnauthorized flag with `zowe auto-init` command

namanpatel1112:
Resolves https://github.com/zowe/zowe-cli/issues/1109


t1m0thyj (changes requested):
Thanks for fixing this, the code change looks good 🙂

Please update tests to verify that the CLI option `rejectUnauthorized` is handled by the auto-init command, and update the changelog to mention #1109 being fixed.

t1m0thyj (commented):
Thanks for adding tests and updating the changelog!

Could you please remove the changelogs for the "cli-test-utils" and "core" packages, since the files changed in this PR were only inside the "packages/cli" directory?

If the changelog check fails for any packages that you haven't modified, there's no need to update the changelog for those packages, and we should file it as a bug 🙂

namanpatel1112:
> Thanks for adding tests and updating the changelog!
> 
> Could you please remove the changelogs for the "cli-test-utils" and "core" packages, since the files changed in this PR were only inside the "packages/cli" directory?
> 
> If the changelog check fails for any packages that you haven't modified, there's no need to update the changelog for those packages, and we should file it as a bug 🙂

@t1m0thyj Alright, I removed those changes.

awharn (approved):

t1m0thyj (approved):

zFernand0 (approved):

PR #1263, opened 2022-01-11
Make EXE executable on Posix.

gejohnston:
Set the executable bit when 'zowe daemon enable' creates the EXE file.


gejohnston:
Confirmed the behavior on Linux with a build from source.

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Set executable attribute on zowe executable file on Windows and Mac.
```
awharn, line 7: Linux and Mac

awharn (commented):
One small comment. 😄

packages/cli/src/daemon/enable/Enable.handler.ts
```diff
@@ -192,8 +192,14 @@ export default class EnableDaemonHandler implements ICommandHandler {
                 .pipe(new tar.Parse())
                 .on('entry', function(entry: any) {
                     if (entry.type == "File" && (entry.path as string).includes(fileToExtract)) {
+                        const sysInfo: ISystemInfo = ProcessUtils.getBasicSystemInfo();
+                        let fileCreateOpts: any = {};
+                        if (sysInfo.platform == "linux" || sysInfo.platform == "darwin") {
```
t1m0thyj, line 197: Curious about this: 
`sysInfo.platform == "linux" || sysInfo.platform == "darwin"`
vs this:
`sysInfo.platform !== "win32"`

It probably doesn't matter, as I don't we think support of the other systems that may be returned by [process.platform](https://nodejs.org/api/process.html#processplatform): `aix`, `freebsd`, `openbsd`, `sunos`

packages/cli/CHANGELOG.md
gejohnston: Duh !!!
Thanks.

packages/cli/src/daemon/enable/Enable.handler.ts
awharn: That might be shorter. I just suggested that line because Windows is working right now. It might be that we don't even need to do a check and can just set the permission bits to 0o750 for every OS.

zFernand0 (approved):

awharn (approved):

t1m0thyj (commented):
Before I approve, not sure why some integration tests failed in the Jenkins pipeline?

gejohnston:
I don't know why the integration tests are failing either.  I just started looking into it.

awharn:
Well, we just set the executable bit, so maybe it thinks the zowe executable is a bash script now?

gejohnston:
I just now manually ran the integration tests successfully on Windows and in a Linux VM, so I still do not know why the Jenkins pipeline is unhappy.

awharn:
@gejohnston I just took a look at the logic for that test and noticed something. The Linux zowe "executable" (the NPM bin) is a symlink to main.js (which statSync.size says is 2963 because it follows the symlink). I don't know if that might be screwing up the integration test, but it might be a lead.

gejohnston:
I am not sure why these tests just started failing now, but I will increase the max size test for a script from 2000 to 5000 and see what happens.

awharn:
The only other thing I can think if is maybe we need to await rimraf or call rimraf.sync in the beforeEach?

gejohnston:
Once we set the -x bit, it changed the nature of the error that occurs when daemon-enable runs a fake EXE. The test-case failed due the change in error output. By removing a test that expects stderr to be empty, the tests pass. I do not know why the tests ran on Linux manually, but failed in the Jenkins pipeline. The test was unnecessary, so removing it seemed reasonable.

Now, both Github Actions and Jenkins are successful. This PR can be merged after it has been approved.

t1m0thyj (approved):

zFernand0 (approved):

awharn (approved):

PR #1271, opened 2022-01-12
[next] Fix daemon parsing of double-byte chars

t1m0thyj:
Signed-off-by: Timothy Johnson <timothy.johnson@broadcom.com>

### DIFF
--- a/packages/cli/__tests__/daemon/__unit__/DaemonClient.unit.test.ts
+++ b/packages/cli/__tests__/daemon/__unit__/DaemonClient.unit.test.ts
@@ -50,9 +50,9 @@ describe("DaemonClient tests", () => {
 
         const daemonClient = new DaemonClient(client as any, server, "fake");
         const daemonResponse: IDaemonResponse = {
-            argv: ["feed", "dog"],
+            argv: ["feed", "🐶"],
             cwd: "fake",
-            env: {},
+            env: { FORCE_COLOR: "0" },
             stdinLength: 0,
             stdin: null,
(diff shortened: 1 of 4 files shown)


packages/cli/__tests__/daemon/__unit__/DaemonClient.unit.test.ts
```diff
@@ -50,9 +50,9 @@ describe("DaemonClient tests", () => {
 
         const daemonClient = new DaemonClient(client as any, server, "fake");
         const daemonResponse: IDaemonResponse = {
-            argv: ["feed", "dog"],
+            argv: ["feed", "🐶"],
             cwd: "fake",
-            env: {},
+            env: { FORCE_COLOR: "0" },
```
awharn, line 55: Quick question, but is there a reason we are forcing color here, and not in the other tests?

gejohnston (approved):

packages/cli/__tests__/daemon/__unit__/DaemonClient.unit.test.ts
t1m0thyj: No particular reason, just wanted to set an environment variable in at least one test to make sure it shows up correctly in the snapshot.

The name/value of the env var doesn't matter here, and we could add it to all tests in this file if you want 🙂

awharn: No, I was just curious.

awharn (approved):

zFernand0 (approved):

PR #1272, opened 2022-01-12
[next] Add linting to Rust code

t1m0thyj:
* Add `cargo clippy` command to the build process for Rust EXE. Clippy is the default Rust linter shipped with `rustup`.
* Add "watch:exe" NPM script that runs `cargo watch` to automatically rebuild the binary when main.rs changes on disk.
* Replace placeholder unit test in Rust code with a meaningful one.
* Address all of Clippy's complaints. I've opened this PR in draft mode, so we can discuss whether or not we want these changes. 
  * Most of the changes suggested by the linter are stylistic and I don't see a problem with accepting them unless we dislike the style it enforces.
  * One change that stands out as more than just stylistic is `stream.write` -> `stream.write_all`. This seems like a good change that fixes a potential bug, based on the documentation of the linting rule [here](https://github.com/rust-lang/rust-clippy/blob/60e68d68c655ad0d29343fcc4c0e1deddc4d596d/clippy_lints/src/unused_io_amount.rs#L8-L18).

### DIFF
--- a/zowex/src/main.rs
+++ b/zowex/src/main.rs
@@ -11,7 +11,7 @@
 
 use std::collections::HashMap;
 use std::env;
-use std::ffi::OsString;
+
 use std::io;
 use std::io::prelude::*;
 use std::io::BufReader;
@@ -102,11 +102,9 @@ fn main() -> io::Result<()> {
     _args.drain(..1); // remove first (exe name)
 
     // Do we only need to display our version?
-    if _args.len() >= 1 {
-        if _args[0] == "--version-exe" {
-            println!("{}", env!("CARGO_PKG_VERSION"));
-            std::process::exit(EXIT_CODE_SUCCESS);
-        }
+    if !_args.is_empty() && _args[0] == "--version-exe" {
+        println!("{}", env!("CARGO_PKG_VERSION"));
+        std::process::exit(EXIT_CODE_SUCCESS);
     }
 
     // daemon commands that overwrite our executable cannot be run by our executable
@@ -161,7 +159,7 @@ fn main() -> io::Result<()> {
  *      The user-supplied command line arguments to the zowe command.
  *      Each argument is in its own vector element.
  */
-fn exit_when_alt_cmd_needed(cmd_line_args: &Vec<String>) {
+fn exit_when_alt_cmd_needed(cmd_line_args: &[String]) {
     // commands other than daemon commands can be run by our exe
     if cmd_line_args.len() < 2 {
         return;
@@ -192,7 +190,7 @@ fn exit_when_alt_cmd_needed(cmd_line_args: &Vec<String>) {
         // We use COMSPEC so that the command will work for CMD and PowerShell
         match env::var("COMSPEC") {
             Ok(comspec_val) => {
-                if comspec_val.len() > 0 {
+                if !comspec_val.is_empty() {
                     if comspec_val.contains(' ') {
                         zowe_cmd_to_show.push('"');
                         zowe_cmd_to_show.push_str(&comspec_val);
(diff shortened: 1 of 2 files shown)


zFernand0 (approved):
Thanks for updating the npm script 👍

awharn (approved):

zowex/src/main.rs
```diff
@@ -229,7 +227,7 @@ fn exit_when_alt_cmd_needed(cmd_line_args: &Vec<String>) {
  * @returns
  *      A String containing all of the command line arguments.
  */
-fn arg_vec_to_string(arg_vec: &Vec<String>) -> String {
+fn arg_vec_to_string(arg_vec: &[String]) -> String {
```
gejohnston, line 230: Since this function no longer takes a Vec type of parameter, should we change the name of the function?

zowex/src/main.rs
```diff
@@ -241,18 +239,18 @@ fn arg_vec_to_string(arg_vec: &Vec<String>) -> String {
...
         } else {
             arg_string.push_str(next_arg);
         }
 
-        arg_count = arg_count + 1;
+        arg_count += 1;
     }
 
-    return arg_string;
+    arg_string
```
gejohnston, line 253: I am not a fan of the practice of NOT having a 'return' statement and just placing a variable name as the last statement in a function. But if that is something which the Rust linter complains about, we should probably conform.

gejohnston (approved):
I just recorded a couple of opinions.

zowex/src/main.rs
t1m0thyj: The name for the Rust datatype `&[String]` seems to be "slice" instead of "vector".

I'm not sure what to rename the function to. I don't think `arg_array_to_string` or `arg_list_to_string` would be technically accurate. And `arg_slice_to_string` or `arg_collection_to_string` might be, but they sound kind of weird IMO.

If there is a name that stands out to you as a good option, I'm happy to change it. Otherwise I'm inclined to leave it as is, since we're still passing a `Vec` parameter into these methods.

t1m0thyj:
A tip for anyone developing with Rust in VSCode 😋 You can enable real-time linting for Rust by changing the following setting to "on" if you have the [Rust extension](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust) installed.
image

zowex/src/main.rs
t1m0thyj: The category of this linting problem is "style:warn". We could disable this specific warning, or perhaps all "style" warnings if we don't care about how "idiomatic" our Rust code is? 😛 
| Category              | Description                                                                         | Default level |
| --------------------- | ----------------------------------------------------------------------------------- | ------------- |
| `clippy::all`         | all lints that are on by default (correctness, suspicious, style, complexity, perf) | **warn/deny** |
| `clippy::correctness` | code that is outright wrong or useless                                              | **deny**      |
| `clippy::suspicious`  | code that is most likely wrong or useless                                           | **warn**      |
| `clippy::style`       | code that should be written in a more idiomatic way                                 | **warn**      |
| `clippy::complexity`  | code that does something simple but in a complex way                                | **warn**      |
| `clippy::perf`        | code that can be written to run faster                                              | **warn**      |
| `clippy::pedantic`    | lints which are rather strict or have occasional false positives                    | allow         |
| `clippy::nursery`     | new lints that are still under development                                          | allow         |
| `clippy::cargo`       | lints for the cargo manifest                                                        | allow         |

zowex/src/main.rs
gejohnston: I agree that we can leave it as it is.

gejohnston (approved):

zFernand0 (approved):

awharn (approved):

Issue #1270, opened 2022-01-12
Daemon-disable does not kill background daemon process on Linux & Mac


gejohnston:
The daemon disable command removes the zowe executable from the .zowe/bin directory, and is supposed to also kill the background daemon process: node ...../zowe (our script) --daemon

The background process is not killed on Linux and Mac. It is killed on Windows.

This is not particularly harmful, but it is incorrect.

awharn:
I think I found the culprit(s).

On my dev machine, I have an alias pointed at the main.js built from GitHub. If the daemon is launched from there, here is my output from find-process:

```
{
      pid: 1549857,
      ppid: 1369504,
      uid: 1000,
      gid: 1000,
      name: 'node',
      bin: 'node',
      cmd: 'node /home/awharn/GitHub/zowe-cli-next/packages/cli/lib/main.js --daemon'
    }
```

This may not be relevant to this issue in particular, but it would be nice if it worked in this kind of environment. However, when installing globally from NPM, if you run the daemon, you get this output from find-process:

```
{
      pid: 1255,
      ppid: 345,
      uid: 1000,
      gid: 1000,
      name: 'node',
      bin: 'node',
      cmd: 'node /home/jenkins/.nvm/versions/node/v16.13.2/bin/zowe --daemon'
    }
```

The regex [here](https://github.com/zowe/zowe-cli/blob/8f19e12f929c58e07f6c6887054efb07e62bcd41/packages/cli/src/daemon/disable/Disable.handler.ts#L112) does not match. Granted I am using NVM, but either way, it points to the file in the NPM bin directory, not the CLI node module's lib directory.

PR #1274, opened 2022-01-14
Next kill daemon

gejohnston:
This PR fixes a bug where the `daemon disable` command did not kill a running daemon on Linux and Mac.

- Closes issue: Daemon-disable does not kill background daemon process on Linux & Mac [#1270](https://github.com/zowe/zowe-cli/issues/1270)
- Added a message to `daemon enable` and `disable` telling the user to open a new terminal window.
  - This is done for bash/zsh to refresh its command-path-cache when a new zowe executable is created or deleted.
  - It will also be needed in the future when we add .zowe/bin to the user's PATH on any OS.
- Added prototype logic to automatically place .zowe/bin into the user's PATH.
  - This new code is actually the majority of the code in the PR. The actual bug fix code is small.
  - This new logic is mostly not called at this time. It is just in place so that it does not fall out of sync with the rest of the code, as we make bug fixes or other enhancements.
  - Since the add-to-PATH code is not called, it makes our code coverage numbers on `Enable.handler.ts` drop quite a bit.
    - This seems like a reasonable trade-off to have the new code in place when we want to implement the add-to-PATH functionality in the (near?) future.


packages/cli/src/daemon/doc/IAnswerQuestions.ts
```diff
@@ -0,0 +1,22 @@
...
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+/**
+ * Specifies wwhether a question can be asked of the user, and if not,
+ * what value should be used by default.
+ */
+export interface IAnswerQuestions {
```
t1m0thyj, line 16: I think the name `IAnswerQuestions` seems rather generic for an interface that is used specifically for daemon mode. Would it make sense to name this interface something more specific like `IDaemonEnablePrompt`? On the other hand, maybe the current name is fine since this interface isn't exported in the top level `index.ts`.

packages/cli/src/daemon/enable/Enable.handler.ts
```diff
@@ -207,4 +221,120 @@ export default class EnableDaemonHandler implements ICommandHandler {
...
+                // don't ask, just use default
+                answer = userQuestions.addBinToPath.defaultVal;
+            }
+
+            if (answer !== null && answer === "y" || answer === "Y") {
+                // user wants us to do it for him/her
+                if ( osPlatform === "win32") {
+                    userInfoMsg += await this.addZoweBinOnWindows(pathToZoweBin);
+                } else {
+                    userInfoMsg += this.addZoweBinOnPosix(pathToZoweBin);
```
t1m0thyj, line 268: Should this condition be commented out? It seems like currently if the user responds "y" to the prompt, on Linux we do nothing.

packages/cli/src/daemon/doc/IAnswerQuestions.ts
gejohnston: I agree with the name change. While in there, I will also simplify the structure. We only need to know if we are allowed to ask the user any questions. We do not need a 'can I ask' property for each possible question.

packages/cli/src/daemon/enable/Enable.handler.ts
gejohnston: I added the structure previously known as IAnswerQuestions, for two reasons:

1.  To enable us to more easily move the enableDaemon( ) function into an SDK if the need arises in the future. If ZoweExplorer ever needed this function, it would want to control whether or not we automatically prompt at the command line.
2. By having the EnableDaemonHandler.process( ) function call the enableDaemon( ) function with properties of `don't ask`  and `don't add-to-path`, the logic that you are pointing to does not have to be commented out, because it will never be run - we won't prompt and we will not attempt to add-to-path for Windows, Linux, or Mac (at least not yet).

I like not having to resurrect many snippets of commented code, when we decide to fully implement the add-to-path logic. With the current approach, we only have to change the structure created in the process( ) function. So, unless you have seen another flaw that I have overlooked in this technique, I would like to leave this piece of code the way that it is.

gejohnston: The only operation that these functions currently perform is to correctly tell the user to open a new terminal when that is required.

zFernand0 (approved):
LGTM!
Works fine on MacOS 😋

packages/cli/src/daemon/enable/Enable.handler.ts
t1m0thyj: Thanks for the explanation, LGTM 🙂

t1m0thyj (approved):

PR #1276, opened 2022-01-19
Make the user field on SSH profiles secure

### LINKED ISSUE #682: User field in SSH profile should be secure
The password field of an SSH profile is secure, but the user field is not.
https://github.com/zowe/zowe-cli/blob/f5d82918db5cce5aa8eec4a4c4fbf63bbcd49503/packages/imperative.ts#L172-L180

Changing a profile field from non-secure to secure could be a breaking change depending on how it is handled. We should consider making this change in a backwards compatible way, that could migrate an existing SSH profile with user stored insecurely to have it stored securely.

awharn:
Resolves #682 


t1m0thyj (approved):
Thanks for deprecating the redundant properties 🙂

t1m0thyj (approved):

Issue #1265, opened 2022-01-11
Jenkins check wrongly fails on missing changes in irrelevant changelogs


namanpatel1112:
When automated tests run on a new PR in zowe-cli repo, Jenkins check wrongly fails with 'missing header' errors in changelog files in packages that haven't been modified. It seems like a bug and needs to be investigated.
Example of this occurring is in this PR: https://github.com/zowe/zowe-cli/pull/1258/

MarkAckert:
Closing this issue, as we are migrating all of our Jenkins automation to Github actions. Thank you.

PR #1279, opened 2022-01-20
Shrinkwrap + Node:16.13.1 on Jenkinsfile + NPM:8 on GHA

zFernand0:


zFernand0:
To fix the Jenkinsfile issue we may need to update the version controller library:
https://github.com/zowe/zowe-cli-version-controller/blob/master/src/org/zowe/pipelines/nodejs/NodeJSPipeline.groovy#L979

packages/cli/package.json
```diff
@@ -41,6 +41,7 @@
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
   "scripts": {
+    "prepublishOnly": "npm i --package-lock-only && npm shrinkwrap",
```
gejohnston, line 44: I do not remember the addition of the prepublishOnly script in the various plugin PRs. Maybe I just overlooked it. If not, why here and not in the plugins? This is not an objection. I am just trying to understand.

gejohnston (approved):

packages/cli/package.json
zFernand0: In the case of the CLI, since we only want the `packages/cli` package in the monorepo to have an npm-shrinkwrap before publishing, we added the prepublishonly in order to create one based on the already installed dependencies thanks to the top level npm-shrinkwrap file 😋

t1m0thyj (approved):

awharn (approved):

zFernand0:
Is the `check-changelog` issue in the `pr-merge` build related to https://github.com/zowe/zowe-cli/issues/1265 ?
Or maybe I should just update all packages' changelog that have imperative as a regular dependency ? 😋

t1m0thyj:
> Is the `check-changelog` issue in the `pr-merge` build related to #1265 ? Or maybe I should just update all packages' changelog that have imperative as a regular dependency ? 😋

Should be ok to ignore the Jenkins changelog check in this case I believe, and trust the smarter GH workflow instead which ignores changes to dependencies in package.json 🙂

PR #1306, opened 2022-02-08
Republish to add optional dependencies to shrinkwrap

t1m0thyj:
Signed-off-by: Timothy Johnson <timothy.johnson@broadcom.com>


zFernand0 (approved):
should the shrinkwrap be included in the version-controller library so that the repo doesn't get out of sync when publishing new versions ?

awharn (approved):

PR #1308, opened 2022-02-08
Replace Daemon Communication Mechanism

awharn:
Changes the Daemon communication mechanism from port based communication to socket and named pipe based communication, allowing for the use of file permissions as an access control mechanism.


gejohnston (dismissed):
Looks good.

packages/cli/__tests__/daemon/__unit__/DaemonDecider.unit.test.ts
```diff
@@ -125,12 +131,195 @@ describe("DaemonDecider tests", () => {
...
         (daemonDecider as any).mParms = ["one", "two", "--daemon"];
-        process.env.ZOWE_DAEMON = testPort;
+        process.env.ZOWE_DAEMON = testSocket;
         (daemonDecider as any).initialParse();
-        expect((daemonDecider as any).mPort).toBe(parseInt(testPort, 10));
+        expect((daemonDecider as any).mSocket).toBe(process.platform === "win32" ? "\\\\.\\pipe\\" + testSocket : testSocket);
+
+    });
+
+    (process.platform !== "win32" ? it : it.skip)("should try to delete an existing socket on *nix", () => {
```
zFernand0, line 142: Would it be ok to group these into a `(process.platform !== "win32" ? describe : describe.skip)` ?
Not a required change though 😋

packages/cli/src/daemon/DaemonDecider.ts
```diff
@@ -142,28 +146,29 @@ export class DaemonDecider {
...
+                        this.mSocket = `\\\\.\\pipe\\${this.mSocket}`;
                     }
+                } else if (process.platform !== "win32") {
+                    this.mSocket = path.join(os.homedir(), ".zowe-daemon.sock");
+                } else {
+                    this.mSocket = `\\\\.\\pipe\\${os.userInfo().username}\\ZoweDaemon`;
                 }
-                Imperative.api.appLogger.debug(`daemon server port ${this.mPort}`);
+
+                Imperative.api.appLogger.debug(`daemon server socket ${this.mSocket}`);
```
zFernand0, line 171: Should we be a bit more generic here ?
instead of socket, maybe say Communication Mechanism (since we are using pipes for win32) ?

zFernand0 (dismissed):
LGTM

gejohnston (approved):

zFernand0 (approved):

gejohnston (dismissed):

zowex/src/main.rs
```diff
@@ -279,30 +290,48 @@ fn run_daemon_command(args: &mut Vec<String>) -> io::Result<()> {
         _resp.append(&mut stdin);
     }
 
-    // form our host, port, and connection strings
-    let daemon_host = "127.0.0.1".to_owned();
-    let port_string = get_port_string();
+    let mut tries = 0;
+    loop {
+        let socket_string = get_socket_string();
```
t1m0thyj, line 295: This could be moved outside of the loop, if we change the `establish_connection` method to take a pointer so that Rust doesn't think the variable passed into it has moved.
L296: `let mut stream = establish_connection(&socket_string)?;`
L326: `fn establish_connection(daemon_socket: &str) -> io::Result<DaemonClient> {`

zowex/src/main.rs
```diff
@@ -279,30 +290,48 @@ fn run_daemon_command(args: &mut Vec<String>) -> io::Result<()> {
...
+            Ok(result) => { return Ok(result); },
+            Err(ref e) if e.kind() == io::ErrorKind::ConnectionReset => {
+                if tries > THREE_MIN_OF_RETRIES {
+                    println!("Terminating after {} connection retries.", THREE_MIN_OF_RETRIES);
+                    std::process::exit(EXIT_CODE_TIMEOUT_CONNECT_TO_RUNNING_DAEMON);
+                }
 
-    let mut stream = establish_connection(daemon_host, port_string)?;
-    talk(&_resp, &mut stream)
+                println!("{} ({} of {})", "The Zowe daemon is in use, retrying", tries, THREE_MIN_OF_RETRIES);
```
t1m0thyj, line 305: Clippy (the Rust linter) doesn't like this line: `literal with an empty format string`

zowex/src/main.rs
```diff
@@ -279,30 +290,48 @@ fn run_daemon_command(args: &mut Vec<String>) -> io::Result<()> {
...
+                
+                // pause between attempts to connect
+                thread::sleep(Duration::from_secs(THREE_SEC_DELAY));
+                tries += 1;
+            },
+            Err(e) => { return Err(e); }
```
t1m0thyj, line 311: Instead of matching the `Ok` and `Err` halves of the Result type separately, do we want to combine them into one?
```
match talk(&_resp, &mut stream) {
    Err(ref e) if e.kind() == io::ErrorKind::ConnectionReset => {
        ...
    },
    result => { return result; },
}
```

zowex/src/main.rs
awharn: Done

zowex/src/main.rs
awharn: Done

zowex/src/main.rs
awharn: Done

t1m0thyj (dismissed):

.github/workflows/zowe-cli.yml
```diff
@@ -129,11 +129,16 @@ jobs:
       env:
         PATH: ${{ github.workspace }}\__tests__\__resources__\application_instances;${{ env.PATH }}
 
-    - name: Integration Tests
-      id: integration
-      if: ${{ always() && steps.build.outcome == 'success' }}
+    - name: Integration Tests (Native)
+      id: integration-native
+      if: ${{ always() && steps.build.outcome == 'success' && !(github.event.inputs.test-type == 'binary' || github.event_name == 'push') }}
```
zFernand0, line 134: Is the `|| github.event_name == 'push'` needed ?
I believe this means that it won't run on a branch push ?

.github/workflows/zowe-cli.yml
```diff
@@ -129,11 +129,16 @@ jobs:
...
       run: npm run test:integration >> integration-tests.txt
 
+    - name: Integration Tests (Daemon)
+      id: integration-daemon
+      if: ${{ always() && steps.build.outcome == 'success' && (github.event.inputs.test-type == 'binary' || github.event_name == 'push') }}
```
zFernand0, line 139: Is the `|| github.event_name == 'push'` needed here ?
Does this mean we want it to run on a branch push if it's not binary ?

zFernand0 (dismissed):
LGTM

.github/workflows/zowe-cli.yml
awharn: The goal of this is to get the native tests to run when the Start Daemon step has not run, and get the daemon tests to run when the daemon was built. I just took the conditional logic for the Build Binary if statement and inverted it (or at least tried to). If there's a better suggestion for accomplishing this behavior (or if my changes' behavior is wrong), I appreciate any suggestions.

zFernand0: I'm afraid I don't have any good suggestions until we see this workflow in action.
 I just wanted to understand how these conditions would affect the integration test 😋

t1m0thyj (dismissed):

zFernand0 (dismissed):

PR #1312, opened 2022-02-14
Fixed third "zowe zos-files create data-set" example

### LINKED ISSUE #1252: Example "zowe zos-files create data-set" command doesn't work
The following example command:
```
  - Create a data set with type LIBRARY:

      $ zowe zos-files create data-set NEW.DATASET --data-set-type LIBRARY
```

fails with the following error:
```
Command Error:
Expect Error: Specify the name of the data set to "allocate like" from.
```

JeffinSiby:
Fixes #1252; Added the '--like' flag to the example

image

Signed-off-by: JeffinSiby <jeffinsiby@gmail.com>

### DIFF
--- a/packages/cli/src/zosfiles/-strings-/en.ts
+++ b/packages/cli/src/zosfiles/-strings-/en.ts
@@ -48,7 +48,7 @@ export default {
                 EXAMPLES: {
                     EX1: "Create a data set with default parameters and like flag",
                     EX2: "Create a data set with default parameters and like flag and lrecl flag",
-                    EX3: "Create a data set with type LIBRARY"
+                    EX3: "Create a data set with default parameters and like flag and type LIBRARY"
                 }
             },
             DATA_SET_PARTITIONED: {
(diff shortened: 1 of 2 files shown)


zFernand0:
Some snapshots may need updated, but LGTM otherwise 😋

packages/cli/__tests__/zosfiles/__integration__/create/ds/__snapshots__/cli.files.create.dataset.integration.test.ts.snap
```diff
@@ -199,9 +199,10 @@ exports[`Create  Data Set should display the help 1`] = `
 
       $ zowe zos-files create data-set NEW.DATASET --like EXISTING.DATASET --lrecl 1024
 
-   - Create a data set with type LIBRARY:
+   - Create a data set with default parameters and like flag and
```
t1m0thyj, line 202: Thanks for updating the example to a working command 🙂 

I'm not sure if updating the example is the best fix for #1252, or if the command should be updated to support the original example.

IMO it's not intuitive for the `zowe zos-files create data-set` command to fail when the `--like` option isn't provided.

It would be nice if the command `zowe zos-files create data-set NEW.DATASET --data-set-type LIBRARY` worked, using default parameters only.

Thoughts about this @MikeBauerCA @zFernand0 ?

zFernand0: Instead of just a reaction I should've used some words 😋 

I do agree that the command should work without the `--like` parameter and get the default parameters based on type of DS being created.

PR #1313, opened 2022-02-14
Update imperative for property name conversion.

gejohnston:
Signed-off-by: Gene Johnston <Eugene.Johnston@broadcom.com>


t1m0thyj (approved):

gejohnston:
> 

Do you think that imperative being included in the Zowe 1.27.2 bundle means that this PR is unnecessary?
Should we just delete this PR?

t1m0thyj:
> Do you think that imperative being included in the Zowe 1.27.2 bundle means that this PR is unnecessary? Should we just delete this PR?

I think the changes in this PR are still worthwhile, maybe we merge it and abort the build if we think it doesn't need to be published right away.

zFernand0 (approved):

PR #1316, opened 2022-02-16
Fix Keytar not present in top level dependencies

### LINKED ISSUE #1314: Keytar not present in top level dependencies
Hi! 

When using zowe-cli version `7.0.0-next.202202041954`, keytar is not present in `node_modules` of the project root directory and this causes some errors when using, in my case, `@zowe/imperative` ProfileInfo API. We tried with different versions to test and verify this and I can say that in one of the next versions `7.0.0-next.202112281543` keytar is present.
To install keytar in my `node_modules` root directory solved the issues I had.

node -v 16.14.0
npm -v 8.3.1

@t1m0thyj is aware of this as he guided us thoroughly to find this issue

t1m0thyj:
Fixes #1314

For non-global installs of the CLI into projects that may have the Keytar dependency located outside of CLI node_modules, we shouldn't be copying Keytar prebuilds at install time.

Also simplifies our preinstall script copied from the Zowe v1 SCS plug-in by configuring `local-prebuilds` in a `.prebuild-installrc` file, rather than copying the prebuild TGZs into `node_modules/keytar/prebuilds`.


Jenkinsfile
```diff
@@ -63,7 +63,7 @@ node('zowe-jenkins-agent-dind') {
     ]
 
     // Initialize the pipeline library, should create 5 steps.
-    pipeline.setup(nodeJsVersion: 'v16.13.1')
+    pipeline.setup(nodeJsVersion: 'v16.13.1', npmVersion: '^8')
```
zFernand0, line 66: 🖤 🦆

zFernand0 (commented):
Changes LGTM

zFernand0 (dismissed):

gejohnston (dismissed):

PR #1327, opened 2022-03-01
Next background daemon enable

gejohnston:
This PR alters the Zowe EXE to run the daemon enable and disable commands in the background, rather than telling the user to run the Zowe Node.js script themselves. By making the EXE run the Node.js script in the background and immedietely exit, the Node.js script can  overwrite or delete the executable file.

The following is sample output when run in a windows CMD terminal window:

```
C:\Users\ej608771> zowe daemon enable
The 'daemon enable' command will run in the background ...

C:\Users\ej608771>
Wait to see the results below ...
Zowe CLI daemon mode is enabled.
Zowe CLI native executable version = 0.8.0

Manually add 'C:\Users\ej608771\.zowe\bin' to your PATH.
Otherwise, you will continue to run the classic Zowe CLI interpreter.

To run further Zowe commands, close this terminal and open a new terminal.
Now press ENTER to see your command prompt.

C:\Users\ej608771>
```


packages/cli/__tests__/daemon/__integration__/cli.zowe.exe.integration.ts
```diff
@@ -34,27 +35,35 @@ describe("Zowe native executable", () => {
...
 
-        // determine executable file extension for our current OS
-        let exeExt: string;
+        // determine executable file name for our current OS
+        const zoweRootDir: string = nodeJsPath.normalize(__dirname + "/../../../../..");
         const sysInfo: ISystemInfo = ProcessUtils.getBasicSystemInfo();
         if (sysInfo.platform == "win32") {
-            exeExt = ".exe";
+            zoweExeFileNm = "zowe.exe";
+        } else if (sysInfo.platform == "linux") {
```
awharn, line 43: Could this else if be dropped and just be the else?

packages/cli/__tests__/daemon/__integration__/disable/cli.daemon.disable.integration.ts
```diff
@@ -44,32 +43,26 @@ describe("daemon disable", () => {
         const sysInfo: ISystemInfo = ProcessUtils.getBasicSystemInfo();
 
         // form our tgz file name
-        let tgzFileName = "zowe-";
         switch (sysInfo.platform) {
             case "darwin": {
```
awharn, line 47: Doesn't need to be implemented, but you could combine the case statements now. 😄

awharn (approved):
These changes look good to me. Added a couple of comments on optional cleanup ideas. 😄

packages/cli/__tests__/daemon/__integration__/cli.zowe.exe.integration.ts
gejohnston: I think in both this case and your later comment, the code set another variable which was different for the different cases. After a refactoring which removed the second variable, I failed to notice the redundancy. These are good changes that I will implement.

packages/cli/__tests__/daemon/__integration__/disable/cli.daemon.disable.integration.ts
gejohnston: Same response as above.

packages/cli/__tests__/daemon/__integration__/cli.zowe.daemon.integration.suite.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+        // Form the path name to our executable.
+        zoweExeDirPath = nodeJsPath.normalize(zoweRootDir + "/zowex/target/release");
+        zoweExePath = nodeJsPath.resolve(zoweExeDirPath, zoweExeFileNm);
+        if (!IO.existsSync(zoweExePath)) {
+            zoweExeDirPath = nodeJsPath.normalize(zoweRootDir + "/zowex/target/debug");
+            zoweExePath = nodeJsPath.resolve(zoweExeDirPath, zoweExeFileNm);
+            if (!IO.existsSync(zoweExePath)) {
+                zoweExePath = "./NoZoweExeExists";
+            }
```
t1m0thyj, line 70: Is this code used? If not, should the `let zoweExePath` on line 27 be removed and replaced with `const zoweExePath` on line 64?

packages/cli/__tests__/daemon/__integration__/disable/cli.daemon.disable.integration.ts
```diff
@@ -44,32 +43,23 @@ describe("daemon disable", () => {
...
             }
             case "win32": {
-                tgzFileName += "windows.tgz";
                 exePath = "zowe.exe";
                 break;
             }
             default: {
-                tgzFileName += "unknownOs.tgz";
                 exePath = "exeForUnknownOs";
                 throw "cli.daemon.disable.integration.test.ts: beforeAll: " + sysInfo.platform + " is not a known OS.";
```
t1m0thyj, line 58: Not sure if it's desirable to hardcode the filename in the tests. When the file is renamed like in this PR it will become out-of-date unless we remember to change it. Feel free to disregard this comment since this is just an error message where the filename doesn't really matter 🙂

If we want to change this, I'd suggest either using the Node.js built-in variable `__filename` or just leaving the filename out if we trust Jest to show us which test file threw an error.

t1m0thyj: The above comment also applies to line 77 of "cli.daemon.enable.integration.ts", which I'm not able to comment on because its not part of the diff.

packages/cli/__tests__/daemon/__integration__/cli.zowe.daemon.integration.suite.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    afterAll(async () => {
+        // Remove any existing prebuilds directory
+        rimrafSync(prebuildsDir);
+    });
+
+    // run all of our daemon-related tests
+    require("./cli.zowe.exe.integration");
+    require("./enable/cli.daemon.enable.integration");
+    require("./disable/cli.daemon.disable.integration");
```
t1m0thyj, line 96: This is a clever way to combine multiple test suites into one 🙂 I'm curious if there are any downsides of having test files where the filename doesn't end in `.test.ts`. None come to mind immediately, since our `checkTestsCompile` and `lint:test` scripts process all files that are in `__tests__` directories.

awharn (dismissed):

zowex/src/main.rs
```diff
@@ -726,6 +767,289 @@ fn run_nodejs_command(cmd_line_args: &mut Vec<String>) -> Result<i32, i32> {
...
+    }
+
+    // add our wait message
+    for next_arg in script_wait_msg.split_whitespace() {
+        script_arg_vec.push(next_arg);
+    }
+    script_arg_vec.push("&&");
+
+    // make script delay so the EXE can exit
+    for next_arg in "ping 127.0.0.1 -n 1 >nul &&".split_whitespace() {
```
t1m0thyj, line 962: Why do we need a `ping` command here and how does it make the script delay? Sorry if I'm missing something obvious 😢

t1m0thyj (commented):
This is a great UX improvement for users of daemon mode 👍 Left a few comments.

packages/cli/__tests__/daemon/__integration__/cli.zowe.daemon.integration.suite.test.ts
gejohnston: I believe the code will be used. zoweExeDirPath is assigned the value "...release" and later the value "...debug". So, the two assignments to zoweExePath are both different. The intent is to use the "release" version of the EXE if it exists, and then fallback to the "debug" version of the EXE, to ensure that the test will also run on a developer's laptop.

However, I will make the declaration of the variables more local to their use. I also noticed that the assignment of "./NoZoweExeExists" was to a variable not used by the later tar command, so I changed that assignment. If no EXE is built during our build, I want the tar command to fail in a very obvious way.

packages/cli/__tests__/daemon/__integration__/disable/cli.daemon.disable.integration.ts
gejohnston: Good thought on the use of __filename.

packages/cli/__tests__/daemon/__integration__/cli.zowe.daemon.integration.suite.test.ts
gejohnston: I considered that checkTestsCompile should confirm that the xxx.non-test.ts files compile, which is desireable. I purposely wanted the required files to not be xxx.test.ts, since jest runs all xxx.test.files. That would mean that the tests run twice, once from the xxx.suite.test.ts file and once from themselves. Some of the individual test files would fail because they would not get some of the setup done in the suite.

Downsides:
- I had to duplicate some logic in the suite and again in the individual tests, because I did not want variables declared in the suite to be used within the individual test files without being declared in the individual test file. It would work, but it would be very confusing.
- You may not be able to run some of the individual test files directly with jest. The suite creates the tgz, so if an individual test file needs the tgz, that test will fail.
    - A decent workaround is that you can always run the suite file, but still specifiy the text of an "it" or "describe" statement to limit the test to one or a few tests.

zowex/src/main.rs
gejohnston: There is no native "sleep" command on Windows that runs in all Windows command shells. My web searches came up with ping as the best always-available command. The ping takes some time to perform its operation. The more request packets that you ask for, the longer the command takes. I read the each request takes about a second. Only doing one request (-n 1) seemed to reliably delay long enough for me. We could increase the value of the -n flag if we find that more time is needed to allow the EXE to exit at other sites.

awharn (approved):

packages/cli/__tests__/daemon/__integration__/cli.zowe.daemon.integration.suite.test.ts
t1m0thyj: The changes in 6a267aecaff36906501a9b44d75bbc5a5313e71a addressed my concerns about unused variables.

zowex/src/main.rs
t1m0thyj: Thanks for the explanation, Windows seems silly to not have a `sleep` command but this makes sense now.

t1m0thyj (approved):

PR #1345, opened 2022-03-20
Next daemon restart

gejohnston:
This PR adds the daemon restart command. The command will stop any running daemon and start a new daemon.

The Rust program was also significantly refactored to achieve the following goals:

- Simplified the main() function
    - Moved most of main's logic into a new function named run_zowe_command(), which determines whether to run a command directly, launch a process to run a command, or transmit a command to a running daemon.
    - Added a new function named run_restart_command().
- Enable the exit code of the Zowe EXE to be the exit code of the command run by the daemon.
    - Changed all run_xxx functions to return an exit code with which main() can then exit the process.
    - Eliminated all calls to panic. Now prints an error message and returns some form of error.
    - Removed a number of calls to exit() from functions, which instead return some form of error.
- Improved the readability of the Rust source
    - Moved all run_xxx functions, and daemon communication functions to the top of the file in the order that they are used. All utility functions now come later in the file. All utility functions are in alphabetic order.
    - This is now the order of functions in the file:
        ```
        main
        run_zowe_command
        run_restart_command
        run_nodejs_command
        run_delayed_zowe_command
        run_daemon_command
        establish_connection
        start_daemon
        talk
        arg_vec_to_string
        form_cmd_script_arg_vec
        form_win_cmd_script_arg_vec
        form_bash_cmd_script_arg_vec
        get_cmd_shell
        get_daemon_process_info
        get_lock_file
        get_lock_string
        get_nodejs_zowe_path
        get_socket_string
        get_zowe_env
        user_wants_daemon
        ```
- I acknowledge and apologize that this level of refactoring and movement of functions make the PR diffs really hard to read. I genuinely think that the changes will make the Rust source file easier to read and maintain going forward.
    - Most of the new functionality resides in the run_xxx functions. For review purposes, I suggest that it might be easier to just directly read the run_xxx functions from the new version of main.rs to evaluate if their behavior makes sense to you.
    - I strongly suggest that we merge this PR before any new Rust modifications are implemented. The merge efforts of not starting with this copy of the Rust code would be significant.
- I did not create any restart tests in Jest.
    - Our runCliScript test utility runs a spawnSync of a bash script, which would launch our EXE's restart command, which would in turn run an asynchronous background process that is our daemon. However, Jest seems to control the OS process during tests. Even though restart runs the daemon as an asynchronous background process, Jest seems to affect the sub-processes such that the Jest test waits for the daemon process to exit, and by design, the daemon does not exit. Thus, the Jest test "hangs" indefinitely.
    - As an alternative, I created restart tests within the Rust code's "test" section. It seems to work pretty well.
    - I added 'Test Rust **platform** EXE and zowe script' actions to the 'Zowe CLI' github workflow, since that pipeline has both our Rust EXE and Node.js Zowe.
    - I separated Rust tests into unit and integration tests, so that I could specify only unit tests in the 'Rust CLI CI' Github workflow (which does not have Node.js zowe available).
- I chose not to add a CHANGELOG entry as part of this PR.
    - This PR does not include the Node.js command definition for the restart command. Thus, no command or help is visible to an end user.
    - The Node.js command will be added in a second PR. That PR should have the CHANGELOG entry.


awharn (approved):
Changes look good to me. 👍

awharn (changes requested):
It ran fine the first time I tried running the restart command, but I noticed all it did was start the daemon (which may or may not be desired behavior if the daemon isn't running). The second time I tried to run the daemon, however, the binary panicked. This is running on Ubuntu via WSL2.

Backtrace:
```
Shutting down the running daemon ...
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("expected value", line: 1, column: 1)', src/main.rs:720:42
```

Stack Backtrace:
```
   0: rust_begin_unwind
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/panicking.rs:116:14
   2: core::result::unwrap_failed
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/result.rs:1690:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/result.rs:1018:23
   4: zowe::talk
             at ./zowex/src/main.rs:720:35
   5: zowe::run_daemon_command
             at ./zowex/src/main.rs:478:15
   6: zowe::run_restart_command
             at ./zowex/src/main.rs:227:31
   7: zowe::run_zowe_command
             at ./zowex/src/main.rs:190:16
   8: zowe::main
             at ./zowex/src/main.rs:154:11
   9: core::ops::function::FnOnce::call_once
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/ops/function.rs:227:5
```

gejohnston:
Not stopping a daemon when none is running (and just starting a new daemon) is the implemented behavior.

Regarding the panic, is it possible that the Node.js Zowe code was not from the same branch as the zowe EXE?

I will try a manual test on a Linux system not running WSL2.

awharn (approved):

gejohnston:
I found a way to catch this daemon communication error and display an error message instead of a panic crash. The command now runs like this:

```
    $ zowex daemon restart
    Shutting down the running daemon ...
    You may be running mismatched versions of Zowe executable and Zowe daemon.
    I/O error during daemon communication.
    Details = expected value at line 1 column 1
    Unable to communicate a command to the Zowe daemon.
    
    $ echo $?
    111

```

zowex/src/main.rs
```diff
@@ -408,15 +709,26 @@ fn talk(message: &[u8], stream: &mut DaemonClient) -> io::Result<()> {
...
                             continue;
                         },
-                        result => result.unwrap(),
+                        result => {
+                            match result {
+                                Ok(ok_val) => {
+                                    p = ok_val;
+                                },
+                                Err(err_val) => {
+                                    eprint!("You may be running mismatched versions of Zowe executable and Zowe daemon.\n");
```
t1m0thyj, line 727: Clippy (Rust linter) doesn't like this line: ``warning: using `eprint!()` with a format string that ends in a single newline``

t1m0thyj (approved):
LGTM! Left a comment about a lint warning I noticed when building the Rust code.

Typically once a file exceeds 1k lines of source code, it makes me question whether we should pull it apart into separate files. Although it could be a significant effort to split up `main.rs` with little reward, particularly if we're not familiar with the process of setting up and working with multi-file Rust projects. I'll leave this to your discretion since you have just refactored the code and may not deem it necessary.

zowex/src/main.rs
gejohnston: I forgot to run clippy once more after my last change. All I knew was that the next line of output after the eprint statement was concatenated onto the end of the eprint output line. Adding the newline to the eprint statement seemed like the obvious fix. I did not even know that 'eprintln' was a thing. This was a good thing to change.

gejohnston: I agree that the file size is big, but getting Rust to compile is so darn hard that I am leery of undertaking that now. We can keep the idea in mind when we undertake another significant change to Rust code. Maybe it could be done in the upcoming fix for detecting the right daemon?

t1m0thyj (approved):

PR #1347, opened 2022-03-21
Export Imperative SDK as namespace

t1m0thyj:
⚠️ Do not merge until https://github.com/zowe/imperative/pull/759 is published and consumed

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -162,10 +162,11 @@ node('zowe-jenkins-agent-dind') {
     // Check Vulnerabilities
     pipeline.checkVulnerabilities()
 
-    pipeline.checkChangelog(
-        file: "CHANGELOG.md",
-        header: "## Recent Changes"
-    )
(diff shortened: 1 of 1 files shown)


packages/cli/src/index.ts
```diff
@@ -20,3 +20,4 @@ export * from "@zowe/zos-tso-for-zowe-sdk";
 export * from "@zowe/zos-uss-for-zowe-sdk";
 export * from "@zowe/zos-workflows-for-zowe-sdk";
 export * from "@zowe/zosmf-for-zowe-sdk";
+export * as imperative from "@zowe/imperative";
```
zFernand0, line 23: Curious if we should do the same in master?

zFernand0 (approved):
LGTM

packages/cli/src/index.ts
zFernand0: `/facepalm` I thought I saw next as the base branch
🤦

PR #1353, opened 2022-03-25
Split Rust code into multiple files

gejohnston:
This PR splits the single rust main.rs source file into multiple source files to ease future development.

I made no functional changes in this PR, The new files contain the same logic and functionality as the original big main.rs file.

The current list of rust source files are now:

```
comm.rs
defs.rs
main.rs
proc.rs
run.rs
test.rs
util.rs
```


zowex/Cargo.lock
```diff
@@ -448,7 +448,7 @@ checksum = "712e227841d057c1ee1cd2fb22fa7e5a5461ae8e48fa2ca79ec42cfc1931183f"
 
 [[package]]
 name = "zowe"
-version = "0.9.0"
+version = "0.9.1"
```
t1m0thyj, line 451: If there are no user-facing changes to daemon behavior in this PR, perhaps there is no need to publish a new binary in the GitHub releases and we should wait to bump the version until a future PR?

t1m0thyj (approved):
Thanks for doing this, it should make maintaining the Rust code a much more pleasant experience in the future 👍 Left a comment regarding the daemon version.

zowex/Cargo.lock
gejohnston: That seems reasonable. I will change it back.

zFernand0 (approved):
LGTM!

awharn (approved):

PR #1360, opened 2022-03-30
Update apiml endpoints

### LINKED ISSUE #1339: `zowe auth login apiml` fails in V2 with HTTP status code 405
I am using the last @next version of the Zowe CLI:

```
$ zowe --version                                                                   
7.0.0-next.202203091358
```

with beta version of the Zowe API ML (Version 2.0.4-SNAPSHOT build # n/a).

The command fails:
```
zowe auth login apiml                                                                         
Enter the port number of your service: ******
Command Error:
z/OSMF REST API Error:
Rest API failure with HTTP(S) status 405

Error Details:
HTTP(S) error status "405" received.

Protocol:  https
Host:      *****.lvn.broadcom.net
Port:      ******
Base Path:
Resource:  /api/v1/gateway/auth/login
Request:   POST
Headers:   undefined
Payload:   undefined
```

awharn:
Updates the APIML endpoints in the CLI to the v2 APIML paths.
Resolves #1339


awharn:
To preemptively answer a question - yes, I did remove a system test, and that test was testing if the auth login command's prompt would timeout. It was waiting for the full 10 minutes. We can restore the test if requested, I just don't see the need to have a sequential test run for 10 minutes to test a timeout.

packages/cli/__tests__/auth/__system__/__scripts__/auth_login_apiml.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -e
 
-zowe auth login apiml
+echo "y" | zowe auth login apiml
```
t1m0thyj, line 4: Curious why this command wasn't prompting before but now it is? Or was this a long-standing bug in our system tests? 😢

t1m0thyj (approved):

packages/cli/__tests__/auth/__system__/__scripts__/auth_login_apiml.sh
awharn: The APIML creates a prompt and asks if the user wants to store their new profile on disk. This seems to be a long standing bug, because otherwise the prompt will continue for about 10 minutes until it times out.

zFernand0 (approved):
LGTM!
Do we need confirmation from Mike before proceeding with this change ?

PR #1362, opened 2022-04-01
Clean up plugins directory in cli-test-utils

t1m0thyj:
Adds code that was missing from the TestEnvironment class: https://github.com/zowe/zowe-cli-sample-plugin/blob/master/__tests__/__src__/environment/TestEnvironment.ts#L95-L98


zFernand0 (approved):
LGTM!

awharn (approved):

PR #1363, opened 2022-04-05
[next] Respect file encoding in USS tags for `files download uss`

### LINKED ISSUE #740: For USS files where the encoding is set, use this when retrieving files
**Problem**
USS tagging is not respected when transfering files

**Proposed solution**
Add a CLI command to retrieve the tags, e.g.
`zowe files chtag request "filename"`
that returns the tagging of the file.  

Also allow `zowe files download uss "filename"` to respect the tagging, so if the USS file is already tagged as ISO8559-1 then transfer it as binary.

Not sure if this should be a default behavior change and the always convert files is an error, or whether it should be a new option on `zowe files download uss "filename" -autocvt` or similar.

**Background**
USS allows files to be tagged which records the codepage used to store the data.  The screen shot belows shows the file ascii.txt that is tagged as ISO8559-1, ebcdic.txt tagged as IBM-1047 and untagged.txt which isn't tagged.

image

In the sample above ascii.txt contains ASCII encoded txt and ebcdic.txt contains EBCDIC text.  

When doing a retrieve of the file from z/OSMF it is possible to set a header to specify the datatype, for example in the first screen shot below the get request returns the ASCII data treating it as EBCDIC encoded

image

If the header X-IBM-Data-Type is set to anything non-null then conversion occurs

image

It is possible to retrieve the encoding of the file using the PUT API and specifying a request body of `{"request":"chtag","action":"list"}` and a Content-Type of application/json

image

t1m0thyj:
Resolves #740

Have tested with the following files:
| File Info | Generated `X-IBM-Data-Type` Header |
| --- | --- |
| `t IBM-850     T=on    ascii.txt` | text;fileEncoding=850 |
| `b binary      T=off   binary.txt` | binary |
| `t IBM-1047    T=on    ebcdic.txt` | text;fileEncoding=1047 |
| `t IBM-1147    T=on    euro.txt` | text;fileEncoding=1147 |
| `t ISO8859-1   T=on    latin1.txt` | binary |
| `m ISO8859-1   T=off   mixed.txt` | binary |
| `- untagged    T=off   untagged.txt` | text |
| `t UTF-8       T=on    utf8.txt` | binary |


packages/zosfiles/src/methods/utilities/Utilities.ts
```diff
@@ -107,14 +108,36 @@ export class Utilities {
...
+    public static async applyTaggedEncoding(session: AbstractSession, USSFileName: string, options: IOptions): Promise<void> {
+        const payload = { request: "chtag", action: "list" };
+        const response = await Utilities.putUSSPayload(session, USSFileName, payload);
+        const jsonObj = JSON.parse(response.toString());
+        if (Object.prototype.hasOwnProperty.call(jsonObj, "stdout")) {
+            const columns = (jsonObj.stdout[0] as string).trim().split(/\s+/);
+            // Tests if binary tag set
+            if (columns[0] === "b" || columns[1].startsWith("UTF-") || columns[1].startsWith("ISO8859-")) {
+                options.binary = true;
+            } else if (columns[1].startsWith("IBM-")) {
```
gejohnston, line 135: How confident are we that the text trailing "IBM-" always represents a valid encoding value for the request?

gejohnston (approved):

awharn (approved):

packages/zosfiles/src/methods/utilities/Utilities.ts
t1m0thyj: I was unable to find a table in IBM docs that provides a comprehensive list of allowed values for this field. However I believe this is a safe assumption to make, given that this seems to be a standard way of representing USS encoding. If you pass only a number to the chtag command (e.g., `chtag -tc 1147 <filename>`), the encoding field will be displayed as "IBM-1147" so the prefix gets added automatically. I suggest that if we discover more cases in the future which we don't yet support, a workaround is to use the `--binary` and `--encoding` options to manually set the encoding, and we can enhance our auto-detection method as these cases are discovered.

gejohnston: 👍

t1m0thyj: I have since found a list with a number of encoding names [here](https://www.ibm.com/docs/en/zos/2.5.0?topic=utilities-code-set-converters-supplied), and some of them include a letter at the end (IBM-932C and IBM-943G). So it seems unsafe to be calling `parseInt` on the remainder of the string after `IBM-`, and I'll need to fix the parsing to make it more robust. Thanks for bringing this up, it prompted me to do some more research 🙂

KevinLoesch1 (approved):

PR #1365, opened 2022-04-06
Next pid for daemon

gejohnston:
This PR enables the Zowe executable to detect if a running daemon belongs to the current user.

- If a daemon is already running but does not belong to the current user, the EXE will know to automatically launch a new daemon.
- When the EXE runs the 'daemon restart' command, it will stop the correct daemon when multiple daemons are running.

How it works:

- Upon daemon startup, our Node.js zowe script will create a file (associated with the current user) which contains the process ID of the daemon process.
- Upon daemon shutdown, our Node.js zowe script will delete that process ID file.
- When the Zowe EXE tries to detect a running daemon, it now additionally compares the process ID of each daemon with the process ID stored in the user's process ID file. Only if the process IDs match have we found the right daemon for this user.

Additional changes:

Since we are creating yet another daemon-related file on disk, I modified existing zowe environment variables to allow users to override default locations with fewer variables than would be required by adding another file and another environment variable.

Previously:

- ZOWE_DAEMON
  - Windows
    - Enabled the user to override the last 2 segments of the name of the communication pipe between the EXE and the daemon.
    - Default = \\.\pipe\the_users_name\ZoweDaemon
  - Linux & Mac
    - Enabled the user to override the complete path to the communication socket between the EXE and the daemon.
    - Default = /the_users_home_dir/.zowe-daemon.sock

- ZOWE_DAEMON_LOCK
  - Windows only
    - Enabled the user to override the complete path to a lock file used to synchronize requests to the daemon when the user simultaneously runs Zowe EXE commands from multiple command prompt terminals.
    - Default = /the_users_home_dir/.zowe-daemon.lock

New changes are:

- Removed ZOWE_DAEMON and ZOWE_DAEMON_LOCK
- ZOWE_DAEMON_DIR
  - All platforms
    - Enables the user to override the complete path to the directory that will hold daemon files related to this user.
    - Default = /the_users_home_dir/.zowe/daemon
    - Can contain these files:
      - daemon.sock
      - daemon.lock
      - daemon_pid.json

- ZOWE_DAEMON_PIPE
  - Windows only
    - Enables the user to override the last 2 segments of the name of the communication pipe between the EXE and the daemon.
    - Default = \\.\pipe\the_users_name\ZoweDaemon


KevinLoesch1 (approved):

awharn (approved):

packages/cli/src/daemon/DaemonUtil.ts
```diff
@@ -0,0 +1,46 @@
...
+     * @returns The absolute path to the daemon directory.
+     */
+    public static getDaemonDir(): string {
+        let daemonDir: string;
+        if (process.env?.ZOWE_DAEMON_DIR?.length > 0) {
+            // user can choose a daemon directory for storing runtime artifacts
+            daemonDir = process.env.ZOWE_DAEMON_DIR;
+        } else {
+            // our default location.
+            daemonDir = path.join(os.homedir(), ".zowe", "daemon");
```
t1m0thyj, line 35: Is it desired behavior that we ignore the ZOWE_CLI_HOME environment variable here? My assumption is yes, since having multiple environment variables (both ZOWE_CLI_HOME and ZOWE_DAEMON_DIR) impact the daemon directory could be confusing.

packages/cli/src/daemon/DaemonDecider.ts
```diff
@@ -118,6 +123,33 @@ export class DaemonDecider {
...
+        const pidForUser: IDaemonPidForUser = {
+            user: os.userInfo().username,
+            pid: process.pid
+        };
+
+        const pidFilePath = path.join(DaemonUtil.getDaemonDir(), "daemon_pid.json");
+        const pidForUserStr = JSON.stringify(pidForUser, null, 2);
+
+        try {
+            fs.writeFileSync(pidFilePath, pidForUserStr);
```
t1m0thyj, line 143: Have we considered setting permission bits for the PID file? Should we use `644` to prevent other users from writing to the file, or is it safe to inherit the default permissions of the `~/.zowe/daemon` directory?

t1m0thyj (approved):

packages/cli/src/daemon/DaemonUtil.ts
gejohnston: I considered using ZOWE_CLI_HOME. We have had discussions about placing plugins into a shared directory. Plugins live under ZOWE_CLI_HOME. It was unclear to me if ZOWE_CLI_HOME would be used for the new shared location, or would continue to point to user-specific data. It just seemed safer for now to ensure that the pid file (and other daemon files) be placed into a location that is specific to the user. For now, when ZOWE_CLI_HOME is left unset, it allows the daemon directory to live under the .zowe directory with all of the other Zowe stuff.

packages/cli/src/daemon/DaemonDecider.ts
gejohnston: We could force a 644 permission on the PID file. However, that would take some control out of the user's hands. The permissions of all new files is controlled by the UMASK setting. I am inclined to let the user determine how new files are created with the UMASK setting (via the umask command).

I could be convinced to explicitly set the file permissions ourselves if most people feel that is the better choice.

packages/cli/src/daemon/DaemonUtil.ts
gejohnston: By the way the existing files that I moved into the daemon directory were located in the user's $HOME directory, So, the new implementation is consistent with what we had before.

PR #1385, opened 2022-04-25
Export profile definition into their respective SDKs

zFernand0:
ℹ️ This should allow ZE to use only SDKs ℹ️


packages/cli/src/Constants.ts
```diff
@@ -9,7 +9,37 @@
...
+    BASE_OPTION_CERT_FILE,
+    BASE_OPTION_CERT_KEY_FILE,
+    BASE_OPTION_HOST,
+    BASE_OPTION_PASSWORD,
+    BASE_OPTION_PORT,
+    BASE_OPTION_REJECT_UNAUTHORIZED,
+    BASE_OPTION_TOKEN_TYPE,
+    BASE_OPTION_TOKEN_VALUE,
+    BASE_OPTION_USER
+} from "@zowe/core-for-zowe-sdk";
```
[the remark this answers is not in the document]
zFernand0: yeah, it will simplify the imports 😋 
Some times people may get more constants than what they really asked for, but that should be ok 😁

gejohnston (approved):

packages/cli/src/zostso/constants/ZosTso.constants.ts
```diff
@@ -1,91 +0,0 @@
...
-
-export const TSO_PROFILE_OPTIONS: ICommandOptionDefinition[] = [
-    TSO_OPTION_ACCOUNT,
-    TSO_OPTION_CHAR_SET,
-    TSO_OPTION_CODE_PAGE,
-    TSO_OPTION_COLUMNS,
-    TSO_OPTION_LOGON_PROCEDURE,
-    TSO_OPTION_REGION_SIZE,
-    TSO_OPTION_ROWS
-];
```
t1m0thyj, line 91: Is removing these exports a breaking change, and should we deprecate them instead?

t1m0thyj (changes requested):

packages/cli/src/zostso/constants/ZosTso.constants.ts
zFernand0: good catch.
fixed now

t1m0thyj: Thanks for fixing the breaking changes 🙂 I was thinking the old definitions (now deprecated) would stay in this file, and only the new definitions would be in the SDK. Up to you whether you want to make that change or leave this file deleted.

zFernand0: for some reason I thought I was editing this file to start with.
I think I got it right this time though

t1m0thyj (approved):

PR #1388, opened 2022-04-27
Remove the call to profiles.get() function

gejohnston:
This PR removes the call to the deprecated function commandParameters.profiles.get() which was made in ZosFilesBase.handler. The call was harmless because the results were unused, but it was also very misleading - it implied that the deprecated function should be called.

Removal of the call to that function has no external effects. It did require that about a million unit tests be modified.


KevinLoesch1 (approved):

zFernand0 (approved):
Changes LGTM! 😋

t1m0thyj (approved):

PR #1402, opened 2022-05-05
Add profiling instructions and clean up dev deps

t1m0thyj:
⚠️ Before merging, would like someone else to verify:
* That the `pprof` instructions work
* That the dependency updates don't break them
---
* Added instructions for profiling Zowe CLI with `pprof`, a profiling tool developed by Google. `pprof` can be used to profile multiple languages, including [Node.js](https://www.npmjs.com/package/pprof) and also [Rust](https://github.com/tikv/pprof-rs) so we could potentially use it with the daemon in the future.
* Cleaned up some dev dependencies in top level package.json:
  * Removed `codepage`, `fs-extra` and `uuid` which are not used at the top level
  * Added `mustache` which is required by Gulp tasks
* Cleaned up some dependencies in cli-test-utils package.json:
  * Changed `@types` packages to dev dependencies since we don't export those types
  * Removed `yocto-queue` dependency (added in https://github.com/zowe/zowe-cli/commit/e89303aca11bc70d62903588e1803f4893278b72 to workaround an npm bug that I think is now fixed)


zFernand0 (approved):
Should we consider deprecating `@zowe/perf-timing` and move this documentation to docs.zowe.org (maybe under the troubleshooting guide page) ?

I do believe this provides way more information out of the box without us having to implement timers with perf-timing in places that we believe take long times.

Thanks for documenting it! 😋 

BTW, I didn't need to do use `npx`
I just did `NODE_OPTIONS=--require pprof zowe -V` and here are the results 😍 
(Please don't laugh at my PC's performance 🤣 )
image

awharn (approved):

t1m0thyj:
> Should we consider deprecating `@zowe/perf-timing` and move this documentation to docs.zowe.org (maybe under the troubleshooting guide page) ?
> 
> I do believe this provides way more information out of the box without us having to implement timers with perf-timing in places that we believe take long times.

I do support the idea of deprecating `@zowe/perf-timing` in the future since as long as it remains unmaintained/unused. However, I don't see any reason to rush doing so, as perf-timing is referenced many places in our code and we've just started trying out `pprof`. Maybe once we have a vNext technical preview for Zowe v3, if we still think perf-timing is unnecessary we can deprecate it?

Regarding moving the documentation to docs.zowe.org, I'm not sure if it belongs there, since there are no references to `perf-timing` on the docs site that I could find. I also don't know if profiling will be very useful for most users, although maybe it would be for CLI plug-in developers. If we do want to add profiling content to the docs site, I suggest we track that as a new story 🙂

gejohnston (approved):

PR #1411, opened 2022-05-10
Add `zowe jobs delete old-jobs` command

### LINKED ISSUE #1285: zowe jobs delete old-jobs command
*This functionality is available in Broadcom's extended jobs plug-in. Broadcom has agreed to contribute this capability and any applicable code to core Zowe CLI as part of Zowe V2 LTS.*

In core, the command will be `zowe zowe-jobs delete old-jobs` with an alias of `oj`.

Command entry in `zowe jobs delete` help
old-jobs | oj - Delete multiple jobs in OUTPUT status

**Command help**

zowe › zos-jobs › delete › old-jobs

Delete (purge) jobs in OUTPUT status. Defaults to deleting all jobs owned by your user ID that are in output status.

Usage
zowe zos-jobs delete old-jobs [options]

Options
--prefix | -p (string)

Only delete jobs with job names that match this prefix. Defaults to deleting all jobs owned by your user ID that are in output status.

... standard profile help ...

Examples
Delete all of your jobs in output status with a job name starting with "ibmuser":

`zowe zos-jobs delete old-jobs -p "ibmuser*"`

t1m0thyj:
Resolves #1285 

The default behavior of the `delete old-jobs` command from Broadcom's external jobs plug-in has been preserved.

Added a `--max-concurrent-requests` option that allows z/OSMF requests to delete jobs to be made in parallel.


packages/cli/__tests__/zosjobs/__unit__/delete/old-jobs/__snapshots__/OldJobs.handler.unit.test.ts.snap
```diff
@@ -0,0 +1,253 @@
...
+    "retcode": "CC 0000",
+    "status": "OUTPUT",
+    "subsystem": "JES2",
+    "type": "job",
+    "url": "www.nowhere.com/restjobs/jobs",
+  },
+]
+`;
+
+exports[`delete old-jobs handler tests should delete all jobs using defaults sequentially 1`] = `"Successfully deleted 2 jobs"`;
```
zFernand0, line 85: It would be nice to move away from using snapshots for things like this.
But that will be something we need to do across the board (if we think it's a good idea) 😋

packages/cli/src/zosjobs/delete/old-jobs/OldJobs.handler.ts
```diff
@@ -0,0 +1,82 @@
...
+            if (job.status === JOB_STATUS.OUTPUT) {
+                deletedJobs.push(job);
+            }
+        }
+        const deleteJobPromise = (job: IJob) => DeleteJobs.deleteJobForJob(this.mSession, job);
+        if (this.arguments.maxConcurrentRequests === 0) {
+            await Promise.all(deletedJobs.map(deleteJobPromise));
+        } else {
+            await asyncPool(this.arguments.maxConcurrentRequests, deletedJobs, deleteJobPromise);
+        }
```
zFernand0, line 67: I love the use of `Promise.all(...)` here 😋 
Question, should we get the API response and provide that as part of the `data.setObj` ?
Not sure if someone might be interesting in scripting based off of the JSON response coming back from this command.

zFernand0 (approved):
LGTM!

packages/cli/src/zosjobs/delete/old-jobs/OldJobs.handler.ts
t1m0thyj: I wanted to implement it this way, as returning the API response from the `DeleteJobs` API would seem better than from the `GetJobs` API. However, the default version of the `modifyVersion` parameter here is `1.0` meaning that the z/OSMF request to delete the job is asynchronous so the API response is undefined.

This is how it was implemented in the original plug-in. I don't know if we'd want to use `modifyVersion = 2.0` instead to obtain an API response, or if that would have drawbacks like requiring a minimum z/OSMF version?

On the other hand, if the z/OSMF request to delete the job is asynchronous, does that mean we don't need to provide a `--max-concurrent-requests` option ourselves? I think it may still be beneficial as I did see some performance improvements with it.

gejohnston (approved):

awharn (approved):

KevinLoesch1 (approved):

PR #1417, opened 2022-05-12
Add the `jobs view all-spool-content` command

### LINKED ISSUE #946: View all spool content of a particular job
We added the `--view-all-spool-content` flag to the `zowe jobs submit ds` command. However, if I want to view all of the spool content of an already completed job, I need to first issue a `zowe jobs list spool-files-by-jobid` command and then run a `zowe jobs view spool-file-by-id` command for each spool file. This enhancement request is for a `zowe jobs view all-spool-content` command that requires a `jobid` and outputs all of the spool content for that particular job.

awharn:
Resolves #946


zFernand0 (approved):

t1m0thyj (approved):
LGTM! Thanks for fixing the bug in original implementation where only the last item in the array was being returned by `--rfj` 🙂

PR #1421, opened 2022-05-16
Remove 'zowe files uf <file> --record' option

KevinLoesch1:
Signed-off-by: KevinLoesch1 <kevin.loesch@broadcom.com>


awharn (approved):

gejohnston (approved):
How dumb that we had the right text in the 'strings' file, but we did not correctly get them in the command definitions. 😒  At least it made the fix easier.👍

t1m0thyj (approved):

PR #1435, opened 2022-06-06
Contribute `zowe files download dataset-matching` command

awharn:
Sample command output:
```
15 data set(s) downloaded successfully to ./
298 data set(s) failed to download:
206 failed because they are archived
    IBMUSER.TEST.DS.ARCHIVED
    ...
15 failed because they are an unsupported type
    IBMUSER.TEST.ZFS
    IBMUSER.TEST.ZFS.DATA
    ...
77 failed because of an uncaught error
    IBMUSER.TEST.DS.BACKUP
    ...

z/OSMF REST API Error:
Rest API failure with HTTP(S) status 500
rc:       4
reason:   3
details:
  - DMS2987 DATA SET CATALOGED TO CA DISK PSEUDO-VOLUME ARCIVE
  - DMS2987 CA DISK HAS ARCHIVED IBMUSER.TEST.DS.BACKUP
  -
  - DMS2971 DO YOU WANT TO RESTORE THE DATA SET? (Y/N)
category: 2
message:  ServletDispatcher failed - received TSO Prompt when expecting TsoServletResponse

...
```


packages/zosfiles/src/methods/download/Download.ts
```diff
@@ -275,6 +290,248 @@ export class Download {
...
+            // Check if data sets matching pattern found
+            if (zosmfResponses.length === 0) {
+                return {
+                    success: false,
+                    commandResponse: ZosFilesMessages.noDataSetsMatchingPattern.message,
+                    apiResponse: []
+                };
+            }
+
+            // Exclude archived data sets
```
gejohnston, line 345: I could be overlooking other logic, but since the 'archived' test comes before the 'exclude' test, might we report that we skipped an archived dataset which the user has already told us to exclude? It seems like excluded datasets would not have to checked for other conditions.

t1m0thyj:
Refactored the APIs based on feedback from team discussion at standup. Before updating all the tests, I'd appreciate reviews to confirm that this implementation matches what we want. 🙂 @gejohnston I believe your comment has also been addressed.

To summarize the new implementation, there are now 2 methods:
* `List.dataSetsMatchingPattern` - Returns an array of data set objects retrieved from the z/OSMF List API. If listing attributes for all matching data sets at once fails for some reason, we fall back to the much slower alternative of listing attributes for each data set individually, and store an error in each data set object for which we failed to get attributes.
* `Download.allDataSets` - This method has been renamed from `Download.dataSetsMatchingPattern`, and now takes in an array of `IZosmfListResponse` data set objects returned from the list API instead of DSLEVEL pattern strings. Instead of reporting data sets as being skipped, the new behavior is:
  * For archived data sets, they are treated as a failure and the names of the data sets are printed.
  * For unsupported types, they are also treated as a failure and the names of the data sets are printed.
  * For excluded patterns, they are excluded by the list API and the download API does not care about them.
  * For empty PO data sets, an empty folder with the data set name is created.

awharn:
I think those all look right, and I think the linked changes all look good. Going to be quite a few tests that need to be updated.

packages/zosfiles/src/methods/download/Download.ts
gejohnston: I think I now see how datasets would be excluded by the datasetMatching handler before this function does its tests.

gejohnston (approved):

packages/zosfiles/src/methods/list/doc/IDsmListOptions.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * This interface defines the options that can be sent into the list data sets matching function
+ */
+export interface IDsmListOptions extends IZosFilesOptions {
+    /**
+     * Exclude data sets that match these DSLEVEL patterns. Any data sets that match
+     * this pattern will not be listed
+     * @example "ibmuser.**.jcl, ibmuser.rexa.*"
+     */
+    excludePatterns?: string[];
```
t1m0thyj, line 24: Should we deprecate the `excludePatterns` property on the `IDownloadOptions` interface? I don't think it is being used anywhere there, and having it there may be misleading since now we have it on this interface instead.

gejohnston (approved):
Looks good.

zFernand0:
FWIW, the pieces (code + tests) that I didn't work on, LGTM! 😋 
Happy to formally approve it if we want to move forward 😉

t1m0thyj:
The Windows builds have started to fail because of a bug in Rust 1.62 (https://github.com/rust-lang/rust/issues/98947). For now we can pin the version of Rust in the pipeline to 1.61 to avoid the issue.

PR #1450, opened 2022-06-15
Files compare datasets terminal

KevinLoesch1:
This pull request adds the new "zos-files compare data-set" command to the CLI. It can be used to compare two datasets and will display the differences to the terminal. 

Below is a sample output of comparing two dataset members with the **noseqnum** option:

image


awharn (commented):
These changes look good, but we may need more tests for coverage if we are aiming for the Gold CI badge.

KevinLoesch1:
It think the testing coverage is an easy fix. I think it is just skipping the code to strip the sequence numbers. If I add that to the test it should reach 100% coverage.

packages/cli/src/zosfiles/compare/ds/Dataset.handler.ts
```diff
@@ -0,0 +1,116 @@
...
+*
+*/
+
+import { AbstractSession, IHandlerParameters, ITaskWithStatus, TaskStage, TextUtils } from "@zowe/imperative";
+import { Get, IZosFilesResponse } from "@zowe/zos-files-for-zowe-sdk";
+import { ZosFilesBaseHandler } from "../../ZosFilesBase.handler";
+import { diff } from "jest-diff";
+
+/**
+ * Handler to view a data set's content
```
gejohnston, line 18: Should this say something about comparing two datasets?

gejohnston (approved):
Just a minor comment about a comment.

packages/cli/src/zosfiles/compare/ds/Dataset.handler.ts
KevinLoesch1: I spotted that yesterday. I have it fixed on my local drive, I will push the change.

packages/cli/__tests__/zosfiles/__integration__/compare/ds/__scripts__/command_compare_data_set_help.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -e
+
+echo "================Z/OS FILES LIST DATA SET HELP==============="
```
t1m0thyj, line 4: `LIST DATA SET` -> `COMPARE DATA SET`

packages/cli/__tests__/zosfiles/__integration__/compare/ds/__scripts__/command_compare_data_set_help.sh
```diff
@@ -0,0 +1,13 @@
...
+zowe zos-files compare ds --help
+if [ $? -gt 0 ]
+then
+    exit $?
+fi
+
+echo "================Z/OS FILES LIST DATA SET HELP WITH RFJ==========="
```
t1m0thyj, line 11: `LIST DATA SET` -> `COMPARE DATA SET`

packages/cli/src/zosfiles/compare/Compare.options.ts
```diff
@@ -0,0 +1,139 @@
...
+        type: "boolean"
+    },
+
+    /**
+     * The contextlines option
+     * @type {ICommandOptionDefinition}
+     */
+
+    contextlines: {
+        name: "contextlines",
```
t1m0thyj, line 134: Should the name be changed to include a hyphen (`context-lines`) to be consistent with other CLI option names?

packages/cli/src/zosfiles/compare/Compare.options.ts
```diff
@@ -0,0 +1,139 @@
...
+        type: "string"
+    },
+
+    /**
+     * The noseqnum option
+     * @type {ICommandOptionDefinition}
+     */
+
+    noseqnum: {
+        name: "noseqnum",
```
t1m0thyj, line 122: Should the name be changed to `seqnum` with a default value of false? I believe this would be consistent with how we handle boolean options elsewhere in the CLI. The "yargs" argument parser will automatically convert `--no-seqnum` to `--seqnum false`.

packages/cli/__tests__/zosfiles/__system__/compare/ds/Compare.ds.system.test.ts
```diff
@@ -0,0 +1,77 @@
...
+import { getUniqueDatasetName} from "../../../../../../../__tests__/__src__/TestUtils";
+
+let REAL_SESSION: Session;
+let testEnvironment: ITestEnvironment<ITestPropertiesSchema>;
+let defaultSystem: ITestPropertiesSchema;
+let dsname: string;
+const TIMEOUT = 20000;
+
+
+describe("View Data Set", () => {
```
t1m0thyj, line 27: The `view data-set` command is mentioned several times throughout this file in the test and script names. Is this intentional, or should it be referring to `compare data-set`?

packages/cli/__tests__/zosfiles/__unit__/compare/ds/Dataset.handler.unit.test.ts
```diff
@@ -0,0 +1,108 @@
...
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import { Get } from "@zowe/zos-files-for-zowe-sdk";
+import { UNIT_TEST_ZOSMF_PROF_OPTS } from "../../../../../../../__tests__/__src__/mocks/ZosmfProfileMock";
+
+describe("View data set handler", () => {
```
t1m0thyj, line 15: `View data set` -> `Compare data set`

KevinLoesch1:
> These changes look good, but we may need more tests for coverage if we are aiming for the Gold CI badge.

The code that is outside the coverage is going to be moved to a common routine in imperative.

t1m0thyj (commented):
Thanks for addressing my earlier comments 🙂 

I resolved most of them, but there are a few comments still applicable.

t1m0thyj (approved):
LGTM!

There are a few comments above which have not been addressed, but they are merely text in comments/test cases so nothing user facing.

PR #1452, opened 2022-06-17
use prepack stage to update licenses in /packages

MarkAckert:
Signed-off-by: MarkAckert <mark.ackert@broadcom.com>

Adds a LICENSE file to each package in the npm `prepack` stage. This stage runs as part of the lifecycle for both the `npm pack` and `npm publish` commands.  `.gitignore` is updated to prevent the package/*/LICENSE files from being added to the repo accidentally; they should always be downstream from the root LICENSE file.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "7.2.3",
+  "version": "7.2.4",
   "command": {
     "publish": {
       "ignoreChanges": [
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -24963,7 +24963,7 @@
     },
     "packages/cli": {
       "name": "@zowe/cli",
-      "version": "7.2.3",
+      "version": "7.2.4",
       "hasInstallScript": true,
       "license": "EPL-2.0",
       "dependencies": {
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## `7.2.4`
+
+- BugFix: Fixed the Zowe Daemon binary exiting with an error if the daemon server does not start within 3 seconds.
+
 ## `7.2.3`
 
 - BugFix: Updated Imperative to address `ProfileInfo` related issues.
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zowe/cli",
-  "version": "7.2.3",
+  "version": "7.2.4",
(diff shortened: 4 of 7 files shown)


MarkAckert:
I chose the `prepack` stage initially, and it's roughly equivalent to `prepublishOnly`. There's a bit of code in the deploy logic which targets prepublishOnly here: https://github.com/zowe/zowe-cli-version-controller/blob/d9665e6a1071f846b3f749c9f5c9924c9349ede7/src/org/zowe/pipelines/nodejs/NodeJSPipeline.groovy#L755 .

 Looks like it rips out the script and runs it separately and I'm not familiar with why its doing this. Any opinion on keeping `prepack` or preferring `prepublishOnly` ?

zFernand0:
No objections with using `npm prepack` 

I don't think it will be a problem to do this when installing git dependencies as well...

>Runs BEFORE a tarball is packed (on "npm pack", "npm publish", and when installing a git dependencies).

More info: https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-pack

--- 

About...

> Looks like it rips out the script and runs it separately

iirc, we wanted to run it before updating the shrinkwrap file since it's possible that a package (like the [zowe-cli](https://github.com/zowe/zowe-cli/blob/master/packages/cli/package.json#L45) needed to dedupe dependencies 😋 

I don't remember the full context though 😅

zFernand0 (approved):
LGTM!

t1m0thyj (commented):
Thanks for adding the `prepack` script 🙂

I'm not sure this will have any effect in the top-level package.json, because when we publish new packages we run `npm pack` inside the `packages/*` directories.

I pulled down the branch and running `cd packages/core && npm pack` produces a TGZ that doesn't contain the license.

We may need to add the `prepack` script to the package.json file for each of the packages inside the monorepo?

awharn (approved):
LGTM!

package.json
```diff
@@ -36,7 +36,8 @@
     "typedoc": "typedoc",
     "typedoc:packages": "lerna run --parallel typedoc",
     "audit:public": "npm audit --registry https://registry.npmjs.org/",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "prepack": "node scripts/prepareLicenses.js"
```
t1m0thyj, line 40: Should the `prepack` script be removed from here, to remind us not to run `npm pack` at the top level?

t1m0thyj (commented):
There's one more package in `__tests__/__packages__/cli-test-utils` that needs the `prepack` script - sorry that's its hidden in a weird place 😢

package.json
t1m0thyj: The `npm publish` command in the Jenkins pipeline only runs in the lower-level package directories, not at the top level, but it shouldn't hurt anything to leave the prepack script here.

t1m0thyj (approved):

PR #1453, opened 2022-06-20
Fix daemon binary startup error

awharn:
Patch Release
Fixes Daemon Binary error if the daemon server does not start within 3 seconds.
Updates error messaging to provide more accurate information during the 3 minute connection loop.


gejohnston (approved):
Looks good. As soon as we publish, we can update the NatWest issue with the availability of the fix and versions.

zFernand0 (approved):
Changes LGTM!
`👍` for Gene's opinion 👆

t1m0thyj (approved):

PR #1473, opened 2022-07-23
Files comparison localfiles datasets

sarthakjdev:
This PR resolves [Issue#1444](https://github.com/zowe/zowe-cli/issues/1444).

Added a command ```zos-files compare local-file-data-set``` to compare the differences between a local file and a dataset.

Sample command: ```zowe files cmp lds "./package.json" "mentee5.public.jcl(test)"  ``` 

It will take two arguments first will be the local file path either relative or absolute and the second one is the data-set name.


packages/workflows/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS workflows SDK package will be documented in this file.
 
+## Recent Changes
```
gejohnston: Should the changelog be updated in zosfiles package instead of the workflows package?

gejohnston (changes requested):

packages/workflows/CHANGELOG.md
sarthakjdev: Hey @gejohnston , 
I have rectified this thing, pls have a look at the updated one.

zFernand0:
Seems like we might be missing some coverage for the handler itself.

Since we plan to combine all Compare handlers into a common CompareBaseHandler in the near future, that's why I'm not too worries about it 😋

zFernand0 (approved):
LGTM! 😋

gejohnston (approved):

packages/cli/src/zosfiles/compare/lds/LocalfileDataset.definition.ts
```diff
@@ -0,0 +1,66 @@
...
+// Does not use the import in anticipation of some internationalization work to be done later.
+const strings = (require("../../-strings-/en").default as typeof i18nTypings).COMPARE;
+
+/**
+ * Compare local files and data sets command definition containing its description, examples and/or options
+ * @type {ICommandDefinition}
+ */
+export const LocalfileDatasetDefinition: ICommandDefinition = {
+    name: "local-file-data-set",
+    aliases: ["lds"],
```
t1m0thyj, line 25: Should the alias be `lfds`?

sarthakjdev: yes, `lfds` sounds more defined for the command let me change it. 
Thanks!

Issue #1445, opened 2022-06-13
CLI diff of USS files in terminal/browser


zFernand0:

Rahulkhinchi03:
Hi @zFernand0,
I'm Rahul Khinchi, applied to the LFX Program at Zowe CLI.

Until then, I am learning ZOWE CLI and would like to work on this issue. In order to understand what enhancements are required for CLI, could you please explain the scope of this issue?

PS: I've read the documentation for Zowe, tested it in my environment, and played with the cli options.

zFernand0:
Hey @Rahulkhinchi03,
It would be amazing if you would like to work on this.
However, I'm afraid that the application process for the LFX Mentorships is already closed and mentees have already been accepted for the various project. You should be hearing back from LFX soon 😋

PR #1474, opened 2022-07-23
Files comparison uss files

sarthakjdev:
This PR resolves [Issue#1445](https://github.com/zowe/zowe-cli/issues/1445).

Added a command ```zos-files compare uss-files``` to compare the differences between two uss-files .

Sample command: ```zowe files cmp uss '/u/mentee5/test/test.jcl' '/u/mentee5/test/test.js' ``` 

**It will take two arguments first will be the path to the first uss-file and the second one is the path to the second uss-file.**

**--bv** options can be used to get the same differences in the browser.


zFernand0:
As expected, code-duplication is only getting worse.
Maybe now is a good time to consider creating that common handler ?

zFernand0 (approved):
I'm a bit curious why it shows the "Launching browser..." message on a headless environment.
Maybe we should check for that and throw an error ?
Otherwise...
LGTM! 😋

gejohnston (approved):
Since you are using the zos-files-for-zowe-sdk/Get.USSFile function, it appears that the compare commands should adhere to the encoding of mainframe files, which has been a consistency concern with our upload and download commands. 

Have you confirmed that encoding is honored with the compare command?

KevinLoesch1:
One small fix. In en.ts, there appears to be an extra space before 'uss file' in both positonal parm descriptions. 

```
                    USSFILEPATH1: "The path of the first  uss file you want to compare.",
                    USSFILEPATH2: "The path of the second  uss file you want to compare."
```

zFernand0:
There is an issue in the `Get` SDK which causes this compare command to ignore encoding.
This also happens for the View command group.
More details:
- https://github.com/zowe/zowe-cli/issues/1495

PR #1478, opened 2022-07-29
Files comparison localfiles uss

sarthakjdev:
This PR resolves [Issue#1446](https://github.com/zowe/zowe-cli/issues/1446).

Added a command ```zos-files compare local-file-uss-file``` to compare the differences between a local file and a uss-file.

Sample command: ```zowe files cmp lfuss "./package.json" "/u/test/**"  ``` 

It will take two arguments first will be the local file path either relative or absolute and the second one is the path to USS file.


packages/cli/src/zosfiles/-strings-/en.ts
```diff
@@ -723,6 +723,19 @@ export default {
...
+            LOCAL_FILE_USS_FILE: {
+                SUMMARY: "Compare content of a local file and a z/os uss file",
+                DESCRIPTION: "Compare content of a local file and a z/os uss file on your terminal (stdout).",
+                POSITIONALS: {
+                    LOCALFILEPATH: "The path of the local file  you want to compare.",
+                    USSFILEPATH: "The path of the uss file set you want to compare."
+                },
+                EXAMPLES: {
+                    EX1: `Compare the contents of the local file and the uss file "./a.txt" and "sys1.samplib(antxtso)"`,
+                    EX2: `Compare the contents of the local file and the uss file "./a.txt" and "sys1.samplib(antxtso)"` +
```
t1m0thyj, line 735: Should these examples be updated to include a USS file path instead of a data set name?

packages/cli/src/zosfiles/compare/lfuss/LocalfileUss.definition.ts
```diff
@@ -0,0 +1,64 @@
...
+// Does not use the import in anticipation of some internationalization work to be done later.
+const strings = (require("../../-strings-/en").default as typeof i18nTypings).COMPARE;
+
+/**
+ * Compare local files and data sets command definition containing its description, examples and/or options
+ * @type {ICommandDefinition}
+ */
+export const LocalfileUssFileDefinition: ICommandDefinition = {
+    name: "local-file-uss-file",
+    aliases: ["lfuss"],
```
t1m0thyj, line 25: I'm probably being picky 😛 but not sure how I feel about this alias. Would `lf-uss` be clearer?

packages/cli/src/zosfiles/compare/lfuss/LocalfileUss.definition.ts
```diff
@@ -0,0 +1,64 @@
...
+    ],
+    examples: [
+        {
+            description: strings.ACTIONS.LOCAL_FILE_USS_FILE.EXAMPLES.EX1,
+            options: `"./a.txt" "sys1.samplib(antxtso)"`
+        },
+        {
+            description: strings.ACTIONS.LOCAL_FILE_USS_FILE.EXAMPLES.EX2,
+            options: `"./a.txt" "sys1.samplib(antxtso)" --no-seqnum`
+        }
```
t1m0thyj, line 62: Same question as above, should the examples be updated to use USS file paths?

packages/cli/src/zosfiles/compare/lfuss/LocalfileUss.handler.ts
```diff
@@ -0,0 +1,128 @@
...
+            } else if (err == null) {
+                throw new ImperativeError({
+                    msg: 'Path given is not a file'
+                });
+            } else {
+                throw new ImperativeError({
+                    msg: 'File not found, please recheck your file path'
+                });
+            }
+        });
```
t1m0thyj, line 54: Currently, there is no guarantee the callback will get executed before the subsequent lines of code execute.

I would suggest refactoring this code to use either the synchronous method `fs.lstatSync` or the promisified method `fs.lstat` from the "fs/promises" module.

t1m0thyj (changes requested):

packages/cli/src/zosfiles/compare/lfuss/LocalfileUss.handler.ts
sarthakjdev: Hey @t1m0thyj 
Refer:  https://github.com/zowe/zowe-cli/pull/1478/commits/0bcde3a5d195f85be348efbe704994df05c1a1e3

This has been resolved now :yum:

zFernand0 (approved):
LGTM! 😋

packages/cli/src/zosfiles/compare/lfuss/LocalfileUss.handler.ts
t1m0thyj: Thank you 👍

packages/cli/src/zosfiles/-strings-/en.ts
t1m0thyj: Thanks for updating the examples 🙂 Should the glob (`/u/**/**/*.ext`) be changed to an example file path like `/u/user/test.txt`?

gejohnston (commented):
See my comment about character sets in PR #1492

t1m0thyj (approved):

gejohnston (approved):

Issue #1480, opened 2022-08-05
Update npm install command in readme


t1m0thyj:
The ["Build Zowe CLI from source"](https://github.com/zowe/zowe-cli#build-zowe-cli-from-source) section of the readme suggests the following command:
```
npm install --@zowe:registry=https://registry.npmjs.org --no-package-lock --force
```

The flags are unnecessary and some may cause problems. The command should be replaced with the following:
```
npm install
```

* `--@zowe:registry=https://registry.npmjs.org` is unnecessary because the `@zowe:registry` defined in .npmrc should work
* `--no-package-lock` ignores the dependencies locked down in shrinkwrap for security/compatiblity reasons
* `--force` should generally not be used unless there is an error that requires force to be bypassed

omkolte17:
Hello, may I update the documentation for the same issue?

t1m0thyj:
@omkolte17 Sure, go ahead 🙂 When creating a pull request, please make sure to follow contribution guidelines including signing off your commits as described [here](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md#sign-all-of-your-git-commits).

PR #1481, opened 2022-08-06
Update npm install command

omkolte17:
Update the command under "Build Zowe CLI from source". #1480  

Signed-off-by: Om Kolte <omkolte17@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,11 +65,6 @@ The first time that you download Zowe CLI from the GitHub repository, issue the 
 npm install
 ```
 
-The flags from previous command are unnecessary now and may cause some problems, if you are wondering what problems might be caused here you go:
-- ```--@zowe:registry=https://registry.npmjs.org``` is unnecessary because the ```@zowe:registry``` defined in .npmrc should work.
-- ```--no-package-lock``` ignores the dependencies locked down in shrinkwrap for security/compatiblity reasons.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,9 +62,14 @@ Zowe CLI requires NPM version 7 to install from source. Before proceeding, check
...
 ```
-npm install --@zowe:registry=https://registry.npmjs.org --no-package-lock --force
+npm install
 ```
 
+The flags from previous command are unnecessary now and may cause some problems, if you are wondering what problems might be caused here you go:
+- ```--@zowe:registry=https://registry.npmjs.org``` is unnecessary because the ```@zowe:registry``` defined in .npmrc should work.
+- ```--no-package-lock``` ignores the dependencies locked down in shrinkwrap for security/compatiblity reasons.
+- ``` --force ``` should generally not be used unless there is an error that requires force to be bypassed.
+
```
t1m0thyj, line 72: Thanks for updating the command 🙂 Please remove these lines, just updating the command above is sufficient. Most people reading this won't know the context of what the previous command was.

omkolte17: Okay sure, done.
I thought it would be helpful to state that there but got what you wanted to say.
Thanks

t1m0thyj (approved):

zFernand0 (approved):

awharn (approved):

PR #1491, opened 2022-08-20
Files comparison spooldds

sarthakjdev:
This PR resolves [Issue#1447](https://github.com/zowe/zowe-cli/issues/1447).

Added a command ```zos-files compare spool-dd``` to compare the differences between two spooldds' 

Sample command: ```zowe files cmp sdd "jobName1:jobId1:spoolId1" "jobName2:jobId2:spoolId2"  ``` 

It will take two arguments containing descriptions for the two spool dds to be compared.
Each argument will contain the job name, job id, and spool id of a spool file separated by `:` for now.


packages/cli/src/zosfiles/-strings-/en.ts
```diff
@@ -736,6 +736,19 @@ export default {
...
+            SPOOL_DD: {
+                SUMMARY: "Compare content of two spool dds",
+                DESCRIPTION: "Compare content of two spool dds on your terminal (stdout).",
+                POSITIONALS: {
+                    SPOOLDDDESCRIPTION1: "The name of the job with the id of spool dd",
+                    SPOOLDDDESCRIPTION2: "The name of the job with the id of spool dd"
+                },
+                EXAMPLES: {
+                    EX1: `Compare the contents of the two spool dds "jobName1:jobId1:spoolId1"  "jobName2:jobId2:spoolId2"`,
+                    EX2: `Compare the contents of the local file and the uss file "jobName1:jobId1:spoolId1"  "jobName2:jobId2:spoolId2"` +
```
zFernand0, line 748: Seems like the text of this example may need to be updated 😋

packages/cli/src/zosfiles/-strings-/en.ts
```diff
@@ -736,6 +736,19 @@ export default {
                      `without sequence numbers`
                 }
             },
+            SPOOL_DD: {
+                SUMMARY: "Compare content of two spool dds",
+                DESCRIPTION: "Compare content of two spool dds on your terminal (stdout).",
+                POSITIONALS: {
+                    SPOOLDDDESCRIPTION1: "The name of the job with the id of spool dd",
+                    SPOOLDDDESCRIPTION2: "The name of the job with the id of spool dd"
```
zFernand0, line 744: should we differentiate them ?
ex: `the name of the first ...`  and `the name of the second ... `

packages/cli/src/zosfiles/compare/sdd/Spooldd.handler.ts
```diff
@@ -0,0 +1,79 @@
...
+        };
+        const descriptionSeperator: string = ":";
+        commandParameters.response.progress.startBar({ task });
+        const spoolDescription1 = commandParameters.arguments.spoolDescription1;
+
+        // retrieving information for first spooldd
+        const spoolDescArr1 = spoolDescription1.split(descriptionSeperator);
+        const jobName1: string = spoolDescArr1[0];
+        const jobId1: string = spoolDescArr1[1];
+        const spoolId1: number = Number(spoolDescArr1[2]);
```
zFernand0, line 35: thoughts on a utility function for this ?
Maybe we can simplify this to `const [jobName1, jobId1, SpoolId1] = getSpoolDescriptor(spoolDescription1)`

zFernand0 (commented):
Left a few comments.
but the functionality LGTM! 😋

packages/cli/src/zosfiles/compare/sdd/Spooldd.handler.ts
sarthakjdev: Should it be in the same handler :thinking:?

zFernand0: No need to do this.
However, if we do it, I think it should be in:
- the zos-files-sdk, or
- the core-sdk (so that in can be shared between the files and jobs SDKs)

zFernand0 (approved):

KevinLoesch1 (approved):

t1m0thyj (approved):

gejohnston (approved):

PR #1492, opened 2022-08-23
Files comparison localfiles spooldd

sarthakjdev:
This PR resolves [Issue#1448](https://github.com/zowe/zowe-cli/issues/1448).

Added a command ```zos-files compare local-file-spool-dd``` to compare the differences between a local-file and a spool-dd' 

Sample command: ```zowe files cmp lfsdd "./package.json" "jobName:jobId:spoolId"  ``` 

It will take two arguments first will be the local file path either relative or absolute and the second one is the description of the spool dd to be compared.
Spooldd description argument will contain the job name, job id, and spool id of a spool file separated by `:` for now.


packages/cli/src/zosfiles/-strings-/en.ts
```diff
@@ -749,6 +749,19 @@ export default {
...
+            LOCAL_FILE_SPOOL_DD: {
+                SUMMARY: "Compare content of a local file and a spool dd",
+                DESCRIPTION: "Compare content of a local-file and a spool-dd on your terminal (stdout).",
+                POSITIONALS: {
+                    LOCALFILEPATH: "The path of the local file  you want to compare.",
+                    SPOOLDDDESCRIPTION: "The name of the job with the id of spool dd"
+                },
+                EXAMPLES: {
+                    EX1: `Compare the contents of the two spool dds "./a.txt"  "jobName:jobId:spoolId"`,
+                    EX2: `Compare the contents of the local file and the uss file "./a.txt"  "jobName:jobId:spoolId"` +
```
zFernand0, line 761: `uss file` -> `spool dd`

packages/cli/src/zosfiles/compare/lfsdd/LocalfileSpooldd.handler.ts
```diff
@@ -0,0 +1,115 @@
...
+
+        // resolving to full path if local path passed is not absolute
+        if (path.isAbsolute(commandParameters.arguments.localFilePath)) {
+            localFile = commandParameters.arguments.localFilePath;
+        } else {
+            localFile = path.resolve(commandParameters.arguments.localFilePath);
+        }
+
+        // check if the path given is of a file or not
+        fs.lstat(localFile, (err, stats) => {
```
zFernand0, line 44: use `fs.lstatSync`

zFernand0 (commented):
left a couple of comments, but everything else LGTM!

gejohnston (commented):
Work is in progress to consistently accommodate the character set (and tagging) for our upload and download commands. Our view and diff commands are believed to have similar character set issues. Some of those new commands have already been released, but for those that have not been released yet, I don't think that we should publish them, because we would be expanding the number of commands with a known problem.

The problems are believed to reside in underlying functions (not in the code in this PR). I would like us to approve these changes, merge them, but not release them. I think perhaps we should merge this PR (and similar PRs) into a "staging" branch, which we do not merge into master until the underlying functions have been enhanced to correctly real with character sets.

Maybe others have better way to approve and merge such PRs without publishing?

packages/cli/src/zosfiles/compare/lfsdd/LocalfileSpooldd.definition.ts
```diff
@@ -0,0 +1,62 @@
...
+// Does not use the import in anticipation of some internationalization work to be done later.
+const strings = (require("../../-strings-/en").default as typeof i18nTypings).COMPARE;
+
+/**
+ * Compare local-file-spool-dd command definition containing its description, examples and/or options
+ * @type {ICommandDefinition}
+ */
+export const LocalfileSpoolddDefinition: ICommandDefinition = {
+    name: "local-file-spool-dd",
+    aliases: ["lfsdd"],
```
t1m0thyj, line 25: Should the alias be `lf-sdd` with a hyphen to be consistent with other new commands that have been added? 🙂

gejohnston (approved):

t1m0thyj (approved):

PR #1512, opened 2022-09-10
Compare command base helper

sarthakjdev:
This PR resolves the problem of excessive code repetition in the compare commands available in the CLI. 

Each compare command has some common operations like setting up option, preparing strings for comparison, and getting responses either in the terminal or browser based upon the options passed to the command. 

In this PR, proposing to add a common base helper for comparing commands, containing utility methods for comparisons.
Also, implemented the cmp ds and cmp uss command using these helper class methods to have a look at the functional behavior of the changes


sarthakjdev:
This is a basic implementation of the helper class as for now. 
Looking for you people suggestions. :slightly_smiling_face: or if there could be any other more convenient way of doing  so.

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
```diff
@@ -0,0 +1,230 @@
...
+*/
+
+import { IHandlerParameters, DiffUtils } from "@zowe/imperative";
+import {  IZosFilesResponse } from "@zowe/zos-files-for-zowe-sdk";
+
+/**
+ * Base helper for compare command to handle common operations through the comparison
+ * @export
+ */
+export default class ComapareBaseHelper {
```
t1m0thyj, line 19: `Comapare` -> `Compare`

packages/cli/src/zosfiles/compare/ds/Dataset.handler.ts
```diff
@@ -1,6 +1,6 @@
 /*
-* This program and the accompanying materials are made available under the terms of the
-* Eclipse Public License v2.0 which accompanies this distribution, and is available at
+* CompareBaseHelper.instance program and the accompanying materials are made available under the terms of the
+* Eclipse Public License v2.0 which accompanies CompareBaseHelper.instance distribution, and is available at
```
t1m0thyj, line 3: Please revert license text back to what it was originally.

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
```diff
@@ -0,0 +1,230 @@
...
+     * @static
+     * @readonly
+     */
+    public static get instance(): ComapareBaseHelper {
+        if(this.hInstance == null){
+            this.hInstance = new ComapareBaseHelper();
+        }
+
+        return this.hInstance;
+    }
```
t1m0thyj, line 41: In the future, Zowe CLI daemon mode may support executing multiple commands in parallel. If multiple handlers running at the same time share a singleton class, this could cause race conditions.

Instead of using a singleton class, I think it would be better to construct a new instance of the class each time it is used in a handler (`new CompareBaseHelper()`).

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
```diff
@@ -0,0 +1,230 @@
...
+     * @memberof ComapareBaseHelper
+     */
+    public record2: boolean;
+
+    /**
+     * volume serial option for file2
+     * @public
+     * @memberof ComapareBaseHelper
+     */
+    public volumeSerial2: string;
```
t1m0thyj, line 104: Instead of defining each option twice (`binary`, `binary2`, `encoding`, `encoding2`, etc), could we create an interface `ICompareFileOptions`? This could remove even more code duplication 🙂 

For example, with an interface like the following:
```
export interface ICompareFileOptions {
    binary: boolean;
    encoding: string;
    record: boolean;
    volumeSerial: string;
}
```

The options could be defined in this class with just 2 lines:
```
public file1Options: ICompareFileOptions;

public file2Options: ICompareFileOptions;
```

t1m0thyj (changes requested):
Thanks for doing this refactoring 🙂 Left a few comments about typos and suggested changes.

sarthakjdev:
hey, 
I made the suggested changes, please have a look.

sarthakjdev:
Also, if this method looks fine, then we can start implementing this base helper in other compare commands and will update and write new tests for the same. :slightly_smiling_face:

gejohnston (dismissed):

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
```diff
@@ -0,0 +1,168 @@
...
+
+import { IHandlerParameters, DiffUtils } from "@zowe/imperative";
+import {  IZosFilesResponse } from "@zowe/zos-files-for-zowe-sdk";
+import { ICompareFileOptions } from "./doc/ICompareFileOptions";
+
+/**
+ * Base helper for compare command to handle common operations through the comparison
+ * @export
+ */
+export default class CompareBaseHelper {
```
t1m0thyj, line 20: Please remove the `default` keyword. I think we generally prefer to avoid using it, and use explicitly named imports/exports instead, except in cases where it is required like for command handler files.

packages/cli/src/zosfiles/compare/doc/ICompareFileOptions.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @exports
+ * @interface ICompareFileOptions
+ */
+export interface ICompareFileOptions {
+    binary?: boolean | false;
+    encoding?: string | "";
+    record?: boolean | false;
+    volumeSerial?: string | "";
+}
```
t1m0thyj, line 24: Can the `| false` and `| ""` be removed from these type definitions? I'm not sure what their intent is, since interface properties can't have default values.

To set default values for these properties, you could do it in the constructor of the class, and optionally include comments in this file that mention the default values as well 🙂

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
```diff
@@ -0,0 +1,168 @@
...
+     * @memberof CompareBaseHelper
+     */
+    public file1Options: ICompareFileOptions = {} as ICompareFileOptions;
+
+    /**
+     * file compare option for file2
+     * @public
+     * @memberof CompareBaseHelper
+     */
+    public file2Options: ICompareFileOptions = {} as ICompareFileOptions;
```
t1m0thyj, line 35: The `as ICompareFileOptions` should not be necessary, because all properties on the interface are optional.

packages/cli/src/zosfiles/compare/ds/Dataset.handler.ts
```diff
@@ -9,114 +9,46 @@
 *
 */
 
-import { AbstractSession, IHandlerParameters, ITaskWithStatus, TaskStage, DiffUtils } from "@zowe/imperative";
+import { AbstractSession, IHandlerParameters, ITaskWithStatus, TaskStage } from "@zowe/imperative";
 import { Get, IZosFilesResponse } from "@zowe/zos-files-for-zowe-sdk";
 import { ZosFilesBaseHandler } from "../../ZosFilesBase.handler";
+import CompreBaseHelper from '../CompareBaseHelper';
```
t1m0thyj, line 15: Please fix the typo and use a non-default import:
`import { CompareBaseHelper } from '../CompareBaseHelper';`

t1m0thyj (changes requested):
Thanks for addressing the comments 🙂 Left a few more.

sarthakjdev:
Made the suggested changes. 
Implemented the compare base helpers in all compare commands.
Added tests for comparing base helpers.

I can't able to test the commands due to my z/OS credentials being expired, figuring out if I could get one somehow, and pls do mention if you people have any leads on how can I get one for development purposes. :slightly_smiling_face:

t1m0thyj:
Please fix the failing test snapshots. We've had an issue recently where local builds of Zowe CLI that have Imperative linked from source produce broken snapshots. I believe @awharn was able to fix this by deleting node modules and starting fresh (`rm -rf node_modules && npm install`). You may need to do something similar to get the snapshot to match what the GitHub workflow is expecting 🙂

zFernand0:
I posted a suggestion in a new branch (based off of this one)
https://github.com/zowe/zowe-cli/compare/compare-command-base-helper...compare-base-handler
I'm hoping to reduce the number of places something could go wrong 😋

sarthakjdev:
Hey, 
Is there anything needed to be updated in this PR? 
I had updated the other handlers as per the base one but wasn't able to test them because of the unauthorized credentials issue.

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
```diff
@@ -0,0 +1,201 @@
...
+                    });
+                }
+            } catch (error) {
+                if (error instanceof ImperativeError) throw error;
+                throw new ImperativeError({
+                    msg: 'Path not found. Please check the path and try again'
+                });
+            }
+            // reading local file as buffer
+            lfContentBuf = fs.readFileSync(localFile);
```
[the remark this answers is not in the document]
zFernand0: This alert is dissmed.
Details: https://github.com/zowe/zowe-cli/security/code-scanning/373

gejohnston (approved):
This eliminates a LOT of duplicated code. Thanks for putting in the effort.

packages/cli/src/zosfiles/compare/CompareBase.handler.ts
```diff
@@ -0,0 +1,77 @@
...
+        helper: CompareBaseHelper,
+        ): Promise<string | Buffer>;
+
+    /**
+     * Abstract method required for compare handlers to determine how they gather the contents of the file/dataset to compare
+     * @param session Gives access to the session object in case the handler needs to submit a request to get the contents
+     * @param args Command arguments required for getting the dataset, filepath, or spool descriptor
+     * @param helper CompareBaseHelper instance to access prepare- functions
+     */
+     public abstract getFile2(
```
awharn, line 72: Lint Warning - Indentation (5 instead of 4)

packages/cli/src/zosfiles/compare/CompareBase.handler.ts
```diff
@@ -0,0 +1,77 @@
...
+     * Abstract method required for compare handlers to determine how they gather the contents of the file/dataset to compare
+     * @param session Gives access to the session object in case the handler needs to submit a request to get the first file
+     * @param args Command arguments required for getting the dataset, filepath, or spool descriptor
+     * @param helper CompareBaseHelper instance to access prepare- functions
+     */
+    public abstract getFile1(
+        session: AbstractSession,
+        args: ICommandArguments,
+        helper: CompareBaseHelper,
+        ): Promise<string | Buffer>;
```
awharn, line 64: Lint Warning - Indentation (8 instead of 4)

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
```diff
@@ -0,0 +1,201 @@
...
+            try {
+                if(!fs.fstatSync(localFileHandle).isFile()){
+                    throw new ImperativeError({
+                        msg: 'Path given is not of a file, do recheck your path again'
+                    });
+                }
+            } catch (error) {
+                if (error instanceof ImperativeError) throw error;
+                throw new ImperativeError({
+                    msg: 'Path not found. Please check the path and try again'
```
awharn, line 144: If the path is not found, the error is not going to surface with the fstat call, it is going to occur with the openSync call. That case does not appear to be handled here.

awharn (changes requested):

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
awharn: openSync is in the outer try so the file not found error should probably be in the outer catch statement. There does not appear to be an outer catch statement. Also, closeSync may fail in the finally block, since the file handle would be undefined.

awharn (changes requested):

zFernand0:
> openSync is in the outer try so the file not found error should probably be in the outer catch statement. There does not appear to be an outer catch statement. Also, closeSync may fail in the finally block, since the file handle would be undefined.

Thanks Andrew.
I've added some tests to help the situation. 😋 

My apologies for not taking a closer look into how the `fileHandler` situation should be handled. 😓

awharn (approved):

Issue #1490, opened 2022-08-19
Expose api to filter jobs based on status. 


kristinochka:
Zowe Explorer uses zowe-cli apis to filter jobs based on user and prefix, or id. ZOSMF api supports filtering based on status as well. Currently zosmf api only supports filtering on "ACTIVE" status. 
https://www.ibm.com/docs/en/zos/2.4.0?topic=interface-list-jobs-owner-prefix-job-id

Zowe-cli need to accept this optional status parameter and use it in the api call to the server.

PR #1513, opened 2022-09-12
Get jobs by status api update

kristinochka:
Related to https://github.com/zowe/zowe-cli/issues/1490


kristinochka:
@zFernand0 FYI

packages/zosjobs/src/GetJobs.ts
```diff
@@ -82,6 +82,22 @@ export class GetJobs {
...
 
+    /**
+     * Get a list of jobs that match various parameters
+     * @static
+     * @param {AbstractSession} session - z/OSMF connection info
+     * @param {string}
+     * @returns {Promise<IJob[]>} - promise that resolves to an array of IJob objects (matching jobs)
+     * @memberof GetJobs
+     */
+    public static async getJobsByParameters(session: AbstractSession, params: IGetJobsParms) {
```
kristinochka, line 93: Currently we have getJobsByOwnerAndPrefix, this can be the new api that gets jibs by any parameter.

kristinochka:
@zFernand0 Are you OK with the function name `getJobsByParameters` accepting `params: IGetJobsParms`?
The idea here is in the future more parameters might be passed.

zFernand0:
> Are you OK with the function name `getJobsByParameters` accepting `params: IGetJobsParms`? The idea here is in the future more parameters might be passed.

No objections here 😋

kristinochka:
@zFernand0 The changelog check says: `Error: The changelog has changed in packages/zosjobs/CHANGELOG.md, but the required header is missing.` It's not clear what it's talking about. 
Once this check passes, it's ready for review.

packages/zosjobs/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS jobs SDK package will be documented in this file.
 
+## `7.2.1`
```
zFernand0, line 5: TL;DR: change ```## `7.2.1` ``` to `## Recent Changes`

the error was complaining about the header.
we don't lock versions down in the PRs
instead, we use `## Recent Changes` (similar to Zowe Exlplorer's `## TBD Release`) 

And in our case, the automation will update the header once a version has been selected by the admins or the repo

kristinochka: Done.

kristinochka:
@zFernand0 Could you look again?

zFernand0 (changes requested):
@kristinochka, the changes look great! 😋 
would you mind fixing the DCO issue ?

kristinochka:
@zFernand0 The issue should be fixed now.

zFernand0:
The sonarcloud token/secret  is not accessible from forked repositories.
Here is a run or the Sonar Scans for this branch
https://github.com/zowe/zowe-cli/actions/runs/3440833426/jobs/5739739970

zFernand0 (approved):
LGTM! 😋

kristinochka:
@t1m0thyj This PR requires 2 approvers before merging. Could you take a look?

kristinochka:
We would like to be able to filter by different statuses in Zowe Explorer. zosmf api only accepts filtering by "active" 
To handle the situation where user passes in status other than "active" we should filter the jobs in zowe-cli and return the ones that match, that way we can support more advanced filtering. I think it would be best if I added those changes as part of this PR.

zFernand0:
> We would like to be able to filter by different statuses in Zowe Explorer. zosmf api only accepts filtering by "active" To handle the situation where user passes in status other than "active" we should filter the jobs in zowe-cli and return the ones that match, that way we can support more advanced filtering. I think it would be best if I added those changes as part of this PR.

Do you plan to pass the query parameter (`status=active`) if the given status-filter is `active`. Otherwise, get everything and filter on the client ? 
No objections from my end if you plan on doing that 😋

kristinochka:
Fernando, yes, (status=active) will be passed if the given status-filter is active. If status is something else, I will add code here in zowe-cli to filter the results based on that status. So then every api call will just work instead of filtering things in Zowe Explorer.

kristinochka:
Other extender APIs are able to filter on more statuses, so in order to provide a unified experience we can filter the results in zowe-cli before passing them to Zowe Explorer or whatever is using this API call.

packages/zosjobs/src/GetJobs.ts
```diff
@@ -393,4 +381,19 @@ export class GetJobs {
         Logger.getAppLogger().info("GetJobs.getSpoolContentCommon() parameters: " + parameters);
         return ZosmfRestClient.getExpectString(session, JobsConstants.RESOURCE + parameters, [Headers.TEXT_PLAIN_UTF8]);
     }
+
+    private static filterResultsByStatuses(jobs: IJob[], params: IGetJobsParms): IJob[] {
```
zFernand0, line 385: Should this be a public method ?

packages/zosjobs/src/GetJobs.ts
```diff
@@ -360,4 +381,19 @@ export class GetJobs {
...
+        if (params?.status && params.status.toLowerCase() != 'active') {
```
zFernand0, line 386: Why not simplify this into `params?.status?.toLowerCase() != "active"` ? 😋

packages/zosjobs/src/GetJobs.ts
```diff
@@ -360,4 +381,19 @@ export class GetJobs {
...
+            const statusJobs = jobs.filter(job => {
```
zFernand0, line 387: Should this filer operation be simplified to reduce cognitive complexity?
Suggestion: `return jobs.filter(job => job.status.toLowerCase() === params.status.toLowerCase())`

packages/zosjobs/src/GetJobs.ts
```diff
@@ -360,4 +381,19 @@ export class GetJobs {
...
+                const paramStatusLower = params.status.toLowerCase();
+                if (resultStatusLower === paramStatusLower) {
+                    return job;
+                }
+                return undefined;
+            });
+            return statusJobs;
+        }
+        return jobs;
+    }
```
zFernand0, line 398: Seems like we are missing some tests for this functionality 😋

packages/zosjobs/src/GetJobs.ts
```diff
@@ -121,55 +134,63 @@ export class GetJobs {
...
         }
 
         let resource = JobsConstants.RESOURCE;
         resource += (query === JobsConstants.QUERY_ID) ? "" : query;
         Logger.getAppLogger().info("GetJobs.getJobsCommon() resource: " + resource);
 
-        return ZosmfRestClient.getExpectJSON<IJob[]>(session, resource);
+        const jobs = await ZosmfRestClient.getExpectJSON<IJob[]>(session, resource);
+
+        return GetJobs.filterResultsByStatuses(jobs, params);
```
zFernand0, line 193: Do you like one-liners? 😋 

`return params?.status == null || && params.status.toLowerCase() === "active" ? jobs : jobs.filter(job => job.status.toLowerCase() === params.status.toLowerCase())`

I'm not suggesting this though. It'll probably make things a bit confusing 😋

packages/zosjobs/src/GetJobs.ts
```diff
@@ -360,4 +381,19 @@ export class GetJobs {
...
         return ZosmfRestClient.getExpectString(session, JobsConstants.RESOURCE + parameters, [Headers.TEXT_PLAIN_UTF8]);
     }
+
+    private static filterResultsByStatuses(jobs: IJob[], params: IGetJobsParms): IJob[] {
+        if (params?.status && params.status.toLowerCase() != 'active') {
+            const statusJobs = jobs.filter(job => {
+                const resultStatusLower = job.status.toLowerCase();
+                const paramStatusLower = params.status.toLowerCase();
+                if (resultStatusLower === paramStatusLower) {
+                    return job;
```
zFernand0, line 391: doesn't the predicate (function) of the `.filter` expects a `boolean`?

zFernand0 (changes requested):
honestly the functionality should work as expected.
I just had some strange suggestions on the code 😋 

The only requirement I do have is for some tests (and maybe considering making the new method public)

packages/zosjobs/src/GetJobs.ts
kristinochka: The users of zowe-cli api are suppose to call getJobsByParameters and that will call zOSMF api and handle filtering results. filterResultsByStatuses is suppose to be an internal function. I separated filtering into a function for easier unit testing and more readable code.

packages/zosjobs/src/GetJobs.ts
kristinochka: Done

packages/zosjobs/src/GetJobs.ts
kristinochka: Done

packages/zosjobs/src/GetJobs.ts
kristinochka: Hmmm....one liners are great when checking one condition, can become confusing the more statements we check. The updated filterResultsByStatuses is much simpler now:
```
    private static filterResultsByStatuses(jobs: IJob[], params: IGetJobsParms): IJob[] {
        if (params?.status?.toLowerCase() != "active") {
            return jobs.filter(job => job.status.toLowerCase() === params.status.toLowerCase());
        }
        return jobs;
    }
```

packages/zosjobs/src/GetJobs.ts
kristinochka: Yes. it worked because job is Truthy. That function has been updated.

kristinochka:
@zFernand0 Thanks for your suggestions. See my updates.

zFernand0 (approved):
LGTM!  😋 
I'm not requesting any more changes.
My comments are just very minor details which won't impact any end-user functionality

packages/zosjobs/__tests__/__unit__/GetJobs.unit.test.ts
```diff
@@ -511,4 +525,171 @@ describe("GetJobs tests", () => {
...
+        });
+        it("should return all status jobs for status *", () => {
+            const jobs = [
+                {
+                    jobid: '1', jobname: 'a', subsystem: 'sub',
+                    owner: 'zowe', status: 'active', type: 't',
+                    class: 'c', retcode: 'r', url: '',
+                    'files-url': '', 'job-correlator': '',
+                    phase: 1, 'phase-name': 'name', 'reason-not-running': 'no'
+                },
```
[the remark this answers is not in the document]
t1m0thyj: The size of this file could be reduced considerably if we remove the extraneous properties.

t1m0thyj (approved):

PR #1497, opened 2022-08-25
Update dev deps to latest versions

t1m0thyj:
Update the following dev dependencies to maintain technical currency and reduce vulnerabilities: ESLint, Jest, Typedoc, TypeScript

Sorry this PR is so big 😢 There were a lot of small changes required in tests to migrate from Jest 24 to 28. Fixed over 500 test compilation errors - most were because type checking has become stricter.

Related to (but not dependent on) https://github.com/zowe/imperative/pull/876 which still has some tests that need fixed.

* [x] Fix build errors
* [x] Fix lint errors
* [x] Fix test errors
* [x] Fix typedoc errors

Issue #1516, opened 2022-09-15
Receiving unexpected packet type with ssh2 from sdk.


fbeyl:
### Discussed in https://github.com/zowe/zowe-cli/discussions/1510

<div type='discussions-op-text'>

<sup>Originally posted by **fbeyl** September  8, 2022</sup>
Tried node sdk 'disk space' example but get ssh2 errors. Every now and then there is a succesfull run but then again several 'type errors'.
```
Outbound: Sending KEXECDH_INIT
Outbound: Sending DISCONNECT (3)
node:events:491
      throw er; // Unhandled 'error' event
      ^
Error: Received unexpected packet type 0
```
Difference I see in client.js is added line 710 proto.start();
When I use ssh2 not from sdk but symlinking to outside ssh2 no more errors occur.
ssh2 version used in sdk is 1.4.0 ---> returns error
outside ssh2 version is 1.11.0 ---> returns ok</div>

t1m0thyj:
Thanks for reporting this! It should be fixed in `@zowe/cli@7.6.1` which is available on npmjs.org with the `@next` tag and will be published to the `@zowe-v2-lts` tag soon.

PR #1517, opened 2022-09-15
Update @types/node to current LTS version

t1m0thyj:
Forgot to update the `@types/node` package in #1497. Updated it to 14.x, I believe this is the best version to use currently. If we used a newer version like 16.x, we might accidentally use a Node.js feature missing in 14.x which we still support.


zFernand0 (approved):

zFernand0:
do you mind adding the `for-review`, `keep`, and `third-party` labels to the `stale.yml` workflow
https://github.com/zowe/zowe-cli/blob/master/.github/workflows/stale.yml#L23

or at least the `keep` 😋

gejohnston (approved):

t1m0thyj:
> do you mind adding the `for-review`, `keep`, and `third-party` labels to the `stale.yml` workflow https://github.com/zowe/zowe-cli/blob/master/.github/workflows/stale.yml#L23
> 
> or at least the `keep` 😋

Sure, we can discuss labels to add at our next standup 🙂 Should I also update the `ssh2` dependency?

zFernand0:
> > do you mind adding the `for-review`, `keep`, and `third-party` labels to the `stale.yml` workflow https://github.com/zowe/zowe-cli/blob/master/.github/workflows/stale.yml#L23
> > or at least the `keep` 😋
> 
> Sure, we can discuss labels to add at our next standup 🙂 Should I also update the `ssh2` dependency?

yeah, maybe we can address this issue 😋 
- #1510 
- #1516

KevinLoesch1 (approved):

awharn (approved):

zFernand0:
Thanks for updating the ssh2 package 🥳

PR #1519, opened 2022-09-16
ensure zowe zos-files create data-set works without --like option

AmandaDErrico:
Signed-off-by: Amanda D'Errico <amanda.derrico@ibm.com>

According to comments in PR [#1312](https://github.com/zowe/zowe-cli/pull/1312), this current PR addresses fixing the example below:

```
   - Create a data set with type LIBRARY:

      $ zowe zos-files create data-set NEW.DATASET --data-set-type LIBRARY
```

Resolves: Issue/Bug [#1252](https://github.com/zowe/zowe-cli/issues/1252)


awharn (commented):
This looks good to me, but there appear to be several linting errors that need to be corrected first.

packages/cli/src/zosfiles/create/ds/ds.handler.ts
```diff
@@ -20,11 +20,23 @@ import { generateZosmfOptions } from "../Create.utils";
...
+        }
+        else {
+            return Create.dataSetLike(
+                session,
+                commandParameters.arguments.dataSetName,
+                commandParameters.arguments.like,
+                generateZosmfOptions(commandParameters.arguments)
+            );
+        }
+
```
t1m0thyj, line 40: For the logic to choose between `Create.dataSet` or `Create.dataSetLike`, I don't think it should be based on the `--data-set-type` argument. There are a number of other arguments such as `--primary-space` or `--record-length` that could also be passed to the command without the `--like` option.

I would suggest using the following logic:
```javascript
if (commandParameters.arguments.like == null) {
  // Create.dataSet
} else {
  // Create.dataSetLike
}
```

AmandaDErrico: Good point, I've changed the logic to use `commandParameters.arguments.like == null`.

packages/cli/__tests__/zosfiles/__system__/create/ds/cli.files.create.ds.system.test.ts
```diff
@@ -39,6 +41,8 @@ describe("Create Data Set", () => {
         REAL_SESSION = TestEnvironment.createZosmfSession(TEST_ENVIRONMENT);
 
         user = defaultSystem.zosmf.user.trim().toUpperCase();
+        pass = defaultSystem.zosmf.password.trim().toUpperCase();
```
awharn, line 44: Will explicitly making password and host uppercase cause an unintended failure?

awharn (commented):
The changes look good to me, I just have one question.

packages/cli/__tests__/zosfiles/__system__/create/ds/cli.files.create.ds.system.test.ts
AmandaDErrico: Thanks for catching, I've updated this section to remove `toUpperCase()`

awharn (approved):

packages/cli/src/zosfiles/create/ds/ds.handler.ts
```diff
@@ -20,11 +20,23 @@ import { generateZosmfOptions } from "../Create.utils";
...
-        return Create.dataSetLike(
-            session,
-            commandParameters.arguments.dataSetName,
-            commandParameters.arguments.like,
-            generateZosmfOptions(commandParameters.arguments)
-        );
+        if (commandParameters.arguments.like == null) {
+            return Create.dataSet(
+                session,
+                CreateDataSetTypeEnum.DATA_SET_BLANK,
```
zFernand0, line 26: Note: this is not a request for changes. 
The code is working as expected and letting z/osmf decide what it needs to do with the information.

I still want to post the following question:
- Should this dsntype be determined based on the provided arguments ?

Based on the help, it feels that someone should be able to specify PDS and a `DATA_SET_PARTITIONED` enum should be selected (and its defaults)

---

The only way I could find out this worked is by creating datasets with all allowed values, and here are my results:
- BASIC: `dsorg: PS`
- EXTPREF: `dsorg: PS-E`
- EXTREQ: `dsorg: PS-E`
- HFS: `dsorg: PS`
- LARGE: `dsorg: PS-L`
- PDS: `dsorg: PS` (I expected this to be `dsorg: PO` - might be a z/osmf issue)
- LIBRARY: `dsorg: PS` (I expected this to be `dsorg: PO` - might be a z/osmf issue)
- PIPE: `dynamic allocation error` 🤷 

---
Help text:
```
   --data-set-type | --dst | --dsntype (string)

      The data set type (BASIC, EXTPREF, EXTREQ, HFS, LARGE, PDS, LIBRARY, PIPE)
```

zFernand0 (approved):
code working as expected.
left a question that I'd like to discuss during standup 😋

PR #1524, opened 2022-09-22
Updated example on zowe profiles create zosmf-profile command

ATorrise:

### DIFF
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to the Zowe CLI package will be documented in this file.
 
 ## Recent Changes
-- BugFix: Updated example on zowe profiles create zosmf-profile command. [#1152](https://github.com/zowe/zowe-cli/issues/1152)
+- BugFix: Updated example for `zowe profiles create zosmf-profile` command. [#1152](https://github.com/zowe/zowe-cli/issues/1152)
 
 ## `7.6.1`
 
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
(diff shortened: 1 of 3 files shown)


MikeBauerCA:
Thanks @ATorrise - this looks good. It was a good call to use the default API Mediation Layer port of 7554 in the example. I think you may need to update the [CLI CHANGELOG](https://github.com/zowe/zowe-cli/blob/master/packages/cli/CHANGELOG.md) as well.

t1m0thyj (approved):
Thanks for updating this example 🙂 I verified the output by running `zowe profiles create zosmf -h` and noticed that it looks kind of weird to have a period before the semicolon at the end:
image

Typically I would suggest to reword the example description to be a single sentence so we can remove the period, but in this case that seems impossible without creating a huge run-on sentence 🤔 I've added some more reviewers to the PR to get more opinions on this.

ATorrise:
> Typically I would suggest to reword the example description to be a single sentence so we can remove the period, but in this case that seems impossible without creating a huge run-on sentence 🤔 I've added some more reviewers to the PR to get more opinions on this.

Would it make more sense to remove the colon instead?

t1m0thyj:
> > Typically I would suggest to reword the example description to be a single sentence so we can remove the period, but in this case that seems impossible without creating a huge run-on sentence 🤔 I've added some more reviewers to the PR to get more opinions on this.
> 
> Would it make more sense to remove the colon instead?

Good idea, I believe the colon is hardcoded at the end of all command examples by Imperative, so unfortunately we can't easily remove it. I suppose we could enhance Imperative so that if the string ends with a period, then it doesn't add a colon at the end. But that's definitely out of scope so I don't want to suggest more work 🙂

gejohnston (approved):
Since the team agreed to remove the period after "profile.:" I think we could merge this after that minor update.

gejohnston (approved):

PR #1525, opened 2022-09-22
Feature/create.ts cmd type

ATorrise:
changed parameter from `cmdType` to `dataSetType` and clarified options description

### DIFF
--- a/packages/zosfiles/CHANGELOG.md
+++ b/packages/zosfiles/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+- Enhancement: Added to the description of Create.ts `cmdType`. [#48](https://github.com/zowe/zowe-cli/issues/48)
+
 ## `7.6.0`
 
 - Enhancement: Updated the `List.fileList` method to support z/OSMF filter parameters such as "mtime" and "name".
--- a/packages/zosfiles/src/methods/create/Create.ts
+++ b/packages/zosfiles/src/methods/create/Create.ts
@@ -34,9 +34,9 @@ export class Create {
     /**
      * Create a data set
      * @param {AbstractSession} session                     - z/OSMF connection info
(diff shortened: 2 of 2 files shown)


packages/zosfiles/src/methods/create/Create.ts
```diff
@@ -34,13 +34,13 @@ export class Create {
     /**
      * Create a data set
      * @param {AbstractSession} session                     - z/OSMF connection info
-     * @param {CreateDataSetTypeEnum} cmdType               - The type of data set we are going to create
+     * @param {CreateDataSetTypeEnum} dataSetType           - The type of data set we are going to create from default options
      * @param {string} dataSetName                          - the name of the data set to create
      * @param {Partial<ICreateDataSetOptions>} [options={}] - additional options for the creation of the data set
```
t1m0thyj, line 39: Should these docstrings be updated based on the changes we discussed?

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+- Enhancement: Added to the description of Create.ts `cmdType`. [#48](https://github.com/zowe/zowe-cli/issues/48)
```
gejohnston, line 6: I suggest rewording the changelog text. We anticipate that our users will read the changlog to know if there are changes of interest to them. The reference to a source file is technically correct, but will not mean much to our users. Try stating the change in terms of the command that has been affected and the result that the end user will see. It still does not have to be very wordy - just a different perspective.

gejohnston (changes requested):
One minor suggestion about the wording of some text.

t1m0thyj (approved):

gejohnston (approved):

packages/zosfiles/CHANGELOG.md
zFernand0: Should this be considered a bugFix / patch since it's not introducing new behavior

zFernand0 (approved):
Thanks for fixing the JSDoc!
left one small comment (which can be address in other PRs 😋

PR #1526, opened 2022-09-22
adding alcunit to zosfilescreateoptions

### LINKED ISSUE #1203: alcunit missing from zowe files create pds
it looks like `alcunit` is missing from the command line options with a default of `CYL` being the only option.  can this option be exposed on the CLI?

ATorrise:

Resolves #1203


packages/cli/src/zosfiles/create/pds/Pds.definition.ts
```diff
@@ -47,6 +47,7 @@ export const PdsDefinition: ICommandDefinition = {
         {...ZosFilesCreateOptions.recfm, defaultValue: CreateDefaults.DATA_SET.PARTITIONED.recfm},
         {...ZosFilesCreateOptions.blksize, defaultValue: CreateDefaults.DATA_SET.PARTITIONED.blksize},
         {...ZosFilesCreateOptions.lrecl, defaultValue: CreateDefaults.DATA_SET.PARTITIONED.lrecl},
+        {...ZosFilesCreateOptions.lrecl, defaultValue: CreateDefaults.DATA_SET.PARTITIONED.alcunit},
```
t1m0thyj, line 50: Should this be `ZosFilesCreateOptions.alcunit`?

t1m0thyj: I think the spread operator and `CreateDefaults` should be kept like below, sorry for the confusion:
```
{...ZosFilesCreateOptions.alcunit, defaultValue: CreateDefaults.DATA_SET.PARTITIONED.alcunit}
```

The reason for keeping the `CreateDefaults` is that it defines a default value for `alcunit` here: https://github.com/zowe/zowe-cli/blob/master/packages/zosfiles/src/methods/create/Create.defaults.ts#L25

gejohnston (commented):
I think that this option should be tested with a live z/OS system before approval.  If system tests exist for this area, we should probably add a new test to confirm that the aclunit option works.

zFernand0 (approved):
🥳 Nice!
It seems that the `generateZosmfOptions` already accepted the `allocationSpaceUnit` option. 
So, sound like this enhancement is ready to go! 😋 

UPDATE: After testing it out on a live system, seems like the `--asu` flag works as expected 
`zdev files create pds --asu TRK test.dataset.name`
resulted in a dataset with `spacu: TRACKS` as one of the attributes (instead of the default `spacu: CYLINDERS`)

awharn (approved):

PR #1527, opened 2022-09-22
Release with GitHub Actions instead of Jenkins

t1m0thyj:
TODO:
* [ ] Verify that prune-shrinkwrap plugin is working correctly
* [ ] Test npm prepublish script for bundling Keytar/daemon
* [ ] Merge octorelease-script PR with major changes and update branch


.github/workflows/zowe-cli.yml
```diff
@@ -121,7 +128,6 @@ jobs:
       run: cargo test ${{ github.event.inputs.binary-type == 'release' && '--release' || '' }} --manifest-path=zowex/Cargo.toml -- --nocapture
 
     - name: Unit Tests
-      id: unit
```
zFernand0, line 124: I'm guessing there's no need to have IDs if we don't ask for their outcome 😋

.github/workflows/zowe-cli.yml
```diff
@@ -26,8 +26,11 @@ on:
 
 jobs:
   test:
-    name: Cross-Platform Test
+    if: (github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository) && !contains(github.event.head_commit.message, '[ci skip]')
```
zFernand0, line 29: Is there a difference in where the `if` is placed ?
Just curious.
I kind of like it being at the very top, makes "mental debugging" easier 😋

scripts/bundleDaemon.sh
```diff
@@ -1,17 +1,15 @@
 #!/bin/bash
-# Usage: bash bundleDaemon.sh <daemonVersion> [githubAuthHeader]
```
zFernand0, line 2: Can we add something here explaining that `$GITHUB_TOKEN` is a requirement if someone wanted to use this script locally ?

scripts/bundleKeytar.sh
```diff
@@ -1,15 +1,13 @@
 #!/bin/bash
-# Usage: bash bundleKeytar.sh <keytarVersion> [githubAuthHeader]
```
zFernand0, line 2: Same comment as before
Do you think it makes sense to add something about the ENV requirement ?

release.config.js
```diff
@@ -0,0 +1,35 @@
...
+        //     name: "next",
+        //     prerelease: true,
+        //     dependencies: { "@zowe/perf-timing": "latest", "@zowe/imperative: "next" }
+        // }
+    ],
+    plugins: [
+        "@octorelease/changelog",
+        ["@octorelease/lerna", {
+            aliasTags: {
+                "latest": ["zowe-v2-lts", "next"]
```
zFernand0, line 23: Q: Do we plan to remove the tag `next` once we decide to uncomment the `branches[2].name == "next"` 😋

packages/cli/package.json
```diff
@@ -42,7 +42,7 @@
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
   "scripts": {
-    "prepublishOnly": "node ./scripts/rewriteShrinkwrap && npm i --package-lock-only --ignore-scripts --no-audit && npm dedupe --production",
+    "prepublishOnly": "bash ../../scripts/bundleDaemon.sh && bash ../../scripts/bundleKeytar.sh",
```
zFernand0, line 45: Will this technically remove the ability for someone to create a zowe-cli.tgz locally ?
I guess we should only rely on the GH workflows, but should we have some sort of backup script equivalent to `prune-shrinkwrap`  ?

.github/workflows/zowe-cli.yml
```diff
@@ -136,26 +142,78 @@ jobs:
...
+
+    - name: Install Dependencies
+      run: npm ci
+
+    - name: Update Dependencies
+      uses: zowe-actions/octorelease-script@update-for-cli
+      env:
+        GIT_COMMITTER_NAME: ${{ secrets.ZOWE_ROBOT_USER }}
+        GIT_COMMITTER_EMAIL: ${{ secrets.ZOWE_ROBOT_EMAIL }}
+        NPM_RESOLUTIONS: ${{ needs.test.outputs.npm-resolutions }}
```
zFernand0, line 204: Offline Q: Does this mean that the `zowe-actions/octorelease-script:npmUpdate` sets the `output.results` to an object containing the `npm-resolutions` ?

(Just trying to understand. Sorry for all the dumb questions. I don't mind posting an update later)

.github/workflows/zowe-cli.yml
t1m0thyj: That was my thought as well 🙂 Doing a bit of cleanup on the workflow like we did for the Imperative one.

.github/workflows/zowe-cli.yml
t1m0thyj: I don't think it makes a difference, more just personal preference and consistency with other repos 😛

scripts/bundleDaemon.sh
t1m0thyj: It's not a requirement though 😛 I've successfully run the script locally without it.

release.config.js
t1m0thyj: That's right, maybe we should add a comment here to remind us?

.github/workflows/zowe-cli.yml
t1m0thyj: That's right, not a dumb question 🙂 The relevant changes are in this PR: https://github.com/zowe-actions/octorelease-script/pull/6

release.config.js
zFernand0: yeah, a comment might be nice! 😋 
Thank you!

KevinLoesch1 (approved):

zFernand0 (approved):
Changes LGTM!
Waiting until after the Zowe 2.4 Code Freeze

gejohnston (approved):

PR #1531, opened 2022-09-26
Updates to the `@zowe/cli-test-utils` package

zFernand0:
In this PR:
- Made all system and integration test shell scripts executable
- Made all system and integration test shell scripts use `#! /bin/bash` 
- Updated the `@zowe/cli-test-utils` package to:
  - Prevent creating profiles when skipping not requested
  - Run scripts directly, allowing them to decide the default shell based on the shebang
    - `win32` continues to use `sh` as the default


KevinLoesch1 (approved):

__tests__/__packages__/cli-test-utils/src/TestUtils.ts
```diff
@@ -35,12 +35,25 @@ export function runCliScript(scriptPath: string, testEnvironment: ITestEnvironme
...
+            // Execute the command synchronously
+            return spawnSync("sh", [`${scriptPath}`].concat(args), {
+                cwd: testEnvironment.workingDir,
+                env: childEnv,
+                encoding: "buffer"
+            });
+        }
+
+        // Check to see if the file is executable
+        try { fs.accessSync(scriptPath, fs.constants.X_OK) } catch {
```
t1m0thyj, line 48: Please fix lint warnings on this line 🙂

gejohnston (approved):

t1m0thyj (approved):
Thanks for making all the shell scripts executable 👍 

Curious about the difference between the shebang lines `#!/usr/bin/env bash` and `#!/bin/bash` 😛 I think I've seen that the first one works in more environments, but the second one works everywhere we need it to, is easier to type, and the one I typically use 🤷

zFernand0:
> difference between the shebang lines `#!/usr/bin/env bash` and `#!/bin/bash`

yeah, before doing this I had to do some ["research" (stackoverflow)](https://stackoverflow.com/questions/16365130/what-is-the-difference-between-usr-bin-env-bash-and-usr-bin-bash). 😋 
TL;DR: 
- `#!/usr/bin/env NAME` searches for the first `NAME` in the `$PATH`
- `#!/bin/NAME` explicitly uses `/bin/NAME` (which could also be the first thing in the `$PATH`)

P.S. Using `#!/usr/bin/env bash` can lead to different results because of possible different interpreters.

awharn (approved):

PR #1537, opened 2022-10-04
zowe files view ds ... --range

dkelosky:
image

### DIFF
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Allow `zowe files view ds ... --range SSS-EEE | SSS,NNN`.
+
 ## `7.6.2`
 
 - BugFix: Updated `minimatch` and `keytar` dependencies for technical currency.
--- a/packages/cli/src/zosfiles/view/ds/Dataset.definition.ts
+++ b/packages/cli/src/zosfiles/view/ds/Dataset.definition.ts
@@ -42,7 +42,8 @@ export const DatasetDefinition: ICommandDefinition = {
         ViewOptions.binary,
         ViewOptions.encoding,
         ViewOptions.record,
-        ViewOptions.volume
+        ViewOptions.volume,
+        ViewOptions.range
     ],
     examples: [
         {
--- a/packages/cli/src/zosfiles/-strings-/en.ts
+++ b/packages/cli/src/zosfiles/-strings-/en.ts
@@ -718,7 +718,9 @@ export default {
                 " to the data. The data transfer process returns each line as-is, without translation. No delimiters are added between records." +
                 " Conflicts with binary.",
             VOLUME: "The volume serial (VOLSER) where the data set resides. You can use this option at any time. However, the VOLSER is required " +
-            "only when the data set is not cataloged on the system. A VOLSER is analogous to a drive name on a PC."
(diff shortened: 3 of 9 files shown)


packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Allow `zowe files view ds ... --range SSS-EEE | SSS,NNN`.
```
zFernand0, line 7: should there be an issue associated with this changelog (and the PR too) ?

zFernand0 (approved):
Changes LGTM! 😋 
- Can we add examples for this command ?
- Can we add a small system test for this ?
  - I don't think there will be any terminal-specific escaping problems with the `sss-eee` and `sss,nnn` syntax

dkelosky:
> Changes LGTM! 😋 (and the functionality works as expected)
> 
> * Can we add examples for this command ?
> * Can we add a small system test for this ?
>   
>   * I don't think there will be any terminal-specific escaping problems with the `sss-eee` and `sss,nnn` syntax

ill try to figure out how to system test 😄 - by examples, do you mean in the help?

packages/cli/CHANGELOG.md
zFernand0: yeah 😋

zFernand0:
> ill try to figure out how to system test 😄 - by examples, do you mean in the help?

yeah, examples in the help text

and thanks for considering adding system tests 😋

packages/cli/src/zosfiles/view/ds/Dataset.definition.ts
```diff
@@ -52,6 +53,14 @@ export const DatasetDefinition: ICommandDefinition = {
         {
             description: strings.ACTIONS.DATA_SET.EXAMPLES.EX2,
             options: `"ibmuser.test.loadlib(main)" --binary`
+        },
+        {
+            description: strings.ACTIONS.DATA_SET.EXAMPLES.EX3,
+            options: `"ibmuser.cntl(iefbr14)" --range 0,2`
```
gejohnston, line 59: As a poor man's UX comment, should we make the range be 1 to n instead of 0 to n - 1, and then convert the user-supplied values into zero-based values for the IBM header?

If other feel that my suggestion is misguided, I am ready to approve the implementation.

gejohnston (commented):
I left one thought for your consideration.

packages/cli/src/zosfiles/view/ds/Dataset.definition.ts
dkelosky: that's a fair point about UX; on that side i have no strong opinion.  

the implementation is a little tricky though, and so i avoided it.  right now, we just pass through whatever the user input was, without validation or modification.  to alter the UX we'd need to distinguish between `SSS,NNN` and `SSS-NNN`.  in the former, decrement only `SSS` by one; in the latter, decrement `SSS` and `NNN` by one.  Preceding zeros are allowed for both `SSS` and `NNN`, and so are other special case situations, e.g. `--range 1-0` (return the entire file except the first row).  so, if the value is zero, we would not decrement.  

given that it makes the implementation more difficult, i lean towards keeping it as is.  what are your thoughts?

gejohnston: I am not a fan of making things more complicated. It frequently adds more edge-case problems. I think we have other commands that retain the characteristics of the underlying zosmf API. I wanted to ensure that we thought about our choices, but I have no objection to retaining the zero-based numbering.

gejohnston (approved):

PR #1541, opened 2022-10-10
Feature/new command zosjobs modify

ATorrise:
Added new command, `zowe jobs modify job [jobid]` with options `--jobclass`, `--hold`, and `--release`. Can now
change the class of a job that has yet to run and also release or hold the running of jobs.

Can choose to change both options (jobclass and hold-status) or just one option. Requires jobid. 

exs:
`zowe jobs modify job JOB87326 --jobclass C --hold`
`zowe jobs modify job JOB87326 --jobclass A`
`zowe jobs modify job JOB87326 --release`

https://www.ibm.com/docs/en/zos/2.4.0?topic=interface-hold-job 
https://www.ibm.com/docs/en/zos/2.4.0?topic=interface-release-job
https://www.ibm.com/docs/en/zos/2.4.0?topic=interface-change-job-class


packages/zosjobs/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS jobs SDK package will be documented in this file.
 
+## Recent Changes 
+- Enhancement: Added new command, `zowe jobs modify job [jobname] [jobid]` with options `--jobclass` and `--hold-status`. Can now
```
gejohnston, line 6: In this directory, instead of the phrase "Added new command", should it be something like "Added new functions to support ... whatever" ?

packages/zosjobs/src/ModifyJobs.ts
```diff
@@ -0,0 +1,107 @@
...
+     * Full version of the API with a parameter object
+     * @static
+     * @param {AbstractSession} session - z/OSMF connection info
+     * @param {IModifyJobParms} parms - parm object (see IModifyJobParms interface for details)
+     * @returns {Promise<undefined|IJobFeedback>} - promise of undefined, or IJobFeedback object returned by API if modifyVersion is 2.0
+     * @memberof ModifyJobs
+     */
+    public static async modifyJobCommon(session: AbstractSession, parms: IModifyJobParms): Promise<undefined|IJobFeedback> {
+        this.log.trace("ModifyJobCommon called with parms %s", JSON.stringify(parms));
+        ImperativeExpect.keysToBeDefinedAndNonBlank(parms, ["jobid"],
```
gejohnston, line 68: This only tests for jobid, but the error message states that jobname and jobid must be supplied. What will happen when no jobname is supplied?

packages/zosjobs/src/ModifyJobs.ts
```diff
@@ -0,0 +1,107 @@
...
+                class: parms.jobclass,
+            };
+            const responseJsonClass = await ZosmfRestClient.putExpectJSON(session, JobsConstants.RESOURCE + parameters, headers, request);
+            response = responseJsonClass as IJobFeedback;
+        }
+
+        // build request to change holdStatus, if defined
+        if (parms.holdstatus != "undefined"){
+            request = {
+                request: parms.holdstatus,
```
gejohnston, line 89: I have not looked at the zosmf API, so this is just a question: is the request property name for holdstatus really named "request" ?

gejohnston (changes requested):
I left a couple of specific questions.
Also, it looks like you must run lint against the source and make the changes it suggests.

ATorrise:
zosmf doesn't support changing `job status` and `job class` at the same time so this PR requires a implementation discussion. Response message gets overridden and current method is to merge response messages into 1.

packages/zosjobs/src/ModifyJobs.ts
ATorrise: yes haha

packages/zosjobs/src/ModifyJobs.ts
ATorrise: Fixed to include jobname in keysToBeDefinedAndNonBlank

packages/zosjobs/src/ModifyJobs.ts
```diff
@@ -0,0 +1,118 @@
...
+ */
+export class ModifyJobs {
+
+    /**
+     * Modify a job
+     * @static
+     * @param {AbstractSession} session - z/OSMF connection info
+     * @param {string} jobname - job name to be translated into parms object
+     * @param {string} jobid - job id to be translated into parms object
+     * @param {string} jobclass - job class to be translated into parms object
```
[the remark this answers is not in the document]
traeok: I would agree that we should probably pass an object to specify options for `modifyJob`, especially since an options object is constructed anyway before `modifyJobCommon` is called. By standardizing to one options object, this would prevent potential mishaps when calling this function internally - such as a missing `undefined` in the parameter list, or if the function was called with `hold/release` passed in the wrong order.

traeok (approved):
See my comment about specifying `modifyJob` parameters - it wouldn't hurt to keep the function parameters as-is, but a comment would serve as a helpful reminder of how to skip parameters for this function. 

Also, for the tests you can probably omit some of the comments for self-describing code (e.g. "Test Environment populated in the beforeAll" in `cli.zos-jobs.modify.job.integration.test.ts`) but  that's not really an important change, just a suggestion.

LGTM

zFernand0:
**TL;DR:** The functionality works as expected, and LGTM! 😋 

---

I agree with the comments of using an interface instead of having to provide null or undefined a ton of times.
I'm curious about the output of the command

```
$ zdev jobs modify job job08733 --jobclass C --hold

Request was successful.
Request was successful.
Class Change: B -> C
Job Held

$ zdev jobs modify job job08733 --jobclass XYZ --release

Request was successful.
Job class invalid
Class Change: C -> XYZ
Job Released
```

It seems that the first `request was successful` is related to the `hold/release` operation, and the second one is for the class change.

Wondering if they should be removed from the API response, or if we could add more details and maybe separate the responses for each request?

---

Example CLI output:
```
$ zdev jobs modify job job08733 --jobclass C --hold

Job hold/release request succeeded .
Job class change request succeeded .
Class Change: B -> C

$ zdev jobs modify job job08733 --jobclass XYZ --release

Job hold/release request succeeded .
Job class change request failed: Job class invalid
Class Change: C -> XYZ
```

---

Suggestion for providing more details from the API:
```typescript
// add an optional property to the IJobFeedback interface (e.g. `responses?: {IModifyJobDetailedResponses}[]`)

// inside IJobFeedback.ts
    /**
     * Detailed responses
     * @type {IModifyJobDetailedResponses[]}
     */
    responses?: IModifyJobDetailedResponses[];

// new interface IModifyJobDetailedResponses.ts
    /**
     * Success
     * @type {boolean}
     */
    success: boolean;
    /**
     * Request type
     * @type {string}
     */
    requestType: string;
    /**
     * Message
     * @type {string}
     */
    message: string;  

// and populate it for the `modifyJob` SDK response.

// inside ModifyJobs.ts
    let response: IJobFeedback
    response.responses = [];
    // ...
    response.responses.push({ success: !exception, request: "holdRelease", message: response.message});
    // ...
    if (parms.jobclass != undefined && !exception){
        // ...
        response.responses.push({ success: !exception, request: "jobClass", message: response.message});
        
```

zFernand0 (changes requested):
Sorry, I did have one requested change 😓

packages/zosjobs/src/ModifyJobs.ts
```diff
@@ -0,0 +1,118 @@
...
+
+        if(parms.hold || parms.release){
+            parms.hold ? request = { request: "hold"} : request = { request: "release"};
+            try{
+                response = await ZosmfRestClient.putExpectJSON(session, JobsConstants.RESOURCE + parameters, headers, request);
+            }
+            catch(err){
+                exception = true;
+                err.mMessage=err.mMessage.concat('Modification Error');
+                return err;
```
zFernand0, line 94: Please consider throwing the error 😋 

---

`When` using the API `ModifyJobs.modifyJob(...)`
`If   `&nbsp;&nbsp; the first request (hold/release) throws an error (e.g ECONNREFUSED)
`Then` the type of the resolved value is not `IJobFeedback` anymore.

This results in an error like the one bellow
```
Unexpected Command Error:
Please review the message and stack below.
Contact the creator of handler:
"/root/gh/zowe/zowe-cli/packages/cli/lib/zosjobs/modify/job/Job.handler"
Message:
Cannot set property message of Error which has only a getter
Stack:
TypeError: Cannot set property message of Error which has only a getter
    at JobHandler.<anonymous> (/root/gh/zowe/zowe-cli/packages/cli/lib/zosjobs/modify/job/Job.handler.js:67:34)
    at Generator.next (<anonymous>)
    at fulfilled (/root/gh/zowe/zowe-cli/packages/cli/lib/zosjobs/modify/job/Job.handler.js:15:58)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

gejohnston (approved):

ATorrise:
> **TL;DR:** The functionality works as expected, and LGTM! 😋
> 
> I agree with the comments of using an interface instead of having to provide null or undefined a ton of times. I'm curious about the output of the command
> 
> ```
> $ zdev jobs modify job job08733 --jobclass C --hold
> 
> Request was successful.
> Request was successful.
> Class Change: B -> C
> Job Held
> 
> $ zdev jobs modify job job08733 --jobclass XYZ --release
> 
> Request was successful.
> Job class invalid
> Class Change: C -> XYZ
> Job Released
> ```
> 

Hey @zFernand0, sorry that I didn't catch this earlier. I've fixed the response so that now, if you were to run the following command, you would get this output: 

$ zdev jobs modify job job08733 --jobclass XYZ --release

Unsuccessful. Job class 'XYZ' invalid
Successful. Job Released

packages/zosjobs/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe z/OS jobs SDK package will be documented in this file.
 
+## Recent Changes 
+- Enhancement: Added new functions to support the changing of a job class and the hold status of a job. Can now call `zowe jobs modify job [jobid]` with options `--jobclass`, `--hold` and `--release`. [#1156](https://github.com/zowe/zowe-cli/issues/1156)
```
t1m0thyj, line 6: For the SDK changelog, we should mention new methods that were added (like `ModifyJobs.modifyJob`), rather than CLI commands which are only applicable to the "@zowe/cli" package.

packages/zosjobs/src/index.ts
```diff
@@ -11,6 +11,8 @@
 
 export * from "./doc/input/ICancelJob";
 export * from "./doc/input/ICancelJobParms";
+export * from "./doc/input/IModifyJob";
+export * from "./doc/input/IModifyJobParms";
```
t1m0thyj, line 15: Please add an export for the new `IModifyJobOptions` interface 🙂

packages/cli/__tests__/zosjobs/__system__/modify/__scripts__/job/real_jobid.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# Max attempt count
+ATTEMPTS=10
```
awharn, line 4: Where are ATTEMPTS and WAIT used in these scripts?

packages/zosjobs/src/ModifyJobs.ts
```diff
@@ -0,0 +1,108 @@
...
+     * Modify a job
+     * @static
+     * @param {AbstractSession} session - z/OSMF connection info
+     * @param {IModifyJobParms} parms - parms object (see IModifyJobParms interface for details)
+     * @param {IModifyJobOptions} options - options object (see IModifyJobOptions interface for details)
+     * @returns {Promise<undefined|IJobFeedback>} - promise of undefined, or IJobFeedback object returned by API if modifyVersion is 2.0
+     * @memberof ModifyJobs
+     */
+    public static async modifyJob(session: AbstractSession,
+        parms: { jobname: string, jobid: string },
```
awharn, line 38: Should this be using IModifyJobParms and IModifyJobOptions instead so they inherit any future changes?

packages/zosjobs/src/ModifyJobs.ts
```diff
@@ -0,0 +1,108 @@
...
+            response = await ZosmfRestClient.putExpectJSON(session, JobsConstants.RESOURCE + parameters, headers, request);
+            mergedMessage = mergedMessage + '\n' + response.message;
+        }
+        response.message = mergedMessage;
+        return response;
+    }
+
+
+    /**
+     * Getter for brightside logger
```
awharn, line 102: This code shouldn't mention brightside. It should just be a logger, since any app could theoretically use Imperative and have its own app logger.

packages/cli/src/zosjobs/modify/job/Job.handler.ts
```diff
@@ -0,0 +1,67 @@
...
+     * Command handler process - invoked by the command processor to handle the "zos-jobs modify job"
+     * @param {IHandlerParameters} params - Command handler parameters
+     * @returns {Promise<void>} - Fulfilled when the command completes successfully OR rejected with imperative error
+     * @memberof JobHandler
+     */
+    public async processCmd(params: IHandlerParameters): Promise<void> {
+        this.arguments = params.arguments;
+
+        // Force yargs parameters to be proper types
+        const jobid: string = this.arguments.jobid + "";
```
awharn, line 41: Just a suggestion, but you could just append `.toString()` to verify this is a string in most cases if you are trying to force them to be the proper type. This doesn't need to be implemented since it is documented what the `+ ""` is doing. 😃

awharn (commented):
A couple of small comments and questions.

packages/zosjobs/src/doc/response/IJob.ts
```diff
@@ -34,6 +34,20 @@ export interface IJob {
...
+     */
+    hold?: boolean;
+
+    /**
+     * flag that releases a held a job for execution
+     * @type {boolean}
+     * @memberof IJob
+     */
+    release?: boolean;
+
```
t1m0thyj, line 50: Is it needed to add the "hold" and "release" properties to the `IJob` interface? I may be missing something here 😢 I'm wondering if we don't need them here since they are defined on `IModifyJobOptions` now.

t1m0thyj (commented):
Please address the code smells reported by SonarCloud. Left one more comment, sorry I didn't notice it before. Everything else LGTM 🙂

awharn (approved):
Aside from the trailing spaces lint errors, this looks good to me! Great new functionality. 🚀

PR #1547, opened 2022-10-24
Release with GitHub Actions instead of Jenkins

t1m0thyj:
Ports #1527 from `master` to `zowe-v1-lts` branch


.github/workflows/zowe-cli.yml
```diff
@@ -22,8 +22,10 @@ on:
         required: false
 jobs:
   test:
-
+    if: (github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository) && !contains(github.event.head_commit.message, '[ci skip]')
     runs-on: ${{ matrix.os }}
+    outputs:
+      npm-resolutions: ${{ steps.npm-update.outputs.result }}
```
zFernand0, line 28: is this a requirement ?
or can the action just set the output without having to explicitly call it out here ?
(the reason why I'm asking is because we might need to add this to the plugins 😋 )

zFernand0 (approved):
LGTM!

.github/workflows/zowe-cli.yml
t1m0thyj: The action can set the output on the step level, but in order to set the output on the job level it has to be explicitly called out like this. See https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs 🙂

gejohnston (approved):

PR #1554, opened 2022-10-31
Remove gulp and ts-node dev dependencies

awharn:
Removed the dependency on gulp, all gulp plug-ins, and ts-node, which was only used by gulp.
Removed 355 dev dependencies.
Removed cleanTypedoc functionality, which looked for a link that is no longer present in typedoc and replaced it.
Removed doc script, which generated CLIReadme, which has not been updated in 16 months and was replaced with the web-help generator and PDF.
Removed the out of date CLIReadme document.
Migrated the update licenses functionality to a plain JavaScript script.


package.json
```diff
@@ -30,9 +30,7 @@
     "watch": "lerna run --parallel watch",
     "watch:exe": "cd zowex && cargo install cargo-watch && cargo watch -x build -x clippy -x test",
     "watch:test": "jest --watch",
-    "doc:clean": "rimraf docs/CLIReadme.md",
-    "doc:generate": "npm run doc:clean && gulp doc",
-    "generateCleanTypedoc": "npm run typedoc && gulp cleanTypedoc",
+    "generateCleanTypedoc": "npm run typedoc",
```
zFernand0, line 33: should we remove this script too since it's basically just calling typedoc 😋

zFernand0 (approved):
LGTM! 😋

package.json
awharn: I didn't want to just in case it was being called somewhere, but we can if you prefer.

t1m0thyj (approved):
🦀 Overrides be gone 🦀

gejohnston (approved):

t1m0thyj:
Should we update the contributing guidelines to remove mention of gulp?
https://github.com/zowe/zowe-cli/blob/763da03c29457c4f642ed88b814da753b60c88c0/CONTRIBUTING.md?plain=1#L182

awharn:
> Should we update the contributing guidelines to remove mention of gulp?
> 
> https://github.com/zowe/zowe-cli/blob/763da03c29457c4f642ed88b814da753b60c88c0/CONTRIBUTING.md?plain=1#L182

I think we can do that, and I have done so on my local branch. Is there anything else that should be inserted instead? Or should we just have `Use build tasks to enforce rules where possible.` on the `Build Process Guidelines` section?

t1m0thyj:
> I think we can do that, and I have done so on my local branch. Is there anything else that should be inserted instead? Or should we just have `Use build tasks to enforce rules where possible.` on the `Build Process Guidelines` section?

I'd be ok with removing the line. Not sure what to insert in place of gulp besides "We use npm scripts", which sounds kind of obvious.

PR #1559, opened 2022-11-03
Fix Sonar Hotspots v1 - Part 1

awharn:
Fixes the pipeline so Sonar's scans can locate all of the hotspots in the v1 branch. Implements fixes in V2 that are known to be hotspots. Once Sonarcloud is reconfigured, this pull request can be merged, but the branch SHOULD NOT be deleted, as a second pull request may be required to fix additional hotspots after the scans are complete.

### DIFF
--- a/.github/workflows/zowe-cli.yml
+++ b/.github/workflows/zowe-cli.yml
@@ -51,7 +51,7 @@ jobs:
         fetch-depth: 0
 
     - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v1
+      uses: actions/setup-node@v3
       with:
         node-version: ${{ matrix.node-version }}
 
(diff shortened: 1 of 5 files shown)


.github/workflows/zowe-cli.yml
```diff
@@ -47,6 +47,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
 
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v1
```
zFernand0, line 54: should we update v1 workflows setup-node to v3 ?

zFernand0 (approved):
LGTM! 😋 
These changes should not affect the overall behavior of the CLI. 👍

gejohnston (approved):

t1m0thyj (approved):

PR #1572, opened 2022-11-21
Update automatic comments on new issues

t1m0thyj:
Thanks @anaxceron for providing the updated wording 🙂

### DIFF
--- a/.github/label-actions.yml
+++ b/.github/label-actions.yml
@@ -1,8 +1,11 @@
 bug:
   comment: |
-    Thank you for creating a bug report. If you haven't already, please ensure you have provided steps to reproduce it and as much context as possible.
+    Thank you for creating a bug report.
+    We will investigate the bug and evaluate its impact on the product.
+    If you haven't already, please ensure you have provided steps to reproduce the bug and as much context as possible.
 enhancement:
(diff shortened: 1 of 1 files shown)


.github/label-actions.yml
```diff
@@ -1,8 +1,11 @@
...
+    We will investigate the bug and evaluate its impact on the product.
+    If you haven't already, please ensure you have provided steps to reproduce the bug and as much context as possible.
 enhancement:
   comment: |
-    Thank you for raising this issue. 
-    The community has 90 days to upvote 👍 the issue.
-    If it receives 5 upvotes, we will move it to our backlog. If not, we will close it.
\ No newline at end of file
+    Thank you for raising this enhancement request.
+    The community has 90 days to vote on it.
```
t1m0thyj, line 9: Should we still include the emoji here (👍) to make it clear what kind of upvotes we are talking about, or is it obvious enough without that?

anaxceron: I think it's clear enough w/o the emoji.

anaxceron (approved):
Thanks for doing this -- all good!

KevinLoesch1 (approved):

PR #1576, opened 2022-12-01
Bug/zosjobs modify

ATorrise:
Addressing the concern found by coverity scan


packages/zosjobs/src/ModifyJobs.ts
```diff
@@ -70,37 +70,34 @@ export class ModifyJobs {
...
-
-        if(options.hold && options.release){
-            throw new ImperativeError({msg: "Parameters `hold` and `release` are in conflict and cannot be specified together"});
-        }
-
-        if(options.hold || options.release){
-            request = { request: options.hold ? "hold" : "release" };
-            try{
-                response = await ZosmfRestClient.putExpectJSON(session, JobsConstants.RESOURCE + parameters, headers, request);
+        if (Object.keys(options).length != 0){
```
awharn, line 73: This may not handle the case if `options.hold` and `options.release` are (individually or both) defined as false, and `options.jobclass` is undefined. In that case, the length will be 1 or 2, and it may try to define `response.message` on an undefined object.

Some validation may need to be done along the lines of `if (options.release || options.hold || options.jobclass != null)` or something of that sort.

ATorrise: true!

awharn: The last check I might add (now that we are onto fixing edge cases) is to check to make sure that `options.jobclass` is not an empty string. I don't think that an empty string would be a valid jobclass. That being said, response would be defined in that case, so I think you have everything covered and tested when it comes to scans.

I will leave this discussion open for others for their input.

traeok (approved):
LGTM! The updated check for the options object should now avoid that Coverity issue.

gejohnston (approved):
As soon as we satisfy Coverity, SonarCloud complains about the corrections that were made to satisfy Coverity. Thanks for chasing down all of the complaints.

packages/zosjobs/src/ModifyJobs.ts
t1m0thyj: I agree that these changes should fix the Coverity issue 👍 To handle all edge cases correctly, I also agree with Andrew's suggestion 🙂 

Would it make sense to make the following changes, to throw an error in the case of an empty string ?

---

https://github.com/zowe/zowe-cli/blob/26b0d3245e923636af31bb5abbd2b9903791f26d/packages/zosjobs/src/ModifyJobs.ts#L72

Suggested change: `if (options.release || options.hold || options.jobclass) { `

So that if jobclass is an empty string, the following exception would be thrown:
`At least one option needs to be selected for modification: 'hold', 'release', 'jobclass'`

---

https://github.com/zowe/zowe-cli/blob/26b0d3245e923636af31bb5abbd2b9903791f26d/packages/zosjobs/src/ModifyJobs.ts#L89

This could then be simplified to: `if (options.jobclass) {`

ATorrise: someone might pass an empty string for jobclass here while release and hold are undefined. there is an eventual error that zosmf throwd for any invalid class, including an empty string: "Job class invalid"

t1m0thyj (approved):

PR #1579, opened 2022-12-08
Bug fix/response timeout

ATorrise:
tiny changes to respect responseTimeout

### DIFF
--- a/packages/zosfiles/src/methods/mount/Mount.ts
+++ b/packages/zosfiles/src/methods/mount/Mount.ts
@@ -58,7 +58,7 @@ export class Mount {
 
         const jsonContent = JSON.stringify(tempOptions);
         const headers = [{"Content-Length": jsonContent.length}, ZosmfHeaders.ACCEPT_ENCODING];
-        if (options.responseTimeout != null) {
+        if (options && options.responseTimeout != null) {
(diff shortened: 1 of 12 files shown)


packages/zosfiles/src/methods/mount/Mount.ts
```diff
@@ -58,7 +58,7 @@ export class Mount {
 
         const jsonContent = JSON.stringify(tempOptions);
         const headers = [{"Content-Length": jsonContent.length}, ZosmfHeaders.ACCEPT_ENCODING];
-        if (options.responseTimeout != null) {
+        if (options && options.responseTimeout != null) {
```
zFernand0, line 61: Thanks for adding the check for the Mount APIs
If you are feeling lazy and don't want to type as much, you could also use the `?.` operator
like: `options?.responseTimeout` 

Will adding this extra check decrease branch coverage for this file? 
🤔

zFernand0 (approved):
LGTM! 😋

packages/zosfiles/src/methods/mount/Mount.ts
ATorrise: same code coverage % woot

awharn (approved):

t1m0thyj (approved):

PR #1580, opened 2022-12-12
Consolidate tests

ATorrise:


zFernand0 (dismissed):
LGTM! 😋

packages/zosjobs/__tests__/__unit__/MonitorJobs.unit.test.ts
```diff
@@ -12,19 +12,22 @@
...
+const LONGER_TIMEOUT = 20000;
+
 describe("MonitorJobs", () => {
     // Use this so that editors don't complain about us accessing private stuff
     const privateMonitorJobs = MonitorJobs as any;
 
     /*************************************************************/
     // Unit tests for the file constants and defaults
-    describe("constant defaults", () => {
+    describe("constant defaults", () => { //GOOD
```
t1m0thyj, line 30: Should these `//GOOD` comments be removed?

packages/zosjobs/__tests__/__unit__/MonitorJobs.test.ts
```diff
@@ -1,1193 +0,0 @@
...
-                } catch (e) {
-                    error = e;
-                }
-                expect(error).toBeDefined();
-                expect(error instanceof ImperativeError).toBe(true);
-                expect(error.message).toMatchSnapshot();
-                expect(mockedGetJobsCommon).toHaveBeenCalledTimes(attempts);
-            });
-
-            it("should expire after the max attempts and the total time should exceed the attempts multiplied by the delay", async () => {
```
t1m0thyj, line 1021: Has this test been migrated? I don't see an equivalent one in `MonitorJobs.unit.test.ts`

t1m0thyj (commented):
Thanks for doing this 🙂 Left a few comments

I noticed a number of tests for `waitForOutputStatus` and `waitForJobOutputStatus` have been removed with no direct equivalent. I believe this should be ok as we have similar tests on `waitForStatusCommon` that achieve an equal amount of code coverage.

packages/zosjobs/__tests__/__unit__/MonitorJobs.unit.test.ts
ATorrise: yes lol

packages/zosjobs/__tests__/__unit__/MonitorJobs.unit.test.ts
```diff
@@ -12,19 +12,22 @@
...
 
-import { ImperativeError } from "@zowe/imperative";
+import { ImperativeError, Session } from "@zowe/imperative";
 import { JOB_STATUS, MonitorJobs, GetJobs } from "../../src";
 import { sleep } from "@zowe/core-for-zowe-sdk";
 import { IMonitorJobWaitForParms } from "../../src/doc/input/IMonitorJobWaitForParms";
 import { IJob } from "../../src/doc/response/IJob";
 
+// Longer timeout for async poll test
+const LONGER_TIMEOUT = 20000;
```
gejohnston, line 22: LONGER_TIMEOUT never seems to be used. Do you still need it?

gejohnston (approved):
I just asked 1 question.

packages/zosjobs/__tests__/__unit__/MonitorJobs.unit.test.ts
```diff
@@ -614,5 +655,368 @@ describe("MonitorJobs", () => {
...
+                GetJobs.getStatusCommon = mockedGetJobs;
+
+                // IJob requires more fields to be populated - fake it out with only the required here.
+                const iJobParms: any = {jobname: "FAKE", jobid: "FAKE"};
+                const response = await MonitorJobs.waitForJobOutputStatus(new Session({hostname: "fake", port: 443}), iJobParms);
+                expect(response).toMatchSnapshot();
+                expect(mockedGetJobs).toHaveBeenCalledTimes(1);
+            });
+
+            it("should return immediately if the initial status is the expected status", async () => {
```
awharn, line 674: Is this the same as the test above?

awharn (commented):
Just have one additional question.

packages/zosjobs/__tests__/__unit__/MonitorJobs.unit.test.ts
```diff
@@ -286,22 +443,129 @@ describe("MonitorJobs", () => {
...
+                        expectError = e;
+                    }
+
+                    expect(expectError).toBeInstanceOf(Error);
+                    expect(expectError.message).toMatchSnapshot();
+                });
 
-                const jobname = "JOB52712";
-                const jobid = "52712";
+                it("should throw the proper error if catching a wrapped error", async () => {
```
t1m0thyj, line 550: I think this test is redundant with "should throw the proper error if catching an ImperativeError", since a "wrapped error" means the same thing as an "Imperative error".

packages/zosjobs/__tests__/__unit__/MonitorJobs.unit.test.ts
```diff
@@ -316,302 +580,563 @@ describe("MonitorJobs", () => {
...
+                    } catch (e) {
+                        expectError = e;
+                    }
+
+                    expect(expectError).toBeDefined();
+                    expect(expectError).toBeInstanceOf(Error);
+                    expect(expectError.message).toMatchSnapshot();
+                });
+
+                it("should throw the proper error if catching a wrapped error", async () => {
```
t1m0thyj, line 672: Same comment as above about test for wrapped error potentially being redundant.

t1m0thyj (commented):
Left a few comments. Also have a few questions - have the following tests been implemented? They are on the master branch, and not sure where they are present in the new consolidated file.
```
 PASS  packages/zosjobs/__tests__/__unit__/MonitorJobs.test.ts (64.714 s)
  Monitor Jobs
    api method wait for status common (1 ms)
      initial status check
        √ should return immediately if the initial status is INPUT and the expected status is INPUT (3 ms)


 PASS  packages/zosjobs/__tests__/__unit__/MonitorJobs.unit.test.ts (8.32 s)
  MonitorJobs
    Public Methods
      waitForJobOutputStatus
        √ should throw an error when job is not passed (1 ms)
```

zFernand0 (commented):
the changes and the test LGTM! 😋 

I have a small request.
Could we try to `expect` certain strings in the responses instead of always falling back to snapshots?

Not saying they are bad, but it's possible that an issue may not get caught in the future because it's not an uncommon practice to run `jest -u` which will update all snapshots (even though the output may not be the desired one)

packages/zosjobs/__tests__/__unit__/MonitorJobs.unit.test.ts
ATorrise: will remove wrapped test

ATorrise:
> Left a few comments. Also have a few questions - have the following tests been implemented? They are on the master branch, and not sure where they are present in the new consolidated file.
> 
> ```
>  PASS  packages/zosjobs/__tests__/__unit__/MonitorJobs.test.ts (64.714 s)
>   Monitor Jobs
>     api method wait for status common (1 ms)
>       initial status check
>         √ should return immediately if the initial status is INPUT and the expected status is INPUT (3 ms)
> 
> 
>  PASS  packages/zosjobs/__tests__/__unit__/MonitorJobs.unit.test.ts (8.32 s)
>   MonitorJobs
>     Public Methods
>       waitForJobOutputStatus
>         √ should throw an error when job is not passed (1 ms)
> ```

Thank you for catching this first INPUT/INPUT test. I've just added it. For the latter, ive changed the wording of the test to be: "should detect missing IJob"

ATorrise:
> the changes and the test LGTM! 😋
> 
> I have a small request. Could we try to `expect` certain strings in the responses instead of always falling back to snapshots?
> 
> Not saying they are bad, but it's possible that an issue may not get caught in the future because it's not an uncommon practice to run `jest -u` which will update all snapshots (even though the output may not be the desired one)

I agree that snapshots can be a nuisance. I'll try to remove some of them, will edit this comment if successful.

packages/zosjobs/__tests__/__unit__/MonitorJobs.unit.test.ts
ATorrise: yes, ive merged the new test title to be: "should return immediately if the initial status is the expected status, ie: OUTPUT/ OUTPUT"

t1m0thyj (approved):

zFernand0 (approved):
LGTM! 😋

PR #1583, opened 2022-12-13
removing line break encodings

ATorrise:


traeok (changes requested):
Thanks for looking into this 👍 Seems like the replace is not matching on my end. Here is the output I'm getting:

image

I believe it is because the regex needs to check for characters before/after the line breaks or they won't be matched:

image

@t1m0thyj suggested that the function `IO.processNewLines` from imperative can be useful here, that should help match line endings between both strings.

Or the regex could be adjusted - also, we probably just want to replace the Windows line breaks w/ the UNIX `\n` instead of removing them, since we are still displaying that output in the console. Kind of a verbose string but an example:

image

traeok (approved):
Working as expected - LGTM! Thanks @ATorrise for the fix

zFernand0 (approved):

PR #1588, opened 2022-12-15
Repurpose in-mem creds doc to NodeJS.Crypto tech

gejohnston:
The in-memory credentials design documentation has been repurposed to be a record of the research into NodeJS crypto techniques that could be used to store secure properties to disk.

Many of the previous use-cases were modified. The use of daemon mode was removed. The record of our investigation into the underlying crypto technology remains.

These updates will allow this documentation to remain as a relevant historical document, as we pursue another approach to storing secure data.


docs/Design_for_securing_CLI_properties_with_NodeJS_Crypto.md
```diff
@@ -0,0 +1,310 @@
...
+### **Support for USS** ###
+
+Another scenario for which a NodeJS Crypto solution could provide value is the use of Zowe CLI on USS. Because no Keytar binary is available for USS, customers must currently store credentials in plain text. A Zowe NodeJS Crypto solution will use the same NodeJS-supplied technology on every platform, including USS.
+
+## **Shortcomings of a NodeJS Crypto approach**
+Because a different approach is used for encrypting data, consumers will experience some operational changes after a NodeJS Crypto solution is delivered.
+
+### **Consumers must re-enter all secure values after upgrade** ###
+
+Because secure values will no longer be stored in the same vault, consumers must re-enter their values one time so that they are available for future use by Zowe CLI commands.
```
traeok, line 34: If we adopt the Node.js crypto functions in the future, we could probably use the Rust library to migrate these credentials and avoid re-entry (at least for the operating systems that the Rust implementation is compatible with).

LGTM!

traeok (approved):

zFernand0 (approved):

PR #1594, opened 2023-01-04
Attempt to fix the daemon broken pipe issue on Windows

### LINKED ISSUE #1538: I/O error during daemon communication - zowe files view ds large.data.set
For `zowe --version` = `7.6.1` if I run `KELDA16.PUBLIC.ASMPGM(CAPAAVL)`, I get the full file but then an error at the end:

```
I/O error during daemon communication.
Details = The pipe has been ended. (os error 109)
```

It seems like it fails for large files, ~5k lines long; however, I can't reproduce the problem on some exact data set file length.  

**Describe the bug**

**Describe your environment**

**Additional context**

awharn:
Fixes #1538 

Stops the daemon client from polling the server once it receives an exit code, in addition to when the daemon receives a zero byte response from the server. This fixes an obscure situation in Windows where:

1) The daemon reads the named pipe until it is empty and the last character in the pipe is a form feed (0xC)
2) Windows closes the pipe (because it is empty and Node closed the connection once the pipe was empty)
3) The daemon processes the data it just received up to and including the line feed (final character in the pipe)
4) When the daemon attempts to read from the pipe again, instead a zero byte response, it gets a broken pipe error

### DIFF
--- a/.github/workflows/rust-cli-publish.yml
+++ b/.github/workflows/rust-cli-publish.yml
@@ -168,6 +168,8 @@ jobs:
 
     - name: Build
       run: cargo build --verbose --release
+      env:
+        RUSTFLAGS: "-Ctarget-feature=+crt-static"
 
     - name: Get Upload URL
       id: get_upload_url
--- a/.github/workflows/rust-cli.yml
+++ b/.github/workflows/rust-cli.yml
@@ -100,6 +100,8 @@ jobs:
 
     - name: Build
       run: cargo build --verbose --manifest-path=zowex/Cargo.toml
+      env:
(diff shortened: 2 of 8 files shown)


gejohnston (approved):

traeok (approved):

traeok (approved):

KevinLoesch1 (approved):

zFernand0:
Hey @anaxceron,
We may want to update the last bullet item in this section of the docs-site
https://docs.zowe.org/stable/user-guide/cli-using-using-daemon-mode#preparing-for-installation

We should mention that the VCRUNTIME140.dll is only required when using a CLI version 7.9.x or older

PR #1600, opened 2023-01-09
Don't require sleep member for zosjobs system tests

### LINKED ISSUE #1598: Add missing system test properties to default_properties.yaml
The `zosjobs.sleepMember` and `zosjobs.modifiedJobclass` properties were added in https://github.com/zowe/zowe-cli/commit/c14107f3469f69b478bf2b011aab85e537585892.

They should be added to this file: https://github.com/zowe/zowe-cli/blob/master/__tests__/__resources__/properties/default_properties.yaml#L92

t1m0thyj:
Resolves #1598 to fix our internal system test pipeline 🙂 

* Adds `zosjobs.modifiedJobclass` to default_properties.yaml
* Removes `zosjobs.sleepMember` from tests that used it because we can reuse IEFBR14 job (no sleep job is needed since z/OSMF is able to modify a job once its in OUTPUT stage)


packages/zosjobs/__tests__/__system__/ModifyJobs.system.test.ts
```diff
@@ -34,47 +33,46 @@ describe("Modify Jobs - System Tests", () => {
         systemProps = testEnvironment.systemTestProperties;
         REAL_SESSION = TestEnvironment.createZosmfSession(testEnvironment);
         account = systemProps.tso.account;
-        jobname = testEnvironment.systemTestProperties.zosmf.user;
         jobclass = testEnvironment.systemTestProperties.zosjobs.jobclass;
         modifiedJobClass = testEnvironment.systemTestProperties.zosjobs.modifiedJobclass;
```
zFernand0, line 37: Should we do something about the new properties added to the `default_parameters.yaml`?

For example, if it's missing from the `custom_...`, should we try to pull them from the `default_...`?
One way we could do it is in the `testEnvironment`... we could merge the systemTestProperties from `{...default, ...custom}` 😋

zFernand0 (approved):
LGTM! 😋

awharn (approved):

traeok (approved):

gejohnston (approved):

PR #1604, opened 2023-01-19
fix in employing context-lines feature for all diff/compare commands

ATorrise:
BugFix: Fix in employing `--context-lines` option for all diff/compare commands. Fixed broken `--seqnum` option implementation. Corrected and added unit, system and integration tests


KevinLoesch1 (approved):

zFernand0 (approved):
Thanks for fixing this.
LGTM! :yum:

t1m0thyj (approved):
Should we add unit tests for these commands verifying that the `contextLines` argument behaves as expected?

I'm also ok with merging the PR as-is since all coverage checks are met 🙂

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+- BugFix: Fix in employing `--context-lines` option for all diff/compare commands. Fixed broken `--seqnum` option implementation. Corrected and added unit, system and integration tests [#1529](https://github.com/zowe/zowe-cli/issues/1529)
```
t1m0thyj, line 6: We mention only user-facing changes in the changelog, so no need to mention "Corrected and added unit, system and integration tests" here 🙂

packages/cli/__tests__/zosfiles/__integration__/compare/ds/__scripts__/bogus_flag.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+zowe zos-files compare ds "AT895452.TESTING.JCL(ECONTXTL)" "AT895452.TESTING.JCL(CONTXTLN)" --bogus-flag
```
t1m0thyj, line 3: Please replace mainframe ID with a generic one like "IBMUSER".

t1m0thyj (commented):
Thanks for adding/fixing tests, left a few comments

packages/cli/__tests__/zosfiles/__system__/compare/ds/Compare.ds.system.test.ts
```diff
@@ -22,12 +22,12 @@ let testEnvironment: ITestEnvironment<ITestPropertiesSchema>;
...
 let dsname: string;
 const TIMEOUT = 20000;
-
+const buffer = Buffer.from("buff");
 
 describe("Compare Data Sets", () => {
     beforeAll(async () => {
         testEnvironment = await TestEnvironment.setUp({
-            installPlugin: true,
+            // installPlugin: true,
```
zFernand0, line 30: Is there a reason why we want to keep this here commented?

packages/cli/__tests__/zosfiles/__unit__/compare/ds/Dataset.handler.unit.test.ts
```diff
@@ -108,6 +110,45 @@ describe("Compare data set handler", () => {
...
+                }
+            };
+            const options: IDiffOptions = {
+                contextLinesArg,
+                outputFormat: "terminal"
+            };
+
+            DiffUtils.getDiffString = jest.fn(async () => {
+                return "compared string";
+            });
```
zFernand0, line 129: I know that subsequent tests are not looking for any output of the called functions, however please keep in mind that this stile of mocking may result in unexpected test failues.
For example, someone adding tests (after this one) might expect the real behavior of the `getDiffString` method and they will get a mocked version instead.
Suggestion: For future tests, please try using `jest.spyOn` (if possible) or `jest.mock` when dealing with completed modules 😋

zFernand0 (commented):
The code changes and tests LGTM!
I do agree with Timothy's comment on the changelog and the sh script.
Happy to approve once those are addressed.

packages/cli/__tests__/zosfiles/__unit__/compare/ds/Dataset.handler.unit.test.ts
ATorrise: I've added this mocked method approach

packages/cli/src/zosfiles/compare/lf-sdd/LocalfileSpooldd.handler.ts
```diff
@@ -59,22 +59,21 @@
~            }
~
~            // reading local file as buffer
~            lfContentBuf = fs.readFileSync(localFileHandle);
~        } finally {
             fs.closeSync(localFileHandle);
         }
 
+        const lfContentBuf = fs.readFileSync(localFile);
```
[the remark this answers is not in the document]
ATorrise: whoops

packages/cli/src/zosfiles/compare/lf-sdd/LocalfileSpooldd.handler.ts
```diff
@@ -59,22 +59,21 @@
~            }
~
~            // reading local file as buffer
~            lfContentBuf = fs.readFileSync(localFileHandle);
~        } finally {
             fs.closeSync(localFileHandle);
         }
 
+        lfContentBuf = fs.readFileSync(localFile);
```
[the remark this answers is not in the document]
ATorrise: whoops pt 2

t1m0thyj (approved):
Thanks for fixing both the `--context-lines` and `--seqnum` options, and increasing coverage and removing snapshots in the process 🙂

zFernand0 (approved):
LGTM

PR #1606, opened 2023-01-19
change the default version from 1.0 to 2.0 across all our jobs commands

ATorrise:
PR also includes check-latest option to Github workflow


zFernand0 (dismissed):
Will this be considered a breaking change since now all of these request will go from being asynchronous to being synchronous?

Not saying we should not do this.
I believe there are advantages to using v2.0 comparted to the v1.0 version of these API.
For example there is some extra permissions that any user needs with v1.0 (I believe it's called "CSM authorized"?) which is not a requirement for the v2.0.

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+- BugFix: Fixed the default version from "1.0" to "2.0" across job commands [#1459](https://github.com/zowe/zowe-cli/issues/1459)
```
t1m0thyj, line 6: We should add a comment explaining how this may affect users as we discussed at standup 🙂 

The old default behavior was asynchronous - returned immediately after the request to cancel/delete the job was issued. The new behavior is synchronous - waits for confirmation from the server that the job has successfully been cancelled/deleted before returning.

t1m0thyj (changes requested):
Please also update the z/OS Jobs SDK (and its changelog) to change the default version there as well:
https://github.com/zowe/zowe-cli/blob/d47097c2834f13711c39f1acd574109dd5769f02/packages/zosjobs/src/JobsConstants.ts#L150
https://github.com/zowe/zowe-cli/blob/d47097c2834f13711c39f1acd574109dd5769f02/packages/zosjobs/src/CancelJobs.ts#L71-L73
https://github.com/zowe/zowe-cli/blob/d47097c2834f13711c39f1acd574109dd5769f02/packages/zosjobs/src/DeleteJobs.ts#L72-L76

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+- BugFix: "Changed default value for modify-jobs option in the zowe jobs command group to 2.0. This change results in calls to z/OSMF becoming synchronous, and a successful response from the modify, cancel, and delete commands indicates the requested action was completed successfully. [#1459](https://github.com/zowe/zowe-cli/issues/1459)
```
t1m0thyj, line 6: Unnecessary quotation mark

packages/zosjobs/CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 # Change Log
 
 All notable changes to the Zowe z/OS jobs SDK package will be documented in this file.
-
+## Recent Changes
+- BugFix: "Changed default value for modify-jobs option in the zowe jobs command group to 2.0. This change results in calls to z/OSMF becoming synchronous, and a successful response from the modify, cancel, and delete commands indicates the requested action was completed successfully. [#1459](https://github.com/zowe/zowe-cli/issues/1459)
```
t1m0thyj, line 5: Instead of "zowe jobs command group", should we say "CancelJobs and DeleteJobs classes" to be more SDK specific?

Also, unnecessary quotation mark (same as above)

t1m0thyj (dismissed):
Thanks for updating the SDK methods 🙂 Please fix failing unit and integration tests. Left a few minor comments about changelog.

zFernand0 (commented):
Code changes LGTM! 😋 
just waiting on the snapshots to approve 👍

gejohnston (approved):

zFernand0 (approved):
Seems like we just need to remove some obsolete snapshots 😋 
LGTM! 😋

zFernand0 (approved):
I like the clean way you are reducing duplicate code 🥳 
LGTM! 😋

PR #1614, opened 2023-01-25
Add CodeQL security and quality scan

t1m0thyj:
**Note:** Rust is currently not a supported language for CodeQL scans.

Also update Imperative and fix broken integration test snapshots.

Thanks @awharn for addressing high severity issues that were found 🙂


awharn (approved):

gejohnston (approved):

traeok (approved):

packages/cli/src/zosfiles/compare/lf-ds/LocalfileDataset.handler.ts
```diff
@@ -39,9 +39,11 @@ export default class LocalfileDatasetHandler extends ZosFilesBaseHandler {
             localFile = path.resolve(commandParameters.arguments.localFilePath);
         }
 
+        let localFileHandle: number;
         // check if the path given is of a file or not
         try {
-            if(!fs.lstatSync(localFile).isFile()){
+            localFileHandle = fs.openSync(localFile, 'r');
+            if(!fs.fstatSync(localFileHandle).isFile()){
```
traeok, line 46: I think this is currently being updated in the other CodeQL QRs, but we probably want to use a `finally` block here - (as you suggested earlier) to call `closeSync` in scenarios where `fstatSync` fails, but the file is still opened

traeok (approved):

zFernand0 (approved):

PR #1616, opened 2023-01-27
Doc on merging config files

gejohnston:
- Added a new document describing how team configuration files are merged.
    - Included doc on where AutoStored properties are stored.
- Also revised an existing document on team configuration. 
    - Removed the phrase "Early Access" from the document.
    - Replaced the term "global configuration" with "team configuration". 

Target Review-by date: Thu Feb 2


traeok (approved):
LGTM! Just left a couple comments regarding phrasing and the flowchart

docs/How_config_files_are_merged.md
```diff
@@ -0,0 +1,576 @@
...
+
+When a team configuration and a user configuration are placed at the same level (both files are in the ZOWE_CLI_HOME directory or both are in the same project directory), the merging of profile properties are done property-by-property. A profile property value in the user config will override a value in the team config file. If a profile property value exists in one config file but not the other, then that value will be also included in the merged set of properties used by a CLI command.
+
+When configuration files (either team or user) exist at both the global level and in a project directory, a profile from a project config completely replaces a profile of the same name from the global config. All profiles that exist in one config file but **NOT** in the other config file are included in the resulting configuration that is used by a CLI command.
+
+Potentially, four configuration file can be in used at once:
```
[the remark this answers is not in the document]
gejohnston: Thanks. Both of your comments are correct. I will correct those items.

awharn (approved):

docs/How_config_files_are_merged.md
```diff
@@ -0,0 +1,576 @@
...
+
+It is when you are using a Zowe CLI configuration in a project directory that you are most likely to use a user configuration. A user configuration has the same structure as a team configuration (although you typically would only specify a small subset of properties that are specific to your needs.) For example, perhaps you need to debug a special test instance of some service, while the rest of your team might need to use a production instance of that same service. You could place the port number of your special test instance into your user config file. Everyone shares all of the other properties from the project team configuration, but only your user config file will override the port number to point to your test instance. With such a scheme, every user could connect to their own test instance of a service, by each user specifying a unique port number in their own user config file.
+
+While you commit your project team configuration file into your source control tool, you **NEVER** commit your user config file into source control. This is what enables the team to share most properties, but still have a few personal overrides when needed.
```
t1m0thyj, line 83: In addition to this warning, should we recommend adding the user config file to an ignore list for their VCS tool (such as `.gitignore`)?

For folks who aren't super familiar with how Git works, it may not be obvious to them that the file can be added to such an ignore list. We don't want them to think that for every commit, they need to be concerned about the risk of accidentally staging the user config file in their list of changes.

docs/How_config_files_are_merged.md
```diff
@@ -0,0 +1,576 @@
...
+
+The config file's ability to understand comments (// or /* */) can make such quick edits even easier. You can comment-out several alternatives and then just uncomment the one you want to use, rather than having to remember the correct value and retyping that alternative.
+
+Since a global configuration exists under your HOME directory, it belongs to you personally. Thus, you may not have a great need for a user configuration file in this scenario - all of the values in a global configuration already belong to you. If you want to periodically share your configuration with colleagues, you could choose to place your standard configuration in a normal team configuration file (zowe.config.json) and place your specialty alternatives in your user configuration (zowe.config.user.json). If you receive an ad-hoc request from a colleague for your configuration, you could just send your zowe.config.json file to that person. Thus you could avoid scrubbing your configuration file before sending it to someone else.
+
+### **Advantages of only using a project configuration**
```
t1m0thyj, line 71: Another potential advantage is that you can easily switch between v1 and v2 profiles depending on which directory you run `zowe` commands in. I'll defer to you on whether this is worth mentioning 🙂

docs/How_config_files_are_merged.md
```diff
@@ -0,0 +1,576 @@
...
+
+- When a user name or password is not specified in the configuration, Zowe Explorer will prompt for the user name or password. If the configuration does not specify a default base profile, Zowe Explorer will display the errors below. The CLI runs successfully in such a configuration. To avoid the following errors in Zowe Explorer, the default "base" profile name must have a value. When `autoStore` is false, Zowe Explorer does not even try to save the values for which it prompted, so the base profile itself does not have to exist.
```
t1m0thyj, line 152: Do we believe this is intended behavior of Zowe Explorer? Both items sound like potential bugs to me.

t1m0thyj (approved):
Thanks for researching and documenting this topic so thoroughly. Left a few comments.

Also, this is the first time I have seen a Mermaid flowchart used - pretty cool 🙂

docs/Using Team Configuration.md
```diff
@@ -1,30 +1,28 @@
-# Using global profile configuration with Zowe CLI <!-- omit in toc -->
-
-**Early access feature:** Global profiles are available in the `@next` version of Zowe CLI. If you already installed the supported version `@zowe-v1-lts`, switch versions to try this feature. The functionality will be included in the next major Zowe release, V2.0.0-LTS.
+# Using team profile configuration with Zowe CLI <!-- omit in toc -->
 
 **Table of Contents:**
 - [Feature overview](#feature-overview)
   - [Benefits](#benefits)
   - [Changes to secure credential storage](#changes-to-secure-credential-storage)
 - [Installing @next version](#installing-next-version)
```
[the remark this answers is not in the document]
gejohnston: I failed to search for the term "next". Will do.

docs/Using Team Configuration.md
```diff
@@ -1,30 +1,28 @@
...
-- [Initializing global configuration](#initializing-global-configuration)
+- [Initializing team configuration](#initializing-team-configuration)
 - [(Optional) Initializing user-specific configuration](#optional-initializing-user-specific-configuration)
 - [Editing configuration](#editing-configuration)
 - [Managing credential security](#managing-credential-security)
 - [Tips for efficient configuration](#tips-for-efficient-configuration)
   - [Tips for using the base array](#tips-for-using-the-base-array)
-- [Sharing global configuration](#sharing-global-configuration)
+- [Sharing team configuration](#sharing-team-configuration)
 - [Example configurations](#example-configurations)
```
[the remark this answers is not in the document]
gejohnston: Yes. I will change them.

docs/How_config_files_are_merged.md
gejohnston: Good idea.

docs/How_config_files_are_merged.md
gejohnston: It is both an advantage and a potential point of confusion, so it is worth mentioning.

docs/How_config_files_are_merged.md
gejohnston: I did not know what behaviors to expect, so I wrote down every result so that we have a record that we can refer to. If a customer reports such a behavior, at least we know that it is expected 'bad' behavior. The saving grace is that the 'config init' command creates a base profile and sets it as the default base profile. Thus, customers are unlikely to encounter this error. I only got the error because I had been hacking up my config file by hand, and removed the base profile as "unnecessary" for what I was doing. Apparently it was necessary.

I agree that it would be reasonable to open a GitHub issue about this behavior. I am inclined to keep this documentation until the buggy behavior is fixed.

gejohnston: I created a ZE GitHub issue, and added a reference to that issue in this document.

zFernand0 (approved):
Thanks for updating the existing `using team config` file (removing the plugins-array reference and adding autoStore)

Issue #729, opened 2020-05-28
Job submit --wait-for-output outputting to directory using -d does not output.


davleop:
**How to replicate** </br>
`zowe jobs sub ds 'path.to.dataset(ds)' --wfo -d output` </br>
or locally </br>
`zowe jobs sub lf path/to/file --wfo -d output` </br>

This occurs with and without the directory `output` present.

**Result** </br>
```
jobid:   JOB#####
retcode: CC 0000
jobname: NAME
status:  OUTPUT
Successfully downloaded output to ./output/JOB#####
``` 
Then nothing is outputted.

**Desired Behavior**
- Output should be similar behavior as if I ran `zowe jobs dl o JOB#####`

**Environment**
`zowe --version`: 6.13.0 </br>
`System:    Kernel: 4.4.0-17763-Microsoft x86_64 bits: 64 compiler: gcc v: 5.4.0 Desktop: N/A`</br>`Distro: Ubuntu 20.04 LTS (Focal Fossa)`

billpereira:
@zFernand0 @dkelosky i was willing to help on it, but Is it a bug of code or maybe documentation? looks it is happening cause if the wfo parm is present it is not checking for directory parm... 

```
Submit.Jobs.ts
if (parms.viewAllSpoolContent || parms.waitForOutput) {
...
} else if (parms.directory) {
```

But when using the parm directory the flag --wfo lose the sense for me as it already wait the job completion to download it?

dkelosky:
hey @billpereira - at a glance I think two things could be done, one for the `api` and one for the `cli`.
- in `SubtmitJobs.checkSubmitOptions` we could add `ImperativeExpect` calls to assert that either `waitForOutput` is specified or `directory` (not both).  See other examples of `ImperativeExpect` on how this could be done and perhaps add some doc that this options are mutually exclusive. 
- in `Submit.definition.ts`, perhaps `[onlyOneOf](https://github.com/zowe/imperative/blob/master/packages/cmd/src/doc/ICommandDefinition.ts)` could be used to say that `-d` and `-wfo` are mutually exclusive

Side note (nothing to do just some observations),  `SubmitJobs.ts` is in the `/api` folder of Zowe CLI.  I think the `SubmitJobs.checkSubmitOptions` method is confusing in retrospect: 
- the method name makes it seem like its doing a check of options (perhaps validating options), but it's actually handling certain option conditions  
[...]

PR #1626, opened 2023-02-08
making agreed upon changes to conditional statements in addition to s…

ATorrise:
…witching the ordering of conditional statements, without doing the latter, output is still not downloaded

----------------------------------------------------------------------------------------------------------------------------------

<b>BUG FIX: Enabled option to download output from submitted job with the -d flag. [#729](https://github.com/zowe/zowe-cli/issues/729)</b>


gejohnston:
Just a general comment: Nothing in the title or description states what this pull request intends to accomplish. This is not a request for change, since I figured out the objective by reading the original issue (which is only mentioned in the changelog update).

packages/zosjobs/src/SubmitJobs.ts
```diff
@@ -250,6 +250,29 @@ export class SubmitJobs {
...
+            }
+            if (parms.task != null) {
+                parms.task.statusMessage = "Downloading spool content for " + job.jobid +
+                    (job.retcode == null ? "" : ", " + job.retcode);
+                parms.task.percentComplete = TaskProgress.SEVENTY_PERCENT;
+            }
+            (await DownloadJobs.downloadAllSpoolContentCommon(session, downloadParms));
+            return job;
+        }
         if (parms.viewAllSpoolContent || parms.waitForOutput) {
```
gejohnston, line 276: If someone specifies both --directory and --wait-for-output, will the following sequence of somewhat repeated messages be displayed?

"Waiting for " + responseJobInfo.jobid + " to enter OUTPUT"
"Downloading spool content for " + job.jobid
"Waiting for " + responseJobInfo.jobid + " to enter OUTPUT"
"Retrieving spool content for " + job.jobid

gejohnston (changes requested):

packages/zosjobs/src/SubmitJobs.ts
ATorrise: yes, thank you for catching this - i changed this so that "Waiting for " + responseJobInfo.jobid + " to enter OUTPUT" will only appear once

gejohnston (approved):

t1m0thyj (changes requested):
The DownloadJobs API that is called by SubmitJobs is currently ignoring the `--extension` option.

Will approve once this is fixed and tests have been added for `--directory` and `--extension` 🙂

packages/zosjobs/__tests__/__unit__/SubmitJobs.unit.test.ts
```diff
@@ -255,6 +340,54 @@
...
+                task: {
+                    percentComplete: 70,
+                    statusMessage:"Waiting for " + fakeJobID + " to enter OUTPUT",
+                    stageName: TaskStage.IN_PROGRESS
+                } as ITaskWithStatus
+            };
+            checkSubmitOptionsSpy.mockReturnValueOnce(sampleJob as IJob);
+            SubmitJobs.checkSubmitOptions = jest.fn(async (fakeSession, parms, responseJobInfo): Promise <IJob | ISpoolFile[]> => {
+                parms = submitParms;
+                responseJobInfo = sampleJob;
```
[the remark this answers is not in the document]
zFernand0: I kind of agree with codeQL's assessment here.
Unless the intention was to update the `parms` or the `responseJobInfo` properties (given that object properties are treated as references), we should be good to remove these two assigments 😋

packages/zosjobs/src/DownloadJobs.ts
```diff
@@ -82,10 +82,7 @@ export class DownloadJobs {
             }
             await DownloadJobs.downloadSpoolContentCommon(session, {
                 jobFile: {...file, ddname: uniqueDDName},
-                outDir: parms.outDir,
-                omitJobidDirectory: parms.omitJobidDirectory,
-                binary: parms.binary,
-                record: parms.record
+                ...parms
             });
```
zFernand0, line 86: Smart way to avoid having to type new properties as they come in.

I guess I do have one small concern.
What if a new property is added to `IDownloadAllSpoolContentParms` called `jobFile`?

We could flip the order in which we configure these parameters, to make sure that we don't overwrite the current `file` and `uniqueDDName` values.
```typescript
await DownloadJobs.downloadSpoolContentCommon(session, {
    ...parms
    jobFile: {...file, ddname: uniqueDDName},
});
```

---

This is just something to keep in mind for future object construction and not something we need to be super concerned about in this particular case because if how the interfaces where combined  😋

packages/zosjobs/src/SubmitJobs.ts
```diff
@@ -248,14 +247,30 @@ export class SubmitJobs {
...
         }
-        // if viewAppSpoolContent option passed, it waits till job status is output
-        // then get content of each spool file and return array of ISpoolFiles object
-        if (parms.viewAllSpoolContent || parms.waitForOutput) {
+        // wait for job status to be OUTPUT for all other situations
+        if (parms.task != null) {
+            parms.task.statusMessage = "Waiting for " + responseJobInfo.jobid + " to enter OUTPUT";
+            parms.task.percentComplete = TaskProgress.THIRTY_PERCENT;
+        }
+        const job: IJob = await MonitorJobs.waitForJobOutputStatus(session, responseJobInfo);
```
t1m0thyj, line 255: This logic needs to stay inside the "if" statements below, so that it only runs when `--directory` or `--wait-for-output` are specified. Moving it out will cause the `Waiting for <jobid> to enter OUTPUT` message to appear on all "submit jobs" commands.

packages/zosjobs/src/DownloadJobs.ts
```diff
@@ -104,11 +101,7 @@ export class DownloadJobs {
         ImperativeExpect.keysToBeDefined(parms, ["jobFile"], "You must specify a job file on your 'parms' parameter" +
             " object to the downloadSpoolContentCommon API.");
 
-        if (parms.outDir == null) {
-            parms.outDir = DownloadJobs.DEFAULT_JOBS_OUTPUT_DIR;
-        }
-
-        const file = DownloadJobs.getSpoolDownloadFile(parms.jobFile, parms.omitJobidDirectory, parms.outDir);
+        const file = DownloadJobs.getSpoolDownloadFilePath(parms);
```
zFernand0, line 104: Smart way to deprecate the `getSpoolDownloadFile` method. 👍 
I was initially concerned about the outDir behavior possibly changing, but it makes sense to more that login into the new version of the function

packages/zosjobs/src/SubmitJobs.ts
```diff
@@ -248,14 +247,30 @@ export class SubmitJobs {
...
-                parms.task.percentComplete = TaskProgress.THIRTY_PERCENT;
+                parms.task.statusMessage = "Downloading spool content for " + job.jobid +
+                    (job.retcode == null ? "" : ", " + job.retcode);
+                parms.task.percentComplete = TaskProgress.SEVENTY_PERCENT;
             }
-            const job: IJob = await MonitorJobs.waitForJobOutputStatus(session, responseJobInfo);
+            await DownloadJobs.downloadAllSpoolContentCommon(session, downloadParms);
+            return job;
+        }
+        if (parms.viewAllSpoolContent || parms.waitForOutput) {
```
t1m0thyj, line 273: Should we use `else if` here? The original code had:
```
if (parms.viewAllSpoolContent || parms.waitForOutput) {
...
} else if (parms.directory) {
```

Flipping the order of these conditions makes sense to fix the linked issue 👍 

I believe we need to keep the `else if` like the following:
```
if (parms.directory) {
...
} else if (parms.viewAllSpoolContent || parms.waitForOutput) {
```

Otherwise we may end up fetching all spool files twice when both `--directory` and `--wait-for-output` are specified.

t1m0thyj (changes requested):
A few comments about changes to the SubmitJobs SDK

packages/zosjobs/src/SubmitJobs.ts
zFernand0, line 255: by removing the `waitForOutput` check, we are forcing every submit command to wait for the output regardless of the options specified (even an explicit `--no-wfo`).
This might be a breaking change as people may be expecting the command to return.
```powershell
> # This is my CLI installed from public NPM
> zowe jobs sub ds 'USERNAME.PUBLIC.CNTL(sleep)' --no-wfo
jobid:   JOB03187
retcode: null
jobname: JOBNAME
status:  INPUT

> # This is my compiled version
> zdev jobs sub ds 'USERNAME.PUBLIC.CNTL(sleep)' --no-wfo
 ███_______| 30%  O | Waiting for JOB03188 to enter OUTPUT
 ```

**Review by zFernand0 — CHANGES_REQUESTED** [review]
The code changed LGTM!
I do have a request in order to prevent a possible change in behavior 😋

File: packages/zosjobs/src/SubmitJobs.ts (hunk 4 above)
Comment (zFernand0) [reply packages/zosjobs/src/SubmitJobs.ts:255]: Apologies for duplicating comments.
This is very similar to Timothy's suggestions.

**Review by zFernand0 — APPROVED** [review]

**Comment by zFernand0** [comment]
Not sure this is important but the waiting message seems to be different.
The command works as expected though 😋 

```powershell
> # The behavior on master seems to wait for job to enter output
> zowe jobs sub ds 'USERNAME.PUBLIC.CNTL(sleep)' --wfo -d test
 ███_______| 30%  o | Waiting for JOB03324 to enter OUTPUT

> # The behavior in this PR seems to wait for the job to be submitted
> zdev jobs sub ds 'USERNAME.PUBLIC.CNTL(sleep)' --wfo -d test
 █_________| 10%  | | Submitting job
```

zFernand0 (approved):

t1m0thyj:
> Not sure this is important but the waiting message seems to be different. The command works as expected though 😋
> 
> ```powershell
> > # The behavior on master seems to wait for job to enter output
> > zowe jobs sub ds 'USERNAME.PUBLIC.CNTL(sleep)' --wfo -d test
>  ███_______| 30%  o | Waiting for JOB03324 to enter OUTPUT
> 
> > # The behavior in this PR seems to wait for the job to be submitted
> > zdev jobs sub ds 'USERNAME.PUBLIC.CNTL(sleep)' --wfo -d test
>  █_________| 10%  | | Submitting job
> ```

Hopefully a simple change 🙂 I believe we want to keep the order of these statements the same as before - updating the task before calling the MonitorJobs API:
```
            if (parms.task != null) {
                parms.task.statusMessage = "Waiting for " + responseJobInfo.jobid + " to enter OUTPUT";
                parms.task.percentComplete = TaskProgress.THIRTY_PERCENT;
            }
            const job: IJob = await MonitorJobs.waitForJobOutputStatus(session, responseJobInfo);
```

t1m0thyj (changes requested):
Will approve once Fernando's comment is addressed 😀

zFernand0 (approved):

t1m0thyj (approved):

PR #1640, opened 2023-02-24
fixes #1639

### LINKED ISSUE #1639: Add missing recfm types for create.
I cannot create a Seq DS of "VBS" recfm type. It is missing in the case statement within the code. I can create this type of DS outside of Zowe CLI. Please fix. 

See ISubmitJclParms.ts:

                        switch (tempOptions.recfm.toUpperCase()) {
                            case "F":
                            case "FB":
                            case "V":
                            case "VB":
                            case "VBS":    
                            case "U":
                                break;
                            default:
                                throw new ImperativeError({ msg: ZosFilesMessages.invalidRecfmOption.message + tempOptions.recfm });
                        }
                        break;

frankgiordano:


packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## `7.10.4`
```
t1m0thyj, line 5: Please change the header to `## Recent Changes`. Our deployment automation will handle updating the version number when a new version is published.

packages/zosfiles/src/methods/create/Create.ts
```diff
@@ -272,10 +272,18 @@ export class Create {
                         // F, V, or U are required; B, A, M, S, T or additional
                         // VBA works on mainframe but not via zOSMF
                         switch (tempOptions.recfm.toUpperCase()) {
+                            case "D":
```
t1m0thyj, line 275: Please add unit tests to cover the updated lines. The unit tests for this file are in `packages/zosfiles/__tests__/__unit__/methods/create/Create.unit.test.ts`

You could add a test that validates all the "recfm" values similar to this one for testing "dsntype":
https://github.com/zowe/zowe-cli/blob/master/packages/zosfiles/__tests__/__unit__/methods/create/Create.unit.test.ts#L101-L119

t1m0thyj (changes requested):
Thanks for your contribution! 🙂 

Left a few comments. Please also fix your commits that are not signed off to make the DCO check happy - see [here](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md#sign-all-of-your-git-commits) for details on how to do so.

PR #1648, opened 2023-03-07
Update README.md

zFernand0:
More info on how to add sonarcloud badges
- https://stackoverflow.com/a/62847411


traeok (approved):
wonder why the coverage is different between the two 🤔 

LGTM, thanks @zFernand0

t1m0thyj (commented):
Thanks for doing this 👍 

Also curious why the numbers are different - should we investigate before merging or plan to resolve it in the future?

zFernand0:
Sonar cloud coverage is a bit out of date.
Looking into it now
👀 https://github.com/zowe-actions/octorelease-script/blob/master/scripts/sonarConfig.ts#L41

t1m0thyj:
Closed the PR because it seems like Sonar is receiving accurate data but reports coverage differently from other tools.

> SonarCloud.io shows a breakdown of the coverage info:
> image
> 
> The line coverage of 92% and branch coverage of 82% pretty much match what Jest itself reports, so I think the overall coverage number is calculated by Sonar using some formula that is essentially a weighted average of the other coverage numbers.

Thanks @traeok for tracking down their coverage formula here: https://docs.sonarcloud.io/digging-deeper/metric-definitions/

PR #1652, opened 2023-03-09
daemon (Windows): coerce username to lowercase

traeok:
This should fix the bug where the same Windows user attempts to run the daemon (already running under their account), but they are blocked because of the username check:

`User name of 'testuser' in file 'C:\Users\TestUser\.zowe\daemon\daemon_pid.json' does not match current user = 'TestUser'.`

Since Windows user names are case insensitive, this check should pass. This PR will coerce the username (Windows only) to a lowercase string for the daemon and `DaemonDecider` operations.

### DIFF
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Solved daemon issue where Windows usernames were treated as case-sensitive when checking the daemon process owner during Zowe commands.
+
 ## `7.11.0`
 - Enhancement: Added support for a CLI specific environment variable file. [#1484](https://github.com/zowe/zowe-cli/issues/1484)
 - BugFix: Enabled option to download output from a submitted job with the -d flag. The -e flag now enables changes to file extension as originally intended. [#729](https://github.com/zowe/zowe-cli/issues/729)
--- a/zowex/Cargo.toml
+++ b/zowex/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zowe"
-version = "1.0.2"
+version = "1.0.3"
 authors = ["Zowe Project"]
 edition = "2018"
 license = "EPL-2.0"
--- a/zowex/src/comm.rs
+++ b/zowex/src/comm.rs
@@ -12,17 +12,14 @@
 // Functions related to daemon cummunication.
 
 use std::io;
-use std::io::BufReader;
 use std::io::prelude::*;
+use std::io::BufReader;
 use std::str;
 use std::thread;
 use std::time::Duration;
 
 #[cfg(target_family = "unix")]
-use {
-    std::net::Shutdown,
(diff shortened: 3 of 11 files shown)


awharn (approved):

awharn:
Do we still need to manually increment the daemon version in cargo?

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Solved daemon issue where Windows usernames were treated as case-sensitive when checking the daemon process owner during Zowe commands.
```
zFernand0, line 7: I know it may be redundant, but whenever there isn't an issue to link in the changelog, we should at least link the PR.
That way it's easier for the docs-squad to find this PR and read its description to understand the use-case. 😋

zowex/src/comm.rs
```diff
@@ -65,7 +60,10 @@ use crate::proc::*;
  *      A Result containing a stream upon success.
  *      This function exits the process upon error.
  */
-pub fn comm_establish_connection(njs_zowe_path: &str, daemon_socket: &str) -> io::Result<DaemonClient> {
+pub fn comm_establish_connection(
+    njs_zowe_path: &str,
+    daemon_socket: &str,
+) -> io::Result<DaemonClient> {
```
zFernand0, line 66: Curious if we should have prettier implemented in the repo. 🤔

zowex/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "zowe"
-version = "1.0.2"
+version = "1.0.3"
```
zFernand0, line 3: Dumd questions:
- Curious if we really need to publish a new version of the daemon when we are mostly reformatting the code (with rust-analyzer)?
  - Is the reason that this PR will remove the `clippy` warning that comes from `util_get_zowe_env`?
    - Does the end user see this warning?

zFernand0 (commented):
LGTM! 😋 
I have a very small request to update the changelog with a link to the PR.
And I also had a few questions that will help me understand the daemon a bit better.

zowex/Cargo.toml
t1m0thyj: There is a bugfix in the Rust code in the first commit: https://github.com/zowe/zowe-cli/pull/1652/commits/7a17ce6c1d4f9288ff5da56e1b23c28dec192456
I believe that warrants the patch release 😛

zFernand0 (approved):
LGTM! 😋

zowex/Cargo.toml
zFernand0: thanks for clarifying! 🥳

t1m0thyj (approved):

zowex/src/comm.rs
traeok: I formatted w/ `rust-analyzer` here, but we could use the [Prettier plugin for Rust](https://github.com/jinxdash/prettier-plugin-rust) if we decide to go that route 😁

PR #1653, opened 2023-03-10
Update exempt stale labels

zFernand0:
In this PR:
- Added `priority-critical` and `priority-high` to the exempt issue labels
- Remove `third-party` label from the list since we only had 1 issue (and to prevent it from being closed, I added `keep`)
- Copied the `stale` and `close` messages from issues to PRs

Question: 
- Should we use the default `Stale` label instead of our own `stale` one?


gejohnston (approved):
Unless we are aware of the default "Stale" label being used for some other contradictory purpose, I think it makes sense to use "Stale".

Will have to alter a lot of issues manually, or maybe experience some dual-use transition period?

zFernand0:
> Will have to alter a lot of issues manually, or maybe experience some dual-use transition period? 

I believe that we can just rename the existing one an it will apply to all labels 😋 
And if there are no objections, I can remove the following lines and rename the label
```yaml
stale-issue-label: "stale" # Should we use the default: `Stale`?
stale-pr-label: "stale" # Should we use the default: `Stale`?
```

awharn (approved):

ATorrise (approved):

zFernand0:
`stale` label renamed to `Stale`
https://github.com/zowe/zowe-cli/issues?q=label%3AStale+is%3Aclosed

PR #1655, opened 2023-03-15
Update RUST dependencies and audit workflow

awharn:
Updates the daemon dependencies and the RUST workflow to find additional issues, like yanked crates.
Workflow result before: https://github.com/zowe/zowe-cli/actions/runs/4426805429
Workflow result after: https://github.com/zowe/zowe-cli/actions/runs/4427115394/jobs/7764266287

### DIFF
--- a/.github/workflows/rust-cli-publish.yml
+++ b/.github/workflows/rust-cli-publish.yml
@@ -58,9 +58,7 @@ jobs:
 
     steps:
     - name: Install Rust toolchain
-      run: |
-        rustup toolchain install stable --profile minimal
-        rustup default stable
+      uses: dtolnay/rust-toolchain@stable
 
     - uses: actions/download-artifact@v3
       with:
@@ -103,9 +101,7 @@ jobs:
     steps:
 
     - name: Install Rust toolchain
-      run: |
-        rustup toolchain install stable --profile minimal
-        rustup default stable
+      uses: dtolnay/rust-toolchain@stable
 
     - uses: actions/download-artifact@v3
       with:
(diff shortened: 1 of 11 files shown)


zFernand0 (approved):
Thanks for updating `bumpalo` 🥳 
And also for updating the workflow 😋 

LGTM! 😋

t1m0thyj (changes requested):
Will approve once additional warnings that appear with `--deny warnings` flag have been addressed 🙂

traeok (approved):
LGTM, thanks @awharn !

gejohnston (approved):

.github/workflows/rust-cli-publish.yml
```diff
@@ -110,39 +103,36 @@ jobs:
...
-    - run: echo ${{ steps.get_upload_url.outputs.upload_url }}
+    - name: Get version
+      id: get-version
+      working-directory: zowex
+      run: echo "ZOWEX_VERSION=$(cargo metadata --no-deps | jq -r .packages[0].version)" >> $GITHUB_OUTPUT
 
     - name: Create Archive
       run: |
         cd target/release
-        gtar -cvzf zowe.tgz zowe
```
t1m0thyj, line 135: Previously we had to use `gtar` to work around a bug where `tar` on MacOS included an extra file in the archive: https://github.com/zowe/zowe-cli/pull/1311

This seems to no longer be a problem, so I'm cool with using `tar` 🙂

.github/workflows/rust-cli.yml
t1m0thyj, line 1: Noticed that we have duplicate runs of this workflow on `push` and `pull_request` events. Do we want to add the same condition we have in other workflows to skip duplicate runs?
`if: (github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository) && !contains(github.event.head_commit.message, '[ci skip]')`

t1m0thyj (approved):

t1m0thyj (approved):

PR #1659, opened 2023-03-17
Fix publish pipeline

awharn:
Remove the working directory changes where they should not be present


t1m0thyj (approved):
LGTM, thanks @awharn!

traeok (approved):

PR #1663, opened 2023-03-22
Fix URI encoding for Jobs commands

### LINKED ISSUE #1596: Special characters in JCL when submitting a job not correctly encoded
**Describe the bug**
When submitting a a JCL job it is possible to use the `MonitorJobs` [link](https://github.com/zowe/zowe-cli/blob/master/packages/zosjobs/src/MonitorJobs.ts) class to check for the job result. Under water, this class is also used by the `SubmitJobs.submitJclNotifyCommon` method.
When submitting a job with a special character, for example `#`, the `MonitorJobs.waitForStatusCommon` method will throw an `HTTP 400: Bad request` error. This is due to the special character not being escaped properly in the [AbstractRestClient](https://github.com/zowe/imperative/blob/ca981d476b7a97fa6129de523c63e1a58000c139/packages/rest/src/client/AbstractRestClient.ts#L416).

The expected result is that the `MonitorJobs.waitForStatusCommon` method returns an IJob object. The actual result is an error when submitting JCL with a special character in the jobcard: `Error obtaining status for jobname "Jobname" jobid "Jobid"`

When submitting JCL without a special character in the jobcard, it works as expected.

**Describe your environment**
In our environment, it is a standard to name jobs executed by a user in the following format: `<prefix>#<jobname>`, for example `KT#JZOWE`

**Additional context**
[I have created a repo to troubleshoot this issue](https://github.com/juleskreutzer/ZoweSubmitJclNotifyCommon). When executing the VSCode extension locally, follow the following steps:
- Run the extension in debug mode
- Execute the command "Hello World" --> You should see a message that the job has been submitted
- Execute the command "Another Hello World" --> You should see a message that the call failed and the error is logged to the console and the Zowe log folder (~/.zowe/zowe/logs)

Change the jobcard to your requirements, but make sure that the jobname includes a special character like `#`

I already discussed this with @VitGottwald. It is possible that he has another private repo with settings tailored to Broadcom's environment.

awharn:
Resolves #1596


gejohnston (approved):
The code changes look fine. I just have a question.

I recall some discussion about how encoding requirements differed between APIML and a direct connection to zosmf. Just to help me understand what we discovered, can you provide a comment on how we concluded that these changes will fix the problem for all cases?

KevinLoesch1 (approved):

packages/zosjobs/__tests__/__system__/CancelJobs.system.test.ts
```diff
@@ -43,80 +43,80 @@ describe("CancelJobs System tests", () => {
         const ACCOUNT = systemProps.tso.account;
         const maxStepNum = 255;  // Use lots of steps to make the job stay in INPUT status longer
```
t1m0thyj, line 44: I know we need the job to stay running for at least 10-20 seconds to give us adequate time to cancel it. 👍 

Should we consider making `maxStepNum` smaller? Currently if a CancelJobs test fails, it could leave the job running for 255*10 seconds ~= 42 minutes, which seems longer than necessary.

packages/zosjobs/__tests__/__system__/CancelJobs.system.test.ts
```diff
@@ -178,3 +178,61 @@ describe("CancelJobs System tests", () => {
...
+            user: systemProps.zosmf.user,
+            password: systemProps.zosmf.password,
+            hostname: systemProps.zosmf.host,
+            port: systemProps.zosmf.port,
+            type: "basic",
+            rejectUnauthorized: systemProps.zosmf.rejectUnauthorized
+        });
+
+        const ACCOUNT = systemProps.tso.account;
+        const maxStepNum = 255;  // Use lots of steps to make the job stay in INPUT status longer
```
t1m0thyj, line 200: Same question as above about `maxStepNum`

t1m0thyj (commented):
Left a few questions about tests 🙂

awharn:
> The code changes look fine. I just have a question.
> 
> I recall some discussion about how encoding requirements differed between APIML and a direct connection to zosmf. Just to help me understand what we discovered, can you provide a comment on how we concluded that these changes will fix the problem for all cases?

@gejohnston We ran all system tests against z/OSMF on the mainframe, then through the APIML, both before and after making changes to the API methods. Prior to changing the API methods, any system test that contained a # in the job name failed when being submitted through the APIML, but succeeded when passed through z/OSMF. After the changes, all system tests succeeded against both the APIML and z/OSMF. It was determined that z/OSMF can handle the requests with or without encoding, but APIML requires it. Therefore, all user input that ends up as part of the URI is now being encoded.

packages/zosjobs/__tests__/__system__/CancelJobs.system.test.ts
awharn: I am going to change this to run for, at most, 60 seconds, just in case.

t1m0thyj (approved):
LGTM, interesting that adding `cross-spawn` dependency made the lock file bigger for v1 and smaller for v2 😛

traeok (approved):

PR #1665, opened 2023-03-23
Fix encoding, add range, warning

### LINKED ISSUE #1495: `Get.USSFile` SDK method does not handle `encoding`
https://github.com/zowe/zowe-cli/blob/217bae8691f84a3c92189cc8bee5ccb1628d2360/packages/zosfiles/src/methods/get/Get.ts#L65

Suggestion: Add encoding header to the request 😋 
```typescript
if (options.encoding) {
  const encodingHeader = ZosmfHeaders.X_IBM_TEXT;
  const encodingHeaderKey = Object.keys(encodingHeader)[0];
  const encodingHeaderText = encodingHeader[encodingHeaderKey] + ZosmfHeaders.X_IBM_TEXT_ENCODING + options.encoding;
  encodingHeader[encodingHeaderKey] = encodingHeaderText;
  reqHeaders.push(encodingHeader);
}
```

awharn:
Resolves #1495 
Adds range option to `zowe zos-files view uss-file` command
Adds conflict warnings for `binary`, `record`, and `encoding` options on `zowe zos-files view` commands


gejohnston (approved):
Does range only apply when the record option is specified? The description of range implies that, but does not say it explicitly.

zFernand0 (approved):
LGTM! :yum:

I saw the dev process and the implementation turned out better than expected. :yum:

Thanks for adding the conflicting details to the option definition for the encoding flag.

awharn:
The consensus after the team discussion was that we will call the second range byteRange, and make that available on USS only. We will not implement that as part of this PR, but will instead implement it when requested.

t1m0thyj (approved):
LGTM, thanks @awharn!

Before closing #1495, do we want to create a separate issue to track [checking USS tag for encoding before retrieving contents](https://github.com/zowe/zowe-cli/issues/1495#issuecomment-1264965448) as an enhancement request?

PR #1672, opened 2023-03-29
new command: zowe files edit [ds, uss] [file location] [--editor]  

ATorrise:
Adding ability to edit mainframe USS and DS files locally
- users can set their default editor   
- requires user to interact with the terminal
    - users can make edits directly to file and return to terminal if no chosen editor
- has logic for editing conflicts
    - utilizes zOSMF ETAGs
    - allows for a temporary stash
- leverages preexisting CLI diff features, including browser view

### DIFF
--- a/packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
+++ b/packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
@@ -123,7 +123,7 @@ export class CompareBaseHelper {
 
     /**
      * Helper function that compare-related handlers will use to get the contents of a local file
-     * @param filePath Path to the file to compate against
+     * @param filePath Path to the file to compare against
      * @returns Buffer with the contents of the file
      */
     public prepareLocalFile(filePath: string): Buffer {
@@ -171,15 +171,15 @@ export class CompareBaseHelper {
     }
 
     /**
-     * To get the difference string in ternninal or in browser
-     * @param {string} string1 - string of file 1 comtent
-     * @param  {string} string2 - string of file 2 comtent
+     * To get the difference string in terminal or in browser
+     * @param {string} string1 - string of file 1 content
+     * @param  {string} string2 - string of file 2 content
      * @returns {IZosFilesResponse}
      * @public
      * @memberof CompareBaseHelper
      */
     public async getResponse(string1: string, string2: string): Promise<IZosFilesResponse>{
-        //  CHECHKING IIF THE BROWSER VIEW IS TRUE, OPEN UP THE DIFFS IN BROWSER
+        //  CHECKING IF THE BROWSER VIEW IS TRUE, OPEN UP THE DIFFS IN BROWSER
         if (this.browserView) {
             await DiffUtils.openDiffInbrowser(string1, string2);
             return {
--- a/packages/cli/src/zosfiles/edit/Edit.options.ts
+++ b/packages/cli/src/zosfiles/edit/Edit.options.ts
@@ -29,6 +29,7 @@ export const EditOptions: { [key: string]: ICommandOptionDefinition } = {
         name: "editor",
         aliases: ["e"],
         description: strings.EDITOR,
-        type: "string"
+        type: "string",
+        required: false
     }
 };
--- a/packages/cli/src/zosfiles/edit/Edit.options.ts
+++ b/packages/cli/src/zosfiles/edit/Edit.options.ts
@@ -1,0 +1,34 @@
+/*
+* This program and the accompanying materials are made available under the terms of the
+* Eclipse Public License v2.0 which accompanies this distribution, and is available at
+* https://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import { ICommandOptionDefinition } from "@zowe/imperative";
+
+import i18nTypings from "../-strings-/en";
+
+// Does not use the import in anticipation of some internationalization work to be done later.
+const strings = (require("../-strings-/en").default as typeof i18nTypings).EDIT.OPTIONS;
+
+/**
+ * Object containing all options to be used by the View API
+ */
+export const EditOptions: { [key: string]: ICommandOptionDefinition } = {
+
+    /**
+     * The option to set a default editor
+     * @type {ICommandOptionDefinition}
+     */
+    editor: {
+        name: "editor",
+        aliases: ["e"],
+        description: strings.EDITOR,
+        type: "string"
+    }
+};
--- a/packages/cli/src/zosfiles/edit/Edit.utils.ts
+++ b/packages/cli/src/zosfiles/edit/Edit.utils.ts
@@ -37,11 +37,11 @@ export class EditUtilities {
         if (isUssFile){
             // Hash in a repeatable way if uss fileName
             const crypto = require("crypto");
-            return tmpdir() +"/" + crypto.createHash("shake256", { outputLength: 8 })
+            return tmpdir() +"\\" + crypto.createHash("shake256", { outputLength: 8 })
                 .update(fileName)
                 .digest("hex");
         }else{
-            return tmpdir() + "/" + fileName;
+            return tmpdir() + "\\" + fileName + ".txt";
         }
     }
 
@@ -55,6 +55,7 @@ export class EditUtilities {
             }
         } catch(err) {
             //imperative error probably?
+            return false;
         }
     }
 
@@ -64,7 +65,7 @@ export class EditUtilities {
             case Prompt.useStash:
                 input = await CliUtils.readPrompt("Keep and continue editing found stash? Y/n");
                 if (input === null) {
-                    // abort the command ... maybe do something w esc
+                    // abort the command ... maybe do something with esc
                 }
                 if (input == lowerCase("y")){
                     // keep stash
@@ -100,8 +101,9 @@ export class EditUtilities {
     public static async fileComparison(session: AbstractSession, commandParameters: IHandlerParameters, lfFile: File): Promise<IZosFilesResponse>{
         const handler = new LocalfileDatasetHandler;
         const helper = new CompareBaseHelper(commandParameters);
+        helper.browserView = true;
 
-        commandParameters.arguments.localFilePath = lfFile.path;
+        commandParameters.arguments.localFilePath = lfFile.path; //why do i do this? idr
         const lf = await handler.getFile1(session, commandParameters.arguments, helper);
         const mfds = await handler.getFile2(session, commandParameters.arguments, helper);
         // Editor will open with local file if default editor was set
@@ -130,7 +132,7 @@ export class EditUtilities {
             if (response.errorMessage){
                 if (response.errorMessage.includes("etag")){ //or error 412
                     //alert user that the version of document they've been editing has changed
-                    //ask if they want to continue working w their stash (local file)
+                    //ask if they want to continue working with their stash (local file)
                     const continueToEdit: boolean = await this.promptUser(Prompt.continueEditing, lfFile.path);
                     if (continueToEdit){
                         // Download dataset again, refresh the etag of lfFile
--- a/packages/cli/src/zosfiles/edit/Edit.utils.ts
+++ b/packages/cli/src/zosfiles/edit/Edit.utils.ts
@@ -127,21 +127,53 @@ export class File {
         return await helper.getResponse(helper.prepareContent(lf), helper.prepareContent(mfds));
     }
 
+    public static async makeEdits(session: AbstractSession, commandParameters: IHandlerParameters, lfFile: File): Promise<void>{
+        await this.fileComparison(session, commandParameters, lfFile);
+        // 5a. check for default editor and headless environment
+        // 5b. open lf in editor or tell user to open up on their own if headless or no set default
+        // 4ba. perform file comparison, show output in terminal
+        await this.promptUser(Prompt.doneEditing);
+        // 7. once input recieved, upload tmp file with saved ETAG
+    }
 
 
-
-    public static async uploadEdits(session: AbstractSession, lfFile: File): Promise<IZosFilesResponse>{
+    public static async uploadEdits(session: AbstractSession, commandParameters: IHandlerParameters, lfFile: File, mfFile: File): Promise<boolean>{
     // 7. once input recieved, upload tmp file with saved ETAG
     // 7a. if matching ETAG: sucessful upload, destroy tmp file -> END
     // 7a. if non-matching ETAG: unsucessful upload -> 4a
         let response: IZosFilesResponse;
         try{
             response = await Upload.fileToDataset(session, lfFile.path, lfFile.name, {etag: lfFile.etag});
-            // successful upload
-            return response;
+            if (response.success){
+                // 7a. if matching ETAG & successful upload, destroy tmp file -> END
+                await this.destroyTempFile(lfFile.path);
+                return true;
+            }
+            if (response.errorMessage){
+                if (response.errorMessage.includes("etag")){ //or error 412
+                    //alert user that the version of document theyve been editing has changed.
+                    //ask if they want to continue working w this file
+                    let continueToEdit: boolean = await this.promptUser(Prompt.continueEditing, lfFile.path);
+                    if (continueToEdit){
+                        // if yes, 
+                        // 1. download dataset again, refresh the etag of lfFile
+                        mfFile.data = await Download.dataSet(session, mfFile.name, {returnEtag: true});
+                        // [TO DO - set etag]
+                        lfFile.etag = mfFile.etag ;
+                        // 2. then perform a file comparision:
+                        await this.makeEdits(session, commandParameters, lfFile);
+                        // 3. perform file comparision w mfds and lf(file youve been editing) with updated etag
+                        return false;
+                    }
+                }
+                throw new ImperativeError({
+                    msg: `Failed to save edits because remote has changed since last downloading file. Edits have been stored locally: ${lfFile.path}`
+                });
+                return false;
+            }
         }catch(err){
             // unsuccessful upload, potential mismatched etag
-            return response;
+            return false;
         }
     }
 
--- a/packages/cli/src/zosfiles/edit/Edit.utils.ts
+++ b/packages/cli/src/zosfiles/edit/Edit.utils.ts
@@ -1,0 +1,42 @@
+/*
+* This program and the accompanying materials are made available under the terms of the
+* Eclipse Public License v2.0 which accompanies this distribution, and is available at
+* https://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import { IImperativeConfig, Imperative, ProcessUtils } from "@zowe/imperative";
+import { PathLike } from "fs";
+import { tmpdir } from "os";
+import path = require("path");
+import { ICompareFileOptions } from "../compare/doc/ICompareFileOptions";
+
+/**
+ * Get the Imperative config object which defines properties of the CLI.
+ * This allows it to be accessed without calling Imperative.init.
+ */
+export function getImperativeConfig(): IImperativeConfig {
+    return require("./imperative");
+}
+
+// 2. build tmp_dir
+export function buildTempDir(sessionInfo, fileName, isUssFile): PathLike{
+    return tmpdir
+}
+// 2a. hash uss file name
+// 3. check for tmp_dir's existance as stash
+// 4a. if prexisting tmp_dir: override stash
+// 4b. if prexisting tmp_dir: use stash
+// 4ba. perform file comparison, show output in terminal
+// 4bb. overwrite ETAG 
+// 5a. check for default editor and headless environment
+ProcessUtils.isGuiAvailable
+// 5b. open lf in editor or tell user to open up on their own if headless or no set default
+// 6. wait for user input to continue
+// 7. once input recieved, upload tmp file with saved ETAG
+// 7a. if matching ETAG: sucessful upload, destroy tmp file -> END
+// 7a. if non-matching ETAG: unsucessful upload -> 4a
--- a/packages/cli/src/zosfiles/edit/Edit.utils.ts
+++ b/packages/cli/src/zosfiles/edit/Edit.utils.ts
@@ -21,43 +21,40 @@ import { unlink, existsSync } from "fs";
 
 export enum Prompt {
     useStash,
-    doneEditing, 
+    doneEditing,
     continueEditing
 }
 
 export class File {
     path: string;
     fileName: string;
-    apiData: IZosFilesResponse;
+    zosFilesResp: IZosFilesResponse;
 }
-  
-  
-  export class EditUtilities {
+
+export class EditUtilities {
     // Build tmp_dir
     public static async buildTempDir(fileName: PathLike, isUssFile?: boolean): Promise<string>{
-        let tmpDir = tmpdir();
-
         if (isUssFile){
             // Hash in a repeatable way if uss fileName
             const crypto = require("crypto");
-            return tmpDir +"/" + crypto.createHash("shake256", { outputLength: 8 })
+            return tmpdir() +"/" + crypto.createHash("shake256", { outputLength: 8 })
                 .update(fileName)
                 .digest("hex");
         }else{
-            return tmpDir + "/" + fileName;
+            return tmpdir() + "/" + fileName;
         }
     }
 
     // Check for tmp_dir's existance as stash
     public static async checkForStash(tmpDir: string): Promise<boolean>{
         try {
-          if (existsSync(tmpDir)) {
-            return true;
-          }else{
-            return false;
-          }
+            if (existsSync(tmpDir)) {
+                return true;
+            }else{
+                return false;
+            }
         } catch(err) {
-          console.error(err); //change to imperative error probably?
+            //imperative error probably?
         }
     }
 
@@ -68,7 +65,7 @@ export class File {
                 input = await CliUtils.readPrompt("Keep and continue editing found stash? Y/n");
                 if (input === null) {
                     // abort the command ... maybe do something w esc
-                } 
+                }
                 if (input == lowerCase("y")){
                     // keep stash
                     return true;
@@ -83,12 +80,13 @@ export class File {
                 }else{
                     return true;
                 }
+                break;
             case Prompt.continueEditing:
-                ``
-                input = await CliUtils.readPrompt("The version of the document you were editing has changed. Continue to make changes? Y/n\n" + fileInfo);
+                input = await CliUtils.readPrompt("The version of the document you were editing has changed." +
+                    "Continue to make changes? Y/n\n" + fileInfo);
                 if (input === null) {
                     // abort the command
-                } 
+                }
                 if (input == lowerCase("y")){
                     // keep stash
                     return true;
@@ -97,7 +95,6 @@ export class File {
                     return false;
                 }
         }
-
     }
 
     public static async fileComparison(session: AbstractSession, commandParameters: IHandlerParameters, lfFile: File): Promise<IZosFilesResponse>{
@@ -117,7 +114,6 @@ export class File {
         await this.promptUser(Prompt.doneEditing);
     }
 
-
     public static async uploadEdits(session: AbstractSession, commandParameters: IHandlerParameters, lfFile: File, mfFile: File): Promise<boolean>{
     // !! WIP !!
     // Once input recieved, upload tmp file with saved etag
@@ -125,7 +121,7 @@ export class File {
     // if non-matching etag: unsucessful upload -> perform file comparison/edit again with new etag
         let response: IZosFilesResponse;
         try{
-            response = await Upload.fileToDataset(session, lfFile.path, lfFile.fileName, {etag: lfFile.apiData.apiResponse.etag});
+            response = await Upload.fileToDataset(session, lfFile.path, lfFile.fileName, {etag: lfFile.zosFilesResp.apiResponse.etag});
             if (response.success){
                 // If matching etag & successful upload, destroy tmp file -> END
                 await this.destroyTempFile(lfFile.path);
@@ -135,17 +131,25 @@ export class File {
                 if (response.errorMessage.includes("etag")){ //or error 412
                     //alert user that the version of document they've been editing has changed
                     //ask if they want to continue working w their stash (local file)
-                    let continueToEdit: boolean = await this.promptUser(Prompt.continueEditing, lfFile.path);
+                    const continueToEdit: boolean = await this.promptUser(Prompt.continueEditing, lfFile.path);
                     if (continueToEdit){
                         // Download dataset again, refresh the etag of lfFile
-                        mfFile.apiData, lfFile.apiData = await Download.dataSet(session, lfFile.fileName, {returnEtag: true, file: lfFile.path, overwrite: false});
-                        // Then perform file comparision w mfds and lf(file youve been editing) with updated etag
+                        mfFile.zosFilesResp, lfFile.zosFilesResp = await Download.dataSet(session, lfFile.fileName,
+                            {returnEtag: true, file: lfFile.path, overwrite: false});
+                        // Then perform file comparision with mfds and lf(file youve been editing) with updated etag
+                        await this.makeEdits(session, commandParameters, lfFile);
+                        return false;
+                    }else{
+                        // Renew stash based on updated file version
+                        mfFile.zosFilesResp, lfFile.zosFilesResp = await Download.dataSet(session, lfFile.fileName,
+                            {returnEtag: true, file: lfFile.path});
                         await this.makeEdits(session, commandParameters, lfFile);
                         return false;
                     }
                 }
                 throw new ImperativeError({
-                    msg: `Failed to save edits because remote has changed since last downloading file. Edits have been stored locally: ${lfFile.path}`
+                    msg: `Failed to save edits because remote has changed since last downloading file.` +
+                    `Edits have been stored locally: ${lfFile.path}`
                 });
             }
         }catch(err){
@@ -159,6 +163,6 @@ export class File {
             if (err) throw new ImperativeError({
                 msg: `Temporary file could not be deleted: ${tmpDir}`
             });
-        })
+        });
     }
-  }
+}
--- a/packages/cli/src/zosfiles/edit/Edit.utils.ts
+++ b/packages/cli/src/zosfiles/edit/Edit.utils.ts
@@ -9,11 +9,10 @@
 *
 */
 
-import { AbstractSession, ICommandArguments, IHandlerParameters, ImperativeError, ProcessUtils } from "@zowe/imperative";
+import { AbstractSession, IHandlerParameters, ImperativeError } from "@zowe/imperative";
 import { PathLike } from "fs";
 import { tmpdir } from "os";
-import path = require("path");
-import { Download, Upload, IZosFilesOptions, IZosFilesResponse } from "@zowe/zos-files-for-zowe-sdk";
+import { Download, Upload, IZosFilesResponse } from "@zowe/zos-files-for-zowe-sdk";
 import LocalfileDatasetHandler from "../compare/lf-ds/LocalfileDataset.handler";
 import { CompareBaseHelper } from "../compare/CompareBaseHelper";
 import { CliUtils } from "@zowe/imperative";
@@ -28,20 +27,18 @@ export enum Prompt {
 
 export class File {
     path: string;
-    name: string;
-    data: IZosFilesResponse;
-    etag: string;
+    fileName: string;
+    apiData: IZosFilesResponse;
 }
   
   
   export class EditUtilities {
-
-    // 2. build tmp_dir
-    public static async buildTempDir(session: AbstractSession, fileName: PathLike, isUssFile?: boolean): Promise<string>{
+    // Build tmp_dir
+    public static async buildTempDir(fileName: PathLike, isUssFile?: boolean): Promise<string>{
         let tmpDir = tmpdir();
 
         if (isUssFile){
-            // 2a. hash if uss fileName
+            // Hash in a repeatable way if uss fileName
             const crypto = require("crypto");
             return tmpDir +"/" + crypto.createHash("shake256", { outputLength: 8 })
                 .update(fileName)
@@ -51,7 +48,7 @@ export class File {
         }
     }
 
-    // 3. check for tmp_dir's existance as stash
+    // Check for tmp_dir's existance as stash
     public static async checkForStash(tmpDir: string): Promise<boolean>{
         try {
           if (existsSync(tmpDir)) {
@@ -64,21 +61,13 @@ export class File {
         }
     }
 
-    // override stash?
-    public static async newStash(session: AbstractSession, fileName: string, isUssFile?: boolean): Promise<IZosFilesResponse>{
-        if (isUssFile){
-            return await Download.ussFile(session, fileName, {returnEtag: true});
-        }
-        return await Download.dataSet(session, fileName, {returnEtag: true});
-    }
-
     public static async promptUser(prompt: Prompt, fileInfo?: string): Promise<boolean>{
         let input;
         switch (prompt){
             case Prompt.useStash:
                 input = await CliUtils.readPrompt("Keep and continue editing found stash? Y/n");
                 if (input === null) {
-                    // abort the command
+                    // abort the command ... maybe do something w esc
                 } 
                 if (input == lowerCase("y")){
                     // keep stash
@@ -115,64 +104,52 @@ export class File {
         const handler = new LocalfileDatasetHandler;
         const helper = new CompareBaseHelper(commandParameters);
 
-        // 4ba. perform file comparison, show output in terminal
         commandParameters.arguments.localFilePath = lfFile.path;
         const lf = await handler.getFile1(session, commandParameters.arguments, helper);
         const mfds = await handler.getFile2(session, commandParameters.arguments, helper);
-
-        // ProcessUtils.openInDefaultApp(lfFile.path)
-
-        // // 5a. check for default editor and headless environment
-        // // 5b. open lf in editor or tell user to open up on their own if headless or no set default
+        // Editor will open with local file if default editor was set
         return await helper.getResponse(helper.prepareContent(lf), helper.prepareContent(mfds));
     }
 
     public static async makeEdits(session: AbstractSession, commandParameters: IHandlerParameters, lfFile: File): Promise<void>{
+        // Perform file comparison: show diff in terminal, open lf in editor
         await this.fileComparison(session, commandParameters, lfFile);
-        // 5a. check for default editor and headless environment
-        // 5b. open lf in editor or tell user to open up on their own if headless or no set default
-        // 4ba. perform file comparison, show output in terminal
         await this.promptUser(Prompt.doneEditing);
-        // 7. once input recieved, upload tmp file with saved ETAG
     }
 
 
     public static async uploadEdits(session: AbstractSession, commandParameters: IHandlerParameters, lfFile: File, mfFile: File): Promise<boolean>{
-    // 7. once input recieved, upload tmp file with saved ETAG
-    // 7a. if matching ETAG: sucessful upload, destroy tmp file -> END
-    // 7a. if non-matching ETAG: unsucessful upload -> 4a
+    // !! WIP !!
+    // Once input recieved, upload tmp file with saved etag
+    // if matching etag: sucessful upload, destroy tmp file -> END
+    // if non-matching etag: unsucessful upload -> perform file comparison/edit again with new etag
         let response: IZosFilesResponse;
         try{
-            response = await Upload.fileToDataset(session, lfFile.path, lfFile.name, {etag: lfFile.etag});
+            response = await Upload.fileToDataset(session, lfFile.path, lfFile.fileName, {etag: lfFile.apiData.apiResponse.etag});
             if (response.success){
-                // 7a. if matching ETAG & successful upload, destroy tmp file -> END
+                // If matching etag & successful upload, destroy tmp file -> END
                 await this.destroyTempFile(lfFile.path);
                 return true;
             }
             if (response.errorMessage){
                 if (response.errorMessage.includes("etag")){ //or error 412
-                    //alert user that the version of document theyve been editing has changed.
-                    //ask if they want to continue working w this file
+                    //alert user that the version of document they've been editing has changed
+                    //ask if they want to continue working w their stash (local file)
                     let continueToEdit: boolean = await this.promptUser(Prompt.continueEditing, lfFile.path);
                     if (continueToEdit){
-                        // if yes, 
-                        // 1. download dataset again, refresh the etag of lfFile
-                        mfFile.data = await Download.dataSet(session, mfFile.name, {returnEtag: true});
-                        // [TO DO - set etag]
-                        lfFile.etag = mfFile.etag ;
-                        // 2. then perform a file comparision:
+                        // Download dataset again, refresh the etag of lfFile
+                        mfFile.apiData, lfFile.apiData = await Download.dataSet(session, lfFile.fileName, {returnEtag: true, file: lfFile.path, overwrite: false});
+                        // Then perform file comparision w mfds and lf(file youve been editing) with updated etag
                         await this.makeEdits(session, commandParameters, lfFile);
-                        // 3. perform file comparision w mfds and lf(file youve been editing) with updated etag
                         return false;
                     }
                 }
                 throw new ImperativeError({
                     msg: `Failed to save edits because remote has changed since last downloading file. Edits have been stored locally: ${lfFile.path}`
                 });
-                return false;
             }
         }catch(err){
-            // unsuccessful upload, potential mismatched etag
+            // do some error handling here
             return false;
         }
     }
--- a/packages/cli/src/zosfiles/edit/Edit.utils.ts
+++ b/packages/cli/src/zosfiles/edit/Edit.utils.ts
@@ -9,34 +9,147 @@
 *
 */
 
-import { IImperativeConfig, Imperative, ProcessUtils } from "@zowe/imperative";
+import { AbstractSession, ICommandArguments, IHandlerParameters, ImperativeError, ProcessUtils } from "@zowe/imperative";
 import { PathLike } from "fs";
 import { tmpdir } from "os";
 import path = require("path");
-import { ICompareFileOptions } from "../compare/doc/ICompareFileOptions";
+import { Download, Upload, IZosFilesOptions, IZosFilesResponse } from "@zowe/zos-files-for-zowe-sdk";
+import LocalfileDatasetHandler from "../compare/lf-ds/LocalfileDataset.handler";
+import { CompareBaseHelper } from "../compare/CompareBaseHelper";
+import { CliUtils } from "@zowe/imperative";
+import { lowerCase } from "lodash";
+import { unlink, existsSync } from "fs";
 
-/**
- * Get the Imperative config object which defines properties of the CLI.
- * This allows it to be accessed without calling Imperative.init.
- */
-export function getImperativeConfig(): IImperativeConfig {
-    return require("./imperative");
+export enum Prompt {
+    useStash,
+    doneEditing, 
+    continueEditing
 }
 
-// 2. build tmp_dir
-export function buildTempDir(sessionInfo, fileName, isUssFile): PathLike{
-    return tmpdir
+export class File {
+    path: string;
+    name: string;
+    data: IZosFilesResponse;
+    etag: string;
 }
-// 2a. hash uss file name
-// 3. check for tmp_dir's existance as stash
-// 4a. if prexisting tmp_dir: override stash
-// 4b. if prexisting tmp_dir: use stash
-// 4ba. perform file comparison, show output in terminal
-// 4bb. overwrite ETAG 
-// 5a. check for default editor and headless environment
-ProcessUtils.isGuiAvailable
-// 5b. open lf in editor or tell user to open up on their own if headless or no set default
-// 6. wait for user input to continue
-// 7. once input recieved, upload tmp file with saved ETAG
-// 7a. if matching ETAG: sucessful upload, destroy tmp file -> END
-// 7a. if non-matching ETAG: unsucessful upload -> 4a
+  
+  
+  export class EditUtilities {
+
+    // 2. build tmp_dir
+    public static async buildTempDir(session: AbstractSession, fileName: PathLike, isUssFile?: boolean): Promise<string>{
+        let tmpDir = tmpdir();
+
+        if (isUssFile){
+            // 2a. hash if uss fileName
+            const crypto = require("crypto");
+            return tmpDir +"/" + crypto.createHash("shake256", { outputLength: 8 })
+                .update(fileName)
+                .digest("hex");
+        }else{
+            return tmpDir + "/" + fileName;
+        }
+    }
+
+    // 3. check for tmp_dir's existance as stash
+    public static async checkForStash(tmpDir: string): Promise<boolean>{
+        try {
+          if (existsSync(tmpDir)) {
+            return true;
+          }else{
+            return false;
+          }
+        } catch(err) {
+          console.error(err); //change to imperative error probably?
+        }
+    }
+
+    // override stash?
+    public static async newStash(session: AbstractSession, fileName: string, isUssFile?: boolean): Promise<IZosFilesResponse>{
+        if (isUssFile){
+            return await Download.ussFile(session, fileName, {returnEtag: true});
+        }
+        return await Download.dataSet(session, fileName, {returnEtag: true});
+    }
+
+    public static async promptUser(prompt: Prompt, fileInfo?: string): Promise<boolean>{
+        let input;
+        switch (prompt){
+            case Prompt.useStash:
+                input = await CliUtils.readPrompt("Keep and continue editing found stash? Y/n");
+                if (input === null) {
+                    // abort the command
+                } 
+                if (input == lowerCase("y")){
+                    // keep stash
+                    return true;
+                } else {
+                    // override stash
+                    return false;
+                }
+            case Prompt.doneEditing:
+                input = await CliUtils.readPrompt("Enter any value in terminal once finished editing temporary file");
+                if (input === null) {
+                    // abort the command
+                }else{
+                    return true;
+                }
+            case Prompt.continueEditing:
+                ``
+                input = await CliUtils.readPrompt("The version of the document you were editing has changed. Continue to make changes? Y/n\n" + fileInfo);
+                if (input === null) {
+                    // abort the command
+                } 
+                if (input == lowerCase("y")){
+                    // keep stash
+                    return true;
+                } else {
+                    // override stash
+                    return false;
+                }
+        }
+
+    }
+
+    public static async fileComparison(session: AbstractSession, commandParameters: IHandlerParameters, lfFile: File): Promise<IZosFilesResponse>{
+        const handler = new LocalfileDatasetHandler;
+        const helper = new CompareBaseHelper(commandParameters);
+
+        // 4ba. perform file comparison, show output in terminal
+        commandParameters.arguments.localFilePath = lfFile.path;
+        const lf = await handler.getFile1(session, commandParameters.arguments, helper);
+        const mfds = await handler.getFile2(session, commandParameters.arguments, helper);
+
+        // ProcessUtils.openInDefaultApp(lfFile.path)
+
+        // // 5a. check for default editor and headless environment
+        // // 5b. open lf in editor or tell user to open up on their own if headless or no set default
+        return await helper.getResponse(helper.prepareContent(lf), helper.prepareContent(mfds));
+    }
+
+
+
+
+    public static async uploadEdits(session: AbstractSession, lfFile: File): Promise<IZosFilesResponse>{
+    // 7. once input recieved, upload tmp file with saved ETAG
+    // 7a. if matching ETAG: sucessful upload, destroy tmp file -> END
+    // 7a. if non-matching ETAG: unsucessful upload -> 4a
+        let response: IZosFilesResponse;
+        try{
+            response = await Upload.fileToDataset(session, lfFile.path, lfFile.name, {etag: lfFile.etag});
+            // successful upload
+            return response;
+        }catch(err){
+            // unsuccessful upload, potential mismatched etag
+            return response;
+        }
+    }
+
+    public static async destroyTempFile(tmpDir:string): Promise<void>{
+        unlink (tmpDir, (err) => {
+            if (err) throw new ImperativeError({
+                msg: `Temporary file could not be deleted: ${tmpDir}`
+            });
+        })
+    }
+  }
--- a/packages/cli/src/zosfiles/edit/ds/Dataset.definition.ts
+++ b/packages/cli/src/zosfiles/edit/ds/Dataset.definition.ts
@@ -1,0 +1,54 @@
+/*
+* This program and the accompanying materials are made available under the terms of the
+* Eclipse Public License v2.0 which accompanies this distribution, and is available at
+* https://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import {ICommandDefinition} from "@zowe/imperative";
+import { EditOptions } from "../Edit.options";
+import i18nTypings from "../../-strings-/en";
+
+// Does not use the import in anticipation of some internationalization work to be done later.
+const strings = (require("../../-strings-/en").default as typeof i18nTypings).EDIT;
+
+/**
+ * View data set command definition containing its description, examples and/or options
+ * @type {ICommandDefinition}
+ */
+export const DatasetDefinition: ICommandDefinition = {
+    name: "data-set",
+    aliases: ["ds"],
+    summary: strings.ACTIONS.DATA_SET.SUMMARY,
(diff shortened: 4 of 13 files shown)


packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,225 @@
...
+            if (err) throw new ImperativeError({
+                msg: `Temporary file could not be deleted: ${tmpDir}`,
+                causeErrors: err
+            });
+        });
+    }
+
+    public static async errorHandler(error: RestClientError): Promise<void>{
+        if (error.causeErrors.code == 'ENOTFOUND'){
+            throw new ImperativeError({
```
ATorrise, line 215: Decided to make a function that both USS and DS handlers can use so error description is only written once. This makes errors pretty long whereas when I would through it in the catch of the handler it was quite short. Unsure if i should keep this method or just throw errors back in the handlers and have some repeated code. What do you guys think?

Longer error from uss/ds shared method:
image

Concise error and repeated code:
image

packages/cli/src/zosfiles/compare/lf-ds/LocalfileDataset.handler.ts
```diff
@@ -25,4 +25,8 @@ export default class LocalfileDatasetHandler extends CompareBaseHandler {
     public async getFile2(session: AbstractSession, args: ICommandArguments, helper: CompareBaseHelper): Promise<Buffer> {
         return await Get.dataSet(session, args.dataSetName, { ...helper.file2Options, task: helper.task });
     }
+    public async getFile3(session: AbstractSession, args: ICommandArguments, helper: CompareBaseHelper): Promise<Buffer> {
```
gejohnston, line 28: I know that this is entirely semantics, but here goes:

When I first saw a new function named getFile3, my first thought was "are 3 files somehow involved"?
I then realized that your were probably just extending the existing function naming convention.

If I correctly understand that we still only have 2 files involved, but you are extending this class to support a different type of file, my suggestion would be to call your new function "getUssFile2".

Geez, this is a verbose comment for such a minor suggestion.

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,227 @@
...
+    // Build tmp path
+    public static async buildTmpDir(commandParameters: IHandlerParameters): Promise<string>{
+        const ext = (commandParameters.arguments.extension ? commandParameters.arguments.extension : ".txt");
+        const hashLen = 8;
+        let fileName: string;
+        if (commandParameters.positionals.includes('uss')){
+            // Hash in a repeatable way if uss fileName (to get around any potential special characters in name)
+            fileName = commandParameters.arguments.file;
+            const crypto = require("crypto");
+            const hash = crypto.createHash('sha1').update(fileName).digest('hex').substring(0,hashLen);
```
gejohnston, line 40: What is our confidence that the 1st eight characters of the digest will be unique among any other files that might be placed in your temp directory? This may not be a problem – just asking.

packages/cli/src/zosfiles/compare/lf-ds/LocalfileDataset.handler.ts
ATorrise: Thank you for pointing this out. After looking at things more critically, I decided to import a second handler, lf-uss (which already has the correct get method for its file 2), rather than extending lf-ds. I didn't realize at the time but my creation of getFile3 really doesn't make sense with the base class these handlers are extending (CompareBase.Handler.ts)

packages/cli/src/zosfiles/edit/Edit.utils.ts
ATorrise: image
I'm not sure if this is a good answer but the above image shows that the hashing is repeatable and very different even when file names differ just by 1 change

packages/cli/src/zosfiles/edit/Edit.utils.ts
gejohnston: I think that the longer errors (and their stack trace) make it appear as if the command crashed, instead of its actual behavior in which it detected some problem, reported the problem, and then gracefully ended.

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,229 @@
...
+        const tmpDir = commandParameters.arguments.localFilePath;
+        // Perform file comparison: show diff in terminal, open lf in editor
+        if (commandParameters.arguments.editor){
+            await ProcessUtils.openInEditor(tmpDir, commandParameters.arguments.editor);
+        }
+        await this.promptUser(Prompt.doneEditing, tmpDir);
+
+    }
+
+    public static async uploadEdits(session: AbstractSession, commandParameters: IHandlerParameters, lfFile: LocalFile): Promise<boolean>{
```
gejohnston, line 146: I know that you asked for an early review focused on the programming logic, but I will make this comment anyway:

I really like your line comments before individual paragraphs of logic. I think that you should additionally place typedoc-style block comments before every function.

packages/cli/src/zosfiles/edit/Edit.utils.ts
ATorrise: Oh I actually reread this and realized I read it wrong initially. I don't have an actual confidence level. Maybe prepending 'zoweTemp' to the hash will be good enough to prevent any collisions?  So `C:\Users\user\AppData\Local\Temp\08deae8d.c` becomes `C:\Users\user\AppData\Local\Temp\zoweTemp08deae8d.c`

packages/cli/src/zosfiles/edit/Edit.utils.ts
ATorrise: Oh yes! Thank you and sorry I forgot to add those in. Will fix!

packages/cli/src/zosfiles/edit/Edit.utils.ts
traeok: @ATorrise You might be able to leverage `sha256` instead of `sha1` here, as I believe no collisions have been found with that algorithm yet.

packages/cli/src/zosfiles/edit/Edit.options.ts
```diff
@@ -0,0 +1,42 @@
...
+    editor: {
+        name: "editor",
+        aliases: ["ed"],
+        description: strings.EDITOR,
+        type: "string",
+        required: false
+    },
+    extension: {
+        name: "extension",
+        aliases: ["e"],
```
t1m0thyj, line 37: The alias `--e` is convenient but also ambiguous since users might expect `--e` to be an alias for `--editor`.

Would it make sense to use the alias `--ext` here instead?

packages/cli/src/zosfiles/edit/Edit.utils.ts
ATorrise: ty trae!

packages/cli/src/zosfiles/edit/Edit.options.ts
ATorrise: image

Accidentally attaching the largest picture ever but I went with --e because other commands implement this. If changing to --ext, should i edit all other instances of --e->--ext to be consistent?

packages/zosfiles/src/methods/edit/Edit.ts
```diff
@@ -0,0 +1,92 @@
...
+import { ZosmfRestClient, ZosmfHeaders, IHeaderContent } from "@zowe/core-for-zowe-sdk";
+import { IEditOptions } from "./doc/IEditOptions";
+import { ZosFilesUtils } from "../../utils/ZosFilesUtils";
+
+/**
+ * This class holds helper functions that are used to edit the content of data sets or USS files through the z/OSMF APIs
+ * @export
+ * @class Get
+ */
+export class Edit {
```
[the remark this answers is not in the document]
ATorrise: Yeah I don't think I've changed anything in this file. I'm not sure if its needed. Will remove

packages/cli/src/zosfiles/edit/Edit.options.ts
t1m0thyj: Sorry, didn't realize that other commands implement this as `-e`. To maintain backwards compatibility and consistency, I agree with leaving it as-is 👍

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
```diff
@@ -171,28 +171,40 @@ export class CompareBaseHelper {
...
+     * @param {string} string1 - string of file 1 content
+     * @param {string} string2 - string of file 2 content
+     * @param {IDiffOptions} options
      * @returns {IZosFilesResponse}
      * @public
      * @memberof CompareBaseHelper
      */
-    public async getResponse(string1: string, string2: string): Promise<IZosFilesResponse>{
-        //  CHECHKING IIF THE BROWSER VIEW IS TRUE, OPEN UP THE DIFFS IN BROWSER
+    public async getResponse(string1: string, string2: string, options?: IDiffOptions): Promise<IZosFilesResponse>{
```
traeok, line 182: This is more of a style choice and this will come off as nitpicking, but I don't find the parameter naming convention to be clear here. Within the context of a diff, I think someone can infer what `string1` and `string2` are, but self-describing parameters would help clear up any confusion for others that might incorporate/use this compare class.

Maybe something like `oldContent` and `newContent` would work here? If you want to maintain consistency with `IDiffOptions`, then feel free to disregard this comment, just a suggestion 😅 This could be something to consider down the road

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,252 @@
...
+    /**
+     * Builds a temp path where local file will be saved. If uss file, file name will be hashed
+     * to prevent any conflicts with file naming. A given filename will always result in the
+     * same unique file path.
+     * @param {IHandlerParameters} commandParameters - parameters supplied by args
+     * @returns {Promise<string>} - returns unique file path for temp file
+     * @memberof EditUtilities
+     */
+    public static async buildTempPath(commandParameters: IHandlerParameters): Promise<string>{
+        const ext = (commandParameters.arguments.extension ? commandParameters.arguments.extension : ".txt");
```
traeok, line 47: Parentheses can be omitted on line 47

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,252 @@
...
+        let fileName: string;
+        if (commandParameters.positionals.includes('uss')){
+            // Hash in a repeatable way if uss fileName (to get around any potential special characters in name)
+            fileName = commandParameters.arguments.file;
+            const crypto = require("crypto");
+            const hash = crypto.createHash('sha256').update(fileName).digest('hex');
+            return tmpdir() +"\\" + hash  + "." + ext;
+        }else{
```
traeok, line 55: Should be able to remove the `else` branch here, since you will return early anyway if the logic runs in the `if` statement

packages/cli/src/zosfiles/edit/ds/Dataset.handler.ts
```diff
@@ -0,0 +1,89 @@
...
+        } while (!uploaded);
+
+        return {
+            success: true,
+            commandResponse: TextUtils.chalk.green(
+                "Successfully uploaded edited file to mainframe"
+            )
+        };
+    }
+}
```
traeok, line 89: I noticed part of the logic for this handler is similar to what is implemented in the USS handler. Could we lift this out into a common function to minimize code duplication?

traeok (commented):
Everything is looking good so far - thanks @ATorrise

packages/cli/src/zosfiles/edit/ds/Dataset.definition.ts
```diff
@@ -0,0 +1,59 @@
...
+            options: `"ibmuser.cntl(iefbr14)" "--editor" "notepad"`
+        },
+        {
+            description: strings.ACTIONS.DATA_SET.EXAMPLES.EX1,
+            options: `"ibmuser.cntl(iefbr14)" "--editor" "C:\\Windows\\System32\\Notepad.exe"`
+        },
+        {
+            description: strings.ACTIONS.USS_FILE.EXAMPLES.EX2,
+            options: `"ibmuser.jcl(iefbr14)" "--editor" "notepad" "--extension" "jcl"`
+        }
```
adam-wolfe, line 57: It would probably be better to remove the quotation marks around --editor and --extension in these examples.

packages/cli/src/zosfiles/edit/Edit.utils.ts
ATorrise: ooo cool thank you!

packages/cli/src/zosfiles/edit/ds/Dataset.definition.ts
ATorrise: good catch thank you!

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,328 @@
...
+    uss = 'uss',
+    ds = 'ds'
+}
+
+/**
+ * A class to hold pertinent information about the local file during the editing process
+ * @export
+ * @class
+ */
+export class LocalFile {
```
t1m0thyj, line 51: Since `LocalFile` defines only properties and not methods, an `interface` might be a better fit than a `class`.

packages/cli/src/zosfiles/edit/Edit.handler.ts
```diff
@@ -0,0 +1,85 @@
...
+            if (error.causeErrors && error.causeErrors.code == 'ENOTFOUND'){
+                throw new ImperativeError({
+                    msg: TextUtils.chalk.red(`ENOTFOUND: Unable to connect to mainframe.`),
+                    causeErrors: error
+                });
+            }
+            throw new ImperativeError({
+                msg: TextUtils.chalk.red(`File not found on mainframe. Command terminated.`),
+                causeErrors: error
+            });
```
t1m0thyj, line 65: For error handling, we want to be careful about assuming what an error means so that our error message isn't misleading. Also wrapping an error that may already be an ImperativeError in a new ImperativeError may obscure the original error message.

If we want to make the error message more useful for data sets that are not found, how about something like this that catches 404 errors to handle them specially, and for other error types simply rethrows them?
```typescript
if (error instanceof ImperativeError && error.errorCode === RestConstants.HTTP_STATUS_404) {
    throw new ImperativeError({
        msg: TextUtils.chalk.red(`File not found on mainframe. Command terminated.`),
        causeErrors: error
    });
} else {
    throw error;
}
```

packages/cli/src/zosfiles/edit/Edit.options.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * The option to set a default editor
+     * @type {ICommandOptionDefinition}
+     */
+    editor: {
+        name: "editor",
+        aliases: ["ed"],
+        description: strings.EDITOR,
+        defaultValue: process.env['ZOWE_EDITOR'] ?? defaultEditor,
```
t1m0thyj, line 32: I think we were considering to deprecate `ZOWE_EDITOR` in other commands that currently use it.
Do we need to support `ZOWE_EDITOR` for this command?
Or is it sufficient that `ZOWE_OPT_EDITOR` is already supported?

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,328 @@
...
+
+/**
+ * An enum to supply more structure within {@link LocalFile}
+ * @export
+ * @enum
+ */
+export enum EditFileType {
+    uss = 'uss',
+    ds = 'ds'
+}
```
t1m0thyj, line 44: Since there are only 2 possible values that are both string literals, a union type might be a better fit than an `enum`. For example: `export type EditFileType = "uss" | "ds";`

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,328 @@
...
+     * @memberof EditUtilities
+     */
+    public static async promptUser(prompt: Prompt, tempPath?: string): Promise<boolean>{
+        let input;
+        switch (prompt){
+            case Prompt.useStash:
+                input = await CliUtils.readPrompt(TextUtils.chalk.green(`Keep and continue editing found stash? Y/n`));
+                if (input === null) {
+                    throw new ImperativeError({
+                        msg: `No input provided. Command terminated.`
```
t1m0thyj, line 115: Should this message be wrapped in `TextUtils.chalk.red` like the other error messages below?

I'm kind on the fence about whether forcing error messages to be in red text is a good idea 🙂 Typically we already output errors in red so making the message red here too may be a bit redundant.

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,328 @@
...
+            {
+                returnEtag: true,
+                file: tempPath
+            }
+        ];
+        if(lfFile.fileType === 'uss'){
+            lfFile.zosResp = await Download.ussFile(...args);
+        }
+        lfFile.zosResp = await Download.dataSet(...args);
+        // TODO: get rid of the fake temp file generated by the useStash case (wouldnt have to do this if the option {overwrite: false} worked)
```
t1m0thyj, line 173: Good catch on the bug with `overwrite = False` not working - seems like something worth fixing in this PR if its a quick fix, or otherwise creating a separate issue for it 👍

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,328 @@
...
+        if (commandParameters.positionals.includes('ds')){
+            mfds = await handlerDs.getFile2(session, commandParameters.arguments, helper);
+        }else{
+            mfds = await handlerUss.getFile2(session, commandParameters.arguments, helper);
+        }
+
+        //if browserview, open diff in browser, otherwise print diff in terminal
+        const response = await helper.getResponse(helper.prepareContent(lf), helper.prepareContent(mfds), options);
+        if (!helper.browserView){
+            console.log('\n'+response.commandResponse);
```
t1m0thyj, line 211: The linter is complaining here because we don't allow `console.log` in our codebase. I think this could be refactored to move the logging to console into the command handler and use `params.response.console.log` on the IHandlerParameters API.

packages/cli/src/zosfiles/edit/uss/USSFile.definition.ts
```diff
@@ -0,0 +1,59 @@
...
+            description: strings.ACTIONS.USS_FILE.EXAMPLES.EX1,
+            options: `"/a/ibmuser/my_text.txt" --editor "notepad"`
+        },
+        {
+            description: strings.ACTIONS.USS_FILE.EXAMPLES.EX1,
+            options: `"/a/ibmuser/my_text.txt" --editor "C:\\Windows\\System32\\Notepad.exe"`
+        },
+        {
+            description: strings.ACTIONS.USS_FILE.EXAMPLES.EX2,
+            options: `"C:\\Windows\\System32\\Notepad.exe" "jcl" --editor "notepad" --extension "jcl"`
```
t1m0thyj, line 56: Same comment as above regarding quote marks not being needed in command examples.

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+- Enhancement: Created edit commands to edit a dataset or uss file locally [PR #1672](https://github.com/zowe/zowe-cli/pull/1672)
```
t1m0thyj, line 6: The changelog entry can be removed from this file if nothing has changed in the zosfiles SDK in this PR 🙂

adam-wolfe:
Not sure if this is a known issue, but "zowe files edit **ds** DATASET" works as expected, however, with "zowe files edit **data-set** DATASET", I get the following error:

```
Command Error:        
z/OSMF REST API Error:
Rest API failure with HTTP(S) status 404
category: 1
rc:       8
reason:   9
message:  Error executing command. exit_code=1
details:
  - /bin/chtag -p /DATASET
  - chtag: FSUM6180 file \"/DATASET\": EDC5129I No such file or directory. (errno2=0x053B006C)


Error Details:
HTTP(S) error status "404" received.

Protocol:  https
Host:      host
Port:      port
Base Path: /ibmzosmf/api/v1
Resource:  /zosmf/restfiles/fs/DATASET
Request:   PUT
Headers:   [{"Content-Type":"application/json"},{"Content-Length":"35"},{"Accept-Encoding":"gzip"},{"X-CSRF-ZOSMF-HEADER":true}]
Payload:   { request: 'chtag', action: 'list' }
```

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,323 @@
...
+            lfFile.fileName,
+            {
+                returnEtag: true,
+                file: tempPath
+            }
+        ];
+        if(lfFile.fileType === 'uss'){
+            lfFile.zosResp = await Download.ussFile(...args);
+        }
+        lfFile.zosResp = await Download.dataSet(...args);
```
gejohnston: Was this the spot where you said that you knew you needed an else-if statement?

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,323 @@
...
+                    if (input === null) {
+                        throw new ImperativeError({
+                            msg: TextUtils.chalk.red(`No input provided. Command terminated. Stashed file will persist: ${tempPath}`)
+                        });
+                    }
+                    return true;
+                }
+            case Prompt.continueToUpload:
+                input = await CliUtils.readPrompt(TextUtils.chalk.green(`Continue uploading edits despite changes on remote? Y/n`));
+                if (input === null) {
```
gejohnston, line 130: Are you testing for a trinary decision with the "input == null" tests in this function? That is to say:  yes, no, or abort?

If you really only want a yes/no answer, you could loop on the prompts until you get a yes/not-yes response, and never throw an exception.

gejohnston:
If you are genuinely ready to complete this work, I suggest that you merge the master branch into this PR and remove this PR from draft status.

adam-wolfe:
I would suggest enabling the "uf" alias for USS files, it seems like most of the zosfiles commands use: `aliases: ["uf", "uss"],`

adam-wolfe:
Some thoughts on the done message: 

- Would it make sense to make "done" a different color to make that stand out in the message that appears when the user closes their editor?
- Should a user have the option to instead enter something like "cancel" to prevent changes from being uploaded. The only way to get out of the message is to use ctrl+c.

adam-wolfe:
When trying to edit a larger dataset (2.3MB text file), notepad does not know where to save the file. If I delete all of the data from the dataset, notepad works normally (ctrl-s saves to the right place). This does not happen with vim, gvim, or vscode...

packages/cli/src/zosfiles/edit/Edit.utils.ts
ATorrise: If null that means prompt was timed out, which is the cue to cancel the command rather than leaving it running in someones terminal forever. but maybe we never want to time out?

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+- Enhancement: Created zos-files edit commands to edit a dataset or uss file locally [PR #1672](https://github.com/zowe/zowe-cli/pull/1672)
+
+## `7.15.0`
```
gejohnston, line 8: This entry for 7.15.0 seems to be a duplicate of the same entry below.

traeok (commented):
I was able to edit both a data set member and USS file using Notepad 👍 this will be really helpful for quickly updating content on the mainframe 🙂 

However, I ran into some build errors when trying to build and test locally. Once making a couple changes I was able to build the CLI and test the edit command.

In `CompareBaseHelper.ts`:
image

In `Edit.utils.ts`:
image

t1m0thyj:
> I was able to edit both a data set member and USS file using Notepad 👍 this will be really helpful for quickly updating content on the mainframe 🙂
> 
> However, I ran into some build errors when trying to build and test locally. Once making a couple changes I was able to build the CLI and test the edit command.

In order to build and test this branch, I believe it is necessary to pull down the associated Imperative branch and build/link it to the CLI: https://github.com/zowe/imperative/pull/967

traeok:
> > I was able to edit both a data set member and USS file using Notepad 👍 this will be really helpful for quickly updating content on the mainframe 🙂
> > However, I ran into some build errors when trying to build and test locally. Once making a couple changes I was able to build the CLI and test the edit command.
> 
> @traeok In order to build and test this branch, I believe it is necessary to pull down the associated Imperative branch and build/link it to the CLI: [zowe/imperative#967](https://github.com/zowe/imperative/pull/967)

Thanks for pointing out that PR, I thought there was an imperative one that was tied together but I overlooked the title 😅

traeok (changes requested):
I was able to test now that I linked the PR from imperative during building. Sorry for the confusion.
I tested a few scenarios, some with conflicts and some without, different editors and extensions, etc. All of those seem good to go except for the following:

- `zowe files edit uss /z/OKTAID/Demo/polling.sh`, with introduced conflict on mainframe before save

When trying to view the diff to see the conflict, a command error appears as there is an issue with the e-tag:
image

Here's the output I get when dealing with a data set conflict, which seems to be the intended behavior:
image

Related question: For the option "Continue uploading edits despite changes on remote? Y/n", do we want to keep "yes" as the default option? I'm receptive to either as the default, but a user might accidentally replace their remote changes by pressing Enter. Granted, if the user is editing a file, they will likely want to overwrite those remote changes, so this can go either way.

---

Also, when viewing the data set diff, I noticed that the diff shows all lines have changed from local to remote, but the only character that was added was the "h" to "sh":
image

I think this might have to deal with the file names. Since the file names are "local file" and "remote file" for the diff viewer, the diff utility interprets it as the file being renamed, so the diff shows the whole file changing as a result. If the name was the same for `name1` and `name2`, it should show the true diff.

ATorrise:
> Related question: For the option "Continue uploading edits despite changes on remote? Y/n", do we want to keep "yes" as the default option? I'm receptive to either as the default, but a user might accidentally replace their remote changes by pressing Enter. Granted, if the user is editing a file, they will likely want to overwrite those remote changes, so this can go either way.

Thank you for your review @traeok I appreciate it!

The default is "no" or "false" if just pressing enter : `return input.toLowerCase() === 'y'`

Since it's probably bad to assume anything, I've changed the code to only accept ay or n value, which looks like this: 
image

Thoughts?

traeok:
> > Related question: For the option "Continue uploading edits despite changes on remote? Y/n", do we want to keep "yes" as the default option? I'm receptive to either as the default, but a user might accidentally replace their remote changes by pressing Enter. Granted, if the user is editing a file, they will likely want to overwrite those remote changes, so this can go either way.
> 
> Thank you for your review @traeok I appreciate it!
> 
> The default is "no" or "false" if just pressing enter : `return input.toLowerCase() === 'y'`
> 
> Since it's probably bad to assume anything, I've changed the code to only accept ay or n value, which looks like this: image
> 
> Thoughts?

No problem!

In hindsight, I misunderstood the prompt. I was under the impression that since the "Y" was capitalized, that it would be the default answer if I pressed enter. If the default answer _is_ no, then you might want to update the message so the "N" is capitalized and the "y" is lower-case to clarify that for the end-user. But, the check that you mentioned (`return input.toLowerCase() === 'y'`) looks good to me 🙂

ATorrise:
@traeok would you mind seeing if you get your reported errors again? pretty sure I've fixed the diff and the command error. thanks!

packages/cli/src/zosfiles/edit/Edit.utils.ts
ATorrise: after talking this through with Fernando, I've realized the functionality was intended to work in a different way than I thought and expected so it seems there's no issue/bug to be fixed

traeok (approved):
Tested again w/ the "USS w/ conflict" test case and looks like the issue has been fixed 👍 

LGTM! Thanks @ATorrise 😀

I noticed that one of the dialogs shows `y/n` - if we have a default option (either "y" or "n"), it should probably be capitalized so users are aware of the default.

image

gejohnston (changes requested):
All of my comments have been addressed, but I think we have to correct the build failures before we approve this PR.

gejohnston:
Are the builds breaking because you need the changes from imperative to merged first?

packages/cli/__tests__/zosfiles/__integration__/edit/uss/__scripts__/edit_uss_default.sh
t1m0thyj, line 1: This script doesn't appear to be used by any tests, can it be removed?

packages/cli/__tests__/zosfiles/__system__/edit/ds/Edit.ds.system.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+        dsname = getUniqueDatasetName(defaultSystem.zosmf.user);
+    });
+
+    afterAll(async () => {
+        await TestEnvironment.cleanUp(testEnvironment);
+    });
+
+    // describe("Success scenarios", () => {
+    //     //this test hangs because requires user input mocking - deciding to put this in unit test
```
t1m0thyj, line 49: Is there a success scenario we could test that has only a single prompt? If so, we could fake out user input fairly easily in a test script: `echo y | zowe config edit...`  If not and the tests would require responding to multiple prompts, then feel free to disregard this comment.

packages/cli/__tests__/zosfiles/__system__/edit/uss/Edit.uss.system.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    });
+
+    // describe("Success scenarios", () => {
+    //     //this test hangs because requires user input mocking - deciding to put this in unit test
+    //     //it("should edit uss file (should upload edited local file with the correct/expected etag)", async () => {})
+    // });
+    describe("Expected failures", () => {
+        it("should fail if specified uss file doesn't exist", async () => {
+            const shellScript = path.join(__dirname, "__scripts__", "command", "edit_nonexistent_uss.sh");
+            const response = runCliScript(shellScript, testEnvironment, [dsname + ".dummy"]);
```
t1m0thyj, line 61: Should this test use `ussname` instead of `dsname`?

packages/cli/__tests__/zosfiles/__unit__/edit/ds/Dataset.definition.unit.test.ts
```diff
@@ -0,0 +1,25 @@
...
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import i18nEnglish from "../../../../../src/zosfiles/-strings-/en";
+import { DatasetDefinition } from "../../../../../src/zosfiles/edit/ds/Dataset.definition";
+
+describe("DatasetDefinition", () => {
+    it("should use the correct object for strings", () => {
+        const strings = i18nEnglish.COPY.ACTIONS.DATA_SET;
```
t1m0thyj, line 17: Should this be `i18nEnglish.EDIT.ACTIONS.DATA_SET`?

packages/cli/src/zosfiles/compare/lf-uss/LocalfileUss.handler.ts
```diff
@@ -23,6 +23,6 @@ export default class LocalfileUssHandler extends CompareBaseHandler {
         return helper.prepareLocalFile(args.localFilePath);
     }
     public async getFile2(session: AbstractSession, args: ICommandArguments, helper: CompareBaseHelper): Promise<string | Buffer> {
-        return await Get.USSFile(session, args.ussFilePath, { ...helper.file2Options, task: helper.task });
+        return await Get.USSFile(session, args.file, { ...helper.file2Options, task: helper.task });
```
t1m0thyj, line 26: Since the command definition for `zowe files compare lf-uss` defines the positional argument as `ussFilePath`, I think we need to keep that name here: https://github.com/zowe/zowe-cli/blob/e7f4484bc132ca0ef6a7cd2249c1330f5dfcafee/packages/cli/src/zosfiles/compare/lf-uss/LocalfileUss.definition.ts#L40-L45

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,341 @@
...
+            //alert user that the version of document they've been editing has changed
+            //ask if they want to see changes on the remote file before continuing
+            const viewUpdatedRemote: boolean = await this.promptUser(Prompt.viewUpdatedRemote);
+            if (viewUpdatedRemote){
+                await this.fileComparison(session, commandParameters);
+            }
+            //ask if they want to keep working with their stash (local file) or upload despite changes to remote
+            const continueToUpload: boolean = await this.promptUser(Prompt.continueToUpload, tempPath);
+            // refresh etag, keep stash
+            lfFile = await this.localDownload(session, lfFile, true);
```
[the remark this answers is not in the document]
t1m0thyj: Please fix this code smell 🙂

t1m0thyj (changes requested):
Left a few comments.

I agree with @gejohnston, since building this PR depends on https://github.com/zowe/imperative/pull/967 we need to merge that PR first. Then we can update Imperative in this PR so the build will pass 👍

packages/cli/src/zosfiles/edit/Edit.options.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * The option to set a default editor
+     * @type {ICommandOptionDefinition}
+     */
+    editor: {
+        name: "editor",
+        aliases: ["ed"],
+        description: strings.EDITOR,
+        defaultValue: process.env['ZOWE_OPT_EDITOR'] ?? defaultEditor,
```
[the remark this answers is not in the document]
ATorrise: so ` IO.getDefaultTextEditor()` is enough here?

packages/cli/src/zosfiles/compare/lf-uss/LocalfileUss.handler.ts
ATorrise: image

ATorrise: image
image
dar
image

packages/cli/src/zosfiles/edit/Edit.options.ts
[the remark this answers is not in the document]
t1m0thyj: Yes it should be 👍

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,349 @@
...
+     * @param {IHandlerParameters} commandParameters - parameters supplied by args
+     * @returns {Promise<IZosFilesResponse>} - the response generated by {@link CompareBaseHelper.getResponse}
+     * @memberof EditUtilities
+     */
+    public static async fileComparison(session: AbstractSession, commandParameters: IHandlerParameters): Promise<IZosFilesResponse>{
+        const handlerDs = new LocalfileDatasetHandler;
+        const handlerUss = new LocalfileUssHandler;
+        const helper = new CompareBaseHelper(commandParameters);
+        const gui = ProcessUtils.isGuiAvailable();
+        const options: IDiffOptions = {
```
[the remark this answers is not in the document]
ATorrise: This is because a change was also made in imperative, you'll have to pull those recent changes too to test https://github.com/zowe/imperative/pull/967

ATorrise:
@t1m0thyj the code smell you are referring to has been addressed, it just doesn't look like it on the conversation for this pr for some reason

t1m0thyj (changes requested):
The `zowe files edit uss` command works for plain-text (EBCDIC) but not binary (ASCII) USS files.

For USS files, I think we need to detect their encoding from the tag in the USS file attributes, and upload the file back to the mainframe with the same encoding.

packages/cli/src/zosfiles/edit/ds/Dataset.definition.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Edit data set command definition containing its description, examples and/or options
+ * @type {ICommandDefinition}
+ */
+export const DatasetDefinition: ICommandDefinition = {
+    name: "data-set",
+    aliases: ["ds"],
+    summary: strings.ACTIONS.DATA_SET.SUMMARY,
+    description: strings.ACTIONS.DATA_SET.DESCRIPTION,
+    type: "command",
+    handler: __dirname.substring(0, __dirname.length-("ds").length) + "Edit.handler",
```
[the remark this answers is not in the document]
ATorrise: thanks!

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,354 @@
...
+     * Builds a temp path where local file will be saved. If uss file, file name will be hashed
+     * to prevent any conflicts with file naming. A given filename will always result in the
+     * same unique file path.
+     * @param {ILocalFile} lfFile - object containing pertinent information about the local file during the editing process
+     * @returns {Promise<string>} - returns unique file path for temp file
+     * @memberof EditUtilities
+     */
+    public static async buildTempPath(lfFile: ILocalFile, commandParameters: IHandlerParameters): Promise<string>{
+        // find the appropriate extension for either uss or ds in one long operation
+        const ext = '.'  + (lfFile.fileType === 'uss'? lfFile.fileName.split(".").pop() : (commandParameters.arguments.extension ?? "txt"));
```
adam-wolfe, line 73: If the USS file has no extension, the temp file path generated here will be invalid (at least in Windows) and the file will be inaccessible.

ATorrise: Thanks Adam!

adam-wolfe: Still doing something weird:
image

t1m0thyj:
I pulled down the latest version of this branch and the command no longer works for me to open USS files.

For example when I run `zowe files edit uss` to edit a file with the `.yml` extension, the file is downloaded to `%TEMP%/0b91582bc29a9e6dcc3ac2b5c9d1f5800f8ab0b96be59d93b93de590a6183fc9.yml` but is not opened in any editor.

I would expect to see a Windows Explorer prompt asking me which editor to use for the `.yml` extension. This seems like a bug in Zowe CLI, because I verified that the Imperative method behaves correctly with the following test script:
```
const { ImperativeConfig, ProcessUtils } = require("@zowe/imperative");
ImperativeConfig.instance.loadedConfig = { envVariablePrefix: "ZOWE" };
ProcessUtils.openInEditor(require("os").tmpdir() + "/0b91582bc29a9e6dcc3ac2b5c9d1f5800f8ab0b96be59d93b93de590a6183fc9.yml");
```

image

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    /**
+     * Enable user to make their edits and wait for user input to indicate editing is complete
+     * @param {string} tempPath - parameters supplied by args
+     * @param {string} editor - optional parameter originally supplied by args
+     * @memberof EditUtilities
+     */
+    public static async makeEdits(tempPath: string, editor?: string): Promise<void>{
+        if (editor){
```
t1m0thyj, line 249: Should this `if` block be removed? It's fine if the `editor` param passed to `ProcessUtils.openInEditor` is undefined - that just means to use the default editor.

packages/cli/src/zosfiles/edit/ds/Dataset.definition.ts
```diff
@@ -0,0 +1,59 @@
...
+            name: "dataSetName",
+            description: strings.ACTIONS.DATA_SET.POSITIONALS.DATASETNAME,
+            type: "string",
+            required: true
+        },
+    ],
+    options: [
+        EditOptions.editor,
+        EditOptions.extension
+    ],
```
t1m0thyj, line 44: Should this command support `--binary`, `--encoding`, and `--record` options same as `zowe files view ds`?
(`--range` probably doesn't apply, `--volser` might but is probably less used)
image

packages/cli/src/zosfiles/edit/uss/USSFile.definition.ts
```diff
@@ -0,0 +1,59 @@
...
+            name: "file",
+            description: strings.ACTIONS.USS_FILE.POSITIONALS.USSFILE,
+            type: "string",
+            required: true
+        },
+    ],
+    options: [
+        EditOptions.editor,
+        EditOptions.extension
+    ],
```
t1m0thyj, line 44: Should this command support `--binary` and `--encoding` options same as `zowe files view uss`?
(`--range` probably doesn't apply)
image

packages/cli/src/zosfiles/edit/ds/Dataset.definition.ts
```diff
@@ -0,0 +1,59 @@
...
+    positionals: [
+        {
+            name: "dataSetName",
+            description: strings.ACTIONS.DATA_SET.POSITIONALS.DATASETNAME,
+            type: "string",
+            required: true
+        },
+    ],
+    options: [
+        EditOptions.editor,
```
t1m0thyj, line 42: Should we create an issue in the Imperative repo to remind us to add an `--editor` option on the `zowe config edit` command and remove support for custom env var `ZOWE_EDITOR` in v3?

packages/cli/CHANGELOG.md
t1m0thyj, line 1: Question about the "Requested Documentation" for this PR:
> Once this command is added in to GA, I'd like documentation to include a note about disabling the gui:
> > If you don't want to use your gui editor, you can remove your system's stored `$DISPLAY` value

I'm a bit hesitant to recommend that users unset their system's `$DISPLAY` variable because it could have unintended consequences for other apps. If we document this, would it apply only to WSL environments where someone wants to use a command-line editor?

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,353 @@
...
+    public static async buildTempPath(lfFile: ILocalFile, commandParameters: IHandlerParameters): Promise<string>{
+        // find the appropriate extension for either uss or ds in two (long) operations
+        const ussExt = (lfFile.fileType === 'uss' && lfFile.fileName.includes(".")) ? lfFile.fileName.split(".").pop() : "";
+        const ext = "."  + (lfFile.fileType === 'uss' ? ussExt : (commandParameters.arguments.extension ?? "txt"));
+        if (lfFile.fileType === 'uss'){
+            // Hash in a repeatable way if uss fileName (incase there are special characters in name)
+            const crypto = require("crypto");
+            const hash = crypto.createHash('sha256').update(lfFile.fileName).digest('hex');
+
+            return path.join(tmpdir(), hash + ext);
```
adam-wolfe, line 79: If a USS file has no extension, this will still append a period to the end of the filename of the temporary file. This prevents the user from being able to open the temporary file.

adam-wolfe (changes requested):

t1m0thyj (changes requested):
When I answer `y` to the following question, an empty diff is opened in the web browser which seems unexpected.
```
Keep and continue editing found local copy? y/n
```

ATorrise:
> When I answer y to the following question, an empty diff is opened in the web browser which seems unexpected.

If you haven't made any edits then theres no difference between what you have locally vs what's on the mainframe. the diff is there to show you the current state of your local file compared to what's on the mainframe and knowing your temp file has yet to change is valuable info no?

packages/cli/CHANGELOG.md
ATorrise: Yes but in previous discussions it was suggested not to bring up WSL and the implication that we don't support it

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,354 @@
...
+     * Builds a temp path where local file will be saved. If uss file, file name will be hashed
+     * to prevent any conflicts with file naming. A given filename will always result in the
+     * same unique file path.
+     * @param {ILocalFile} lfFile - object containing pertinent information about the local file during the editing process
+     * @returns {Promise<string>} - returns unique file path for temp file
+     * @memberof EditUtilities
+     */
+    public static async buildTempPath(lfFile: ILocalFile, commandParameters: IHandlerParameters): Promise<string>{
+        // find the appropriate extension for either uss or ds in two (long) operations
+        const ussExt = (lfFile.fileType === 'uss' && lfFile.fileName.includes(".")) ? lfFile.fileName.split(".").pop() : "";
```
adam-wolfe, line 72: The --extension option gets ignored for USS files. I think ussExt should be set based on the --extension option provided by the user. If no --extension is specified, then use the extension of the file from its USS file name (how it works currently).

adam-wolfe (changes requested):

packages/cli/__tests__/help/__integration__/__snapshots__/BrightHelpTests.integration.test.ts.snap
```diff
@@ -25,22 +25,25 @@ exports[`Root level help tests top level help should contain support link 1`] =
...
+   plugins                            Install and manage plug-ins.             
+   profiles                           Create and manage configuration profiles.
+                                      (deprecated)                             
+   provisioning | pv                  Perform z/OSMF provisioning tasks        
+   zos-console | console              Issue z/OS console commands and collect  
+                                      responses                                
+   zos-files | files                  Manage z/OS data sets                    
+   zos-jobs | jobs                    Manage z/OS jobs                         
+   zos-logs | logs                    Interact with z/OS logs                  
+   zos-ssh | ssh | zos-uss | uss | uf Issue z/OS USS commands and receive
```
awharn, line 42: Can.... someone check that this is correct please?

packages/cli/src/zosuss/ZosUss.definition.ts
```diff
@@ -15,7 +15,7 @@ import { SshSession } from "@zowe/zos-uss-for-zowe-sdk";
 
 export const definition: ICommandDefinition = {
     name: "zos-ssh",
-    aliases: ["ssh", "zos-uss", "uss"],
+    aliases: ["ssh", "zos-uss", "uss", "uf"],
```
awharn, line 18: Is this appropriate for the zos-ssh and zos-uss group? `uf` tends to imply a uss file, which this group does not manipulate (except maybe through ssh commands), and is more appropriate for the `zos-files` group.

KevinLoesch1 (commented):
In USSFile.definition.ts, the one example does not look correct. 

        {
            description: strings.ACTIONS.USS_FILE.EXAMPLES.EX2,
            options: `**C:\\Windows\\System32\\Notepad.exe** --editor notepad --extension jcl`
        }

Shouldn't that bolded part be the file that is to be edited? Not the editor path.

gejohnston (approved):
Once the still-open comments from Timothy, Adam, Andrew, and Kevin are explained or changed, I have no additional comments. It should then be ok to merge.

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
```diff
@@ -146,7 +146,7 @@ export class CompareBaseHelper {
                 });
             }
             // reading local file as buffer
-            lfContentBuf = fs.readFileSync(localFileHandle);
+            lfContentBuf = IO.readFileSync(localFile, true);
```
awharn, line 149: Is there a reason that we need to normalize the newlines? I am asking because `fs` is going through the process of locking the file and setting the file handle, but if we use IO.readFileSync, we are then opening the file again, which may have changed on disk.

Alternatively, we can enhance IO in Imperative to accept file handles (numbers) and strings, and move forward with that. But that would require an Imperative change and version bump.

traeok (changes requested):
I tried the scenario where I would edit a USS file, introduce a change on the mainframe, and type in "done" to trigger the conflict resolution process.

image

I viewed the diff, and then entered "n" to prevent the CLI from uploading my local edits. After doing so, the command reprints the message "Downloaded to temporary file: ...",  _but_ the contents of the local file still contain the outdated file content.

In addition, after entering "done" the second time, the local file (with the old contents) will be uploaded and as a result, it will overwrite the USS file on the mainframe. 

It seems like the e-tag is being tracked to reflect the latest mainframe data, but the local file generated from this command has not been updated to contain the new changes from the mainframe:

image

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
ATorrise: I don't remember the reason but this fixed an issue and was recommended by Fernando.

t1m0thyj: The reason might have been to address this issue that Trae mentioned where all lines where shown as changed instead of just one: https://github.com/zowe/zowe-cli/pull/1672#pullrequestreview-1417375164

I don't think we need to use `IO.readFileSync` necessarily - perhaps we could call `IO.processNewlines(fs.readFileSync(localFileHandle))` here?

packages/cli/src/zosfiles/edit/uss/USSFile.definition.ts
```diff
@@ -0,0 +1,59 @@
...
+            description: strings.ACTIONS.USS_FILE.EXAMPLES.EX1,
+            options: `/a/ibmuser/my_text.txt --editor notepad`
+        },
+        {
+            description: strings.ACTIONS.USS_FILE.EXAMPLES.EX1,
+            options: `/a/ibmuser/my_text.txt --editor C:\\Windows\\System32\\Notepad.exe`
+        },
+        {
+            description: strings.ACTIONS.USS_FILE.EXAMPLES.EX2,
+            options: `C:\\Windows\\System32\\Notepad.exe --editor notepad --extension jcl`
```
t1m0thyj, line 56: The first option should be a USS file path instead of the local path to Notepad.exe

packages/cli/src/zosfiles/edit/Edit.utils.ts
adam-wolfe: > > I understand --extension to mainly be a flag for datasets not uss files
> 
> That's fair, but if `--extension` doesn't apply to the `files edit uss` command, then I think we should remove that option from this file:
> 
> https://github.com/zowe/zowe-cli/blob/987abaf2a1718dc16d0fd4494e128b98b8188038/packages/cli/src/zosfiles/edit/uss/USSFile.definition.ts#L43

Sounds good

t1m0thyj (changes requested):
I'm seeing an issue with binary files in the web diff:
* `zowe files edit uss` to edit a USS file tagged as binary
* Make changes to the remote version of the file on the mainframe
* Make conflicting changes to the local version of the file in the editor
* Type "done" to upload the changes, then type "y" when prompted to open a diff
* The right side of the diff is garbled because the binary file is not decoded
image

packages/cli/src/zosfiles/compare/lf-uss/LocalfileUss.handler.ts
t1m0thyj: Changing from `ussFilePath` to `file` makes the new command work, but breaks the `files compare lf-uss` command:
```
zowe files compare lf-uss <localFile> <ussFile>
Command Error:
Expect Error: Specify the USS file name.
```

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
awharn: If we could do that, it would be appreciated. In the mean time, I might open a PR on Imperative to allow numbers to be passed in.

packages/cli/__tests__/zosfiles/__unit__/edit/Edit.utils.unit.test.ts
```diff
@@ -0,0 +1,494 @@
...
+    beforeEach(async () => {
+        jest.resetAllMocks();
+    });
+    describe("buildTempPath()", () => {
+        it("should be able to build the correct temp path with ext argument - uss", async () => {
+            // check that the returned string (string.split('.')[0]) contains only numbers and letters
+            const response = await EditUtilities.buildTempPath(localFileUSS, commandParametersUss);
+            expect(response).toContain(".jcl");
+            if (response.includes('/')){
+                expect(response.split('/').pop()?.split('.')[0]).toMatch(/^[A-Za-z0-9]*$/);
```
awharn, line 83: Would it be beneficial to replace this with `path.sep`?

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,369 @@
...
+        let ext = "."  + (lfFile.fileType === 'uss' ? ussExt : (commandParameters.arguments.extension ?? "txt"));
+        ext = (ext === "." ? "" : ext);
+        if (lfFile.fileType === 'uss'){
+            // Hash in a repeatable way if uss fileName (incase there are special characters in name)
+            const crypto = require("crypto");
+            let hash = crypto.createHash('sha256').update(lfFile.fileName).digest('hex');
+            // shorten hash
+            const hashLen = 10;
+            hash = hash.slice(0, hashLen);
+            return path.join(tmpdir(), lfFile.fileName.split('.')[0] + '_' + hash + ext);
```
t1m0thyj, line 82: `lfFile.fileName.split('.')[0]` will return the entire file path up to the extension (e.g., "/tmp/user/filename"), is that the desired behavior? Or do we want to use `path.parse(lfFile.fileName).name` to return only the base filename without the extension (e.g., "filename")? The name should still be unique because the hash was generated from the full path.

ATorrise: I changed this and during my squashing, i lost changes :( not sure of how much was lost but this was one of the lost changes

traeok (changes requested):
I noticed a couple issues while testing the latest changes:

1. I think we should have a default option for the "Overwrite remote with local edits" prompt - this would maintain consistency with other yes/no prompts in the CLI. Otherwise, the prompt will loop when no input is provided:

no default for prompt

~2. As @adam-wolfe mentioned in the group chat, the diff still seems unclear. Ideally, it would only highlight the changed lines rather than the whole document:~ **This has been resolved with recent changes**

diff unclear

~This seems like it might be related to normalizing line endings between the local and remote files.~

3. When asked to "ignore conflicts and overwrite remote", after entering "y" the edit command exits and the local changes are not uploaded to the mainframe:

answering yes to ignore conflicts does nothing

4. If the edit command found a local temp file, and the user chooses to use this temp file for editing, it informs the user that the "remote was downloaded to temp" even though the temp file has not changed. I think the current behavior is working as intended, but the message in this scenario is misleading as the remote file was not downloaded again:

did not use found temp file

zowe-cli
t1m0thyj, line 1: Can we remove this extra file?

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -0,0 +1,381 @@
...
+*
+*/
+
+import { Download, Upload, IZosFilesResponse, IDownloadOptions, IUploadOptions } from "@zowe/zos-files-for-zowe-sdk";
+import { AbstractSession, IHandlerParameters, ImperativeError, ProcessUtils, GuiResult,
+    TextUtils, IDiffNameOptions, CliUtils } from "@zowe/imperative";
+import { CompareBaseHelper } from "../compare/CompareBaseHelper";
+import { existsSync, unlinkSync } from "fs";
+import { tmpdir } from "os";
+import path = require("path");
```
t1m0thyj, line 18: Please update to use newer import style with `import` instead of `require`: `import * as path from "path"`

traeok (changes requested):
I'm running into a couple unexpected behaviors when testing with this command:
`zowe files edit uss <file_path> --editor code`

- After entering "y" to overwrite remote w/ local edits, the "Conflict: Remote has changed" message appears. Answering "y" to this message does nothing for the first response, but the prompt is triggered once more. Answering "y" this time will then bring up the web diff.
  - _Expected behavior:_ Answering "y" to the first conflict prompt will being up the web diff.
scenario 1 - first answer to view updated file does nothing

- After addressing the web diff, the "Conflict: Ignore conflicts and overwrite remote" message appears. I answered "n" to this message to continue with my local file. However, answering "y" to the new "overwrite remote" prompt results in the command exiting early, with the message "Exiting now. Temp file persists for editing."
  - _Expected behavior:_ Answering "y" to the new overwrite prompt will upload the file and print the success message.
scenario 2 - after answering yes to overwrite, the command exits
  - **More info:** I think this is happening because the `uploadEdits` command is returning true for the "second boolean", which is saved into the `canceled` variable defined in the `EditHandler.processWithSession` function. If `canceled` is true, then the above exit message is printed. This _likely_ also means that `uploaded` is returning `false` from the function, even though it should be uploaded after answering "y" to the overwrite prompt. 

Aside from those two scenarios, everything else is working as expected! If you have any questions or concerns w/ these cases, feel free to reach out - would be happy to assist in any way that I can 😅

zFernand0 (approved):
It seems that all comments have been addressed.
It just needs a few tests fixed, but other than that...
LGTM! :yum:

KevinLoesch1:
I did notice one small thing. If you edit a fixed block dataset and make changes that exceed the record size, the command will fail with the following error:

Command Error:
Command terminated. Issue uploading stash. Temp file will persist

The problem is that the file actually saves on the mainframe side. It just truncates the data. This error might be misleading or confusing if the user is not aware of the cause of the error.

ATorrise:
> Only issue I am seeing is that if there is a conflict that I choose not to ignore, the next time I try to overwrite the remote with the local edits, it will not prompt me to do something about the conflict despite the conflict still existing.
> 
> image

@adam-wolfe This behavior is intended, the wording can be changed for whichever prompt however. The only problem there is keeping it general because the 'overwrite remote' prompt gets used even in situations without a conflict. 

After a conflict is detected, you're given the opportunity to view a diff. Then you are able to overwrite right then and there, despite conflicts, or to continue with your edits. Answering no to 'CONFLICT: Ignore conflicts and overwrite remote?' is interpreted as you needing another opportunity to make your edits since you can escape/terminate the command at any point. The 'overwrite' prompt happens after you've been told the remote is different from your local file so I don't see an issue with not prepending CONFLICT to the overwrite message but I'll find a way to add it if the consensus is that makes the most sense.

This edit command doesn't offer conflict resolution so if you don't want to continue with your outdated local version, the way forward is to terminate the command, start it up again and answer no to the prompt about keeping your temp file/local version. This will pull down a copy of the current remote and get rid of any conflicts, giving you a fresh version to edit.

traeok (approved):
I was able to confirm that the issues I mentioned above were resolved 👍 

I do have a couple suggestions based on today's comments in the internal stand-up: The overwrite/conflict prompts may confuse the user, depending on how they interpret the messages.

Keep in mind that these are merely suggestions and not requests for changes, so feel free to implement or disregard them depending on what you think works best:

- We might want to print an info/warning message if the user chooses to ignore the first remote conflict. Otherwise, users may think that this conflict message will appear again, if new remote changes are made between the time of ignoring the last conflict and before their next upload attempt. Something like "Conflicts ignored; future upload attempts will ignore the latest changes on the mainframe" would suffice. This would also make it clear that the command is not checking for conflicts every time the user tries to upload with this command.
- "CONFLICT: Remote has changed. View updated mainframe file?" might make some users think that the mainframe file will show up in their editor, or that only that file will appear. Should this be changed to be more representative of the action that's happening, e.g. "View diff between local and mainframe files?"

Aside from the comments above, everything LGTM - thanks @ATorrise !

package.json
```diff
@@ -38,8 +38,10 @@
     "prepare": "husky install"
   },
   "dependencies": {
+    "@zowe/cli": "^7.16.6",
     "@zowe/imperative": "5.15.1",
-    "@zowe/perf-timing": "1.0.7"
+    "@zowe/perf-timing": "1.0.7",
+    "cd": "^0.3.3"
```
t1m0thyj, line 44: Is the "cd" package used in this PR? It appears to be over 10 years and old and no longer maintained, so wondering if it should be removed from the "dependencies" list here in package.json

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
```diff
@@ -146,11 +146,11 @@ export class CompareBaseHelper {
                 });
             }
             // reading local file as buffer
-            lfContentBuf = fs.readFileSync(localFileHandle);
+            lfContentStr = fs.readFileSync(localFileHandle, "utf-8");
         } finally {
             if (localFileHandle != null) fs.closeSync(localFileHandle);
         }
-        return lfContentBuf;
+        return Buffer.from(lfContentStr);
```
t1m0thyj, line 153: Can you please restore the original implementation of this function?
```
let lfContentStr: Buffer;
lfContentBuf = fs.readFileSync(localFileHandle);
return lfContentBuf;
```

It should behave the same but more efficiently. Without an encoding specified, the `readFileSync` method returns a buffer, so I don't think its necessary to load the file as a string and then convert it to a buffer.

zowe-cli
t1m0thyj, line 1: Please remove this submodule

t1m0thyj (changes requested):

package.json
```diff
@@ -38,6 +38,7 @@
     "prepare": "husky install"
   },
   "dependencies": {
+    "@zowe/cli": "^7.16.6",
```
t1m0thyj, line 41: Please remove, sorry that I missed this before. We don't want the CLI to be a dependency of itself 😋

packages/cli/__tests__/zosfiles/__unit__/compare/CompareBaseHelper.unit.test.ts
```diff
@@ -63,15 +63,15 @@ describe("Comapare Base Helper", () => {
...
             jest.spyOn(fs, "openSync").mockReturnValue(0);
             jest.spyOn(fs, "fstatSync").mockReturnValue({isFile: () => true} as any);
             jest.spyOn(fs, "readFileSync").mockReturnValue("test");
             jest.spyOn(fs, "closeSync").mockImplementation();
 
             const response = helper.prepareLocalFile("/absolute/path/to/real/file");
 
-            expect(response).toEqual("test");
+            expect(response).toEqual(Buffer.from("test"));
         });
```
t1m0thyj, line 75: I think the changes to this test can be reverted to use a string rather than a buffer, since you restored the implementation of the `CompareBaseHelper` method to the previous implementation 🙂

t1m0thyj (dismissed):

npm-shrinkwrap.json
t1m0thyj, line 1: The shrinkwrap is outdated since it still includes `@zowe/cli` and `cd` as dependencies.

Could you please run `npm install` locally to update it? If you don't want to overwrite your local node_modules (for example your linked copy of Imperative), you can run `npm install --package-lock-only` to only rebuild this lockfile and do nothing else.

packages/cli/__tests__/zosfiles/__unit__/edit/ds/Dataset.definition.unit.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+describe("DatasetDefinition", () => {
+    it("should use the correct object for strings", () => {
+        const strings = i18nEnglish.EDIT.ACTIONS.DATA_SET;
+
+        expect(DatasetDefinition.description).toBe(strings.DESCRIPTION);
+        expect(DatasetDefinition.children).toBeUndefined();
+        expect(DatasetDefinition.options).toMatchSnapshot();
+        expect(DatasetDefinition.examples).toMatchSnapshot();
+        expect(DatasetDefinition.description).toBe(strings.DESCRIPTION);
```
t1m0thyj, line 23: This line is a duplicate of line 19

packages/cli/__tests__/zosfiles/__unit__/edit/uss/UssFile.definition.unit.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+describe("DatasetDefinition", () => {
+    it("should use the correct object for strings", () => {
+        const strings = i18nEnglish.EDIT.ACTIONS.USS_FILE;
+
+        expect(USSFileDefinition.description).toBe(strings.DESCRIPTION);
+        expect(USSFileDefinition.children).toBeUndefined();
+        expect(USSFileDefinition.options).toMatchSnapshot();
+        expect(USSFileDefinition.examples).toMatchSnapshot();
+        expect(USSFileDefinition.description).toBe(strings.DESCRIPTION);
```
t1m0thyj, line 23: This line is a duplicate of line 19 (same comment as above)

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+- Enhancement: Created zos-files edit commands to edit a dataset or uss file locally [PR #1672](https://github.com/zowe/zowe-cli/pull/1672)
```
t1m0thyj, line 6: For the z/OS Files SDK changelog, can you replace this with the items I had listed in #1729?

> - Enhancement: Added streaming capabilities to the `Download.dataSet` and `Download.ussFile` methods.
> - BugFix: Fixed `Get.USSFile` API not respecting USS file tags.

We want to mention the new edit commands in the CLI changelog, but they are not relevant for the SDK changelog - only changes that were made to the APIs that are inside the SDK.

t1m0thyj (dismissed):
Left hopefully my final set of comments 🙂 

Tested local/remote conflict handling with the commands locally, LGTM 👍

packages/zosfiles/CHANGELOG.md
ATorrise: oops! thanks for catching this

gejohnston (approved):

adam-wolfe (approved):
LGTM

t1m0thyj (approved):
LGTM, thanks @ATorrise!

PR #1685, opened 2023-04-19
Merger in copy-cross-lpar enhancements

KevinLoesch1:


packages/cli/src/zosfiles/copy/dsclp/Dsclp.handler.ts
```diff
@@ -0,0 +1,65 @@
...
+        );
+    }
+
+    /**
+     * Private function to prompt user if they wish to overwrite an existing dataset.
+     */
+    private promptForOverwrite(console: IHandlerResponseConsoleApi) {
+        return async (targetDSN: string) => {
+            const answer: string = await console.prompt(
+                `The dataset '${targetDSN}' already exists on the target system. Do you wish to overwrite it? [y/N]: `);
```
gejohnston, line 61: Minor nit: Years ago, Tech Writing & QA had a guideline that we use the word "want" instead of "wish". Not a request for change, just something to consider.

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -84,4 +92,239 @@ export class Copy {
...
+            const task2: ITaskWithStatus = {
+                percentComplete: 0,
+                statusMessage: "Retrieving data set contents",
+                stageName: TaskStage.IN_PROGRESS
+            };
+
+            /**
+             *  Get the dataset and store it in a buffer
+             *
+             *  We want to download and upload in binary, as this command is copying from z/OS to z/OS system.
```
gejohnston, line 181: We had team conversations about tagging (and loss of tagging) if a binary transfer is specified. Does that concern apply to datasets? If tagging is applicable, would it mean that the target would not have the same tag?

gejohnston (approved):
I posed a few questions that will not necessarily result in any request for change.

packages/cli/src/zosfiles/copy/Copy.options.ts
```diff
@@ -0,0 +1,23 @@
+/*
```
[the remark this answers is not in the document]
KevinLoesch1: This was a early change that should have been removed. I have deleted it from the project.

packages/cli/src/zosfiles/copy/dsclp/Dsclp.handler.ts
KevinLoesch1: This was an early change that is no longer needed. I have removed it from the project.

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -84,4 +92,239 @@ export class Copy {
...
+            const targetMember  = toMemberName;
+            let targetDataSetObj: IZosmfListResponse;
+            let targetFound: boolean = false;
+            let targetMemberFound: boolean = false;
+            let overwriteTarget: boolean = options.replace;
+
+            /**
+             * If no console is passed and no overwrite is passed then set to owerwrite to false
+             */
+            if(console == undefined && overwriteTarget == undefined){
```
t1m0thyj, line 138: I believe the `console == undefined` check is left behind from when this method accepted a `console` parameter.

We should remove it now - it will always evaluate to `false` because the JS console API is always defined.

packages/zosfiles/src/methods/copy/Copy.ts
t1m0thyj: I believe tagging to specify encoding only applies to USS files.

t1m0thyj (changes requested):

packages/zosfiles/src/constants/ZosFiles.messages.ts
```diff
@@ -632,6 +632,46 @@ export const ZosFilesMessages: { [key: string]: IMessageDefinition } = {
...
+    datasetCopiedAborted: {
+        message: "Data set copied aborted. The existing target dataset was not overwritten."
+    },
+
+    /**
+     * Message indicating that the data set member was copied was aborted
+     * @type {IMessageDefinition}
+     */
+    datasetMemberCopiedAborted: {
+        message: "Data set copied aborted. The existing target dataset member was not overwritten."
```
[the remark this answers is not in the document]
KevinLoesch1: I've made the change. It is now spelled consistently.

packages/cli/__tests__/zosfiles/__system__/copy/dsclp/__snapshots__/cli.files.copy.dsclp.system.test.ts.snap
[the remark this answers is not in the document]
KevinLoesch1: I've made the suggested changes.

packages/zosfiles/src/methods/copy/Copy.ts
KevinLoesch1: This check has been removed.

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -84,4 +92,232 @@ export class Copy {
...
+                            overwriteTarget = await options.promptFn(targetDataset +"(" + targetMember + ")");
+                        }
+                    }
+                }
+            }
+
+            /*
+            *  Don't overwrite an existing dataset or member if overwrite is false
+            */
+            if(overwriteTarget == undefined || overwriteTarget == true || targetFound == false){
```
t1m0thyj, line 264: If the default behavior of the API is to _not_ replace files, should we remove this condition? `overwriteTarget == undefined`

t1m0thyj (commented):
Thanks for addressing earlier feedback, left one more question.

Will approve once tests are passing and merge conflicts are resolved 🙂

packages/zosfiles/src/methods/copy/Copy.ts
KevinLoesch1: I removed the check, as it is no longer relevant.

t1m0thyj (approved):
LGTM, thanks @KevinLoesch1!

PR #1706, opened 2023-04-28
Improve error message when logging in with token and without base path

t1m0thyj:
Fixes confusing error message "Token is not valid or expired" when APIML token is used to connect direct-to-service with `ZosmfRestClient` (https://github.com/zowe/imperative/issues/978)

| Zowe CLI | _List data sets with "zosmf-direct" profile:_<br>`zowe files ls ds SYS1.PARMLIB --zosmf-p zosmf-direct` |
| --- | --- |
| _Before_ | image |
| _After_ | image |


gejohnston (approved):
Looks good.

awharn (approved):

PR #1710, opened 2023-05-03
feat(daemon): Remove `named_pipe` in favor of `tokio`

traeok:
The `named_pipe` library is no longer maintained, and the recommendation from the library's author is to use `tokio`'s implementation of named pipes instead.

This PR replaces the `named_pipe` library with `tokio`, and also uses `tokio::net::UnixStream` instead of the standard library's `UnixStream` so that logic is consistent across all platforms for the communication functions.

It also updates the daemon logic so that the Rust client looks for the Node.js server process before trying to connect to the socket/pipe. This prevents race conditions where the server process was started by the daemon, but the Rust client tried to attach to the pipe before it was initialized by the server.


zowex/Cargo.toml
```diff
@@ -19,9 +19,13 @@ serde_json = "1.0.94"
 simple-error = "0.2.3"
 supports-color = "2.0.0"
 sysinfo = "0.28.2"
+tokio = { version = "1", features = ["io-util", "macros", "net", "rt-multi-thread"] }
```
awharn, line 22: Can you specify the full version so we can lock it down?

awharn (commented):
Just one question

zowex/Cargo.toml
traeok: yep! fixed in incoming commit

awharn (approved):

zowex/src/comm.rs
```diff
@@ -71,12 +70,6 @@ pub fn comm_establish_connection(
     let mut cmd_to_show: String = String::new();
 
     let stream = loop {
-        let conn_result = DaemonClient::connect(daemon_socket);
-        if let Ok(good_stream) = conn_result {
-            // We made our connection. Break with the actual stream value
-            break good_stream;
-        }
```
t1m0thyj, line 78: I think it was intentional to have this block of code first for maximum performance. If the daemon server is already running which is likely to be true for the majority of requests, then we can save time by skipping the check to determine if the process is running.

Have we benchmarked the daemon before and after to compare performance? If the difference is negligible, then I have no objection to switching to LBYL (Look Before You Leap) instead of EAFP (Easier to Ask for Forgiveness than Permission).

zowex/src/comm.rs
```diff
@@ -155,36 +170,44 @@ pub fn comm_establish_connection(
...
+    loop {
+        match stream.try_write(message) {
+            Ok(_n) => break,
+            // If the write operation would need to block, try again as we do not want to discard any messages
+            Err(e) if e.kind() == io::ErrorKind::WouldBlock => continue,
+            Err(e) => {
+                return Err(e.into());
+            }
+        }
+    }
```
t1m0thyj, line 189: Instead of the loop, could we use `write_all` here from Tokio io-util?
* https://docs.rs/tokio/latest/tokio/net/struct.UnixStream.html#method.write_all
* https://docs.rs/tokio/latest/tokio/net/windows/named_pipe/struct.NamedPipeClient.html#method.write_all

zowex/Cargo.toml
t1m0thyj, line 1: With the added dependencies, the daemon binaries are significantly larger. Just an observation - I think the benefits of this PR outweigh the extra size, and we've already limited the size as much as possible by limiting the Tokio features we pull in. 🙂 

image
_^ Zowe 2.8.0 bundle_

image
_^ Artifacts from this PR_

t1m0thyj (commented):
Thanks for migrating to `tokio` for technical currency 👍 Left a few questions 🙂

zowex/Cargo.toml
t1m0thyj: Sorry, feel free to ignore this comment - I realized this is not a valid comparison as release builds (in the Zowe bundle) are much smaller than debug builds (in the PR artifacts).

zowex/src/comm.rs
```diff
@@ -261,23 +284,21 @@ pub fn comm_talk(message: &[u8], stream: &mut DaemonClient) -> io::Result<i32> {
...
 
     // Terminate connection. Ignore NotConnected errors returned on macOS.
     // https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.shutdown
     #[cfg(target_family = "unix")]
-    match stream.shutdown(Shutdown::Read) {
+    match reader.shutdown().await {
         Err(ref e) if e.kind() == io::ErrorKind::NotConnected => (),
         result => result?,
     }
-    #[cfg(target_family = "unix")]
```
gejohnston, line 275: As I recall, this step was added to avoid some weird error about connection terminated on some platform. If we have confirmed that the extra shutdown is no longer needed anywhere, then all is good.

gejohnston (approved):
I agree with Timothy's comment that it will be useful to confirm whether there is any performance impact related to connecting first or checking for daemon first.

I also had one question about confirming the behavior during shutdown of the communication stream.

Perhaps neither question will require any change. Everything else looks good to me.

zowex/src/comm.rs
traeok: yep! fixed in incoming commit

KevinLoesch1 (approved):

zowex/src/comm.rs
traeok: I updated the import to use `tokio::io::BufReader` for the read stage, so I can still access the functions from the stream object. I was going to keep this call, but since `stream` was moved when initializing a `BufReader`, I no longer could use it. `reader.shutdown` will end up calling `stream.shutdown`, so that functionality still remains 🙂

zowex/src/comm.rs
traeok: Not going to benchmark, as it's clear that EAFP will be faster than LBYL, especially since it loops over system processes to find the daemon server. Addressed in upcoming commit 👍

t1m0thyj (approved):

PR #1714, opened 2023-05-04
Update workflows v1

awharn:
Updates zowe-v1-lts workflows to supported versions and remove Node 14 support

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -14,12 +14,12 @@ jobs:
 
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
-    - name: Use Node.js 14.x
+    - name: Use Node.js 16.x
       uses: actions/setup-node@v3
       with:
-        node-version: 14.x
+        node-version: 16.x
 
     - name: Check Vulnerabilities
       run: npm audit --production --audit-level=moderate
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -16,10 +16,10 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v3
 
-    - name: Use Node.js 16.x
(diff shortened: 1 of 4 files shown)


.github/workflows/audit.yml
zFernand0, line 1: |If it's ok, can we add the `zowe-v?-lts` branch to the following workflows:
- audit
- changelog
- linter (just needs to change the `v1` with `v?` 😋 

Here is an example
```yaml
on:
  pull_request:
    branches:
      - master
      - zowe-v?-lts
```

zFernand0 (changes requested):
Changes LGTM!
Just left one small comment 😋

t1m0thyj (approved):

.github/workflows/audit.yml
```diff
@@ -14,12 +14,12 @@ jobs:
...
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
-    - name: Use Node.js 14.x
+    - name: Use Node.js 16.x
       uses: actions/setup-node@v3
       with:
-        node-version: 14.x
```
zFernand0, line 22: should we use `lts/*`?

.github/workflows/audit.yml
zFernand0: this same comment may apply to the master workflows 😋

zFernand0 (changes requested):

zFernand0 (changes requested):
should we add the `zowe-v?-lts` to the `audit` and `changelog` workflows?

awharn:
> should we add the `zowe-v?-lts` to the `audit` and `changelog` workflows?

Wouldn't we want to make that change on the master branch instead of this one? That is the branch's workflows that would apply to zowe-v2-lts, zowe-v3-lts, etc.

zFernand0 (approved):

Issue #1439, opened 2022-06-08
Invalid block size is used by `zowe files create data-set-sequential`


plavjanik:
When I allocate a new dataset without specifying a block size is uses a default that is not a valid size and the allocated dataset is not usable.

```
zowe files create data-set-sequential DSN --recfm VB --lrecl 32756
Data set created successfully.
```

When I use the dataset in a JCL, it fails:
```
11.05.34 JOB98529  IEC141I 013-20,IGG0191A,PLAPECHW,,ZLDATA,8F26,WRKD22,  441
```

The explanation of the code is:
```
    S013-20
        An OPEN macro instruction was issued for a sequential data set using
        queued access. BLKSIZE is not a multiple of LRECL, or LRECL is 0 with
        RECFM=FB or RECFM=FBS, or BLKSIZE is not at least four bytes greater
        than LRECL with RECFM=V.
```

It is allocated as:
<img width="191" alt="image" src="https://user-images.githubusercontent.com/7644758/172581357-81a0cc1c-a9d2-4777-9bd6-3ee875b82503.png">

I would expect to do what ISPF allocates in this case (block size is 32760 which is record length + 4 because it is VB) :
<img width="508" alt="image" src="https://user-images.githubusercontent.com/7644758/172579412-ef903487-59eb-4ce9-aa42-ad9abdc3ee04.png">

KevinLoesch1:
Fixed in the v2.10 release.

PR #1727, opened 2023-06-05
Check for invalid block size

KevinLoesch1:
**What It Does**
This is a fix for issue https://github.com/zowe/zowe-cli/issues/1439. It checks the block size specified by the user, or the default, against the logical record length. If the block size is too small then it is changed to a valid size. 

**How to Test**
Attempt create a dataset with a lrec size greater than 6160 and no block size specific, or specify a lrec that is larger than the specified block size. If the dataset format is a variable block, it will add an extra four bytes on to the calculated block size. 

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/packages/zosfiles/__tests__/__unit__/methods/create/Create.unit.test.ts
+++ b/packages/zosfiles/__tests__/__unit__/methods/create/Create.unit.test.ts
@@ -251,6 +251,38 @@ describe("Create data set", () => {
             );
         });
 
+        it("should be able to create a sequential data set using the a block size that is too small", async () => {
+            const custOptions = {
+                dsorg: "PS",
+                alcunit: "CYL",
+                primary: 20,
+                secondary: 1,
+                recfm: "VB",
+                blksize: 100,
+                lrecl: 1000
+            };
+            const response = await Create.dataSet(dummySession, CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL, dataSetName, custOptions);
+
+            expect(response.success).toBe(true);
+            expect(response.commandResponse).toContain("created successfully");
+            expect(mySpy).toHaveBeenCalledWith(
+                dummySession,
+                endpoint,
+                [ZosmfHeaders.ACCEPT_ENCODING],
+                JSON.stringify({
+                    ...{
+                        alcunit: "CYL",
+                        dsorg: "PS",
+                        primary: 20,
+                        recfm: "VB",
+                        blksize: 1004,
+                        lrecl: 1000,
+                        secondary: 1
+                    }
+                })
+            );
+        });
+
         it("should be able to create a sequential data set using the primary allocation and default the secondary allocation", async () => {
             const custOptions = {
                 dsorg: "PS",
--- a/packages/zosfiles/__tests__/__unit__/methods/create/Create.unit.test.ts
+++ b/packages/zosfiles/__tests__/__unit__/methods/create/Create.unit.test.ts
@@ -251,7 +251,7 @@ describe("Create data set", () => {
             );
         });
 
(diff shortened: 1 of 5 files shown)


traeok (changes requested):
I had to make a change locally to get the fix to work, but once I did that, it works as expected 👍

packages/zosfiles/src/methods/create/Create.ts
```diff
@@ -197,8 +197,8 @@ export class Create {
                         break;
 
                     case "blksize":
-                    /* 
-                    *  This is a fix for issue https://github.com/zowe/zowe-cli/issues/1439. 
+                    /*
+                    *  This is a fix for issue https://github.com/zowe/zowe-cli/issues/1439.
                     *
                     */
                         if (isNullOrUndefined(tempOptions.blksize)) {
```
ATorrise, line 204: `util.isNullOrUndefined` pops up the following deprecated warning in my editor: 
```@deprecated — since v4.0.0 - use value === null || value === undefined instead```
I'm not sure if you'll want to fix this though since there are 17 instances of its use in this file

**Review by traeok — COMMENTED** [review]
Sorry for the confusion from my previous review - after pulling down the branch again and rebuilding, it works as expected. Not sure what happened there. 

Will approve once @ATorrise's comment is resolved 👍

File: packages/zosfiles/__tests__/__unit__/methods/create/Create.unit.test.ts [hunk 2]
```diff
@@ -251,6 +251,38 @@ describe("Create data set", () => {
             );
         });
 
+        it("should be able to create a sequential data set using the a block size that is too small", async () => {
```
awharn, line 254: Would it be good to add more of these tests to ensure that the other RECFM options specified in the case statement also behave the same? It says the only statement that isn't covered is the RECFM being "V", which is a bit odd.

t1m0thyj (commented):
Thanks @KevinLoesch1 for the fix! Tested this PR and have a question about one of the scenarios below:
| Sample Code | Old Behavior | New Behavior | Comments |
| --- | --- | --- | --- |
| `Create.dataSet(session, CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL, "<dsname>", { recfm: "VB", lrecl: 6160 });` | z/OSMF throws error: `For a V file, the LRECL must be greater than 4 bytes and must be at least 4 bytes smaller than the BLKSIZE.` | We use 6164 as default `blksize` value | 👍 |
| `Create.dataSet(session, CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL, "<dsname>", { recfm: "VB", lrecl: 32756 });` | z/OSMF does not throw an error?! | We use 32760 as default `blksize` value | 👍 |
| `Create.dataSet(session, CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL, "<dsname>", { recfm: "VB", lrecl: 32756, blksize: 6160 });` | z/OSMF does not throw an error?! | We override the `blksize` value and use 32760 | _Should we keep whatever `blksize` is specified by the user, even if it is invalid, and allow z/OSMF to handle any errors?_ |

packages/zosfiles/src/methods/create/Create.ts
KevinLoesch1: Updated with the latest update.

KevinLoesch1:
@t1m0thyj The logic behind this part of the change was to make the behavior consistent. If someone specifies a block size that is too small for the lrecl on a fixed block dataset, we automatically adjust it to the correct minimum size. We're doing the same thing for variable records. The only difference is what constitutes a valid block size.

gejohnston (approved):
Looks good.

awharn (approved):

PR #1738, opened 2023-06-28
Reformats messages that are displayed for REST request errors

gejohnston:
**What It Does**

The intent is to highlight the information that is relevant to the end user, and to appropriately categorize the additional details that are displayed. The enhancement was motivated by issue [#935](https://github.com/zowe/zowe-cli/issues/935).

The user enables this new behavior by setting the environment variable named ZOWE_V3_ERR_FORMAT to "true".

Much of this feature comes from an updated version of imperative, which populates data structures and displays the messages in a different format. This CLI PR further populates data structures for errors occurring in our ZosmfRestClient so that the appropriate information is displayed in the right category. It also populates data structures for errors occurring in the GetJobs API.

**How to Test**

One command that generates an error specific to this CLI PR is `zowe zos-jobs list spool-files-by-jobid JOB00000`.

Running that command will display the following results.

image

Set ZOWE_V3_ERR_FORMAT to "true". Running the same command again will display the following results.

image

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

fixes https://github.com/zowe/zowe-cli/issues/935


packages/core/__tests__/rest/__unit__/ZosmfRestClient.unit.test.ts
```diff
@@ -54,6 +62,72 @@ describe("ZosmfRestClient tests", () => {
...
+            (zosmfRestClient as any).mResponse = {
+                statusCode: RestConstants.HTTP_STATUS_401
+            };
+            const error: IImperativeError = {
+                msg: "Rest API failure with HTTP(S) status 401",
+                causeErrors: "",
+                additionalDetails: 'Received HTTP(S) error 401 = Unauthorized.\n\n' +
+                    'Host:              some.company.com\n' +
+                    'Port:              1234\n' +
+                    'Allow Unauth Cert: true' +
```
t1m0thyj, line 82: Do we think that users may find "Allow Unauth Cert" confusing, because it is the opposite of the `--reject-unauthorized` option?

packages/core/src/rest/ZosmfRestClient.ts
```diff
@@ -51,53 +54,103 @@ export class ZosmfRestClient extends RestClient {
...
-            // if we didn't get an error trying to parse json, check if there is a stack
-            // on the JSON error and delete it
-            if (json.stack != null) {
-                this.log.error("An error was encountered in z/OSMF with a stack." +
-                    " Here is the full error before deleting the stack:\n%s", JSON.stringify(json));
-                this.log.error("The stack has been deleted from the error before displaying the error to the user");
-                delete json.stack; // remove the stack field
-            }
+            // don't try to parse an empty string
+            if (original.causeErrors !== "") {
```
t1m0thyj, line 66: Should we also handle the case of `original.causeErrors === undefined`, since "causeErrors" appears to be an optional property of IImperativeError?

gejohnston: This logic existed in the original implementation. With all of the complication of the dual-path logic, I made an effort to NOT change any existing logic. I can see the concern. I will add guards for what could be existing errors.

packages/core/__tests__/rest/__unit__/ZosmfRestClient.unit.test.ts
gejohnston: While the "Allow Unauth Cert" text exists in zowe-cli/.../GetJobs module, the same text is produced by the master version of imperative/.../AbstractRestClient. If we want to to alter this text, we will need to open a Git issue against imperative and modify the imperative repo first.

t1m0thyj: I don't feel strongly enough about this to warrant a separate PR on Imperative for it.

t1m0thyj (approved):
LGTM! We may want to hold off on merging since publishing the CLI is currently failing until this fix is applied to Zowe Artifactory: https://jfrog.atlassian.net/browse/RTFACT-29785

zFernand0 (commented):
LGTM! 😋 
I believe the Zowe Artifactory issue was resolved
Ok to merge @t1m0thyj ?

zFernand0 (approved):
LGTM! 😋 
I believe the Zowe Artifactory issue was resolved
Ok to merge @t1m0thyj ?

PR #1753, opened 2023-07-20
Fix error when listing ds members that include '"'

t1m0thyj:
**What It Does**
Fixes a bug in the implementation of https://github.com/zowe/zowe-cli/pull/1730 - error when listing data set members that include double quote in the name.

**How to Test**
There isn't really a way to test the changes manually. The error is only reproducible in the unit test ["should list members from given data set that contains a member with an invalid name"](https://github.com/zowe/zowe-cli/blob/master/packages/zosfiles/__tests__/__unit__/methods/list/List.unit.test.ts#L169) which generates random member names to test with. The member name that was previously failing the test is `XK"¶J9	^`, and I've verified that it now passes with this member name.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


traeok (approved):
LGTM, thanks @t1m0thyj

ATorrise (approved):

KevinLoesch1 (approved):

PR #1759, opened 2023-07-25
feat(pkg): Add secrets SDK and `keyring` module

traeok:
**What It Does**

This adds [`keytar-rs`](https://github.com/traeok/keytar-rs) into the CLI as a new package: `@zowe/secrets-for-zowe-sdk`. It also removes references to the old `node-keytar` library as it is no longer used.

- [x] Added the secrets SDK as a new package to the `zowe-cli` monorepo
- [x] Added `keyring` module to said secrets SDK
- [x] Added test and publish stages for secrets SDK to prepare for publishing

**How to Test**

**Note:** The test stages will fail - 2 integration tests are failing on the daemon because Imperative is not using the Secrets SDK yet. Once the Imperative branch is merged into the codebase, CLI will pass with the new version.

Reviewers should be able to clone the repo, `cd` into `packages/secrets`, and run `npm install` to build from source. In addition, reviewers should be able to run `npm run test` in this folder to verify that all tests are passing for the `keyring` module.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

Sayonara, `node-keytar` :wave:


.github/workflows/secrets-sdk.yml
```diff
@@ -0,0 +1,275 @@
...
+    paths:
+      - "packages/secrets/**"
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+jobs:
+  build:
+    defaults:
+      run:
+        working-directory: packages/secrets
```
t1m0thyj, line 21: This `defaults` section could be moved up above `jobs`, so that we only have to define it once for all jobs, rather than twice inside each job (`build` and `test`).

.github/workflows/secrets-sdk.yml
```diff
@@ -0,0 +1,275 @@
...
+      - name: Setup and run tests
+        uses: addnab/docker-run-action@v3
+        if: ${{ matrix.settings.host == 'ubuntu-latest' && !endsWith(matrix.settings.target, 'musl') }}
+        with:
+          image: ${{ format('node:{0}-slim', matrix.node) }}
+          options: "-v ${{ github.workspace }}:/build -w /build --cap-add=IPC_LOCK ${{ matrix.settings.platform && format('--platform={0}', matrix.settings.platform) }}"
+          run: |
+            set -e
+            apt update -y && apt install -y gnome-keyring
+            dbus-run-session -- bash packages/secrets/scripts/linux-test.sh
```
t1m0thyj, line 265: For the purpose of being consistent with other scripts like "prebuildify.sh", and to make it easier for users to run the script locally, would it make sense to remove the `cd` command in "linux-test.sh"?

Then this command could be replaced with:
`cd packages/secrets && dbus-run-session -- bash scripts/linux-test.sh`

.github/workflows/secrets-sdk.yml
```diff
@@ -0,0 +1,275 @@
...
+      - name: Setup and run tests (MUSL)
+        uses: addnab/docker-run-action@v3
+        if: ${{ matrix.settings.host == 'ubuntu-latest' && endsWith(matrix.settings.target, 'musl') }}
+        with:
+          image: ${{ format('node:{0}-alpine', matrix.node) }}
+          options: "-v ${{ github.workspace }}:/build -w /build --cap-add=IPC_LOCK ${{ matrix.settings.platform && format('--platform={0}', matrix.settings.platform) }}"
+          run: |
+            set -e
+            apk add dbus gnome-keyring libsecret
+            dbus-run-session -- sh packages/secrets/scripts/linux-test.sh
```
t1m0thyj, line 275: Same comment as above - this command could be replaced with:
`cd packages/secrets && dbus-run-session -- sh scripts/linux-test.sh`

packages/secrets/scripts/prebuildify.sh
t1m0thyj, line 1: I think there is a mismatch between where this script downloads prebuilds, and where "index.js" loads them from. Either this script needs to be updated to download prebuilds into `src/keyring/prebuilds`, or "index.js" needs to be updated to look 2 levels up in `path.join(__dirname, "..", "..", "prebuilds")`.

Not sure which fix is better? I guess I'd like having prebuilds at the top level because they don't really seem to belong in the "src" folder.

t1m0thyj:
Is it right that we expect 2 CLI tests to be failing for now, until we're able to update Imperative to use the secrets SDK?

traeok:
> @traeok Is it right that we expect 2 CLI tests to be failing for now, until we can update Imperative to use the secrets SDK?

Yes - once imperative has the new secrets SDK, those integration tests should pass.

packages/secrets/src/keyring/.npmignore
t1m0thyj, line 1: Should we delete this file? I don't think it will have any effect here since it's not at the package root level.
And I don't think we need an ".npmignore" since we've added a "files" property in package.json to filter the files.

packages/secrets/src/keyring/README.md
```diff
@@ -0,0 +1,169 @@
+# keyring
+
+keyring is a native Node.js module for accessing and managing OS credential storage.
+
+## Compatibility
```
t1m0thyj, line 5: Should we update the compatibility matrix to be current?
We have added an x86 build for Linux.
Also, we don't yet have an s390x build.

packages/secrets/src/keyring/src/build.rs
t1m0thyj, line 1: This seems to be a duplicate of the "build.rs" file one level up - do we need both?

ATorrise (commented):
image


Hi - compiles! when building on windows, I get some unrelated output though

ATorrise (approved):
Those warnings went away, looks good :)

awharn (approved):

t1m0thyj (changes requested):
Will approve once the UTF-8 encoding issue on Windows has been addressed 😋

packages/secrets/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@napi-rs/cli": "^2.16.2",
+    "@types/node": "^20.4.5",
+    "ava": "^4.3.3",
+    "typescript": "^4.9.5"
+  },
+  "ava": {
+    "timeout": "3m"
+  },
+  "engines": {
+    "node": ">= 10"
```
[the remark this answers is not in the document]
awharn: I think we should

packages/secrets/src/keyring/README.md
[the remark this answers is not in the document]
traeok: At first glance, I'm not sure if `armv7l` is referring to `armv7-unknown-linux-gnueabi` or `armv7-unknown-linux-gnueabihf`. My guess is it's referring to the former, and if that's the case, I can update our existing armv7 target - or I can add support for this one as well 🙂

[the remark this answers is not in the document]
t1m0thyj: I believe `armv7l` is the same as `arm-linux-gnueabihf` - see https://github.com/atom/node-keytar/pull/446/files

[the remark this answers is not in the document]
traeok: Added to compat. matrix in upcoming commit 👍

t1m0thyj (approved):

PR #1765, opened 2023-07-27
Implement uppercase windows user fix

awharn:
**What It Does**
Fixes the error with the Daemon on Windows that results in the user being unable to connect to the daemon server if their username contains a capital letter.

**How to Test**
Unfortunately, there is no way to easily test this, considering any integration tests may run in the daemon. The change itself is relatively simple, so perhaps additional testing may not be required. However, this is the only location in the code that will write to the daemon file with a username with any capital letters in Windows.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


traeok (approved):
LGTM, thanks @awharn 😃

packages/cli/src/daemon/DaemonDecider.ts
t1m0thyj, line 1: On line 205 in `initialParse` there is another reference to `os.userInfo().username` which is not lower-cased.

We can't easily replace this other occurrence with `this.mUser` like we did for the first one because it's not defined yet.

I think its ok to leave as-is because it's only used to build the pipe path which should be case insensitive on Windows?

awharn: It might be okay to leave as is, but I do not like relying entirely on the OS's case insensitive behavior for this purpose. I think I am going to modify that to use `this.mUser` as well.

traeok: Since Windows paths are case insensitive I think we can leave it as-is - or, we could move the assignment of `this.mUser` into the `initialParse` function :yum:

t1m0thyj (approved):
LGTM, thanks @awharn!

gejohnston (approved):
In the file zowe-cli\packages\cli\__tests__\daemon\__integration__\enable\cli.daemon.enable.integration.ts

there is a function named willRunNodeJsZowe(). It is a bit of a hack, but that function is used within that file to only run a test when we are running our tests with the NodeJS version of our zowe command. I think you could easily reverse the logic to create a function named willRunDaemon() that might enable you to write a daemon-only test of your fix. 

This test idea does not not seem essential. I just thought that I would mention the idea.

zFernand0 (approved):
Tested with the one command that I knew had to go through this logic. `zowe daemon restart` 😋 

Without the fix:
```
C:\Users\WDAGUtilityAccount>zowe daemon restart
User name of 'WDAGUtilityAccount' in file 'C:\Users\WDAGUtilityAccount\.zowe\daemon\
daemon_pid.json' does not match current user = 'wdagutilityaccount'.
```

With the fix:
```
C:\Users\WDAGUtilityAccount>
C:\Users\WDAGUtilityAccount>zowe daemon restart
Starting a background process to increase performance ...
A new daemon has started.

C:\Users\WDAGUtilityAccount>zowe daemon restart
Starting a background process to increase performance ...
A new daemon has started.
```

ATorrise (approved):

PR #1766, opened 2023-07-27
Add Apple Silicon support for Zowe daemon

### LINKED ISSUE #1764: Support Zowe daemon for Apple Silicon
The MacOS daemon binary that we currently publish is only supported on x86 Mac and not on ARM Mac.
If the `zowe daemon enable` command is run on ARM Mac it appears to succeed, but the prebuilt binary won't run.
Ideally we could create a universal build that supports both architectures: https://github.com/rust-lang/cargo/issues/8875

t1m0thyj:
**What It Does**
Resolves #1764

**How to Test**
* Download the MacOS artifact from this PR's workflow run: https://github.com/zowe/zowe-cli/actions/runs/5684509878
* Extract the TGZ from the downloaded ZIP and rename it to "zowe-macos.tgz"
* Navigate to the directory where Zowe CLI is installed and replace the TGZ in the "prebuilds" folder
* Run `zowe daemon enable` followed by a CLI command to verify that the new daemon works

**Review Checklist**
I certify that I have:
- [x] tested my changes
  - [x] Test on ARM Mac - tested on my local machine
  - [x] Test on x86_64 Mac - tested in [this workflow run](https://github.com/zowe/zowe-cli/actions/runs/5717741634)
- [x] updated the changelog

**Additional Comments**
The universal binary is about twice as large as the old one, but I think the benefits make it worthwhile.


zowex/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "zowe"
-version = "1.1.0"
```
t1m0thyj, line 3: Should we bump to version 1.1.1 or 1.2.0? I'm kind of on the fence - it is an enhancement but only for Mac. Not sure whether a minor bump is warranted since nothing has changed for Windows and Linux users?

zFernand0: I think it makes sense to update the version to 1.2.0 😋

zFernand0 (approved):
I don't have a Mac at hand, but I appreciate you taking the time IRL to walk me through the fix.
LGTM! 😋

awharn (approved):

Issue #1724, opened 2023-05-27
zos-ssh command returns even the prompt and issuing command in stdout


weinfurt:
**Describe the bug**
When issuing zos ssh commands I always get the prompt character and the issuing command in the stdout. Even the `--response-format-json` flag does not help. Sometimes, when the command is long it gets trimmed. Here are some examples:

![image](https://github.com/zowe/zowe-cli/assets/10806904/c28e8a94-e738-453a-8a44-b363a1cf7f00)

Here is the output even more corrupted:
![image](https://github.com/zowe/zowe-cli/assets/10806904/7a3de9b6-502f-4be4-b9d3-2b87a63cfc39)

**Expected and actual results**
only the command output should be returned and no prompt. Like running the command on the system itself.
![image](https://github.com/zowe/zowe-cli/assets/10806904/80df6bb4-a33d-44c7-9978-5de78fe7bd47)

**Describe your environment**
Zowe v7.16.4
Node v19.9.0
Npm v9.6.3
Ubuntu bash in WSL

**Additional context**

weinfurt:
I spent some time to investigate it and maybe workaround this and I figured out that the output is dependent on some setup in my `.profile` config file in my home directory. I am using `bash` by default so this is the last command in my `.profile` file. If I remove the `bash` command and also the configuration of my default text editor (vi), then I will get output with just the default `$` prompt and then the expected command output. It is not perfect but at least I am not getting the clutter from the command line itself.
This is what I am getting with disabled bash and default editor:
![image](https://github.com/zowe/zowe-cli/assets/10806904/458fd56a-1312-4bad-8ade-fb147f1fdbf0)

The only extraneous character in the output is the `$` prompt sign and some white spaces

I can share my .profile file if needed.

adam-wolfe:
This should be resolved by #1789. Please feel free to comment if you still see this behavior.

PR #1776, opened 2023-08-04
only showing command output in stdout for `zowe uss issue cmd`

ATorrise:
**What It Does**
Fixes Issue (#1724 

**How to Test**
Run any command with `zowe uss issue cmd "chosenCommand" --rfj` and verify that `$chosenCommand` isn't present in the stdout of the json response. 

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**
Newline padding has been removed from the front of command output but not the end. This is because line endings appear to vary across systems and rather than attempt to remove these endings and cause some unintended break, it seems easier to have the end user remove this information if they choose.

### DIFF
--- a/packages/zosuss/src/Shell.ts
+++ b/packages/zosuss/src/Shell.ts
@@ -94,8 +94,10 @@ export class Shell {
                                 dataToPrint = "";
                                 isUserCommand = false;
                             } else if (isUserCommand) {
-                                // print out the user command result
-                                stdoutHandler(dataToPrint);
+                                // don't print out command, just print result
+                                if (!dataToPrint.match(new RegExp("\\$"))){
+                                    stdoutHandler(dataToPrint);
+                                }
                                 dataToPrint = "";
                             }
                         }
--- a/packages/zosuss/src/Shell.ts
+++ b/packages/zosuss/src/Shell.ts
@@ -96,7 +96,11 @@ export class Shell {
                             } else if (isUserCommand) {
                                 // don't print out command, just print result
                                 if (!dataToPrint.match(new RegExp("\\$"))){
-                                    stdoutHandler(dataToPrint);
+                                    if (dataToPrint.startsWith("\r\n")){
+                                        stdoutHandler(dataToPrint.split("\r\n")[1]);
+                                    }else{
+                                        stdoutHandler(dataToPrint);
+                                    }
                                 }
                                 dataToPrint = "";
                             }
--- a/packages/zosuss/CHANGELOG.md
+++ b/packages/zosuss/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe z/OS USS SDK package will be documented in this file.
 
+## Recent Changes
+- BugFix: Updated `zowe zos-ssh issue cmd` to return just the command output in `stdout` instead of both the command and its output. [#1724](https://github.com/zowe/zowe-cli/issues/1724)
+
 ## `7.6.1`
 
 - BugFix: Updated `ssh2` dependency to fix "Received unexpected packet type" error. [#1516](https://github.com/zowe/zowe-cli/issues/1516)
--- a/packages/zosuss/__tests__/__unit__/Shell.unit.test.ts
+++ b/packages/zosuss/__tests__/__unit__/Shell.unit.test.ts
@@ -61,11 +61,12 @@ function checkMockFunctionsWithCommand(command: string) {
     expect(mockConnect).toBeCalled();
     expect(mockShell).toBeCalled();
 
-    // Check the stream.end() fucntion is called with an argument containing the SSH command
+    // Check the stream.end() function is called with an argument containing the SSH command
     expect(mockStreamWrite.mock.calls[0][0]).toMatch(command);
     expect(mockStreamEnd).toHaveBeenCalled();
     expect(stdoutHandler).toHaveBeenCalledWith("stdout data\n");
     expect(stdoutHandler).toHaveBeenCalledWith("\rerror");
+    expect(stdoutHandler).not.toContain("$")
 }
 
 describe("Shell", () => {
--- a/packages/zosuss/__tests__/__unit__/Shell.unit.test.ts
+++ b/packages/zosuss/__tests__/__unit__/Shell.unit.test.ts
@@ -66,7 +66,7 @@ function checkMockFunctionsWithCommand(command: string) {
     expect(mockStreamEnd).toHaveBeenCalled();
     expect(stdoutHandler).toHaveBeenCalledWith("stdout data\n");
     expect(stdoutHandler).toHaveBeenCalledWith("\rerror");
-    expect(stdoutHandler).not.toContain("$")
+    expect(stdoutHandler).not.toContain("$");
 }
 
 describe("Shell", () => {


packages/zosuss/src/Shell.ts
```diff
@@ -94,8 +94,14 @@ export class Shell {
                                 dataToPrint = "";
                                 isUserCommand = false;
                             } else if (isUserCommand) {
-                                // print out the user command result
-                                stdoutHandler(dataToPrint);
+                                // don't print out command (starts with `$`) just print result
+                                if (!dataToPrint.match(/\$/)){
+                                    if (dataToPrint.startsWith("\r\n")){
+                                        stdoutHandler(dataToPrint.split("\r\n")[1]);
```
zFernand0, line 100: I'm afraid this might be removing any new line endings coming back, which messes up with the output 😢 

<details>
<summary>Before and After with the current implementation</summary>
Before:

```
zFernand0 @ SSH:~/gh# zowe uss issue cmd "ls -al clitest/test" 

$ total 48
drwxr-xr-x   3 OMVSKERN SYS1        8192 Jul 26 15:07 .
drwxr-xr-x   8 OMVSKERN SYS1        8192 Jul 27 11:54 ..
drwxr-xr-x   4 OMVSKERN SYS1        8192 Jul 26 17:49 usstest
zFernand0 @ SSH:~/gh# zowe uss issue cmd "ls -al clitest/test" --rfj
{
  "success": true,
  "exitCode": 0,
  "message": "",
  "stdout": "\r\n$ total 48\r\ndrwxr-xr-x   3 OMVSKERN SYS1        8192 Jul 26 15:07 .\r\ndrwxr-xr-x   8 OMVSKERN SYS1        8192 Jul 27 11:54 ..\r\ndrwxr-xr-x   4 OMVSKERN SYS1        8192 Jul 26 17:49 usstest\n",
  "stderr": "",
  "data": {}
}
```

After

```
zFernand0 @ SSH:~/gh# zdev uss issue cmd "ls -al clitest/test" 
drwxr-xr-x   3 OMVSKERN SYS1        8192 Jul 26 15:07 .
zFernand0 @ SSH:~/gh# zdev uss issue cmd "ls -al clitest/test" --rfj
{
  "success": true,
  "exitCode": 0,
  "message": "",
  "stdout": "drwxr-xr-x   3 OMVSKERN SYS1        8192 Jul 26 15:07 .\n",
  "stderr": "",
  "data": {}
}
```
</details>

I believe what we wanted to do is to remove the first line that starts with `\r\n`?
perhaps `slice()` (and re-`join`ing the array) would help 😋 

<details>
<summary>Using the `slice` and `join` trick</summary>
Before:

```
zFernand0 @ SSH:~/gh# zowe uss issue cmd "ls -al clitest/test" 

$ total 48
drwxr-xr-x   3 OMVSKERN SYS1        8192 Jul 26 15:07 .
drwxr-xr-x   8 OMVSKERN SYS1        8192 Jul 27 11:54 ..
drwxr-xr-x   4 OMVSKERN SYS1        8192 Jul 26 17:49 usstest
zFernand0 @ SSH:~/gh# zowe uss issue cmd "ls -al clitest/test" --rfj
{
  "success": true,
  "exitCode": 0,
  "message": "",
  "stdout": "\r\n$ total 48\r\ndrwxr-xr-x   3 OMVSKERN SYS1        8192 Jul 26 15:07 .\r\ndrwxr-xr-x   8 OMVSKERN SYS1        8192 Jul 27 11:54 ..\r\ndrwxr-xr-x   4 OMVSKERN SYS1        8192 Jul 26 17:49 usstest\n",
  "stderr": "",
  "data": {}
}
```

After

```
zFernand0 @ SSH:~/gh# zdev uss issue cmd "ls -al clitest/test" 

drwxr-xr-x   3 OMVSKERN SYS1        8192 Jul 26 15:07 .
drwxr-xr-x   8 OMVSKERN SYS1        8192 Jul 27 11:54 ..
drwxr-xr-x   4 OMVSKERN SYS1        8192 Jul 26 17:49 usstest
zFernand0 @ SSH:~/gh# zdev uss issue cmd "ls -al clitest/test" --rfj
{
  "success": true,
  "exitCode": 0,
  "message": "",
  "stdout": "\r\ndrwxr-xr-x   3 OMVSKERN SYS1        8192 Jul 26 15:07 .\r\ndrwxr-xr-x   8 OMVSKERN SYS1        8192 Jul 27 11:54 ..\r\ndrwxr-xr-x   4 OMVSKERN SYS1        8192 Jul 26 17:49 usstest\n",
  "stderr": "",
  "data": {}
}
```
</details>


The only problem with that is that we can't assume that removing the first line is the correct solution either. Let's try the command `pwd`.

<details>
<summary>Using the `pwd` command as an example</summary>
Before:

```
zFernand0 @ SSH:~/gh# zowe uss issue cmd "pwd" 

$ /u/users/rijfe01
zFernand0 @ SSH:~/gh# zowe uss issue cmd "pwd" --rfj
{
  "success": true,
  "exitCode": 0,
  "message": "",
  "stdout": "\r\n$ /u/users/rijfe01\n",
  "stderr": "",
  "data": {}
}
```

After

```
zFernand0 @ SSH:~/gh# zdev uss issue cmd "pwd" 

zFernand0 @ SSH:~/gh# zdev uss issue cmd "pwd" --rfj
{
  "success": true,
  "exitCode": 0,
  "message": "",
  "stdout": "\n",
  "stderr": "",
  "data": {}
}
```
</details>

So, we may need to find a better way to handle these special cases 😋

packages/zosuss/src/Shell.ts
```diff
@@ -94,8 +94,14 @@ export class Shell {
                                 dataToPrint = "";
                                 isUserCommand = false;
                             } else if (isUserCommand) {
-                                // print out the user command result
-                                stdoutHandler(dataToPrint);
+                                // don't print out command (starts with `$`) just print result
+                                if (!dataToPrint.match(/\$/)){
```
zFernand0, line 98: also, what if my shell on USS has a different shell character (`>` for example)

zFernand0 (changes requested):
left a few comments,
Not sure what to do about the `$` since I can't seem to find how to change it on my USS session.

packages/zosuss/src/Shell.ts
```diff
@@ -94,8 +94,14 @@ export class Shell {
...
+                                    if (dataToPrint.startsWith("\r\n")){
```
t1m0thyj, line 99: The issue with the original command being included in the response does not occur on all systems.
Not sure why but I haven't been able to reproduce it.
When I test this PR locally, it breaks the SSH command by removing the first line of command output.

We should only strip off the first line of data in situations where we are confident it can be discarded without losing data.
Perhaps we can isolate the types of systems where the issue occurs? For example maybe it is specific to Windows.
Or we could make this logic more specific by checking whether `dataToPrint` begins with the command that the user entered.

t1m0thyj (changes requested):

packages/zosuss/src/Shell.ts
ATorrise: 😅

t1m0thyj: `zowe zos-ssh issue cmd "pwd"` works fine now 🙂 
But if I run a command that outputs multiple lines like `pwd && echo hello`, the 2nd line of output is lost.

I think some additional logic may be needed so that we check for the command to be present only for the first line of output, rather than for all lines?

ATorrise: Should work now, try again. I removed handling new line characters since we really can't anticipate how users will interact with this command

adam-wolfe (changes requested):
When I build and install this branch I am not seeing any output from issuing ssh commands.

With this branch:
```
PS C:\Users\me\GitRepos\zowe-cli\packages\cli> npm install -g

removed 278 packages, changed 1 package, and audited 3 packages in 5s

found 0 vulnerabilities
PS C:\Users\me\GitRepos\zowe-cli\packages\cli> zowe ssh issue cmd "echo hello;echo hello"

PS C:\Users\me\GitRepos\zowe-cli\packages\cli> zowe ssh issue cmd "echo hello;echo hello" --rfj
{
  "success": true,
  "exitCode": 0,
  "message": "",
  "stdout": "\n",
  "stderr": "",
  "data": {}
}
```

With latest from npmjs:
```
PS C:\Users\me\GitRepos\zowe-cli\packages\cli> npm install -g @zowe/cli

added 278 packages, changed 1 package, and audited 280 packages in 23s

13 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
PS C:\Users\me\GitRepos\zowe-cli\packages\cli> zowe ssh issue cmd "echo hello;echo hello"

$ hello
hello
PS C:\Users\me\GitRepos\zowe-cli\packages\cli> zowe ssh issue cmd "echo hello;echo hello" --rfj
{
  "success": true,
  "exitCode": 0,
  "message": "",
  "stdout": "\r\n$ hello\r\nhello\n",
  "stderr": "",
  "data": {}
}
```

packages/zosuss/src/Shell.ts
```diff
@@ -93,8 +94,11 @@ export class Shell {
                                 stdoutHandler(dataToPrint);
                                 dataToPrint = "";
                                 isUserCommand = false;
-                            } else if (isUserCommand) {
-                                // print out the user command result
+                            }
+                            else if (dataToPrint.includes(command)) {
```
t1m0thyj, line 98: I was able to reproduce the original issue by editing `~/.profile` to change my login shell from `sh` to `bash`.

The way this is currently written, if a user has `sh` as their login shell which does not return the command at the start of its output, we may end up stripping out text that we shouldn't be.
For example if a user runs `zowe uss issue cmd "pwd"` and their home directory is `/z/pwdaniels`, the "pwd" would be stripped from the output resulting in `/z/aniels`.

Should we make this check more strict? Perhaps something like the following:
`if (dataToPrint.trimStart().startsWith("$ " + command))`

t1m0thyj (approved):
I tested a number of cases with both `sh` and `bash` and they all worked 👍 
1. Command that outputs multiple lines: `pwd && echo hello`
2. Command that contains itself in output: `cat test.sh` where the contents of "test.sh" are:
    ```sh
    cat pre-test.sh
    cat test.sh
    cat post-test.sh
    ```
3. Long command >80 characters: `echo 'The quick brown fox jumps over the lazy dog' && echo 'The fast smug hare loses to the persistent tortoise'`

adam-wolfe (approved):

gejohnston (approved):

packages/zosuss/src/Shell.ts
```diff
@@ -87,16 +91,22 @@ export class Shell {
...
                                 stdoutHandler(dataToPrint);
                                 dataToPrint = "";
                                 isUserCommand = false;
-                            } else if (isUserCommand) {
-                                // print out the user command result
-                                stdoutHandler(dataToPrint);
-                                dataToPrint = "";
+                            }
+                            else if (dataToPrint.includes(Shell.startCmdFlag)) {
+                                commandFound = true;
```
t1m0thyj, line 102: The `commandFound` variable seems like it may be redundant with `isUserCommand`. Curious if it could be removed?

PR #1780, opened 2023-08-08
feat(test-utils): Enhance the CLI Test Utils to further support works…

zFernand0:
…paces

**What It Does**

Allows for monorepos (like the [CICS-for-Zowe-Client](https://github.com/zowe/cics-for-zowe-client)) to have its test results and test resources in the individual packages of the monorepo instead of at the root of the project.

**Note:** This is based on the assumption that `lerna.json#command.version.private === true`

**How to Test**

- Clone the https://github.com/zowe/cics-for-zowe-client
- Checkout the `v2-final` branch
- `cd packages/cli`
- `npm run test:unit`

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog: **N/A**

**Additional Comments**

This may (or may not) be a blocker to the CICS Monorepo work :yum:

Also, we could get this changes published with whatever CLI PR goes out next 😋

### DIFF
--- a/__tests__/__packages__/cli-test-utils/src/TestConstants.ts
+++ b/__tests__/__packages__/cli-test-utils/src/TestConstants.ts
@@ -13,19 +13,46 @@ import * as fs from "fs";
 import * as nodePath from "path";
 import * as findUp from "find-up";
 
+export let TEST_USING_WORKSPACE = false;
+
 function projectRootDir() {
     // First look for lerna.json to handle monorepos with tests at top level
-    const lernaJson = findUp.sync("lerna.json");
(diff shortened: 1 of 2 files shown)


t1m0thyj (approved):
Clever solution to support different project structures for monorepos 😋

awharn (approved):

awharn (changes requested):
Just noticed all of the tests failing in exactly the same way, should investigate before any kind of merge is done.

t1m0thyj (approved):

awharn (approved):

PR #1782, opened 2023-08-09
Update release config to create GH releases with changelog

t1m0thyj:
Other repos like [npm/cli](https://github.com/npm/cli) have pretty GitHub releases that include a list of changes and contributors (see [example](https://github.com/npm/cli/releases/tag/v9.8.1)).
We already use [zowe-actions/octorelease](https://github.com/zowe-actions/octorelease) which can generate release notes and create GitHub releases.
This PR enables these features, so that in the future when we publish Zowe CLI or SDKs a release will be created like [this](https://github.com/t1m0thyj/zowe-cli/releases/tag/v7.17.1).
Thanks @ATorrise for suggesting this 🙂 Here's an example of what a release would look like:

image

**Note:** The test release was made on my fork, and doesn't show a list of contributors because the owner of the repo (me) was the only contributor. When there are multiple contributors for a release, then it will automatically appear 🚀


.github/workflows/rust-cli-publish.yml
```diff
@@ -38,6 +38,7 @@ jobs:
           Native Zowe CLI client which communicates with a "daemon" version of Zowe CLI.
         draft: false
         prerelease: false
+        makeLatest: false
```
zFernand0, line 41: Will we have to go in and manually change the latest GH Release after every PR is merged (and published)?

t1m0thyj: I don't think we should have to manually set any releases to latest.
The `native-X` tags that we release for the Rust binary in this workflow should never be "latest" if we want the latest release to point to a release of the Zowe CLI package that is created by Octorelease.

ATorrise (approved):

awharn (approved):

release.config.js
```diff
@@ -17,7 +17,22 @@ module.exports = {
...
+                "zosconsole": "z/OS Console SDK",
+                "zosfiles": "z/OS Files SDK",
+                "zosjobs": "z/OS Jobs SDK",
+                "zoslogs": "z/OS Logs SDK",
+                "provisioning": "Provisioning SDK",
+                "zostso": "z/OS TSO SDK",
+                "zosuss": "z/OS USS SDK",
+                "workflows": "Workflows SDK",
+                "zosmf": "z/OSMF SDK",
+                "cli-test-utils": "CLI Test Utils"
```
traeok, line 33: Since the `secrets` package is published separately, I'm guessing it can't be added here?

traeok (approved):
LGTM, thanks @t1m0thyj - just left a question regarding `release.config.js` out of curiosity 😅

release.config.js
t1m0thyj: Good catch, I don't think it will hurt to add it now so that we don't forget to later 🙂 It just won't show up in our release notes until we stop publishing it separately and remove the `private: true` from its package.json.

PR #1789, opened 2023-08-24
Remove cmd from stdout

ATorrise:
**What It Does**
Fixes Issue #1724 

**How to Test**
Run any command with `zowe uss issue cmd "chosenCommand" --rfj` and verify that `$chosenCommand` isn't present in the stdout of the json response. 

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**
Newline padding removal has been left out of this fix. This is because line endings appear to vary across systems and commands and rather than attempt to remove these endings and cause some unintended break, it seems easier to have the end user remove this information if they choose.


packages/zosuss/__tests__/__unit__/Shell.unit.test.ts
```diff
@@ -95,6 +112,13 @@ describe("Shell", () => {
         checkMockFunctionsWithCommand(command);
     });
 
+    it("Should execute ssh command when `$ exit` found", async () => {
```
t1m0thyj, line 115: Do we need 2 separate unit tests or can they be combined into 1? I believe the only difference between them is that one calls `mockStream.emit("exit", 0);` and the other doesn't

t1m0thyj (approved):
LGTM, thanks @ATorrise!

packages/zosuss/src/Shell.ts
```diff
@@ -45,6 +45,9 @@ export class Shell {
                     let isUserCommand = false;
                     let rc: number;
 
+                    //isolate the command
+                    const cmd = command.split(' ')[0];
```
t1m0thyj, line 49: Sorry I realized one of the tests I tried on the previous PR isn't working anymore when my SSH shell is `sh`:
> 2. Command that contains itself in output: `cat test.sh` where the contents of "test.sh" are:
>     ```sh
>     cat pre-test.sh
>     cat test.sh
>     cat post-test.sh
>     ```

I think instead of just grabbing the first word of the command (in this case "cat"), maybe we want to grab up to the first 80 characters (before it would wrap)?
`const cmd = command.slice(0, 80)`

t1m0thyj (changes requested):

packages/zosuss/src/Shell.ts
ATorrise: doesn't work for "echo 'kjsdf kahdfkhjsd kahsdfkjas aksdjhfkajsdhfkhjasdf asdkfjhajksdhfkahjsdf asdkjfhajksd'" in bash

t1m0thyj: After discussing in person we concluded that first line of the command seems to get wrapped at 75 characters.

This is probably because the `ssh2` package has a default of 80 columns for terminal width.
We subtract 5 from that because of `\r\n$ ` at the beginning plus a line ending character at the end.

t1m0thyj (approved):

awharn (approved):

traeok (approved):
LGTM, thanks @ATorrise!

zFernand0 (approved):
LGTM! 😋

PR #1791, opened 2023-08-25
feat(secrets): Bundle static CRT for Windows builds

traeok:
**What It Does**

This PR adds a Rust flag so that users can run the Secrets SDK on Windows without needing the MSVC redistributables.

**How to Test**

Try running Zowe CLI in Windows Sandbox or an environment without the MSVC redistributables. It should still be able to store/retrieve secure credentials.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/packages/secrets/src/keyring/.cargo/config.toml
+++ b/packages/secrets/src/keyring/.cargo/config.toml
@@ -1,0 +1,5 @@
+[target.i686-pc-windows-msvc]
+rustflags = ["-Ctarget-feature=+crt-static"]
+
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-Ctarget-feature=+crt-static"]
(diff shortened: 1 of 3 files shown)


awharn (approved):

packages/secrets/src/keyring/.cargo/config.toml
t1m0thyj, line 1: Do we also need to add the "crt-static" flag for `aarch64-pc-windows-msvc`?

Also rather than adding a 3rd section to this file, I'm curious if we can combine them all into one using this syntax:
`[target.'cfg(all(windows, target_env = "msvc"))']`

zFernand0 (approved):

packages/secrets/src/keyring/.cargo/config.toml
awharn: For the first part, I think so. For the second, not sure.

traeok: confirmed, we can combine them 😁 resolved in upcoming commit

t1m0thyj (approved):

PR #1799, opened 2023-08-31
ci(secrets): Fix CI workflow to use prebuilds from current branch

traeok:
**What It Does**

The CI workflow was previously grabbing the latest prebuilds from the master branch. This PR changes this behavior so it will use the current branch that the publish workflow is running on.

This PR also prevents lerna from auto-versioning the Secrets SDK, as it is versioned and published separately.


awharn (approved):

t1m0thyj (approved):
LGTM, the correct prebuilds and `.cargo/config.toml` are now bundled 👍 
image

PR #1800, opened 2023-09-01
chore: Update CHANGELOG versions for CLI and Secrets SDK

traeok:
**What It Does**

This PR fixes the versions for the changelog in both Zowe CLI and the Secrets SDK.
It also:

- adds a version number (1.0.0) and license to `keyring/Cargo.toml`.
- adds a fallback to `npm@9` in case it cannot find a version and tries to use `npm@10`, which is suffering from a `EBADENGINE` error on Windows.

### DIFF
--- a/.github/workflows/zowe-cli.yml
+++ b/.github/workflows/zowe-cli.yml
@@ -78,7 +78,7 @@ jobs:
 
     - name: Use Original NPM Version
       id: original-npm-version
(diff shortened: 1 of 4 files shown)


awharn (approved):

traeok:
Looks like the release of npm v10 broke Node 16 on Windows 🙃 we might have to use npm@9 for now

KevinLoesch1 (approved):

PR #1840, opened 2023-10-30
Restore Jest sonar reporter and bump workflow timeout

awharn:
**What It Does**
Restore Jest-Sonar scans and increase workflow timeout

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**


gejohnston (approved):
Thanks for your continued efforts to keep this stuff running.

t1m0thyj (approved):
LGTM, thanks @awharn!

jest.config.js
```diff
@@ -134,6 +134,7 @@ module.exports = {
         }],
         ["github-actions", { "silent": false } ]
     ],
+    "testResultsProcessor": "jest-sonar-reporter",
```
zFernand0, line 137: Should the "jest-sonar-reporter" be added to the reporters list above?

I'm kind of worried about the `testResultsProcessor` property, since it had a tendency to override some reporters (probably in older versions of jest). IDK if you remember, but that's why Dan's `jest-stare` reporter had a list of `additionalReporters` where we had the `jest-junit` and `jest-sonar`.

If this is not something to be concerned about in newer versions of Jest, then I'm fine with leaving it as-is.

package.json
```diff
@@ -80,5 +81,8 @@
     "ts-jest": "^29.0.0",
     "typedoc": "^0.23.10",
     "typescript": "^4.0.0"
+  },
+  "jestSonar": {
```
zFernand0, line 85: If we add the `jest-sonar-reporter`to the reporters list, we can move this configuration item there too :yum:

zFernand0 (changes requested):
Changes LGTM.
I had one comment/question

jest.config.js
awharn: I tried that and it doesn't work. jest-sonar-reporter does not support the newer interface, and has been archived since 2019. Nothing else seems to exist outside of beta for providing Jest results to Sonar.

package.json
awharn: Again, tried that, Jest threw errors because the jest-sonar-reporter API wasn't compatible.

zFernand0 (approved):

PR #1859, opened 2023-11-10
Only one error format for V3 & remove references to bright

gejohnston:
**What It Does**

The revised format of Zowe CLI error messages (which was an option controlled by the environment variable ZOWE_V3_ERR_FORMAT during V2) is the only error format used by Zowe CLI in V3.

Also remove all references to 'bright' and 'brightside' within the app.

Closes [#1792](https://github.com/zowe/zowe-cli/issues/1792)

**How to Test**

You can test that the new error format is automatically being used by issuing the following erroneous command.

```
zowe zos-files list all-members "DATASET.DOES.NOT.EXIST"
```
The output will begin with the text:

```
Unable to perform this operation due to the following problem.
ISRZ002 Data set not cataloged - 'DATASET.DOES.NOT.EXIST' was not found in catalog.
```

You can confirm that the ```bright``` command is gone by installing zowe-cli into a clean environment. You then will get an error like the following when using the ```bright``` command.

```
> bright --version
bright: The term 'bright' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] added/updated automated tests
- [x] updated the changelog
- [x] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

**Additional Comments**

PR #1961, opened 2023-11-13
Fix peer dependencies in Next

awharn:
**What It Does**
The peer deps are not updated in Next. This should fix that.

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/__tests__/__packages__/cli-test-utils/package.json
+++ b/__tests__/__packages__/cli-test-utils/package.json
@@ -46,6 +46,6 @@
     "@zowe/imperative": "8.0.0-next.202311132045"
   },
   "peerDependencies": {
-    "@zowe/imperative": "^5.0.0"
+    "@zowe/imperative": "^8.0.0"
   }
 }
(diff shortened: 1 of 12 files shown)


__tests__/__packages__/cli-test-utils/package.json
```diff
@@ -46,6 +46,6 @@
     "@zowe/imperative": "8.0.0-next.202311132045"
   },
   "peerDependencies": {
-    "@zowe/imperative": "^5.0.0"
+    "@zowe/imperative": "^8.0.0"
```
t1m0thyj, line 49: According to the [npm semver calculator](https://semver.npmjs.com/), `^8.0.0` doesn't match versions with a prerelease string like `8.0.0-next.XXX`.

I think `^8.0.0-next` would work for now. In the future once after we've successfully published a `next` tag, we may want to switch to `>=8.0.0-next.XXX <9.0.0` and reference a specific timestamp - I believe this is what we settled on doing back when v2 was vNext.

t1m0thyj (approved):
LGTM, thanks @awharn!

zFernand0 (approved):
Even after doing something similar to this, I had to pass the `force: true` flag on the `scripts/rewriteShrinkwrap.js` script during a test I was doing on another PR 😋 
LGTM! 😋

zFernand0:
combining with:
- #1859

Issue #1972, opened 2023-11-23
Add data set creation template for COBOL source codes and compiler listings


patricktiu:
**Is your feature or enhancement request related to a problem or limitation? Please describe**

There are no data set templates to create PDSE data sets to hold a large number of COBOL source code, copybooks, and compiler listings.

**Describe your enhancement idea**
- Provide a default COBOL source code data set creation template that creates a PDSE type of data set to hold a large number of COBOL source codes and copybooks
- Provide a default compiler listing data set creation template that can store the COBOL compiler listings.

**Describe alternatives you've considered**

**Provide any additional context**

adam-wolfe:
Thanks for your contribution, but as mentioned in the PR, we have decided to take a different approach.

Issue #1699, opened 2023-04-24
v3: Remove validation of `RECFM` and return z/OSMF error


zFernand0:
Eliminating unnecessary validation of RECFM in the Zowe CLI can provide greater flexibility and compatibility for mainframe systems. Not all mainframe datasets have a specific record format, and some may have unique formats that are not recognized by the Zowe CLI. By removing this validation, the CLI can support a wider range of datasets and be more compatible with different mainframe systems. This can provide a better experience for system administrators and developers, who may require greater flexibility in their dataset management processes.

https://github.com/zowe/zowe-cli/blob/ea9c047daa99bfa41d4f99184146a5da3dd99be3/packages/zosfiles/src/methods/create/Create.ts#L266

Issue #1679, opened 2023-03-30
Open solution for default file allocations settings or templates


adam-wolfe:
Adapted from: https://github.com/zowe/vscode-extension-for-zowe/issues/1425:

**Is your feature request related to a problem? Please describe.**

User's have no way of specifying templates to be used for allocating data sets. They can select from a list of predefined data set templates (e.g., `data-set-binary`, `data-set-c`, `data-set-classic`, etc.) and specify options on the command to override default allocation settings. However, some users are interested in adding to the list of available data set templates.

**Describe the solution you'd like**

Allow users to define templates that are stored in a "zowe.templates.json" or similarly named file in the user's Zowe home directory that Zowe CLI can use to add additional data set templates under the `zowe zos-files create` command.

Example of a "zowe.templates.json" from https://github.com/zowe/vscode-extension-for-zowe/issues/1425:
```
    "Zowe-Default-Dataset-Allocations": [
       "PDS-FB-80": {
          "name": "Fixed blocked LRECL=80 library",
          "dsorg": "LIBRARY",
          "alcunit": "CYL",
          "primary": 1,
          "secondary": 1,
          "dirblk": 0,
          "recfm": "FB",
          "blksize": 0,
          "lrecl": 80
       },
       "PDS-VB-256": {
          "name": "Variable blocked LRECL=256 library",
          "dsorg": "LIBRARY",
          "alcunit": "CYL",
          "primary": 1,
          "secondary": 1,
          "dirblk": 0,
          "recfm": "FB",
          "blksize": 0,
          "lrecl": 80
       },
       "Loadlib": {
          "name": "Loadlib for program object",
          "dsorg": "LIBRARY",
          "alcunit": "CYL",
          "primary": 10,
          "secondary": 10,
          "dirblk": 0,
          "recfm": "U",
          "blksize": 32760,
          "lrecl": 27998
       },
       "Report": {
        "name": "Sysout capture for LRECL=133",
        "dsorg": "PS",
        "alcunit": "CYL",
        "primary": 10,
        "secondary": 10,
        "dirblk": 0,
        "recfm": "FBA",
        "blksize": 0,
        "lrecl": 133
     }
  ]
```
**Describe alternatives you've considered**

Users could write a script to feed the allocation options to Zowe CLI or they could keep their allocation commands in a text file and copy and paste them into a terminal as needed, however it would be much simpler for users to just enter a `zowe files create` command using their own templates.

**Additional context**

[...]

PR #1973, opened 2023-11-23
Add data set templates for COBOL source codes and compiler listings

patricktiu:
**What It Does**

Provide a default COBOL source code data set creation template that creates a PDSE type of data set to hold a large number of COBOL source codes and copybooks and a default compiler listing data set creation template that can store the COBOL compiler listings. 

Resolve https://github.com/zowe/zowe-cli/issues/1972

**How to Test**

Run automated tests

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


adam-wolfe:
Hi Patrick. First, I want to acknowledge that you have put together an impressively thorough pull request and we realize that this must have taken significant effort. However, after discussing this at length with the Zowe CLI Squad, we have decided to go a different direction. Our main concern is that, by adding to our existing list of data set templates, we could be opening the door to future pull requests adding any number of additional templates which might be too site-specific or user-specific to be broadly useful. 

Instead, we think it may be better to give users a way to specify their own templates (https://github.com/zowe/zowe-cli/issues/1679). One thing I noticed in your PR is that you added FBA as a record format. We do plan to remove our checks for recfm, so the user would be able to specify any record format (with Zowe CLI)) and let z/OSMF handle validation (https://github.com/zowe/zowe-cli/issues/1699). It may also be helpful for users to have a built-in PDSE template to select from the `zowe files create` command (https://github.com/zowe/zowe-cli/issues/1988), which is something we will consider for the future.

PR #1977, opened 2023-11-28
Deprecate old getDataSet in v2 LTS, move to zosfiles SDK

awharn:
**What It Does**
Copies getDataSet to the zosFiles SDK under the new name, and deprecates the original function

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/cli/src/zosfiles/ZosFiles.utils.ts
+++ b/packages/cli/src/zosfiles/ZosFiles.utils.ts
@@ -14,6 +14,7 @@
 import { IDataSet } from "@zowe/zos-files-for-zowe-sdk";
 
 /**
+ * @deprecated - use @zowe/zos-files-for-zowe-sdk's ZosFilesUtils.getDataSetFromName instead
  * Converts the name of a data set to an IDataSet
  * @param {string} name  - the name in the form USER.DATA.SET | USER.DATA.SET(mem1)
  */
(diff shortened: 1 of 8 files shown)


packages/cli/src/zosfiles/ZosFiles.utils.ts
```diff
@@ -14,6 +14,7 @@
 import { IDataSet } from "@zowe/zos-files-for-zowe-sdk";
 
 /**
+ * @deprecated - use @zowe/zos-files-for-zowe-sdk's ZosFilesUtils.getDataSetFromName instead
  * Converts the name of a data set to an IDataSet
  * @param {string} name  - the name in the form USER.DATA.SET | USER.DATA.SET(mem1)
  */
```
zFernand0, line 20: Alongside the deprecation message, should we call the `getDataSetFromName` method here?
That way, if there is ever a bug in the method, we can fix in in one place? 😋 

I don't think doing this will cause a circular dependency since the CLI package already imports the files sdk 😋

zFernand0 (changes requested):

packages/cli/src/zosfiles/ZosFiles.utils.ts
gejohnston: 👍

gejohnston (approved):
Approve after Fernando's comment is addressed.

packages/cli/src/zosfiles/ZosFiles.utils.ts
awharn: Done

zFernand0 (approved):

Issue #1992, opened 2023-12-19
Incorrect output/results when issuing zowe zos-files compare on Windows using the --seqnum false option.


KevinLoesch1:
When executing the **zowe zos-files compare** on Windows using the **--seqnum false**, erroneous data may be included in the output and could potentially give an incorrect compare result. 

![image](https://github.com/zowe/zowe-cli/assets/102045561/24aca5e1-7345-4025-aa5e-10499279e78a)

- Operating system and version: Windows

adam-wolfe:
Resolved by #1993.

Issue #1754, opened 2023-07-20
zowe files compare option description text is incorrect for uss-files


adam-wolfe:
**Describe the bug**

The description for `zowe files compare uss-files` incorrectly says that the command will compare the contents of a local file with that of a z/OS file. This is incorrect and appears to have been copied from the `local-file-uss-file` option.

**Expected and actual results**

Description for the uss-files | uss | uf option should be "Compare content of two z/OS USS files".

The capitalization for z/OS and USS for these option descriptions should also be more consistent, so maybe it would be best to change z/os to z/OS and uss to USS where they appear in the option descriptions for the zowe files compare command.

**Describe your environment**

Zowe CLI Version: 7.17.0

**Additional context**
Copy of help text:
```
\> zowe files compare

 DESCRIPTION
 -----------

   Compare the contents of a two data set members on your terminal
   (stdout).

 USAGE
 -----

   zowe zos-files compare <command>

   Where <command> is one of the following:

 COMMANDS
 --------

   data-set | ds                Compare content of two z/OS data sets
   local-file-data-set | lf-ds  Compare content of a local file and a
                                z/os dataset
   local-file-spool-dd | lf-sdd Compare content of a local file and a
                                spool dd
   local-file-uss-file | lf-uss Compare content of a local file and a
                                z/os uss file
   spool-dd | sdd               Compare content of two spool dds
   uss-files | uss | uf         Compare content of a local file and a
                                z/os uss files
```

adam-wolfe:
Resolved by #1993

PR #1993, opened 2023-12-19
Fix for compare sequence numbers and online help

KevinLoesch1:
**What It Does**
This pull request corrects two bugs. The first one is to correct the output and compare when **zowe files compare** is issued on Windows and the **--noseq false** option is specified. The second fix is for the **zowe files compare** online help. 

https://github.com/zowe/zowe-cli/issues/1754
https://github.com/zowe/zowe-cli/issues/1992

**How to Test**
Run **zowe files compare** with the **--noseq false** on Windows against files that have sequence numbers. The first sequence number should not be the last character of the line. 

**Review Checklist**
I certify that I have:
- [X] tested my changes
- [X] added/updated automated tests
- [X] updated the changelog
- [X] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)


gejohnston (approved):

packages/cli/__tests__/zosfiles/__unit__/compare/ds/Dataset.handler.unit.test.ts
```diff
@@ -178,6 +178,41 @@ describe("Compare data set handler", () => {
             expect(getDiffStringSpy).toHaveBeenCalledWith("compared", "compared", options);
         });
 
+        it("should compare two data sets containing carriage returns in terminal with --seqnum specified(Windows Specific)", async () => {
+            if(process.platform === "win32")
```
zFernand0, line 182: should we do `(process.platform === "win32" ? it : it.skip)("should compare ...` ?
That way we don't have an empty test on other operating systems?

I don't recall if this is the exact syntax though.

zFernand0 (approved):
Left a small comment, but changes LGTM! 😋

packages/cli/src/zosfiles/compare/CompareBaseHelper.ts
```diff
@@ -164,7 +164,11 @@ export class CompareBaseHelper {
     public prepareContent(content: string | Buffer): string {
         let contentString = content.toString();
         if(this.seqnum === false) {
-            const seqnumlen = 8;
+            let seqnumlen = 8;
+            /* If Windows, account for the carriage return */
+            if(process.platform === "win32" && content.toString().endsWith("\r\n")){
```
t1m0thyj, line 169: Is checking for Windows necessary? It's possible to have a local file with Windows line endings (CRLF) on any OS.

t1m0thyj: I tried commenting out the process check and testing on my Mac with a local file that has CRLF. It seems like the first sequence number still shows up:
```
zowe files compare lf-uss temp.jcl "/tmp/XXX/iefbr14.jcl" --no-seqnum
- Removed
+ Added

- //IEFBR14J JOB  (IZUACCT),'TIMOTHY',CLASS=A                             0
- //STEP1    EXEC PGM=IEFBR14                                             0
+ //IEFBR14J JOB  (IZUACCT),'TIMOTHY',CLASS=A                             
+ //STEP1    EXEC PGM=IEFBR14                                             
```

KevinLoesch1: I was thinking about that possibility. I will make the change.

PR #1996, opened 2024-01-02
adding test-specific tsconfig files in appropriate dirs

ATorrise:
**What It Does**
Configures typescript test files to include jest type information, reducing the amount of distracting, not-real-yet-still-highlighted-as-such errors when working in test files

**How to Test**
Pull branch and start opening up some test files, making sure there are no errors like: 
```
Cannot find name 'describe'. Do you need to install type definitions for a test runner? Try `npm i --save-dev @types/jest` or `npm i --save-dev @types/mocha`.ts(2582) 
```

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog


packages/cli/__tests__/tsconfig.json
t1m0thyj, line 1: Can we reduce duplication in the `packages/**/__tests__/tsconfig.json` files by having them extend from `__tests__/tsconfig.json`?

Similar to how `packages/cli/tsconfig.json` doesn't need to define any compiler options because it extends from the top-level `ts-config.json`

zFernand0 (approved):
TL;DR: LGTM! 😋 

---

I like this solution a lot.
It resolved the problem of having to import `@types/jest` in a random file for those Jest functions (`describe, afterAll, ...`) to be recognized as actual functions.

I'm curious if we could do some technique to reduce duplication. Something like...
```
"extends": "../../tsconfig.json",
```

One thing to keep in mind that this only fixes tests that are under the `packages/PKG/__tests__/` directory. For example, this won't help with 
`packages/imperative/src/utilities/__tests__/CliUtils.unit.test.ts` since the `tsconfig.json` is under `packages/imperative/__tests__/` 😋

zFernand0:
>One thing to keep in mind that this only fixes tests that are under the packages/PKG/__tests__/ directory. For example, this won't help with
packages/imperative/src/utilities/__tests__/CliUtils.unit.test.ts since the tsconfig.json is under packages/imperative/__tests__/ 😋

I take that back.
Even files from within the `imperative/src/utilities/__tests__/` work now 😋

traeok (approved):
LGTM, thanks @ATorrise!

PR #2001, opened 2024-01-03
feat(cfg): Schema management using `ProfileInfo`

traeok:
**What It Does**

This PR implements:
- **Schema management** APIs in Imperative's `ProfileInfo` class, primarily to allow schema contributions from extenders and plug-ins of Zowe client applications. Also provides functionality for the following use cases:
  - _Retrieve the schema for a specific profile type_
  - _Rebuild the entire schema using all available profile types, or rebuild the schema with types sourced from a specific application_
  - _Retrieve the list of sources for a specific profile type_
  - _Add a profile type to a layer in the configuration file (with defaults, if any exist)_
- The concept of **schema versioning**, enabling the following:
  - _Replace a schema on-disk with a newer version, if provided_
  - _Recognize conflicts between schemas based on their versions_
  - _Warn users when a schema has been upgraded to a major version_
  - _Warn users if an extension/plug-in tries to register with an older major schema version than the one on-disk_

**How to Test**

Create a test plug-in or extension that uses `ProfileInfo` and try one (or more) of the newly-implemented APIs:
- `addProfileToConfig`
- `addProfileTypeToSchema`
- `getProfileTypes`
- `buildSchema`
- `getSchemaForType`

Alternatively, build and run tests, then do black-box testing to determine whether default behavior has remained the same. Testers should not notice any obvious differences when using the CLI build in this PR.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog


traeok:
I've marked this as ready for early reviews - going to do some manual testing to ensure some scenarios are handled, but figured anyone can start reviewing as the APIs are mostly finalized.

gejohnston (approved):
Looks good to me.

zFernand0 (commented):
LGTM! 😋 
Curious if you had planned to support reading an equivalent "zowe.extender.json" file from the project directory in this PR, or as a follow up enhancement if requested by extenders?

Or maybe I missed it here 😋 

Right now I'm trying to test each one of the statements in your PR descriptions.
Thanks for describing them all 😋

packages/imperative/src/config/src/ConfigBuilder.ts
```diff
@@ -30,31 +31,10 @@ export class ConfigBuilder {
...
-                            properties[k] = v.optionDefinition.defaultValue;
-                        }
-                        if (properties[k] === undefined) {
-                            // Fall back to an empty value
-                            properties[k] = this.getDefaultValue(v.type);
-                        }
-                    }
-                }
-            }
+            const defaultProfile = this.buildDefaultProfile(profile, opts);
```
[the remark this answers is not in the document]
traeok: Yes, good catch! I think I got lucky and avoided a compile error, because `this` would end up being a static `ConfigBuilder` 😅  but best to be explicit. Resolved in upcoming commit 👍

packages/imperative/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to the Imperative package will be documented in this file.
 ## Recent Changes
 
 - BugFix: Fixed error message shown for null option definition to include details about which command caused the error. [#2002](https://github.com/zowe/zowe-cli/issues/2002)
+- Enhancement: Added multiple APIs to the `ProfileInfo` class to help manage schemas between client applications.
```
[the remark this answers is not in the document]
traeok: We do have an issue in the ZE repo, but not this one.
I can create one and link it to the ZE issue for tracking purposes 👍 

Will update the changelog once I've added the issue

traeok:
> LGTM! 😋 Curious if you had planned to support reading an equivalent "zowe.extender.json" file from the project directory in this PR, or as a follow up enhancement if requested by extenders?
> 
> Or maybe I missed it here 😋

I discussed this a bit with @t1m0thyj, and while it seems a bit premature to add handling for a project-level `zowe.extenders.json`, I don't want to risk breaking project-level schemas because of missing checks. However, since I don't see many scenarios where two project directories would have plug-ins with different schema versions, I plan on following up with a couple of internal folks to evaluate the benefits of those checks.

packages/imperative/CHANGELOG.md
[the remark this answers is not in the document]
traeok: Added in 85d418a ✔️

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -1225,6 +1234,301 @@ export class ProfileInfo {
         LoggerUtils.setProfileSchemas(this.mProfileSchemaCache);
     }
 
+    /**
+     * Reads the `extenders.json` file from the CLI home directory.
+     * Called once in `readProfilesFromDisk` and cached to minimize I/O operations.
+     */
+    private readExtendersJsonFromDisk(): void {
```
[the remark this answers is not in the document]
traeok: Unfortunately, since I made it static to access during plugin installation, I cannot access the `ensureReadFromDisk` function. But, the function will ensure that the `extenders.json` file is created if it does not exist at the global level, and it will fall back to the global-level file if it doesn't exist at the project level.

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -1225,6 +1234,312 @@ export class ProfileInfo {
...
+        const layerToUpdate = this.getTeamConfig().mLayers.find((l) => l.path === layerPath);
+        const cacheKey = `${layerPath}:${profileType}`;
+
+        const sameSchemaExists = versionChanged ? false :
+            this.mProfileSchemaCache.has(cacheKey) && lodash.isEqual(this.mProfileSchemaCache.get(cacheKey), schema);
+        // Update the cache with the newest schema for this profile type
+        this.mProfileSchemaCache.set(cacheKey, schema);
+
+        const schemaUri = new url.URL(layerToUpdate.properties.$schema, url.pathToFileURL(layerPath));
+        const schemaPath = url.fileURLToPath(schemaUri);
```
zFernand0, line 1327: I understand we haven't talked much about supporting the use-case of hosting the schema file somewhere.
However, I'm curious if we should do something similar to the `Config#getSchemaInfo()` function.
- https://github.com/zowe/zowe-cli/blob/master/packages/imperative/src/config/src/Config.ts#L380

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -1225,6 +1234,312 @@ export class ProfileInfo {
         LoggerUtils.setProfileSchemas(this.mProfileSchemaCache);
     }
 
+    /**
+     * Reads the `extenders.json` file from the CLI home directory.
+     * Called once in `readProfilesFromDisk` and cached to minimize I/O operations.
+     * @internal
+     */
+    public static readExtendersJsonFromDisk(): IExtendersJsonOpts {
+        const extenderJsonPath = path.join(ImperativeConfig.instance.cliHome, "extenders.json");
```
zFernand0, line 1243: Should we document (maybe with JSDoc) how the `extender.json` file looks like?

We don't have to be too verbose, but something like the Pseudocode examples in this file might help 
- https://github.com/zowe/zowe-cli/blob/master/packages/imperative/src/config/src/ProfileInfo.ts#L61

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -1225,6 +1234,312 @@ export class ProfileInfo {
...
+
+        // Include all profile types from extenders.json if we are not filtering by source
+        if (filteredBySource) {
+            return [...profileTypes].filter((t) => {
+                if (!(t in this.mExtendersJson.profileTypes)) {
+                    return false;
+                }
+
+                return this.mExtendersJson.profileTypes[t].from.some((src) => sources.includes(src));
+            }).sort((a, b) => a.localeCompare(b));
```
zFernand0, line 1514: Considering the case of an invalid schema (i.e. new profile type not yet registered)...
I do think it's the correct thing to return an empty array for this.

However, should we have some extra validation and throw an error if any profileTypes in the provided source list is invalid?

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -1225,6 +1234,312 @@ export class ProfileInfo {
...
+    }
+
+    /**
+     * Returns the schema object belonging to the specified profile type.
+     *
+     * @param {string} profileType The profile type to retrieve the schema from
+     * @returns {IProfileSchema} The schema object provided by the specified profile type
+     */
+    public getSchemaForType(profileType: string): IProfileSchema {
+        let finalSchema: IProfileSchema = undefined;
```
zFernand0, line 1527: The previous function return an empty array, and this one undefined for invalid profileTypes.
I'm happy with the way that we are handling this scenario in both functions.

However, if we decide to throw an error in the `getProfileTypes`, then we probably should do the same here 😋 

I'm not advising we do, though. Just something to consider 😋

zFernand0 (approved):
TL;DR: LGTM! 😋 

Left a few comments, but none of them should stop this PR from being merged 😋 

Nice work! 🥳

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -15,13 +15,15 @@ import * as path from "path";
 import * as url from "url";
 import * as jsonfile from "jsonfile";
```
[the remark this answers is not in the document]
traeok: I can't remember where I saw it being used, but noticed it was a dependency. I've been using it mostly for the `spaces` option to fix the output formatting, otherwise the object is written to disk without any new lines.

But, if `fs.writeFileSync` is efficient enough to do this behind the scenes, I definitely would prefer using that instead 😋 I'll try it out and give an update once confirmed 👍

packages/imperative/src/config/src/ProfileInfo.ts
traeok: Ah, I didn't realize this function existed! 😅 

After today's discussion about only having to update the global-level schema, it probably would make more sense to use `getSchemaInfo`. I will test this out locally to see if this changes the behavior first, but I think that would be a nice change.

packages/imperative/src/config/src/ProfileInfo.ts
traeok: Yes, thanks for catching that - ideally, I'd like to keep knowledge of that structure to developers only, especially since we don't intend to expose this file to users. I think this would fit well in the typedoc somewhere.

I was thinking that I should add that comment above the `IExtenderJsonOpts` type in the `IExtenderOpts.ts` file, especially since I am storing related types in that area 😋

packages/imperative/src/config/src/ProfileInfo.ts
traeok: I'm not sure that we can do much about updating a schema located at a web URL. I feel like that would require some extensive work to support connecting over some interface (SSH, FTP, etc.) or an API to update the contents of a remote schema.

That being said, I added a check that will not update the schema when the built URL does not have the `file` protocol. If it is the `file` protocol, even mangled paths such as `file:///Users/traeok/https:/a.txt` will work fine with `existsSync` (and will just return false).

packages/imperative/src/config/src/ProfileInfo.ts
traeok: Added in [4b72f50](https://github.com/zowe/zowe-cli/pull/2001/commits/4b72f50deeed88c3debf557a0158af890ab0e7a8)

packages/imperative/src/config/src/ProfileInfo.ts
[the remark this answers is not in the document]
traeok: I looked into it a bit, and we were using `jsonfile` before these changes, so I will keep it as-is for now. I also noticed that we are using it in other places within imperative, so maybe these adjustments could be done as a separate enhancement 🙂

packages/imperative/src/config/src/ProfileInfo.ts
traeok: Discussed offline - will file a separate issue to validate schemas when grabbing them in `loadAllSchemas` 👍

packages/imperative/src/profiles/src/doc/config/IProfileTypeConfiguration.ts
```diff
@@ -39,6 +39,10 @@ export interface IProfileTypeConfiguration {
      * @memberof IProfileTypeConfiguration
      */
     schema: IProfileSchema;
+    /**
+     * The version for the JSON schema document (not required).
+     */
+    schemaVersion?: string;
```
t1m0thyj, line 45: If the version is specific to the schema, would it make sense to store this as a `version` property on the `IProfileSchema` object instead?

packages/imperative/src/config/src/doc/IExtenderOpts.ts
```diff
@@ -0,0 +1,59 @@
...
+ *     "profileTypes": {
+ *         "banana": {
+ *             "from": ["@zowe/banana-for-zowe-cli", "Zowe Explorer Banana Extension"],
+ *             "version": "v1.1.0",
+ *             "latestFrom": "Zowe Explorer Banana Extension"
+ *         }
+ *     }
+ * }
+ * ```
+ */
```
t1m0thyj, line 29: Thanks for including an example file in the typedoc 😍

packages/imperative/src/config/__tests__/ProfileInfo.TeamConfig.unit.test.ts
```diff
@@ -1342,4 +1349,316 @@ describe("TeamConfig ProfileInfo tests", () => {
...
+                        },
+                        res: {
+                            success: true
+                        }
+                    }
+                );
+            });
+        });
+        describe("buildSchema", () => {
+            // TODO
```
t1m0thyj, line 1638: Is there an outstanding action associated with this todo?

traeok: Yes, thanks for catching this - resolved in 10c2f9e 👍

packages/imperative/src/profiles/src/doc/config/IProfileTypeConfiguration.ts
traeok: Thanks for the suggestion - I've moved `schemaVersion -> version` in `IProfileSchema` (see 97373db)

PR #2009, opened 2024-01-05
Fix 'create like' block size issues

KevinLoesch1:
**What It Does**
When using the **'Create.dataSetLike()'** File SDK method, the new data set didn't always have the correct or valid block size. This could cause the data set to be unusable. 

**How to Test**
Create a new data set using the **--like** option in the Zowe CLI or the **'Allocate Like'** menu option in Zowe Explorer VS Code extension.  The new data set should always have a block size that matches the 'Like' data set. 

**Review Checklist**
I certify that I have:
- [X ] tested my changes
- [X ] added/updated automated tests
- [X ] updated the changelog
- [X ] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)


awharn (approved):

packages/zosfiles/src/methods/create/Create.ts
```diff
@@ -150,8 +152,36 @@ export class Create {
...
+        *  set the block size, as the zosmf Rest API does not set the block size properly in
+        *  some instances.
+        *
+        */
+        if (tempOptions.blksize === null || tempOptions.blksize === undefined) {
+            let likeDataSetObj: IZosmfListResponse;
+            const likeDataSetList  = await List.dataSet(session, likeDataSetName, {
+                attributes: true, maxLength: 1,
+                start: likeDataSetName,
+                recall: "wait"
```
zFernand0, line 172: Thanks for considering waiting for a recall operation. (source: [_X-IBM-Migrated-Recall_](https://www.ibm.com/docs/en/zos/3.1.0?topic=interface-zos-data-set-member-utilities#IZUHPINFO_API_PutDataSetMemberUtilities__title__5))

Do we know what would happen if the migrated data set is in a volume which requires TSO prompts to be recalled?

This may be a larger endeavor, so I'm not asking for this recall to be handled here.  😋

zFernand0 (approved):
This PR still LGTM! 😋

PR #2013, opened 2024-01-09
Remove the 'zowe profiles' command group

gejohnston:
**What It Does**

This PR removes the command definitions for the `zowe profiles` command group. Thus, end-users will no longer be able to run any command from the `profiles` command group. It also removes the initialization of the profiles directory that previously occurred in the Imperative.init() function.

The majority of the changes in this PR come from re-implementing all automated tests that created a V1 profile in order to perform a desired operation.

This PR also fixes numerous system tests that failed due to changes to the Zowe error message format.

**How to Test**

1. Run a `profiles` command, and see that zowe tells you that such a command does not exist. For example:
    ```
    > zowe profiles create zosmf-profile YourZosmfProfileName --host YourHost --port 123 --user YourUser --pass "YourPassword"
    Command failed due to improper syntax
    Unknown group: profiles
    Did you mean: files create zos-file-system?

    Command entered: "profiles create zosmf-profile YourZosmfProfileName --host YourHost --port 123 --user YourUser --pass YourPassword"
    Use "zowe --help" to view groups, commands, and options.
    ```
2. Place some existing V1 profiles into the ZOWE_CLI_HOME\profiles directory and run a command to successfully convert those profiles to a team config.
   ```
   zowe config convert-profiles
   ```

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


gejohnston:
Regarding the failure of the codecov check: 

All of the identified 'failures' are related to new code that was added to properly pass system test failures. So, clearly that code is being tested. Those tests are just not being counted as part of the code coverage metric.

Further, these updates address fixes related to the new error format. Our new error format is not the objective of this pull request.

For these reasons, I feel justified in not addressing the failure of the codecov check in this Pull request.

packages/core/__tests__/utils/__integration__/__scripts__/exitOnFailure.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# function to exit if we encounter a bad return code
+exitOnFailure () {
```
zFernand0, line 4: yeah, this could definitely be reused if exposed at a higher level than just `cli/auth/system` or `core/utils/integration` 😋

packages/imperative/__tests__/__integration__/cmd/__tests__/integration/cli/auth/__scripts__/exitOnFailure.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# function to exit if we encounter a bad return code
+exitOnFailure () {
```
zFernand0, line 4: I believe we should reuse this code 😋

packages/imperative/__tests__/__integration__/cmd/__tests__/integration/cli/profiles/__scripts__/profiles/banana_profile_and_specify_cli.sh
```diff
@@ -1,25 +1,27 @@
-#!/bin/bash
+#!/bin/sh
 
-color=$1
-description=$2
-moldtype=$3
+profileColor=$1
+profileDescription=$2
+profileMoldType=$3
```
zFernand0, line 5: I know most of these scripts are somewhat specific to the test files that call them, but should we do the same and add some information about the field in case it is missing?
`${1:?"First parm (profileColor) is required."}`

packages/imperative/__tests__/__integration__/imperative/__tests__/__integration__/cli/config/schema/cli.imperative-test-cli.config.schema.integration.test.ts
```diff
@@ -38,8 +38,8 @@ describe("imperative-test-cli config schema", () => {
     });
     it("should print the generated schema", () => {
         const response = runCliScript(__dirname + "/__scripts__/schema.sh", TEST_ENVIRONMENT.workingDir, [""]);
-        expect(JSON.parse(response.stdout.toString())).toEqual(expectedSchemaObject);
         expect(response.stderr.toString()).toEqual("");
         expect(response.error).toBeFalsy();
+        expect(JSON.parse(response.stdout.toString())).toEqual(expectedSchemaObject);
```
zFernand0, line 43: Thanks for moving this down. 
That way we will see right away what the error is, instead of `"" !== { ... object ... }`

packages/imperative/src/imperative/__tests__/plugins/cmd/install/install.handler.unit.test.ts
```diff
@@ -12,6 +12,9 @@
 /* eslint-disable jest/expect-expect */
 import Mock = jest.Mock;
 
+let expectedVal: unknown;
+let returnedVal: unknown;
```
zFernand0, line 16: Not sure we have many variables being defined prior to `import` statements.

I assume we do this because the `jest.doMock(...)` block needs them.

If that's the case (and this is just me being picky), could we move them both (variable definitions and `jest.doMock` block) below imports.

If I remember correctly, only `jest.mock` is treated specially and has to be called very early in the file. However, I believe we can call `jest.doMock` at any point.

This bit of [Jest doc](https://jestjs.io/docs/jest-object#jestdomockmodulename-factory-options) explains how using `babel-jest` hoists all `jest.mock` calls, and that we can use `jest.doMock` to avoid this behavior.

---

That being said, we we need to mock the `path` module before importing any libraries, then I'm fine to keep both (definitions and `doMock` block) where they are now.

zFernand0 (commented):
Changes LGTM so far! 😋 

I'm still missing about 20 files to review, but I should probably provide feedback now and then continue with the rest 😋

packages/cli/__tests__/zostso/__system__/stop/__snapshots__/cli.zos-tso.stop.system.test.ts.snap
```diff
@@ -1,13 +1,17 @@
...
 Command failed due to improper syntax
 Command entered: \\"zos-tso start as --foo-bar\\"
 Available commands are \\"address-space\\".
 Use \\"zowe zos-tso start --help\\" to view groups, commands, and options.
-Error Details:
+
+Response From Service
+Error: Unknown arguments: foo-bar, fooBar
+
+Diagnostic Information
```
t1m0thyj, line 14: It seems odd to me to show the "Unknown arguments" error in the "Response From Service" category. There is no response from the service if client-side validation fails before a request is issued. And the error message is duplicated in the "Diagnostic Information" below. Is this an intended behavior of the new error formatting in v3?

packages/imperative/__tests__/__integration__/imperative/__tests__/__integration__/cli/config/convert-profiles/cli.imperative-test-cli.config.convert-profiles.integration.subtest.ts
```diff
@@ -142,17 +153,20 @@ describe("imperative-test-cli config convert-profiles", () => {
...
-            expect(response.stderr.toString()).toEqual("");
+            expect(response.status).toBe(0);
             expect(fs.existsSync(configJsonPath)).toBe(false);
         });
         it("should not delete profiles if prompt is rejected", () => {
-            runCliScript(__dirname + "/__scripts__/delete_profiles_secured_and_base.sh", TEST_ENVIRONMENT.workingDir);
+            // delete profiles previously created, but leave the profile type definitions
+            let response = runCliScript(__dirname + "/__scripts__/delete_profiles.sh", TEST_ENVIRONMENT.workingDir);
+            expect(response.stdout.toString()).toEqual("");
+            expect(response.stderr.toString()).toEqual("");
```
t1m0thyj, line 167: Should we switch the order of these checks - 1st stderr and 2nd stdout? In case there is an error thrown, this makes it easier to debug tests because the error message logged to stderr will show up in the Jest output.

packages/imperative/__tests__/__integration__/cmd/__tests__/integration/cli/auth/__scripts__/auth_login_cmd_line_cert.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# copy our config file and certificate files
+resourceDir=$myScriptDir/../__resources__
+cp $resourceDir/base_cert.config.json .
+exitOnFailure "Failed to copy config file." $?
+
+cp $resourceDir/$baseCertFile .
+exitOnFailure "Failed to copy certificate file." $?
+
+cp $resourceDir/fakeKey.key .
```
t1m0thyj, line 19: Should this be copying `$resourceDir/$baseCertKey`?

packages/cli/__tests__/zostso/__system__/stop/cli.zos-tso.stop.address-space.system.test.ts
```diff
@@ -32,16 +32,18 @@ describe("zos-tso start address-space", () => {
 
     it("should throw an error if servlet key parameter is not provided", async () => {
         const response = runCliScript(__dirname + "/__scripts__/address-space/as_error_stop.sh", TEST_ENVIRONMENT);
-        expect(response.status).toBe(1);
         expect(response.stdout.toString()).toBe("");
-        expect(response.stderr.toString()).toMatchSnapshot();
+        expect(response.stderr.toString()).toContain('IZUG1126E: z/OSMF cannot correlate the request for ' +
```
t1m0thyj, line 36: Should we flip the order to be 1st stderr and 2nd stdout (same question as above)?

packages/cli/__tests__/zostso/__system__/stop/cli.zos-tso.stop.address-space.system.test.ts
```diff
@@ -32,16 +32,18 @@ describe("zos-tso start address-space", () => {
...
+        'key "ZOSMFAD-55-aaakaaac" with an active z/OS application session');
+        expect(response.status).toBe(1);
     });
 
     it("should throw an error if provided address space is inactive", async () => {
         const response = runCliScript(__dirname + "/__scripts__/address-space/as_error_stop.sh", TEST_ENVIRONMENT);
-        expect(response.status).toBe(1);
         expect(response.stdout.toString()).toBe("");
-        expect(response.stderr.toString()).toMatchSnapshot();
+        expect(response.stderr.toString()).toContain('IZUG1126E: z/OSMF cannot correlate the request for ' +
```
t1m0thyj, line 44: Should we flip the order to be 1st stderr and 2nd stdout (same question as above)?

t1m0thyj (commented):
Thanks @gejohnston and @awharn for all your work on this! Appreciate the improvements to our tests that were made along the way like better error handling in Bash scripts and reducing usage of snapshots 🙂  Left a few comments

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Properly construct workflow error messages to display properly with V3 error formatting.
```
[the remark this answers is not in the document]
gejohnston: Note that ProfilesConstants and CoreMessages are not part of this PR. They will be part of the next PR.

The impactful changes for this PR are in the imperative package, so the that is where I placed the statement that profile commands are gone. I can duplicate that entry from the imperative changelog in the CLI package changelog. Here is that imperative entry:

- LTS Breaking: Remove deprecated V1 'profiles' command group.

We just have to decide whether we want to be technically correct (imperative is where the changes occurred) or feel that we should make the change visible to people reading the CLI changelog.

What do you think?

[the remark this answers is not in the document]
zFernand0: You are right that it's technically on the Imperative package. 😋 

I guess it wouldn't hurt to duplicate the entry this time around (if that isn't too much trouble) 😋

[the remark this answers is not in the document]
gejohnston: Not a problem. Other changes need to be made anyway.

packages/cli/__tests__/auth/__system__/__scripts__/exitOnFailure.sh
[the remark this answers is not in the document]
gejohnston: After I copied the exitOnFailure script a couple of time, I thought that it would be good re-use a single script, but I did not know of a good home for the script.

Also, the scripts that use exitOnFailure will now have to use a long relative path to access exitOnFailure.sh (much like our javascript import statements that have paths which can be hard to determine where they actually lead.)

None-the-less, I tried __tests__/__scripts__/exitOnFailure.sh in one script and it worked just fine. I will do that in all scripts and consolidate down to just one copy of exitOnFailure.sh

packages/cli/src/workflows/create/Create.common.handler.ts
```diff
@@ -154,11 +155,11 @@ export default class CreateCommonHandler extends ZosmfBaseHandler {
                         this.arguments.remoteDirectory
                     );
                 } catch (err) {
-                    error =
-                        "Creating z/OSMF workflow with local file: " +
-                        this.arguments.localFile +
-                        " failed. More details: \n" +
-                        err;
+                    error = new ImperativeError({
```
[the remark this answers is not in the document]
gejohnston: Credit Andrew for all of the fixes to workflow modules.

packages/imperative/__tests__/__integration__/cmd/__tests__/integration/cli/profiles/__scripts__/profiles/banana_profile_and_specify_cli.sh
gejohnston: All of these parameters were already defined in the script when I started work on this script, so I did not think to add the extra parameter testing.

I will be updating this script for the exitOnFailure changes, so I will add parameter tests at the same time.

By the way, the syntax for these parameters is not just human readable information. If the parameter is missing, the text will be displayed on standard error, and the script will terminate immediately. When our Jest scripts test for  standard error to be an empty string, the error about the missing parameter will be displayed in the Jest error report.

packages/core/__tests__/utils/__integration__/__resources__/profiles/base/base_meta.yaml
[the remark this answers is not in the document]
gejohnston: Yes. The approach that I took was that if a test tested the operation to write a V1 profile, then I removed the test. If the test tested the operation to read a V1 profile, I kept the test (with pre-created V1 profile files). Those tests should confirm that the infrastructure still works for our (or extender's) conversion operations.

packages/imperative/src/imperative/__tests__/plugins/cmd/install/install.handler.unit.test.ts
gejohnston: If definition of the two variables are moved after the imports, the test script fails with this error:
```
    ReferenceError: Cannot access 'expectedVal' before initialization
       > 31 |             if (path[0] == expectedVal) {
            |                            ^
```

If the related doMock is then also moved after the definition of the variables, then a test fails with the following error:
```
    Expected: "/some/test/directory/prod-plugins.json"
    Received: "C:\\ourstuff\\repos\\zowe-cli\\prod-plugins.json"

    Number of calls: 1

      172 |      */
      173 |     const wasReadFileSyncCallValid = (expectedFile: string) => {
    > 174 |         expect(mocks.readFileSync).toHaveBeenCalledWith(
          |                                    ^
      175 |             expectedFile
      176 |         );
      177 |     };

```

Instead of providing a test value, the script now seems to return a path to the root of the tester's source directory, which is obviously not what should happen.

Since the location of the two variables was part of a commit from @awharn, he may have some additional insight. Otherwise, I suggest that we leave the working test the way it is.

packages/cli/__tests__/zostso/__system__/stop/__snapshots__/cli.zos-tso.stop.system.test.ts.snap
gejohnston: The error format handling typically takes properties from a REST response (or an imperative error) and places them under the most appropriate titles. When we never get to the remote service, values are filled in with the available information.

In the case of the zos-tso command run in this test, here is the output from running the same command interactively:
```
-> zowe zos-tso start as --foo-bar
Unknown arguments: foo-bar, fooBar
Command failed due to improper syntax
Command entered: "zos-tso start as --foo-bar"
Available commands are "address-space".
Use "zowe zos-tso start --help" to view groups, commands, and options.

Response From Service
Error: Unknown arguments: foo-bar, fooBar

Diagnostic Information
Unknown arguments: foo-bar, fooBar
```
I feel that the objectives of this PR to remove V1 profile definitions (and fix the inability of system test scripts to properly detect newly formatted errors) are being met. I think that additional changes should be handled outside this PR.

It might be valid to open an issue about how the zos-tso command group prepares its error message content (or how the error formatting logic choses to display that content).

gejohnston: This is the output of the command in the old error format:
```
$ zowe zos-tso start as --foo-bar
Command Error:
Unknown arguments: foo-bar, fooBar
Command failed due to improper syntax
Command entered: "zos-tso start as --foo-bar"
Available commands are "address-space".
Use "zowe zos-tso start --help" to view groups, commands, and options.
Error Details:
Unknown arguments: foo-bar, fooBar
```

packages/cli/__tests__/zostso/__system__/stop/cli.zos-tso.stop.address-space.system.test.ts
gejohnston: When we expect no error on stderr and meaningful output on stdout, I agree with that order.
When the test expects an error on stderr and nothing on stdout, I tend to want to see whatever is totally unexpected on stdout to be displayed before the test terminates in a failure.

packages/imperative/src/imperative/__tests__/plugins/cmd/install/install.handler.unit.test.ts
awharn: I have no further insights, just experienced the same behaviors you found, @gejohnston. I wanted to fix the tests and get them back into a working order, and the simplest solution seemed to be moving the let statements to above the imports. I did not put much effort into the changes aside from making sure the tests passed, and the changes made sense. If we want to re-evaluate how the tests are being run and where we are defining variables, perhaps that would be good for another issue (when we are less time constrained). The main purpose of these changes were to just get things working ahead of V3 so we can properly test other changes that we plan on making later without worrying about these tests polluting the results. :)

zFernand0: We can leave this test file (and possibly others like this) alone.
"_If it isn't broken, don't fix it!_" 😋 

I do appreciate you (@gejohnston) for spending some time to evaluate each and every single comment I made 😋

packages/cli/__tests__/zostso/__system__/stop/cli.zos-tso.stop.address-space.system.test.ts
zFernand0: I kind of agree with the logic order as it is.
For a few number of tests (those expecting errors), we may want to know right away if there was some stdout.

That way we see a nice Jest output along the lines of:
```

Failed: ... should throw an error...  
...
"address space stopped successfully"
```

packages/cli/__tests__/zostso/__system__/stop/cli.zos-tso.stop.address-space.system.test.ts
zFernand0: The previous comment also applies in this case 😋

packages/imperative/__tests__/__integration__/cmd/__tests__/integration/cli/auth/__scripts__/auth_login_cmd_line_cert.sh
zFernand0: Seems like we would want to use the required `baseCertKey` here 😋

packages/imperative/src/profiles/src/abstract/AbstractProfileManager.ts
```diff
@@ -233,7 +233,7 @@ export abstract class AbstractProfileManager<T extends IProfileTypeConfiguration
             this.validateConfigurationDocument(config);
         }
         this.mProfileTypeSchema = this.mProfileTypeConfiguration.schema;
-        this.mProfileTypeRootDirectory = this.createProfileTypeDirectory();
+        this.mProfileTypeRootDirectory = this.profileRootDirectory + "/" + this.profileType + "/";
```
zFernand0, line 236: Very important change here.
We still provide protected getter `profileTypeRootDirectory` but at least we are not creating the directory anymore 😋 

I guess we still need the getter for conversion purposes 😋

packages/imperative/src/profiles/src/BasicProfileManager.ts
```diff
@@ -55,77 +51,6 @@ import { ProfileIO } from "./utils";
...
-     * your Imperative configuration documents) and all profile "type" configuration documents and constructs the directories
-     * needed to manage profiles of all types. You must execute this method before beginning to use profiles OR you must
-     * supply all the type configuration documents (normally obtained from your Imperative configuration document) to
-     * the constructor of
-     * @static
-     * @param {IProfileManagerInit} parms
-     * @returns {Promise<IProfileInitialized[]>}
-     * @memberof AbstractProfileManager
-     */
-    public static async initialize(parms: IProfileManagerInit): Promise<IProfileInitialized[]> {
```
zFernand0, line 69: I think this will be impactful to Zowe Explorer (since they use it from the CliProfileManager which extends this class). 
It will be good to produce a ZE build with this vNext version after this PR is merged 😋 

Search for `ZoweExplorerExtender.initForZowe` in the Zowe Explorer codebase

zFernand0 (approved):
LGTM! 😋 

Thanks for reusing the `exitOnFailure.sh` script. It must've been painful to get the right number of `../`, so thank you! 😋 
Also, thanks for adding descriptions to most/all variables in our test `.sh` scripts 😋 

I think the only thing left to do is the very small change to the `.sh` script that Timothy suggested (to use the `baseCertFile` required parameter) 😋 

That being said, I do think this PR is ready to be merged since most/all system test worked the last time they ran (internally).

---

If possible, I would like for this PR to be merged and published on it's own.
That way Zowe Explorer can test with this individual version, unless we think the `master` update and the imperative flattening should be part of the ZE tests around zowe profiles removal.

packages/imperative/__tests__/__integration__/cmd/__tests__/integration/cli/auth/__scripts__/auth_login_cmd_line_cert.sh
gejohnston: Yes. I just forgot to use the supplied variable. Thanks.

packages/imperative/__tests__/__integration__/imperative/__tests__/__integration__/cli/config/convert-profiles/cli.imperative-test-cli.config.convert-profiles.integration.subtest.ts
gejohnston: That seems reasonable. I will switch them.

zFernand0 (approved):
Thanks for addressing Timothy's comments.

Re-approving! LGTM! :yum:

traeok (approved):
LGTM - thanks @gejohnston (and @awharn for your work on this PR)!

PR #2015, opened 2024-01-10
feat(plugins): add/remove from schema during `zowe plugins install/uninstall`

traeok:
Signed-off-by: Trae Yelovich <trae.yelovich@broadcom.com>

**What It Does**

This PR addresses the following:
- BugFix: `zowe plugins install` command no longer rebuilds the schema based on its known profile types
  - Now, the command will only add new types to the schema
- Enhancement: `zowe plugins uninstall` now removes a profile type from the schema if no other client is providing it
  - This command now checks the global `extenders.json` file to see if the profile type is contributed by other sources
  - If the only source is the plug-in, the profile type is removed from the schema during uninstallation

**How to Test**

- Using the latest release of Zowe CLI, uninstall a CLI plug-in that provides a profile type
  - Notice that the profile type was not removed during uninstall
- Then, using the CLI build from this PR:
  - Install the same plug-in providing said profile type
  - Uninstall the plug-in

With these changes, only the schema for that profile type will be removed.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog


packages/imperative/src/imperative/src/plugins/utilities/npm-interface/install.ts
```diff
@@ -12,6 +12,7 @@
 import { PMFConstants } from "../PMFConstants";
 import * as path from "path";
 import * as fs from "fs";
+import * as jsonfile from "jsonfile";
 import { readFileSync, writeFileSync } from "jsonfile";
```
zFernand0, line 16: should we remove one of the two?
I think we are using the same two methods 😋 

Or even better, could we try to use `fs` ?

Same for `uninstall`

packages/imperative/src/imperative/src/plugins/utilities/npm-interface/install.ts
```diff
@@ -134,14 +139,85 @@ export async function install(packageLocation: string, registry: string, install
...
+        if (PMFConstants.instance.PLUGIN_USING_CONFIG)
         {
             // Update the Imperative Configuration to add the profiles introduced by the recently installed plugin
             // This might be needed outside of PLUGIN_USING_CONFIG scenarios, but we haven't had issues with other APIs before
-            if (Array.isArray(pluginImpConfig.profiles)) {
+            const globalLayer = PMFConstants.instance.PLUGIN_CONFIG.layers.find((layer) => layer.global && layer.exists);
+            if (globalLayer && Array.isArray(pluginImpConfig.profiles)) {
                 UpdateImpConfig.addProfiles(pluginImpConfig.profiles);
-                ConfigSchema.updateSchema({ layer: "global" });
+                const schemaUri = new URL(globalLayer.properties.$schema, pathToFileURL(globalLayer.path));
```
zFernand0, line 149: Similar comment to other PRs, I'm always scared to use the `$schema` property directly as it could be a valid URL (without the `file://` protocol).

If possible, could we try to do something similar to the `Config#getSchemaInfo()` method?

Same goes for `uninstall`

packages/imperative/src/imperative/src/plugins/utilities/npm-interface/install.ts
```diff
@@ -134,14 +139,85 @@ export async function install(packageLocation: string, registry: string, install
...
+                                    existingType.schema = profile.schema;
+                                    existingType.schemaVersion = profile.schemaVersion;
+                                }
+                            }
+                            shouldUpdate = shouldUpdate || updateExtendersJson(profile);
+                        }
+
+                        if (shouldUpdate) {
+                            // Update extenders.json (if necessary) after installing the plugin
+                            ProfileInfo.writeExtendersJson(extendersJson);
```
zFernand0, line 214: This seems like a clever way to loop through all profile types and make sure we update the `extendersJson` object.

However, after the first time that the `extenderJson` objectis updated, the `updateExtendersJson` method won't ever be called again.

I might be reading things wrong, but don't we need to call the `updateExtendersJson` for every profile type?

Otherwise, we might be writing an incomplete `extender.json` 😋

packages/imperative/src/imperative/src/plugins/utilities/npm-interface/uninstall.ts
```diff
@@ -84,6 +88,71 @@ export function uninstall(packageName: string): void {
...
+                        iConsole.error("Error when removing profile type for plugin %s: failed to parse schema", npmPackage);
+                    }
+
+                    // Only update global schema if we were able to load it from disk
+                    if (loadedSchema != null) {
+                        if (typesToRemove.length > 0) {
+                            loadedSchema = loadedSchema.filter((typeCfg) => !typesToRemove.includes(typeCfg.type));
+                            const schema = ConfigSchema.buildSchema(loadedSchema);
+                            ConfigSchema.updateSchema({ layer: "global", schema });
+                            jsonfile.writeFileSync(schemaPath, schema, { spaces: 4 });
```
zFernand0, line 149: Nothing to be worried about in this PR, but...

Come to think of it, do we support comments in the schema files?
Not sure if we have ever made that claim before or not 😋 

I can't see us supporting that since we rebuild it very often 😋

zFernand0 (commented):
Looking good so far.

Left a few minor comments that may need to be addressed 😋

packages/imperative/src/imperative/src/plugins/utilities/npm-interface/install.ts
traeok: Ah, good catch! That's correct, I overlooked the OR short-circuiting that could happen with `shouldUpdate`. I think if I set it to `updateExtendersJson(profile) || shouldUpdate`, that should ensure that `updateExtendersJson` is updated for every type.

I will add a couple unit tests to this PR as well to ensure that all the types are added if multiple are provided 👍

traeok: Resolved in [13c1e94](https://github.com/zowe/zowe-cli/pull/2015/commits/13c1e94132a75fd7e50e0be322891d251d59b62b) 🙂

packages/imperative/src/imperative/src/plugins/utilities/npm-interface/install.ts
traeok: I removed the duplicate import in [b412e4d](https://github.com/zowe/zowe-cli/pull/2015/commits/b412e4dd2fb6d54a8856f90814235c0114470279) 👍 

I will test using `fs.write/read`, not sure if it will format the JSON properly but I'll investigate 😅

packages/imperative/src/imperative/src/plugins/utilities/npm-interface/install.ts
traeok: Resolved in 13c1e94

packages/imperative/src/imperative/src/plugins/utilities/npm-interface/install.ts
traeok: In 3f20c48, I've introduced the same `file` protocol check from #2001 to ensure we do not try to update a web URL schema 😁

packages/imperative/src/imperative/src/plugins/utilities/npm-interface/install.ts
traeok: Per my comment on #2001, going to track this as a separate enhancement since we use `jsonfile` in multiple places outside of this context 😋

packages/imperative/src/imperative/src/plugins/utilities/npm-interface/uninstall.ts
traeok: If we do support it, it's through other interfaces, as `ConfigSchema.updateSchema` just writes the schema that's built with `ConfigSchema.buildSchema`.

This might be worth tracking as a separate issue to evaluate where comments are supported in schema operations.

gejohnston (approved):
I am leaving a comment for which I am requesting **NO** changes. It is simply food for thought in future test scripts.

The plugin management facility has a long history of creating involved test-functions that result in simple test cases. The advantage is that there is only one copy of the complicated test logic. The disadvantage is that when someone reads a test case, the test really only conveys that my test-function does what I programmed it to do. It is not obvious what the real expected behavior should be. The reader has to correctly read the test-function logic to understand what the expected behavior really is.

At one point the team had a discussion about whether it might be advantageous to duplicate some logic in each test so that the test expectations were easier to understand. I do not think that any absolute conclusion was reached. I only leave this comment so that newer people are informed of such considerations and make their value judgements with awareness of past discussions.

t1m0thyj (approved):
LGTM, thanks @traeok!

Is it expected that the `extenders.json` file was generated with an empty "profileTypes" array? Profiles types for core CLI and other plugins I have installed were not added.

packages/imperative/src/imperative/src/plugins/utilities/npm-interface/install.ts
```diff
@@ -24,6 +24,40 @@ import { PluginManagementFacility } from "../../PluginManagementFacility";
...
 import { UpdateImpConfig } from "../../../UpdateImpConfig";
 import { CredentialManagerOverride, ICredentialManagerNameMap } from "../../../../../security";
+import { IProfileTypeConfiguration } from "../../../../../profiles";
+import * as semver from "semver";
+import { ProfileInfo } from "../../../../../config";
+import { IExtendersJsonOpts } from "../../../../../config/src/doc/IExtenderOpts";
+
+// Helper function to update extenders.json object during plugin install.
+// Returns true if the object was updated, and false otherwise
+const updateExtendersJson = (
```
zFernand0, line 34: Curious why this functions wasn't exported but its counterpart (in the uninstall.ts file) was exported and defined as a function?

packages/imperative/src/imperative/src/plugins/utilities/npm-interface/uninstall.ts
```diff
@@ -19,8 +19,51 @@ import { ImperativeError } from "../../../../../error";
...
+import { ConfigSchema, ProfileInfo } from "../../../../../config";
+import { IProfileTypeConfiguration } from "../../../../../profiles";
 const npmCmd = findNpmOnPath();
 
+/**
+ * Updates `extenders.json` and returns a list of types to remove from the schema, if applicable.
+ * @param npmPackage The package name for the plug-in that's being uninstalled
+ * @returns A list of types to remove from the schema
+ */
+export function updateAndGetRemovedTypes(npmPackage: string): string[] {
```
zFernand0, line 31: If the intention is to make use of this function from outside imperative, we should modify the `npm-interface/index.ts` to include this new method in the list of exports 😋

zFernand0 (approved):
Changes LGTM! 😋 

Left two questions/suggestions but nothing that should be addressed right away (unless requested in the future) 😋

packages/imperative/src/imperative/src/plugins/utilities/npm-interface/uninstall.ts
traeok: I exported this function for the sake of unit testing, but its not intended to be used outside of imperative. I've updated the style (to match the arrow function defined in `install.ts`) in commit 3ad57860.

traeok: I exported this function for the sake of unit testing, but its not intended to be used outside of imperative. I've updated the style (to match the arrow function defined in install.ts) in commit 3ad57860.

zFernand0:
Sonarcloud issues are related to cognitive complexity. 
Merging 😋

PR #2016, opened 2024-01-11
fix(profInfo): Fix secure property names being returned for wrong profile

t1m0thyj:
**What It Does**
Fixes a small bug in `ConfigSecure.securePropsForProfile` where listing secure property names for "abcdef" profile also included property names for "abc" profile

Also moves top-level methods in `ConfigUtils.ts` inside a class to conform to best practices (thanks @zFernand0 for the suggestion).

**How to Test**
* Create a project config like the one in https://github.com/zowe/vscode-extension-for-zowe/issues/2633#issuecomment-1881841883
* Run a test script like the following and verify that "zosmf-apiml" profile has no secure properties
```javascript
const { Config } = require("@zowe/imperative");
(async () => {
    const config = await Config.load("zowe", { projectDir: <projectDir> });
    for (const profName of ["zosmf", "zosmf-apiml", "base"]) {
        console.log(profName, config.api.secure.securePropsForProfile(profName));
    }
})();
```

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/src/config/index.ts
+++ b/packages/imperative/src/config/index.ts
@@ -14,7 +14,7 @@ export * from "./src/ConfigAutoStore";
 export * from "./src/ConfigConstants";
 export * from "./src/ConfigSchema";
 export * from "./src/ConfigBuilder";
-export * as ConfigUtils from "./src/ConfigUtils";
+export * from "./src/ConfigUtils";
 export * from "./src/ProfileCredentials";
 export * from "./src/ProfileInfo";
 export * from "./src/ProfInfoErr";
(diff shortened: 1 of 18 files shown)


packages/imperative/src/config/index.ts
```diff
@@ -14,7 +14,7 @@ export * from "./src/ConfigAutoStore";
 export * from "./src/ConfigConstants";
 export * from "./src/ConfigSchema";
 export * from "./src/ConfigBuilder";
-export * as ConfigUtils from "./src/ConfigUtils";
```
zFernand0, line 17: I think this is the way in which we will have to eventually deal with the Imperative package.
Little by little, and as non-breaking changes. 😋 
Thanks for moving the functions into a class 😋

packages/imperative/src/config/src/ProfileInfo.ts
zFernand0, line 1: Thanks for addressing these edge cases for the ProfileInfo APIs

zFernand0 (approved):
I learned that using `.startsWith` may not always be a good practice 😋 
LGTM! 😋

adam-wolfe (approved):

PR #2043, opened 2024-02-09
New 'zos-jobs search jobs' command

KevinLoesch1:
**What It Does**
Adds a new **zowe zos-jobs search job** command. This allows the user to search a job or jobs spool files for a specified string or regular expression.  

**How to Test**
The command requires a job name, or wildcard, and either the **--search-string** or **--search-regex** parameter. For example, to search for the string 'RC=0000' in all jobs starting with 'TESTJOB' you would use the following command:

zowe zos-jobs search job 'TESTJOB*' --search-string 'RC=0000'

**Review Checklist**
I certify that I have:
- [X] tested my changes
- [X] added/updated automated tests
- [X] updated the changelog
- [X] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)


packages/cli/src/zosjobs/-strings-/en.ts
```diff
@@ -73,6 +73,43 @@ export default {
...
+                POSITIONALS: {
+                    JOBNAME: "The job name to search. Wildcards are accepted for the job name." +
+                        " You can use, for example \"USERJOB*\" to search all jobs that start" +
+                        " with USERJOB."
+                },
+                OPTIONS: {
+                    SEARCHSTRING: "The string to search for in the spool output",
+                    SEARCHREGEX: "The regular expression to search for in the spool output",
+                    CASEINSENSITIVE: "The search is case insensitive or not",
+                    SEARCHLIMIT: "The maximum number of matching lines to return for an individual spool file",
```
adam-wolfe, line 92: Hi @KevinLoesch1, could you add a period at the end of these option definitions?

KevinLoesch1: I have checked in an update to correct this.

adam-wolfe (approved):
LGTM. Thanks @KevinLoesch1

packages/zosjobs/src/SearchJobs.ts
```diff
@@ -0,0 +1,162 @@
...
+                    }
+                }
+            }
+        }
+
+        return replyBuffer;
+    }
+
+    /**
+     * Get jobs (defaults to the user ID of the session as owner)
```
gejohnston, line 119: Are the references to 'get' (here and below) left over from some early implementation efforts, or maybe a copy/paste error? I would expect them both to refer to 'search'.

packages/zosjobs/src/SearchJobs.ts
```diff
@@ -0,0 +1,162 @@
...
+     * @returns {Promise<IRestClientResponse>} - promise that resolves to spool output and response headers
+     * @memberof searchJobs
+     */
+    private static async searchSpoolContentCommon(session: AbstractSession,
+        searchString : string,
+        searchRegex:string,
+        caseInsensitive:boolean,
+        jobFile: IJobFile,
+        startingLine: number) {
+        Logger.getAppLogger().trace("SearchJobs.getSpoolContentCommon()");
```
gejohnston, line 136: Same comment about 'get' vs 'search'.

gejohnston (changes requested):
Implementation looks good. Thanks for cleaning up all of the 'pa' parameter errors in the system tests.

I left a couple of comments about wording. If I misinterpreted those items, you can ignore my request for change.

packages/zosjobs/src/SearchJobs.ts
KevinLoesch1: These were just a cut and paste remnant. I have corrected them.

packages/zosjobs/src/SearchJobs.ts
KevinLoesch1: They were a copy/paste remnant. Fixed.

gejohnston (approved):

packages/zosjobs/src/SearchJobs.ts
```diff
@@ -0,0 +1,162 @@
...
+        const caseInsensitive: boolean = searchParms.caseInsensitive;
+        const searchLimit: number = searchParms.searchLimit;
+        const fileLimit: number = searchParms.fileLimit;
+
+        Logger.getAppLogger().info("SearchJobs.searchJobs() called!");
+        let replyBuffer:string = "";
+
+        // Validate that a search string or regex parameter was passed
+        if(searchRegex === undefined && searchString === undefined){
+            throw new ImperativeError({ msg: ZosJobsMessages.missingSearchOption.message });
```
zFernand0, line 49: Curious if we should use ImperativeExpect here (and in some of the others below)

I'm not sure if the syntax will be any better since we are talking about two parameters that should be both undefined.
That's why I'm not requesting changes with this comment 😋

packages/zosjobs/src/SearchJobs.ts
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        // Validate that both options are not passed on the same call
+        if(searchRegex !== undefined && searchString !== undefined){
+            throw new ImperativeError({ msg: ZosJobsMessages.missingSearchOption.message });
+        }
+
+        const jobsList: IJob[] = await GetJobs.getJobsByPrefix(session, jobName);
+        let fileCount = 0;
+        for(const job of jobsList )
```
zFernand0, line 59: Not saying that these weren't tested, since there are system, integration, and unit tests (for the handler and definition files)

But I have to agree with Codecov on this one. 
We should probably add one or two unit-level API tests to make sure that these lines are covered.

zFernand0 (changes requested):
Implementation LGTM! Testes included! 🥳 

I do have one small request, and I hate to ask this since the code was thoroughly tested, however there are no functional unit-level tests for the new API added. 😢 
That means that technically the API wasn't touched during our unit test phase, and since we don't gather coverage from integration tests (nor can we gather coverage for system test) I'm requesting for one or two unit-level tests to be added to touch the API code 😢 

This is one really good example that coverage shouldn't matter as much though.

packages/zosjobs/src/SearchJobs.ts
KevinLoesch1: I hope the way I worded the message makes since. I used the same message as it states that you need to specify either one or the other. If neither is specified or both are specified then it is a failing condition.

packages/zosjobs/src/SearchJobs.ts
KevinLoesch1: I can write a API test, but it isn't going to do much, as so much of this code is dependent on responses from zosmf. I can see I can mock them up. We do not have them for ViewJob either.

packages/zosjobs/src/SearchJobs.ts
zFernand0: yeah, I do think it makes sense 😋

packages/zosjobs/src/SearchJobs.ts
zFernand0: I do agree that the tests won't be doing much (compared to the rests of the tests that you already have written in this PR) 😋 

We should probably discuss how we can consider these scenarios in the future (i.e. the code has been thoroughly tested outside of the SDK/API unit-level testing

awharn (approved):

packages/zosjobs/src/SearchJobs.ts
KevinLoesch1: I've added new unit and system test. The code coverage requirements for the API should now pass.

zFernand0 (approved):

zFernand0:
Thanks for adding a couple of new unit tests for these APIs 😋

PR #2045, opened 2024-02-09
fix(ProfileInfo): ignore command-based props during schema comparison

### LINKED ISSUE #2051: Create command definition constants for schema-related updates
Is there any "constants" file that we can grab these strings from?
Otherwise, should we move these into constants to be reused elsewhere? 😋 

Or even better, should we make this array a constant of some sort that indicates the command-specific options that we are adding to the schema? 
  - Although somewhat unlikely, we can prevent having to update this in the future if we add new command-related options to the schema 😋

_Originally posted by @zFernand0 in https://github.com/zowe/zowe-cli/pull/2045#discussion_r1486864579_

traeok:
**What It Does**

Resolves #2051 

Since we allow command-based schemas to be registered through Zowe Explorer, the comparison checks between schemas need to omit any command-related properties to prevent the equality checks from failing.

Otherwise, the warning regarding "two unversioned schemas with different contents" will persist as the equality check will always fail. This PR resolves this by ignoring any command-based properties when comparing schemas.

Ideally, we would make `initForZowe` in Zowe Explorer API use the `IProfileTypeConfiguration` type instead of `ICommandProfileTypeConfiguration` - unfortunately, this would be a breaking change, so the best we can do for now is make comparisons and ignore the properties from the command-based type.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog


gejohnston (approved):

KevinLoesch1 (approved):

packages/imperative/src/config/__tests__/ProfileInfo.TeamConfig.unit.test.ts
```diff
@@ -1432,6 +1432,8 @@ describe("TeamConfig ProfileInfo tests", () => {
                 const blockMocks = getBlockMocks();
                 const profInfo = createNewProfInfo(teamProjDir);
                 await profInfo.readProfilesFromDisk({ homeDir: teamHomeProjDir });
+                jest.spyOn(Map.prototype, "has").mockReturnValue(true);
+                jest.spyOn(lodash, "isEqual").mockReturnValue(true);
```
zFernand0, line 1436: Thi is an interesting way to get to the desired code. 😋 
Glad that we have a `jest.restoreAllMocks()` call in the `afterEach()`

LGTM! :yum:

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -1348,11 +1351,29 @@ export class ProfileInfo {
...
 
+    /**
+     * This helper function removes all command-related properties from the given schema properties object and returns it.
+     * This is so we can easily compare schemas from disk with those that are registered with type ICommandProfileSchema.
+     * It's also been added to avoid a breaking change (as we currently allow ICommandProfileSchema objects to be registered).
+     * @param obj The properties object from the schema
+     * @returns The properties object, but with all of the command-related properties removed
+     */
+    private omitCmdPropsFromSchema(obj: Record<string, any>): Record<string, IProfileProperty> {
+        const result = lodash.omit(obj, ["optionDefinition", "optionDefinitions", "includeInTemplate"]);
```
zFernand0, line 1367: Is there any "constants" file that we can grab these strings from?
Otherwise, should we move these into constants to be reused elsewhere? 😋 

Or even better, should we make this array a constant of some sort that indicates the command-specific options that we are adding to the schema? 
  - Although somewhat unlikely, we can prevent having to update this in the future if we add new command-related options to the schema 😋

zFernand0 (approved):
LGTM! 😋 
left a few small suggestions but nothing that's blocking this PR from being merged 🥳

packages/imperative/src/config/src/ProfileInfo.ts
traeok: I like the idea of moving these into the constants file. I've done so in commit 7ff571b 👍

zFernand0 (approved):
LGTM! 😋

PR #2048, opened 2024-02-12
Rename --root to --name-only and update deps

### LINKED ISSUE #1797: Add --name-only as an alias for --root in zowe config list command
The meaning of the --root option is unclear.

Vit Gottwald suggested that users will typically not understand what an option named --root will accomplish.

> There is something similar in `git`. When you do a `git diff` you get a complete diff. But you can pass `--name-only` to only get a list of files. It also works on `git log -p` and other git commands dealing with commits or diffs.

> That said, `zowe config list --locations --name-only` would sound like something easy to understand, especially for anyone who knows git. Would it be possible to add `--name-only` as an alias to `--root`?

awharn:
**What It Does**

Renames --root to --name-only in `zowe config list`
Resolves technical currency
Resolves #1797 

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -11457,9 +11457,22 @@
         "node": ">=12.0.0"
       }
     },
-    "node_modules/ip": {
-      "version": "2.0.0",
-      "license": "MIT"
+    "node_modules/ip-address": {
+      "version": "9.0.5",
+      "resolved": "https://registry.npmjs.org/ip-address/-/ip-address-9.0.5.tgz",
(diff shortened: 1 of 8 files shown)


zFernand0 (approved):
LGTM! 😋 
Thanks for adding the alias and updating tests 🥳

t1m0thyj (approved):
LGTM, thanks @awharn!

gejohnston (approved):

PR #2053, opened 2024-02-15
Removing profile versioning return value from commands about environment

ATorrise:
**What It Does**
Part of the modernization effort in preparation for v3. Given v1 profiles will no longer be supported, their consideration in certain command logic is no longer necessary. 

**How to Test**
1. Append your command with `--show-inputs-only`. 
2. Test `zowe config report-env`. 
Ensure neither command returns anything related to differentiating your environment between `TeamConfig` of `v1`

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**
If there are any other commands the team can think of that return information about environment for user information that have special logic for v1, the removal of that logic could be appended to this PR.


zFernand0:
Smart to have based your branch off of the `rm-profile-handlers` branch 🥳 
Changes LGTM! 😋 

just gotta address a few failing tests 😅

gejohnston:
Thanks for going beyond just changing the text that is displayed. The removal of V1 logic in report-env is above and beyond the original issue, but a very worthwhile change.

I confirmed that your branch has the same bug fix as the original branch from which you created your branch. So you were right, your test errors are the result of some specific changes in your branch.

packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
```diff
@@ -1679,54 +1618,6 @@ describe("Command Processor", () => {
         expect(commandResponse.data.locations.length).toBeGreaterThan(0);
         expect(commandResponse.data.optionalProfiles[0]).toBe(`banana`);
         expect(commandResponse.data.requiredProfiles).toBeUndefined();
-        expect(commandResponse.data.profileVersion).toBe('TeamConfig');
-    });
-
-
-    it("should mask input value for a secure parm when --show-inputs-only flag is set", async () => {
```
t1m0thyj, line 1686: Is removing this test deliberate? It doesn't look specific to v1 profiles to me, so seems like it would be valid to keep.

packages/imperative/src/cmd/src/CommandProcessor.ts
```diff
@@ -814,28 +805,20 @@ export class CommandProcessor {
...
         /**
-         * Determine if Zowe Team Config is in effect.  If it is, then we will construct
-         * a Set of secure fields from its API.  If it is not, then we will construct
-         * a Set of secure fields from the `ConnectionPropsForSessCfg` defaults.
+         * Construct a Set of secure fields from Zowe Team Config API.
          */
-        const secureInputs: Set<string> =
-        useConfig ?
-            new Set([...configSecureProps]) :
-            new Set([...LoggerUtils.CENSORED_OPTIONS, ...LoggerUtils.SECURE_PROMPT_OPTIONS]);
```
t1m0thyj, line 837: Related to my previous comment - is this change deliberate? I think we'd want to continue censoring options defined in `LoggerUtils.CENSORED_OPTIONS` and `LoggerUtils.SECURE_PROMPT_OPTIONS`. If someone defines user/password in plain text in their config and omits it from the secure array, we may still want to hide it.

t1m0thyj (commented):
Looks pretty good, thanks @ATorrise! Left a few comments about how we handle secure inputs.

packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
zFernand0: Even though it doesn't look like it is looking for V1 profiles, the fact that it was not defining a team_config meant that it was always falling back to V1 profiles. (hint: `commandParameters.definition.profile` isn't populated by the `prepare` method)

Happy to discuss offline 😋

packages/imperative/src/cmd/src/CommandProcessor.ts
ATorrise: deliberate. from my understanding this is determining how to omit from the secure array, if useConfig that means teamConfig and to use first method, otherwise you're using v1 and need to use the latter method

zFernand0: >If someone defines user/password in plain text in their config and omits it from the secure array, we may still want to hide it.

🤔 I would not mind to see user and password in plain text (if I'm running things in plain text), and may even be slightly confused about the user and password not showing up in those scenarios

packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
ATorrise: ```it("should display input value for simple parm when --show-inputs-only flag is set and v1 profile exists", async () => {"```
yes, this test is related to setting --show-inputs-only when using v1 and my understanding is that this PR is about migrating away from v1 cases meaning this situation will no longer be present in v3

zFernand0 (approved):
LGTM! 😋

packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
t1m0thyj: I don't see any test with that name in [this file](https://github.com/zowe/zowe-cli/blob/show-inputs-only/packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts)

packages/imperative/src/cmd/src/CommandProcessor.ts
t1m0thyj: This seems like potentially a breaking change, but I understand the reasoning behind it if we believe that users should have control over which fields are secure. 👍  For v1 profiles there was no equivalent to the secure array so we had to hardcode a list of options, but now we don't anymore 😋

packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
ATorrise: and team config exists****

gejohnston (approved):

packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
ATorrise: ` it("should mask input value for a secure parm when --show-inputs-only flag is set", async () => {` should resolve this conversation. sorry for confusion

packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
```diff
@@ -26,6 +26,7 @@ import { ImperativeConfig } from "../../utilities/src/ImperativeConfig";
 import { setupConfigToLoad } from "../../../__tests__/src/TestUtil";
 import { EnvFileUtils } from "../../utilities";
 import { join } from "path";
+import { boolean } from "yargs";
```
t1m0thyj, line 29: Please fix the lint warning for unused import

packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
```diff
@@ -2423,67 +2429,4 @@ describe("Command Processor", () => {
         });
     });
 
-    describe("profiles", () => {
```
t1m0thyj, line 2426: Is removing these tests deliberate? They appear to have been using v2 team config. Also removing them seems to have dropped our coverage a bit:
> Comparison is base [(`fa037d4`)](https://app.codecov.io/gh/zowe/zowe-cli/commit/fa037d4dd4e43899036f9815e0c793e342ace889?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=zowe) 90.83% compared to head [(`c76c93b`)](https://app.codecov.io/gh/zowe/zowe-cli/pull/2053?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=zowe) 90.70%.

t1m0thyj (changes requested):
Thanks for updating the secure values test to handle team config 👍  Left a few comments.

packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
zFernand0: Sorry, I might have missed this test before approving
or maybe it happened after `¯\_(ツ)_/¯`

I agree that we should probably keep this one

ATorrise:
@gejohnston Are these failing tests acceptable or should I do something about them?

gejohnston:
> Are these failing tests acceptable or should I do something about them?

@ATorrise I think that we should merge this branch into rm-profile-handlers. I think that the error will not exist in the rm-profile-handlers branch. Even if there is a problem, it will be easier to fix in that branch.

packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
```diff
@@ -2329,6 +2334,7 @@ describe("Command Processor", () => {
                 envVariablePrefix: ENV_VAR_PREFIX,
                 definition: SAMPLE_COMMAND_REAL_HANDLER_WITH_OPT,
                 helpGenerator: FAKE_HELP_GENERATOR,
+                profileManagerFactory: FAKE_PROFILE_MANAGER_FACTORY,
```
t1m0thyj, line 2337: I think we may need to remove this line now to fix the failing unit tests?

t1m0thyj (approved):
LGTM if we want to fix the tests in another branch after merge 🙂 

Left a comment about why the test may be failing

PR #2062, opened 2024-02-22
standardizing error handling

ATorrise:
**What It Does**
Brings standard error handling from zosjobs to zosfiles

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


packages/cli/CHANGELOG.md
```diff
@@ -73,22 +76,6 @@ LTS Breaking: Removed the following previously deprecated items: [#1981](https:/
 
 - Major: First major version bump for V3
 
-## `7.23.3`
```
gejohnston, line 76: I do not think that we will want recent V2 changelog entries to be deleted. They should remain.

gejohnston (changes requested):
I suggested a modification in the changelog.

gejohnston:
I am a little confused about the objective of the story itself.

A user specifies a non-existent file and the error message implies that the command has an internal logic failure. It blames the command handler for the error and displays a stack trace.

Should we instead capture the root error message 'no such file or directory' and display it in a format similar to the more user friendly format utilized for REST request errors that we used in a pull request several months ago?

gejohnston:
Note that the current full zos-files error message is:

```
zowe files upload ftds local remote
Node.js File System API error

Response From Service
Error: ENOENT: no such file or directory, lstat 'C:\ourstuff\records\Broadcom\zowe\tech_info\config_files\local'

Diagnostic Information
Error: ENOENT: no such file or directory, lstat 'C:\ourstuff\records\Broadcom\zowe\tech_info\config_files\local'
```

The format above was created in response to another issue about unfriendly error messages.

Perhaps we should make the zos-jobs messages adhere to the format used by zos-files?

adam-wolfe:
After looking at this some more it seems like the best thing to do would be to not show a stack trace for the file not found error. Currently, in the latest versions of V2 and V3, the `jobs submit lf` command still displays a stack trace. I think we would want the jobs command failure to match the `zowe files upload` error message format.

t1m0thyj:
Closing in favor of #2078

PR #2077, opened 2024-03-05
Update prompting logic if port is 0

### LINKED ISSUE #2075: Prompt if Port in Team Config is 0
With Zowe V3, some plug-ins include `port` in the template. Without a default value, it fills the port value with 0, which is invalid. This behavior stops prompting because a port is now specified. If the port is 0, which is both the `includeInTemplate` default and invalid, it should be ignored and prompted for

Prompt for port if port is supplied as 0.
Prompt for host if host is supplied as "".

- Leave port as 0 and expect the end user to change it once the connection fails.

awharn:
**What It Does**

Prompts the user for port if the port retrieved from the user's config is 0 (default for includeInTemplate, invalid)
Resolves #2075 

**How to Test**

Install, set port to 0 in config for a service, observe prompting

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
+++ b/packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
@@ -135,11 +135,13 @@ export class ConnectionPropsForSessCfg {
         }
 
         // check what properties are needed to be prompted
-        if (ConnectionPropsForSessCfg.propHasValue(sessCfgToUse.hostname) === false && !doNotPromptForValues.includes("hostname")) {
+        if ((ConnectionPropsForSessCfg.propHasValue(sessCfgToUse.hostname) === false || sessCfgToUse.hostname === "") &&
+            !doNotPromptForValues.includes("hostname")) {
             promptForValues.push("hostname");
         }
 
-        if (ConnectionPropsForSessCfg.propHasValue(sessCfgToUse.port) === false && !doNotPromptForValues.includes("port")) {
+        if ((ConnectionPropsForSessCfg.propHasValue(sessCfgToUse.port) === false || sessCfgToUse.port === 0) &&
+            !doNotPromptForValues.includes("port")) {
             promptForValues.push("port");
         }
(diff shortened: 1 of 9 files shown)


anaxceron (approved):

adam-wolfe (approved):

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -139,7 +139,8 @@ export class ConnectionPropsForSessCfg {
             promptForValues.push("hostname");
         }
 
-        if (ConnectionPropsForSessCfg.propHasValue(sessCfgToUse.port) === false && !doNotPromptForValues.includes("port")) {
+        if ((ConnectionPropsForSessCfg.propHasValue(sessCfgToUse.port) === false || sessCfgToUse.port === 0) &&
+            !doNotPromptForValues.includes("port")) {
```
zFernand0, line 143: clever way to prompt for port 🥳 

Question:
How would Zowe Explorer react to this new prompt item?

Curious if you could share a screenshot or something like that 😋 

Happy to approve afterwards this ⬆️

awharn: Zowe Explorer doesn't use AddPropsOrPrompt as far as Trae and I are aware. As for ZE, its current behavior is such that if a port is 0 in the team config, it tries using port 443 (for z/OSMF at least).

awharn: For the current CICS plug-in, a value of 0 results in an error.
image

zFernand0: thanks for providing more details around ZE (and related apps) 😋

zFernand0 (approved):
forgot to approve 😅

PR #2078, opened 2024-03-05
Fnf error reporting

ATorrise:
**What It Does**
Makes zosjob command errors more legible

**How to Test**
- In vscode debugger: 
```
    {
        "type": "node",
        "request": "launch",
        "name": "FNF messaging",
        "program": "${workspaceFolder}/zowe-cli/packages/cli/lib/main.js",
        "console": "integratedTerminal",
        // "args": ["files", "upload", "dtp", "src", "ibmuser.src"],
        "args": ["jobs", "submit", "lf", "noexist"],
        // "args": ["files", "upload", "ftds", "local", "remote"],
        "outputCapture": "std",
        "resolveSourceMapLocations": null
    },
```
- compare output via command line: 
`zowe jobs submit lf "noexist"`
`zowe files upload ftds local "remote"`

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


packages/cli/src/zosjobs/submit/Submit.shared.handler.ts
```diff
@@ -31,150 +31,158 @@ export default class SharedSubmitHandler extends ZosmfBaseHandler {
...
+                    try {
+                        const JclString = fs.readFileSync(this.mArguments.localFile).toString();
+                        apiObj = await SubmitJobs.submitJclString(this.mSession, JclString, parms);
+                        source = this.mArguments.localFile;
+                        if (parms.viewAllSpoolContent) {
+                            spoolFilesResponse = apiObj;
+                        }
+                    } catch (err) {
+                        throw new ImperativeError({msg: err});
+                    }
```
t1m0thyj, line 129: If `SubmitJobs.submitJclString` throws an error, then "recasting" the error in the catch block may lose some useful error details. So I don't think we want a "blanket" try-catch here.

Instead you might want to wrap only the first line in a try-catch, to handle if the `fs.readFileSync` call throws an I/O error. This would match the error handling of the `files upload` command that Gene mentioned 😋 
```
try {
    const JclString = fs.readFileSync(this.mArguments.localFile).toString();
} catch (err) {
    throw new ImperativeError({
        msg: ZosFilesMessages.nodeJsFsError.message,
        additionalDetails: err.toString(),
        causeErrors: err
    });
}
```

Note: We'll need to copy over the "nodeJsFsError" message from ZosFilesMessages to ZosJobsMessages if we want to use it here in the z/OS Jobs SDK 😋

packages/cli/src/zosjobs/submit/Submit.shared.handler.ts
```diff
@@ -31,150 +31,158 @@ export default class SharedSubmitHandler extends ZosmfBaseHandler {
...
-            params.response.console.log(`Successfully downloaded output to ${directory}/${apiObj.jobid}`);
+            // Print path where spool content was downloaded
+            if (directory != null && spoolFilesResponse == null) {
+                directory = directory.includes("./") ? directory : `./${directory}`;
+                params.response.console.log(`Successfully downloaded output to ${directory}/${apiObj.jobid}`);
+            }
+            params.response.progress.endBar();
+            this.data.setMessage(`Submitted JCL contained in "${sourceType}": "${source}"`);
+        }catch (err){
+            throw new ImperativeError({msg: err?.mMessage?.message ?? err?.message ?? err?.stderr, causeErrors: err?.causeErrors});
```
t1m0thyj, line 185: If the error is already an instance of ImperativeError, it's probably good to avoid "recasting" it so we don't lose any useful info stored on the object.

To avoid this, I think inside the catch block you could use code similar to what @awharn recently added in the DB2 plugin [here](https://github.com/zowe/zowe-cli-db2-plugin/blob/2ba3155f9e00b2304e5159b82559d2392ff2ae23/src/cli/DB2BaseHandler.ts#L38-L44):
```
if (err instanceof ImperativeError){
    throw err;
} else {
    throw new ImperativeError({msg: err.message, causeErrors: err});
}
```

t1m0thyj (commented):
Looks pretty good, thanks @ATorrise! Left a few comments around how we could make the error details look more similar to output of `files upload` command.

gejohnston (commented):
Thanks for reworking this and reversing the behavior that was originally requested in the issue.

After Timothy's comments are addressed, I would approve the PR.

packages/cli/src/zosjobs/submit/Submit.shared.handler.ts
ATorrise: good suggestions thank you. these suggestions maintained the diagnostic information: 
image

packages/cli/src/zosjobs/submit/Submit.shared.handler.ts
```diff
@@ -31,150 +31,164 @@ export default class SharedSubmitHandler extends ZosmfBaseHandler {
...
+                // Submit the JCL from a local file
+                case "local-file": {
+                    parms.jclSource = this.mArguments.localFile;
+                    try {
+                        const JclString = fs.readFileSync(this.mArguments.localFile).toString();
+                        apiObj = await SubmitJobs.submitJclString(this.mSession, JclString, parms);
+                        source = this.mArguments.localFile;
+                    } catch (err) {
+                        throw new ImperativeError({
+                            msg: ZosFilesMessages.nodeJsFsError.message,
```
t1m0thyj, line 126: Please copy the definition for `nodeJsFsError` and add it to ZosJobsMessages so it can be imported from there instead.

packages/cli/src/zosjobs/submit/Submit.shared.handler.ts
```diff
@@ -31,150 +31,164 @@ export default class SharedSubmitHandler extends ZosmfBaseHandler {
...
+                directory = directory.includes("./") ? directory : `./${directory}`;
+                params.response.console.log(`Successfully downloaded output to ${directory}/${apiObj.jobid}`);
+            }
+            params.response.progress.endBar();
+            this.data.setMessage(`Submitted JCL contained in "${sourceType}": "${source}"`);
+        }catch (err){
+            if (err instanceof ImperativeError){
+                throw err;
+            } else {
+                throw new ImperativeError({msg: err?.mMessage?.message ?? err?.message ?? err?.stderr, causeErrors: err?.causeErrors});
```
t1m0thyj, line 190: I don't think we need all the `??` conditions here. If the `else` block is entered, then the error being thrown is not an ImperativeError so `mMessage` won't be defined. And all standard errors should have the `message` property defined. So just this should work:
`throw new ImperativeError({msg: err.message, causeErrors: err});`

packages/cli/src/zosjobs/submit/Submit.shared.handler.ts
```diff
@@ -31,150 +31,164 @@ export default class SharedSubmitHandler extends ZosmfBaseHandler {
...
-                if (parms.viewAllSpoolContent) {
-                    spoolFilesResponse = apiObj;
-                }
+                    break;
+                // Submit the JCL from a local file
+                case "local-file": {
+                    parms.jclSource = this.mArguments.localFile;
+                    try {
+                        const JclString = fs.readFileSync(this.mArguments.localFile).toString();
+                        apiObj = await SubmitJobs.submitJclString(this.mSession, JclString, parms);
```
t1m0thyj, line 122: Please move this line and the following one outside of (below) the `try-catch` block. We only want to show an FS error if the `fs.readFileSync` call fails. REST API errors thrown by the `SubmitJobs` API can be handled by the outer try-catch block.

packages/cli/__tests__/zosjobs/__integration__/submit/local-file/cli.zos-jobs.submit.local-file.integration.test.ts
```diff
@@ -40,7 +40,6 @@ describe("zos-jobs submit local-file command", () => {
         const response = runCliScript(__dirname + "/__scripts__/submit_invalid_local_file.sh",
             TEST_ENVIRONMENT);
         expect(response.status).toBe(1);
-        expect(response.stderr.toString().toLowerCase()).toContain("error");
```
t1m0thyj, line 43: Looking into why this fails and the snapshot changed, will post an update if I figure it out 🙂

t1m0thyj (changes requested):

packages/cli/__tests__/zosjobs/__unit__/submit/Submit.shared.handler.unit.test.ts
```diff
@@ -585,7 +612,7 @@ describe("submit shared handler", () => {
 
             expect(error).toBeUndefined();
             expect(SubmitJobs.submitJclString).toHaveBeenCalledTimes(1);
-            expect(copy.response.console.log).toHaveBeenCalledTimes(4);
+            // expect(copy.response.console.log).toHaveBeenCalledTimes(4);
```
t1m0thyj, line 615: Looking into why this fails and the snapshot changed, will post an update if I figure it out 🙂

packages/cli/src/zosjobs/submit/Submit.shared.handler.ts
```diff
@@ -31,150 +31,164 @@ export default class SharedSubmitHandler extends ZosmfBaseHandler {
...
 
-                break;
-            // Submit the JCL from a local file
-            case "local-file": {
-                parms.jclSource = this.mArguments.localFile;
-                const JclString = fs.readFileSync(this.mArguments.localFile).toString();
-                apiObj = await SubmitJobs.submitJclString(this.mSession, JclString, parms);
-                source = this.mArguments.localFile;
-                if (parms.viewAllSpoolContent) {
-                    spoolFilesResponse = apiObj;
```
t1m0thyj, line 123: Please add back this `if` block after the `try-catch` 😋  This should fix the failing test

packages/cli/src/zosjobs/submit/Submit.shared.handler.ts
```diff
@@ -31,150 +31,167 @@ export default class SharedSubmitHandler extends ZosmfBaseHandler {
...
+                            msg: ZosJobsMessages.nodeJsFsError.message,
+                            additionalDetails: err.toString(),
+                            causeErrors: err
+                        });
+                    }
+                    if (parms.viewAllSpoolContent) {
+                        spoolFilesResponse = apiObj;
+                    }
+                    apiObj = await SubmitJobs.submitJclString(this.mSession, JclString, parms);
+                    source = this.mArguments.localFile;
```
t1m0thyj, line 134: The `apiObj` definition needs to be moved above the `if` statement so that we don't assign an undefined value to `spoolFilesResponse`

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Removing stack trace for zosjobs errors. [#2078](https://github.com/zowe/zowe-cli/pull/2078)
```
t1m0thyj, line 7: For the changelog, do we want to mention something about fixing prompting behavior for user/password on SSH commands? Will defer to others reviewing this PR 😋

packages/cli/src/zosjobs/submit/Submit.shared.handler.ts
t1m0thyj: After making this change, you should be able to uncomment line 615 in the unit test and have it pass 🙂

packages/cli/__tests__/zosjobs/__unit__/submit/Submit.shared.handler.unit.test.ts
t1m0thyj, line 615: Should this line of the test be restored?

ATorrise: yes sorry

npm-shrinkwrap.json
```diff
@@ -17773,6 +17773,9 @@
       "engines": {
         "node": ">=18.12.0"
       },
+      "engines": {
+        "node": ">=18.12.0"
+      },
```
zFernand0, line 17778: a bit confused as to why this was added twice here 😋

zFernand0 (approved):
changes do make sense!
LGTM! 😋

Issue #1031, opened 2021-06-16
Unexpected command error when using SSH


JirkaAichler:
I am getting the following error that tells me to contact the support:

```
λ zowe zos-uss issue ssh "echo test"
Unexpected Command Error:
Please review the message and stack below.
Contact the creator of handler:
"C:\Users\ja628455\AppData\Roaming\npm\node_modules\@zowe\cli\lib\zosuss\issue\ssh/Ssh.handler"
Message:
Invalid username
Stack:
Error: Invalid username
    at Client.connect (C:\Users\ja628455\AppData\Roaming\npm\node_modules\@zowe\cli\node_modules\ssh2\lib\client.js:192:11)
    at C:\Users\ja628455\AppData\Roaming\npm\node_modules\@zowe\cli\node_modules\@zowe\zos-uss-for-zowe-sdk\lib\Shell.js:102:18
    at new Promise (<anonymous>)
    at Function.executeSsh (C:\Users\ja628455\AppData\Roaming\npm\node_modules\@zowe\cli\node_modules\@zowe\zos-uss-for-zowe-sdk\lib\Shell.js:33:25)
    at Handler.<anonymous> (C:\Users\ja628455\AppData\Roaming\npm\node_modules\@zowe\cli\lib\zosuss\issue\ssh\Ssh.handler.js:38:57)
    at Generator.next (<anonymous>)
    at C:\Users\ja628455\AppData\Roaming\npm\node_modules\@zowe\cli\lib\zosuss\issue\ssh\Ssh.handler.js:18:71
    at new Promise (<anonymous>)
    at __awaiter (C:\Users\ja628455\AppData\Roaming\npm\node_modules\@zowe\cli\lib\zosuss\issue\ssh\Ssh.handler.js:14:12)
    at Handler.processCmd (C:\Users\ja628455\AppData\Roaming\npm\node_modules\@zowe\cli\lib\zosuss\issue\ssh\Ssh.handler.js:31:16)
```

**Steps to reproduce:**
My profiles do not contain a user name and password but contain an APIML JWT token.

Base:
```
  name:     sys1 (default)
  contents:
    host:               sys1.com
    port:               1443
    rejectUnauthorized: true
    tokenType:          apimlAuthenticationToken
    tokenValue:         1234
```

SSH:
```
  name:     sys1 (default)
  contents:
    host: sys1.com
    port: 22
```

```
λ zowe --version
6.32.0
```

MarkAckert:
[...]

PR #2081, opened 2024-03-06
Ssh error

ATorrise:
**What It Does**
Fixes https://github.com/zowe/zowe-cli/issues/1031 :
- enables addition of user/password to ssh profile (which will never support apiml) when apimltoken is present in base
- provides better error message for debugging connectivity to plugins. not possible to be precise as plugins manage themselves differently and it will not always be clear why connectivity is unsuccessful. Failure could be due to any of the 3 options when an apiml token is present: 
  - service failure
  - apiml is not supported by that plugin (not easy for us to programatically discern)
  - invalid/expired credentials or properties

**How to Test**
- For vscode launch.json
```
    {
        "type": "node",
        "request": "launch",
        "name": "ssh error",
        "program": "${workspaceFolder}/zowe-cli/packages/cli/lib/main.js",
        "console": "integratedTerminal",
        "args": ["zos-uss", "issue", "ssh", "echo test"],
        "outputCapture": "std",
        "resolveSourceMapLocations": null
    },
```
- Regular command:
`zowe zos-uss issue ssh "echo test"`

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**
Handling connectivity errors with other plugins that aren't yet supported by apiml should be a topic for discussion amongst team (ie: CICS and MQ)

 

Result of PR: 
image

New error handling for other plugins apiml doesn't support AND that have yet to implement `supportedAuthTypes?: AUTH_TYPE_CHOICES[]`
image


t1m0thyj (changes requested):
To fix #1031 I'm not sure that the error message needs to be changed. I believe for SSH commands we wanted to prompt for user/password if they are missing instead of showing the error (since SSH doesn't support token authentication).

packages/cli/__tests__/zosuss/__system__/issue/cli.issue.ssh.system.test.ts
t1m0thyj, line 1: Thanks for adding this system test to ensure that SSH commands show a nice error message when credentials are invalid. 👍

packages/cli/src/zosuss/issue/ssh/Ssh.handler.ts
t1m0thyj, line 1: I think changes in this file can be reverted, because the SSH base handler already has logic for formatting error messages nicely. I tested reverting these changes locally and the system test above still passes 🙂

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Provide more legible errors to user when they are missing user/password credentials while
+running `zowe zos-uss issue ssh` commands. [#2081](https://github.com/zowe/zowe-cli/pull/2081)
```
t1m0thyj, line 8: Please update the changelog since error formatting wasn't changed in this PR. The fix for CLI is to prompt for user/password on SSH commands when a token is stored in the config.

packages/zosuss/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Zowe z/OS USS SDK package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Provide more legible errors to user when they are missing user/password credentials while
```
t1m0thyj, line 7: Please update changelog (same comment as above 🙂)

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -144,8 +144,11 @@ export class ConnectionPropsForSessCfg {
             promptForValues.push("port");
         }
 
-        if (ConnectionPropsForSessCfg.propHasValue(sessCfgToUse.tokenValue) === false &&
-            ConnectionPropsForSessCfg.propHasValue(sessCfgToUse.cert) === false) {
+        const isTokenIrrelevant = ConnectionPropsForSessCfg.propHasValue(sessCfgToUse.tokenValue) === false ||
+            (connOpts.supportedAuthTypes && !connOpts.supportedAuthTypes.includes("token"));
```
awharn, line 148: Would it be practical to change these strings to their defined constants in SessConstants.ts, like AUTH_TYPE_TOKEN? That way, if we choose to change things in the future, it can be in one place.

awharn (commented):
Just one small comment up for discussion. This implementation looks great! 🚀

t1m0thyj (approved):
LGTM, thanks @ATorrise!

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -269,7 +272,11 @@ export class ConnectionPropsForSessCfg {
             // }
         }
 
-        if (ConnectionPropsForSessCfg.propHasValue(sessCfg.tokenValue)) {
+        const isTokenUsed = ConnectionPropsForSessCfg.propHasValue(sessCfg.tokenValue) &&
+            (connOpts.supportedAuthTypes == null || connOpts.supportedAuthTypes.includes("token"));
+        const isCertUsed = ConnectionPropsForSessCfg.propHasValue(sessCfg.cert) &&
+            (connOpts.supportedAuthTypes == null || connOpts.supportedAuthTypes.includes("cert-pem"));
```
zFernand0, line 278: There are other instances of Andrew's commnet 😋

packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -64,7 +64,11 @@ export abstract class SshBaseHandler implements ICommandHandler {
         }];
         const sshSessCfg: ISshSession = SshSession.createSshSessCfgFromArgs(commandParameters.arguments);
         const sshSessCfgWithCreds = await ConnectionPropsForSessCfg.addPropsOrPrompt<ISshSession>(
-            sshSessCfg, commandParameters.arguments, {parms: commandParameters, propertyOverrides: sshSessCfgOverride}
+            sshSessCfg, commandParameters.arguments, {
+                parms: commandParameters,
+                propertyOverrides: sshSessCfgOverride,
+                supportedAuthTypes: ["basic"]
```
zFernand0, line 70: Should we use the `SessConstants` here as well?
`AUTH_TYPE_BASIC`

zFernand0 (changes requested):
LGTM! 😋 
left a small comment re: Andrew's idea 😋

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
ATorrise: oof thank you sorry for oversight

zFernand0 (approved):
thanks for addressing the comment so quickly! 🥳 
LGTM! 😋

awharn (approved):

PR #2087, opened 2024-03-12
fix(ProfileInfo): update global schema even if project config is present

### LINKED ISSUE #2086: `ProfileInfo` does not update global schema when a project-level config is detected
**Describe the bug**

If a user has a workspace folder opened containing a project-level configuration, and a VS Code extension is installed, `ProfileInfo.addProfileTypeToSchema` does not update the global schema as expected.

_(originally reported by @FrankSu1996 in issue [zowe-explorer-vscode#2773](https://github.com/zowe/zowe-explorer-vscode/issues/2773))_

**Expected:** The profile type and its metadata are added to the global schema even if a project-level config is found.
**Actual:** The schema is not updated.

A workaround is to close the workspace folder, delete `extenders.json` from your Zowe home directory and then try to register the extension again in VS Code.

**Describe your environment**

**Additional context**

traeok:
**What It Does**

Resolves #2086 

**How to Test**

This one is a bit difficult to test on the CLI side because it relies on information from the source that initialized the ProfileInfo instance.

1. Create an instance of `ProfileInfo` and call `readProfilesFromDisk`, specifying both a global dir and project dir
2. While in a directory containing a project-level config and schema, call `ProfileInfo.addProfileTypeToSchema` to add a new profile type to the global schema
3. Notice that the global schema will still be updated even though you are in a directory with a project-level config.

You can also test this through Zowe Explorer, but it requires a bit more setup:
- Run `npm run build && npm pack` in the imperative folder
- Set a resolution in Zowe Explorer's root `package.json` to resolve the local tarball containing the fixed Imperative
- Run `yarn fresh-clone && yarn && yarn build`
- Start debugging Zowe Explorer once the build is complete
- In the extension host, open a workspace folder containing a project-level config file
- Install the FTP extension for Zowe Explorer
- Notice that the global schema is updated with the `zftp` profile type even though you have a workspace folder open (containing a proj. config)

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Fixed issue where `ProfileInfo.addProfileTypeToSchema` did not update the global schema if a project-level configuration was detected. [#2086](https://github.com/zowe/zowe-cli/issues/2086)
+
 ## `5.22.4`
 
 - BugFix: Fixed race condition in `config convert-profiles` command that may fail to delete secure values for old profiles
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
-- BugFix: Fixed issue where `ProfileInfo.addProfileTypeToSchema` did not update the global schema if a project-level configuration was detected. [#2086](https://github.com/zowe/zowe-cli/issues/2086)
+- BugFix: Fixed issue where the `ProfileInfo.addProfileTypeToSchema` function did not update the global schema if a project-level configuration was detected. [#2086](https://github.com/zowe/zowe-cli/issues/2086)
 
 ## `5.22.4`
 
--- a/packages/imperative/src/config/src/ProfileInfo.ts
+++ b/packages/imperative/src/config/src/ProfileInfo.ts
@@ -1338,8 +1338,10 @@ export class ProfileInfo {
         const sameSchemaExists = this.mProfileSchemaCache.has(cacheKey) && lodash.isEqual(transformedSchemaProps, transformedCacheProps);
         // Update the cache with the newest schema for this profile type
         this.mProfileSchemaCache.set(cacheKey, schema);
+        const schemaUri = new url.URL(layer.properties.$schema, url.pathToFileURL(layer.path));
+        const schemaPath = url.fileURLToPath(schemaUri);
 
-        if (!fs.existsSync(layer.path)) {
+        if (!fs.existsSync(schemaPath)) {
             this.mImpLogger.trace(
                 "ProfileInfo.updateSchemaAtLayer returned false: the schema does not exist on disk for this layer."
             );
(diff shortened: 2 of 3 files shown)


traeok:
Should be ready for reviews now - I had to enhance one of the return checks to support updating the global schema while a project dir is open.

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Fixed issue where `ProfileInfo.addProfileTypeToSchema` did not update the global schema if a project-level configuration was detected. [#2086](https://github.com/zowe/zowe-cli/issues/2086)
```
anaxceron, line 7: Can we just add a definition for `ProfileInfo.addProfileTypeToSchema`? Would be better if the note read something like " ... issue where the `ProfileInfo.addProfileTypeToSchema` property/profile/biscuit/whatever did not update ..."

anaxceron (approved):
Left a minor note but other than that, looks good.

packages/imperative/CHANGELOG.md
traeok: Sorry, I should be more specific. I know a similar discussion came up for a changelog in a previous PR.
Thank you for catching this!

gejohnston (approved):

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -1469,11 +1455,58 @@ export class ProfileInfo {
...
+            }
+            wasGlobalUpdated = this.addToGlobalSchema(profileType, typeInfo).success;
+        } else {
+            result = this.addToGlobalSchema(profileType, typeInfo);
+            wasGlobalUpdated = result.success;
+        }
+
+        // If we could not add to the schema, return early
+        if (!result.success && !wasGlobalUpdated) {
+            return result;
```
zFernand0, line 1505: Could you help me understand why we need to return early?

it would seems that if we can't update the global layer's schema (i.e. `wasGlobalUpdated === false`) we won't attempt to update the extenders.json file 🐶

zFernand0 (approved):
I like the clean implementation of `updateProjectSchema`! 
LGTM! 😋

packages/imperative/src/config/src/ProfileInfo.ts
traeok: > it would seem that if we can't update the global layer's schema (i.e. wasGlobalUpdated === false), we won't attempt to update the extenders.json file 🐶

yep, that's correct - if `wasGlobalUpdated` is false, it means one of two scenarios occurred:
- the layer does not exist
- the layer exists, but the schema does not exist on disk

In both cases, we don't want to update `extenders.json` because the schema was left unchanged, and `extenders.json` represents the state of the global schema. For the other scenarios, `wasGlobalUpdated` will be set to `true` and it will then update `extenders.json` as expected 😋 (but only if the contents of `extender.json` have changed)

traeok: Ah, I'm being dumb - I made a truth table to verify the behavior, and confirmed that I can remove this check since it will have the same effect without it. I'll remove it. 😋 Thanks for catching that!!

traeok: Resolved in 5200c18

PR #2104, opened 2024-03-27
Refactor handling of invalid ds member names

t1m0thyj:
**What It Does**

Improves the way that invalid ds member names are handled. Before we tried to parse the invalid member names and replace control characters with �, which was prone to edge case problems. Now we stop parsing members after the first invalid member name is encountered, and add a fake member entry at the end that says "... N more members" for the ones we are unable to display.

**How to Test**

Run against our mainframe test system: `zowe files ls am {Timothy's PMF key}.COPYBOOK.UNTERSED`
This data set contains thousands of members with special characters in the name 🙂 

**Review Checklist**

I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**


packages/zosfiles/src/methods/list/List.ts
```diff
@@ -77,35 +74,23 @@ export class List {
...
+            } catch (err) {
+                const match = /in JSON at position (\d+)/.exec(err.message);
+                if (match != null) {
+                    // Remove invalid member names from end of list and try to parse again
+                    const lineNum = data.slice(0, parseInt(match[1])).split("\n").length - 1;
+                    const lines = data.trim().split("\n");
+                    lines[lineNum - 1] = lines[lineNum - 1].replace(/,$/, "");
+                    lines.splice(lineNum, lines.length - lineNum - 1);
+                    response = JSONUtils.parse(lines.join("\n"));
+                    response.items.push({ "member": `… ${response.returnedRows - response.items.length} more members` });
```
t1m0thyj, line 90: @anaxceron What do we want the message to say to make it clear there was an error loading these members? Something like "N invalid members" or "N members with errors"?

The output will look something like this:

```
$ zowe files list all-members IBMUSER.PDS
MEMBER1
MEMBER2
...3 invalid members
```

packages/zosfiles/__tests__/__unit__/methods/list/List.unit.test.ts
```diff
@@ -173,15 +173,18 @@ describe("z/OS Files - List", () => {
...
             expectStringSpy.mockResolvedValueOnce(`{"items":[\n` +
                 memberNames.map((memName) => `  {"member":"${memName}"}`).join(",\n") + `\n` +
                 `],"returnedRows":${memberNames.length},"JSONversion":1}`);
 
             const expectedListApiResponse = {
-                items: memberNames.map((memName) => ({ member: memName.replace((List as any).CONTROL_CHAR_REGEX, "\ufffd") })),
+                items: expect.arrayContaining([
+                    { member: "m1" },
+                    { member: "m2" },
+                    { member: expect.stringMatching(/… (\d+) more members/) }
```
awharn, line 186: Can this break --rfj output?

awharn: Alternatively, should we consider not adding this as an additional member (since it isn't technically a valid member), and trigger logic to warn if the list of members doesn't match the expected number of members that is returned in the `returnedRows` property?

packages/zosfiles/src/methods/list/List.ts
anaxceron: Just replying to say that we discussed this in a Google Chat and agreed on the message. Thank you!

anaxceron (approved):

gejohnston (approved):

traeok (approved):
LGTM - thanks @t1m0thyj

PR #2108, opened 2024-04-02
Update next branch with changes from master

t1m0thyj:
Ports #2104 - Refactor handling of invalid ds member names


zFernand0 (approved):
Seems like we were a bit behind on porting changed. 😢 
Thanks for updating though 😋

PR #2118, opened 2024-04-12
MVP: Imperative Event Emitter

zFernand0:
**What It Does**

- Introduce client event emission capabilities

**How to Test**

You may need to create a dummy app (e.g. VSCE) in order to test this 😋 
( or use ~`fernando's example VSCE`~ - amber's comment )

UPDATE:
Here is a public instance of the sample VSCE
- https://github.com/zowe/zowe-cli-sample-scripts/tree/add-sample-vsce/client-event-handling/sample-vsce

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

TODO: 
- [ ] Integration/system testing
- [ ] Documentation of the new behavior
- [x] Outline of future plans (see `packages/imperative/src/events/src/ImperativeEventConstants.ts`)

<details>
<summary>Expand this section for a complete list of future enhancements</summary>

  ```
   The following list of event types will only be implemented upon request
   
    Shared events:
      Global:
         - $ZOWE_CLI_HOME/.events/onConfigChanged
         - $ZOWE_CLI_HOME/.events/onSchemaChanged
      Project:
         - $ZOWE_CLI_HOME/.events/<project-hash-based-on-path>/onConfigChanged
         - $ZOWE_CLI_HOME/.events/<project-hash-based-on-path>/onSchemaChanged
    User events:
      Global:
         - ~/.zowe/.events/onUserConfigChanged
      Project:
         - ~/.zowe/.events/<project-hash-based-on-path>/onUserConfigChanged
    Custom events:
      Shared:
        Global:
         - $ZOWE_CLI_HOME/.events/<hash-based-on-app-name>/<event-id>
        Project:
         - $ZOWE_CLI_HOME/.events/<hash-based-on-app-name>/<project-hash-based-on-path>/<event-id>
      User:
        Global:
         - ~/.zowe/.events/<hash-based-on-app-name>/<user-event-id>
        Project:
         - ~/.zowe/.events/<hash-based-on-app-name>/<project-hash-based-on-path>/<user-event-id>
  ```
</details>


packages/imperative/src/events/src/doc/IImperativeEventParms.ts
```diff
@@ -0,0 +1,45 @@
...
+     * @type {string}
+     * @memberof IImperativeEventParms
+     */
+    appName: string;
+    /**
+     * The type of imperative event that occurred
+     * @type {ImperativeEventType}
+     * @memberof IImperativeEventParms
+     */
+    eventName: ImperativeEventType | string
```
traeok, line 32: Missing a semicolon on line 32:

```suggestion
    eventName: ImperativeEventType | string;
```

packages/imperative/src/events/src/doc/IImperativeEventParms.ts
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Specifies whether this is a user event or not
+     * @type {ImperativeEventType}
+     * @memberof IImperativeEventParms
+     */
+    isUser: boolean
```
traeok, line 38:
```suggestion
    isUser: boolean;
```

packages/imperative/src/events/src/ImperativeEventEmitter.ts
```diff
@@ -0,0 +1,288 @@
...
+import { ImperativeEvent } from "./ImperativeEvent";
+import { Logger } from "../../logger/src/Logger";
+import { LoggerManager } from "../../logger/src/LoggerManager";
+import { IImperativeRegisteredAction, IImperativeEventEmitterOpts, IImperativeEventJson } from "./doc";
+import { ConfigUtils } from "../../config/src/ConfigUtils";
+
+export class ImperativeEventEmitter {
+    private static mInstance: ImperativeEventEmitter;
+    private static initialized = false;
+    private subscriptions: Map<string, [fs.FSWatcher, Function[]]>;
```
traeok, line 27: Should we investigate using a stronger type for the callback functions? Otherwise, the current `Function[]` type will allow a user to provide a function with multiple parameters as a callback, even though the current logic does not provide any arguments to these callbacks when fired.

Something like this would suffice:

```ts
type EventCallback = () => void;
private subscriptions: Map<string, [fs.FSWatcher, EventCallback[]]>;
```

---

Also, just a note for awareness: the current implementation will only support synchronous callbacks. I'm not sure if this is intended, but if you wanted to support both, the callback type above could be adjusted to handle both return types:

```ts
type EventCallback = () => void | PromiseLike<void>;

// later, when firing callbacks...
for (const cb in callbacks) {
    await cb();
}
```

Keep in mind that support for `async` callbacks would also involve updating the function passed to the `fs.watch` call.

packages/imperative/src/events/src/ImperativeEventEmitter.ts
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    /**
+     * Check to see if the file path exists, otherwise, create it : )
+     * @param filePath Zowe or User path where we will write the events
+     */
+    private ensureEventFileExists(filePath: string) {
+        try {
+            if (!fs.existsSync(filePath)) {
+                fs.closeSync(fs.openSync(filePath, 'w'));
```
traeok, line 97: Following up on the CodeQL failure, this could be addressed by doing an `openSync` as the first statement in the `try` block and then creating the file in the `catch` block (as it will fail to open if the file does not exist)

traeok (dismissed):
Looks great so far! Going to run some tests locally, but the logic makes sense to me. I left some feedback regarding some of the types and the CodeQL warning.

zFernand0:
Thanks for the review! 🙏🏽 
Moving this back to draft to address these items 😋

packages/imperative/src/events/src/ImperativeEventEmitter.ts
zFernand0: I believe I can allow the callback to be async, however, I don't think we can guarantee that there won't be race conditions  due to how the Node's WatchListener is defined.

I'll do some testing on this again to see what can be done. 😋

traeok: Yep you're right, sorry about that - the current implementation will allow firing `async` callbacks, but as you mentioned the order cannot be guaranteed without `await`ing each callback.

If that's the case and we are okay with the implications of that, then you can disregard the comment about async. callbacks 😋

packages/imperative/src/events/src/ImperativeEventEmitter.ts
zFernand0: I do like the approach as it will help prevent errors thrown due to the file not existing.
However, this technique takes advantage of the `-w` system flag, which will create the file if it doesn't exist or truncate it if it dow ([more details](https://nodejs.org/api/fs.html#file-system-flags))

packages/imperative/src/events/src/ImperativeEventEmitter.ts
```diff
@@ -0,0 +1,265 @@
...
```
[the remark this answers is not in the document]
zFernand0: It is ok if the file system changed by the time we try to open it.
The use of the `-w` system flag helps avoid any issues around the file existing or not on the file system 😋 

Aler dismissed!

packages/imperative/src/events/src/ImperativeEventEmitter.ts
zFernand0: I'd still want to do some testing to see how it reflects on a working VSCE 😋

packages/imperative/src/events/src/ImperativeEventEmitter.ts
gejohnston: In the current logic, I think that openSync will not truncate an existing file because the earlier 'if' test should prevent openSync from being called when the file already exists. Admittedly, I do not know if other parts of this class cares whether existing files are truncated before use.

packages/imperative/src/events/src/ImperativeEventEmitter.ts
```diff
@@ -0,0 +1,288 @@
...
+     * @internal Not implemented in the MVP
+     */
+    public isCustomEvent(eventName: string): eventName is ImperativeEventType {
+        return !this.isUserEvent(eventName) && !this.isSharedEvent(eventName);
+    }
+
+    /**
+     * ZOWE HOME directory to search for system wide ImperativeEvents like `configChanged`
+     */
+    public getSharedEventDir(): string {
```
gejohnston, line 184: Could getSharedEventDir and getUserEventDir be private functions? Would the public function getEventDir provide all that is needed for external callers?

packages/imperative/src/events/src/ImperativeEventEmitter.ts
```diff
@@ -0,0 +1,288 @@
...
+     * @param eventName type of event to which we will create a watcher for
+     * @param callbacks list of all callbacks for this watcher
+     * @returns The FSWatcher instance created
+     */
+    private setupWatcher(eventName: string, callbacks: ImperativeEventCallback[] = []): fs.FSWatcher {
+        const dir = this.getEventDir(eventName);
+        this.ensureEventsDirExists(dir);
+        this.ensureEventFileExists(join(dir, eventName));
+
+        const watcher = fs.watch(join(dir, eventName), (event: "rename" | "change") => {
```
gejohnston, line 137: How do you manage to skip the second optional parameter of fs.watch?
It has a signature like this:
fs.watch( filename[, options][, listener] )

It looks like you only supply the first and third parameters. I am just trying to understand.

zFernand0:
This PR should be ready to review as-is.

There will be many more changes happening once the following PR gets in:
- #2136

gejohnston (approved):
I made all of my comments while this PR was still draft, and most of those comments were just questions.

packages/imperative/src/events/src/ImperativeEventEmitter.ts
zFernand0: This is possible thanks to the multiple declarations (a.k.a. [overload signatures](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads)) of the fs.watch method.
image

gejohnston: Thanks. I looked for another signature, but I did not find one. At least now I know that I am not crazy. 😁

packages/imperative/src/events/src/doc/IImperativeEventJson.ts
```diff
@@ -40,3 +40,5 @@ export interface IImperativeEventJson {
      */
     user?: boolean;
 }
+
+export type ImperativeEventCallback = () => void | Promise<void>;
```
traeok, line 44: We might want to change this to `PromiseLike` in order to support callbacks from a VS Code extension with a `Thenable` return type (as Microsoft's `Thenable` satisfies the `PromiseLike` type constraint, but does not satisfy `Promise`).

traeok (approved):
LGTM, thanks for all your work on this and for addressing feedback 😁  This is a nice stepping stone for Zowe Client apps and extensions to keep in sync with one another 😋

packages/imperative/src/events/src/doc/IImperativeEventJson.ts
zFernand0: Oh, that's right, Thenables (as PromiseLike) don't implement a `catch` method 😋

packages/imperative/src/events/src/ImperativeEventEmitter.ts
zFernand0: As we thought, all async callbacks will be fired in order, but no guarantee of who finishes first.
Also, it's possible that new emissions happen before the async callback is finished, in which case the results may be unpredictable 😋

zFernand0:
Merging in order to get the MVP out in its own prerelease version (`next-<timestamp>`)

packages/imperative/src/events/src/doc/IImperativeRegisteredAction.ts
```diff
@@ -0,0 +1,23 @@
...
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+/**
+ * Imperative Registered Action
+ * @export
+ * @interface IImperativeRegisteredAction
+ */
+export interface IImperativeRegisteredAction {
```
t1m0thyj, line 17: If the intent is to implement a disposable, would it make sense to name this something else like `IDisposableAction`?

I was also discussing with @ATorrise and we were wondering if the `Imperative` prefix is necessary on all the class and interface names related to the event emitter. I don't think we include it in most of our class names except for `ImperativeConfig` 😋

zFernand0: I agree with both points.
Addressed in 9305b3e1a34fa4336718fb4ce481efc2d7898952

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Add client-side event handling capabilities. [#1987](https://github.com/zowe/zowe-cli/pull/1987)
```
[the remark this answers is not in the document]
zFernand0: (－‸ლ)

Addressed in ed5d4684f12c5883b7c766ab20317a551f939701

Issue #2109, opened 2024-04-03
Breaking Change in User/Password/Token Order of Precedence 


dkelosky:
**Describe the bug**
This commit is technically a breaking change for Zowe CLI.  
https://github.com/zowe/imperative/commit/262f4fc84411e075e8e3af77ba2f8e663119be7d#diff-14634f70cce872732119d82fadcf7046800005fd18cca5ce7b7081672b9fdd2aL466

This changes the behavior in authentication order of precedence. 

**Expected and actual results**
Can the change be investigated as to provide an option to revert to the old behavior?  

**Describe your environment**

**Additional context**

gejohnston:
Fixed as part of the pull request "Reinstate token precedence over password in AbstractRestClient" #2119

PR #2119, opened 2024-04-15
Reinstate token precedence over password in AbstractRestClient

gejohnston:
**What It Does**

Modified this description to reflect changes made in response to reviews.

The AbstractRestClient class was originally released with logic that chose a token over user and password. Other commonly used Zowe SDK APIs enforce the order back to PASSWORD_OVER_TOKEN. Consumers which directly extended AbstractRestClient came to rely on the unintended order of TOKEN_OVER_PASSWORD. Later changes in this class to adhere to the Zowe policy of PASSWORD_OVER_TOKEN inadvertently broke such extenders. Until a means is provided for consumers (and/or end-users) to customize their credential order of precedence, we now revert the behavior of AbstractRestClient of back to TOKEN_OVER_PASSWORD.

A new array named authTypeOrder was added to ISession. It contains the order of precedence for authentication types. It can provide a building block for a future, customizable order of precedence. For now, that array is hard-coded in the AbstractRestClient constructor. Logic in another function within AbstractRestClient uses the order of that array to determine the type of authentication to use.

Related issue: [Breaking Change in User/Password/Token Order of Precedence](https://github.com/zowe/zowe-cli/issues/2109)

**How to Test**

Zowe CLI and ZE cannot be used to test the TOKEN_OVER_PASSWORD option. Too many up-stream safeguards ensure that using a token is not even a possibility when user and password are present. We would have to write a specialized test app that confirms that a token can be given precedence over password. Even then we would only confirm that our test app works. I think that everyone is better served if the consuming team confirms that the modifications address their needs.

Therefore, this pull request has been opened as a draft. I propose that we take the following actions:
- The Zowe CLI/ZE team reviews the logic to confirm that it seems correct.
- We create a CLI package containing this update.
- We post the package on Box.
- We notify the affected team about the availability of this "test-fix".
- Upon confirmation, we complete our normal test and publication process of Zowe CLI.

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**


packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -220,15 +221,29 @@ export abstract class AbstractRestClient {
...
      * Creates an instance of AbstractRestClient.
      * @param {AbstractSession} mSession - representing connection to this api
+     * @param {ICredOrder} credOrder
+     *          Specifies the order in which credentials will be used when multiple
+     *          types of credentials exist in the zowe client configuration.
+     *          When unspecified, a token will be used before user & password.
      * @memberof AbstractRestClient
      */
-    constructor(private mSession: AbstractSession) {
+    constructor(private mSession: AbstractSession, credOrder = ICredOrder.TOKEN_OVER_PASSWORD) {
```
dkelosky, line 242: without much thought, I "assumed" this option would have been implemented here: https://github.com/zowe/zowe-cli/blob/master/packages/imperative/src/rest/src/session/doc/ISession.ts

do you think this is simply an alternative place to add this new option?

gejohnston: AbstractRestClient is only called from one file (RestClient). Thus, the correction can be made in 2 existing files (and 1 new file). No other zowe code should see any potential change or capabilities. It would just correct something that was broken.
 
ISession is used in over a dozen other Zowe files. I am a little concerned about how widely visible this change would become.

We have had requests to officially change the credential order of precedence. To do that properly, we may have to introduce the ability to control the order of password, token, certificate, and Personal-Access-Token, (and maybe something else?) Further, we may have to define a way for end-users to set the precedence differently for different sessions. We have not thought through these requirements well enough to start down any particular implementation path.

I suppose that by making a new ISession property optional, it would not break other functions. Further, if RestClient **ALWAYS** sets that property, and AbstractRestClient is the only class to read that property, it could just as effectively revert the breaking change and limit the change to only the broken class. 

Your comment has made me think about making an ISession property more complex so that it might accommodate more than just password and token. A more complex property might provide the means to control the order of more credential types in the future, even though we would only implement password and token now. 

I will try out some of my thoughts and update this pull request if they pan out.

gejohnston: Update: The only way to make this alternative non-breaking is for the AbstractRestClient constructor to **Always** set the Session property to prefer "token". Otherwise a direct caller of AbstractRestClient would have to set the new ISession property before calling AbstractRestClient.constructor. The RestClient already overrides this AbstractRestClient behavior to prefer password, so this should be ok.

gejohnston: Hi @dkelosky - I made revisions as described above. Please re-review.

dkelosky: Hey @gejohnston, thanks for considering this approach. It feels a little more natural to me. Hopefully others feel the same 😃

t1m0thyj (approved):
LGTM, thanks @gejohnston!

AFAICT this should fix the potential breaking change that was reported, and I like the way it sets us up for supporting user-defined auth type order in the future.

When we're ready to build a CLI TGZ for others to test, we should be able to use the `npm run package` script to bundle the Imperative package changes in this PR 😋

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -229,6 +229,27 @@ export abstract class AbstractRestClient {
...
+         * determine the order, until a means is provided for consumers (and/or end-users)
+         * to customize their credential order of precedence, we hard-code the
+         * original order of token over password to correct the breaking change.
+         */
+        this.mSession.ISession.authTypeOrder = [
+            SessConstants.AUTH_TYPE_TOKEN,
+            SessConstants.AUTH_TYPE_BASIC,
+            SessConstants.AUTH_TYPE_BEARER,
+            SessConstants.AUTH_TYPE_CERT_PEM
+        ];
```
zFernand0, line 252: Love this! ❤️

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -459,63 +480,48 @@ export abstract class AbstractRestClient {
...
-                const authentication: string = `${this.session.ISession.tokenType}=${this.session.ISession.tokenValue}`;
-                headerKeys.forEach((property) => {
-                    options.headers[property] = authentication;
-                });
-            } else {
-                throw new ImperativeError({msg: "No credentials for a BASIC or TOKEN type of session."});
+        let credsAreSet: boolean = false;
+        for (const nextAuthType of this.session.ISession.authTypeOrder) {
+            if (nextAuthType === SessConstants.AUTH_TYPE_TOKEN) {
+                credsAreSet ||= this.setTokenAuth(options);
```
zFernand0, line 489: Smart use of the locall-or (falsy assignment) operator here 😋

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -459,63 +480,48 @@ export abstract class AbstractRestClient {
...
-                throw new ImperativeError({
-                    msg: "Failed to open one or more PEM certificate files, the file(s) did not exist.",
-                    causeErrors: err,
-                    additionalDetails: err.message,
-                });
+
+            if (credsAreSet) {
+                break;
             }
+            // else if (this.session.ISession.type === SessConstants.AUTH_TYPE_CERT_PFX) {
```
zFernand0, line 504: I know this is commented out code, but should we move this to its own little `setCertPfxAuth` function and have this `else if` commented out right below the `setCertPemAuth`?

zFernand0 (approved):
LGTM! 😋 
left a small comment/question but nothing that should stop this PR from being merged 😋

traeok (approved):
LGTM, thanks @gejohnston!

packages/imperative/src/rest/src/client/AbstractRestClient.ts
gejohnston: I thought about creating a function for setCertPfxAuth, but I am reluctant to do it for the following reasons:

1. Once the logic is moved into its own function, the conditionals must be changed to achieve the same result. I prefer not to make logic changes into commented code that I cannot test. I do not want to accidentally lose the conditions specified by the original programmer.

2. If a new function is created with a commented-out body, some code-scanning tool is going to tell us what horrible programmers we are.

3. If I instead comment out the whole function, a programmer would have to find 2 blocks of comments that are separated by many lines of code to determine how to update the program to handle the new credential type.

I think a reasonable compromise would be to add some instructional comments at the beginning of the existing commented-out code-block describing how a new function should be created to handle Pfx certs. When a programmer wants to add Pfx certs, all of the information will be in one place, located where the programmer will likely look. I plan to make this change.

zFernand0: I guess I inclined to agree with the reasons. 😅 
It's definitely not idea to have to modify code that we won't be testing, plus tools will start complaining about it 😋 

Thanks for adding the information about the commented out code though 🥳

Issue #2116, opened 2024-04-10
Consolidate the Zowe client log files into the same directory


gejohnston:
**Is your feature or enhancement request related to a problem or limitation? Please describe**

The location of our log files is unintuitive. Our 2 log files are in two separate directories. It is not clear that to find a log file that you must traverse into a directory named '.zowe/imperative' or a directory named '.zowe/zowe'

```
C:\USERS\YouUserName\.ZOWE
│
├───imperative
│   └───logs
│       imperative.log
│
└───zowe
    └───logs
        zowe.log
```

**Describe your enhancement idea**

Consolidate the log files into a directory named '.zowe/logs'. Remove the otherwise empty directories named '.zowe/imperative' and '.zowe/zowe'.

```
C:\USERS\YouUserName\.ZOWE
│
└────logs
     imperative.log
     zowe.log
```

**Describe alternatives you've considered**

**Provide any additional context**

Because the product works ok as-is, this is a low priority. It is possible that changing the location of log files could be considered a breaking change and would have to be completed before V3 if the work is undertaken.

PR #2122, opened 2024-04-24
Move Imperative and CLI log files into same folder

t1m0thyj:
**What It Does**

Fixes https://github.com/zowe/zowe-cli/issues/2116

TODO:
* [x] ~Update default log location in ZE to match~
  * ZE already stores them in the same folder but a different location: `~/.vscode/extensions/zowe.vscode-extension-for-zowe-X.Y.Z/logs`
* [ ] Update default log location in Zowe docs

**How to Test**

Build the CLI from this PR and verify that logs are generated in the new location:
* `~/.zowe/logs/imperative.log`
* `~/.zowe/logs/zowe.log`

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/src/io/src/IO.ts
+++ b/packages/imperative/src/io/src/IO.ts
@@ -30,11 +30,12 @@ export class IO {
 
     /**
      * File delimiter
+     * @deprecated Use `path.sep` or `path.posix.sep` instead
      * @static
      * @type {string}
      * @memberof IO
      */
-    public static readonly FILE_DELIM: string = "/";
+    public static readonly FILE_DELIM: string = path.posix.sep;
 
     /**
      * UTF8 identifier
--- a/packages/imperative/src/io/src/IO.ts
+++ b/packages/imperative/src/io/src/IO.ts
@@ -30,7 +30,7 @@ export class IO {
 
     /**
      * File delimiter
-     * @deprecated Use `path.sep` or `path.posix.sep` instead
+     * @deprecated Use `path.posix.sep` instead or `path.sep` for better cross-platform support
      * @static
      * @type {string}
      * @memberof IO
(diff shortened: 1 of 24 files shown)


zFernand0 (commented):
LGTM so far! 😋

gejohnston:
Should we add something about this in the "end-user functionality" section of the V3 Breaking changes doc?
Note that we are now exporting that Word doc to a PDF file in the same directory.

packages/imperative/src/logger/src/LoggerConfigBuilder.ts
zFernand0, line 1: since we just deprecated `FILE_DELIM`, maybe we should remove its usage across our codebase (in favor of `path.posix.sep`) 😋 

Also, curious if we can deprecate `DEFAULT_LOG_DIR` and other variables that are only exporting something from another class 😅

packages/imperative/src/io/src/IO.ts
```diff
@@ -30,11 +30,12 @@ export class IO {
 
     /**
      * File delimiter
+     * @deprecated Use `path.sep` or `path.posix.sep` instead
```
zFernand0, line 33: since we had hardcoded `/`, should we just mention `path.posix.sep` to avoid a `\\` on windows? 😋 

Not sure that it matters much, but there may be some test failures for plug-in developers if they use `path.sep` on windows. 😋

zFernand0 (approved):
LGTM, thanks @t1m0thyj! 😋 

Left a few comments, but nothing that should stop the PR from being merged 😋

traeok (approved):
LGTM, thanks for consolidating the log files 👍

packages/imperative/src/io/src/IO.ts
t1m0thyj: I think it's better practice to use `path.sep` rather than hardcoding `/` which is less Windows-friendly, but agree that changing it could cause tests fail. Should we mention both, and provide more details in the comment about why you might choose one over the other?

gejohnston (approved):

awharn (approved):

PR #2125, opened 2024-05-01
Restore precedence of token over password in AbstractRestClient - V3

gejohnston:
**What It Does**

Implements the same changes into V3 as the following pull request did in V2:
Reinstate token precedence over password in AbstractRestClient #2119

Related issue : Breaking Change in User/Password/Token Order of Precedence #2109 (already closed)

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/src/rest/src/client/AbstractRestClient.ts
+++ b/packages/imperative/src/rest/src/client/AbstractRestClient.ts
@@ -229,6 +229,27 @@ export abstract class AbstractRestClient {
         ImperativeExpect.toNotBeNullOrUndefined(mSession);
         this.mLogger = Logger.getImperativeLogger();
         this.mIsJson = false;
+
+        /* Set the order of precedence in which available credentials will be used.
+         *
+         * The Zowe SDK policy is to select password credentials over a token.
+         * However, this class was originally released with the token over password.
+         * The commonly-used ConnectionPropsForSessCfg.resolveSessCfgProps enforces the
+         * order of password over token. None-the-less, consumers which directly extended
+         * AbstractRestClient came to rely on the order of token over password.
+         *
+         * Later changes in this class to adhere to Zowe policy inadvertently broke
+         * such extenders. While we now use a generalized authTypeOrder property to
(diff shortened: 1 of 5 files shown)


t1m0thyj (approved):

awharn (approved):

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -229,6 +229,27 @@ export abstract class AbstractRestClient {
...
+         *
+         * Later changes in this class to adhere to Zowe policy inadvertently broke
+         * such extenders. While we now use a generalized authTypeOrder property to
+         * determine the order, until a means is provided for consumers (and/or end-users)
+         * to customize their credential order of precedence, we hard-code the
+         * original order of token over password to correct the breaking change.
+         */
+        this.mSession.ISession.authTypeOrder = [
+            SessConstants.AUTH_TYPE_TOKEN,
+            SessConstants.AUTH_TYPE_BASIC,
```
zFernand0, line 249: Do we want to reverse the order here (basic > Token) since this is against the `next` branch (to be consistent)?
or keep it as is?

zFernand0 (approved):
Changes look identical to the other PR. 🥳 

LGTM! 😋 

Left a small question in case we wanted to reverse the order in the array to `Basic > Token` (for consistency with other implementations) 😋

packages/imperative/src/rest/src/client/AbstractRestClient.ts
gejohnston: Even though we are "allowed" to make breaking changes in V3, I think that the same extenders will hit the same dead-end failure with no practical solution. Their products will be broken.

I think that we will have to keep the current credential order in place until we implement a means for extenders (and maybe end-users) to specify their preferred order of credentials. I don't think that we know when we will undertake such an enhancement. I also cannot predict if that enhancement can be done in a non-breaking fashion after V3 is released. None-the-less, I think we have to stick with the current order for now.

PR #2135, opened 2024-05-09
Fix secure properties not listed outside the active config layer

t1m0thyj:
**What It Does**

Fixes https://github.com/zowe/zowe-explorer-vscode/issues/2633

**How to Test**

Create zowe.config.json and zowe.config.user.json with the contents from https://github.com/zowe/zowe-explorer-vscode/issues/2633#issuecomment-2098383503

Then try to list the secure properties associated with the base profile using the ProfileInfo API:
```javascript
const { ProfileInfo } = require("@zowe/imperative");

(async () => {
    // Load connection info from default z/OSMF profile
    const profInfo = new ProfileInfo("zowe");
    await profInfo.readProfilesFromDisk();
    console.log(profInfo.getTeamConfig().api.secure.securePropsForProfile("base"));
})();
```

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] added/updated automated tests
- [x] updated the changelog
- [x] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

**Additional Comments**

PR #2133, opened 2024-05-09
Show strings instead of regexes for allowed values in CLI help

t1m0thyj:
**What It Does**

Fixes #1863

**How to Test**

1. Run `zowe zos-files list data-set --protocol invalid` and verify that there are no `^` or `$` characters in the command response:
    ```
    You specified:
    invalid
    
    The value must match one of the following regular expressions:
    [ 'http', 'https' ].
    ```
2. Run `zowe auth logout apiml --help` and verify that there is no `^` character before "apimlAuthenticationToken":
    ```
       --token-type | --tt (string)
    
          The type of token to get and use for the API. Omit this option to use the
          default token type, which is provided by 'zowe auth login'.
          
          Allowed values: apimlAuthenticationToken.*, jwtToken, LtpaToken2
    ```

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] added/updated automated tests
- [x] updated the changelog
- [x] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

**Additional Comments**

**Note:** Removing the `^` in front of "apimlAuthenticationToken" is purely cosmetic and should not change the behavior, because it will be added by `SyntaxValidator`:
https://github.com/zowe/zowe-cli/blob/59a4b1ea5fbdba0e7bf133daca25665077fa77fa/packages/imperative/src/cmd/src/syntax/SyntaxValidator.ts#L300-L304

PR #2126, opened 2024-05-01
Fix macOS tgz containing extra `GNUSparseFile.0` directory

t1m0thyj:
**What It Does**

Fixes regression of #1311 (broke in v7.11.2)

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] added/updated automated tests
- [ ] updated the changelog
- [ ] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

**Additional Comments**

PR #2117, opened 2024-04-10
Create dco.yml

zFernand0:
**What It Does**

Adds DCO related configuration.

**How to Test**

We will just have to merge this.

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] added/updated automated tests
- [ ] updated the changelog
- [ ] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

**Additional Comments**

For context, here is the DCO docs on the `allowRemediationCommits` option:
- https://github.com/dcoapp/app?tab=readme-ov-file#third-party-remediation-support

PR #2142, opened 2024-05-15
Merge master into next

awharn:
**What It Does**

Updates Next to contain the following enhancements and bug fixes from Master:
- #2117
- #2126
- #2133
- #2135
- #1822
- #2139

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**


packages/zosjobs/src/GetJobs.ts
```diff
@@ -366,13 +367,14 @@ export class GetJobs {
      * @returns {Promise<string>} - promise that resolves to the spool content
      * @memberof GetJobs
      */
-    public static async getSpoolContentById(session: AbstractSession, jobname: string, jobid: string, spoolId: number) {
+    public static async getSpoolContentById(session: AbstractSession, jobname: string, jobid: string, spoolId: number, encoding?: string) {
```
zFernand0, line 370: Should we move the params to a typed object? (in V3)

packages/zosjobs/src/SubmitJobs.ts
```diff
@@ -106,16 +106,23 @@ export class SubmitJobs {
      * @returns {Promise<IJob>} - Promise that resolves to an IJob document with details about the submitted job
      * @memberof SubmitJobs
      */
-    public static submitJcl(session: AbstractSession, jcl: string, internalReaderRecfm?: string, internalReaderLrecl?: string) {
+    public static submitJcl(session: AbstractSession, jcl: string, internalReaderRecfm?: string,
```
zFernand0, line 109: Same as previous commnet...
Should we move everything after `jcl` (or include `jcl` too) into a typed object (with its own interface)?
This can be done in V3, and we can follow the example of `submitJobCommon` in this file 😋

zFernand0 (approved):
The port of al of those PRs LGTM! 😋 
left a few questions to see if we would want to make a slight breaking change here 😋

packages/zosjobs/src/GetJobs.ts
awharn: I was thinking that any changes to interfaces and parameters would be a different story, since this was primarily meant to get the two branches reconciled. It might be nice to do that change on this API....

packages/zosjobs/src/SubmitJobs.ts
awharn: ...however, on this API, we already call submitJclCommon. Frankly, I think it would be nice to just get rid of some of these redundant APIs.

ATorrise (approved):

traeok (approved):
LGTM, thanks @awharn

t1m0thyj (approved):
LGTM, thanks @awharn!

Issue #1305, opened 2022-02-07
Using zowe-cli behind a corporate proxy


Crosswind:
I stumbled accross #498 from @jguillaumes and believe that his original request was misunderstood and thus I am opening another issue. 

Let me elaborate a bit on the use case:
I am currently working on IBM Z Xplore challenges and am asked to access the system through zowe-cli. The system can be routed from the Internet (204.90.115.200:10443). Our company doesn't allow any direct traffic flow towards Internet facing systems which are not going through port 443 and through the browser (i.e. mostly cli utilities and stuff like IDEs). Therefore, a proxy server was setup which has a whitelist of hosts that can be accessed through the proxy server on non-standard ports and protocols. The Z Xplore is whitelisted on our proxy server (access through VS Code is possible). With tools like VS Code, Maven, npm, etc. I use this proxy server to access all kinds of services, e.g.:
```
# This doesn't work:
npm i -g @zowe/cl

# This works
npm config set proxy http://corp-proxy.telekom.de:8080
npm config set https-proxy http://corp-proxy.telekom.de:8080
npm i -g @zowe/cl
```

For zowe-cli I need to configure the same thing. Using the `--host` `--protocol` and `--port` option changes the actual host that I am trying to connect to (in my case overriding what is specified in the profile). This is not what I want. The original host (200.90.115.200) is the one I am trying to connect to but since my cli cannot reach that directly it needs to route the requests through a proxy.

As @jguillaumes also noted a variety of possibilities exist on how you can configure different utilities to route requests through a proxy to the eventual target host. I imagine that adding an option to the zowe profile would make a lot of sense. Accessing the ML through VS Code works perfectly fine because I have an option to set the proxy and the plugin honors that.

I'd be happy to support with testing in case this gets enough attention and an implemention ;)

Cheers - David

jguillaumes:
Hello,

To be honest, I had completely forgotten I opened that issue two years ago... but I've not had that need again. Until now.

@Crosswind is right on target. I want to connect to my zowe apiml which is in <my internal hostname>:7554/api/v1, THRU my proxy at http://192.168.0.92:3128. ZOWE CLI _should_ honor a system wide proxu setting (in windows) or the http_proxy and/or https_proxy environment variables in unix-like environments. Allowing to override that at the profile level would also be awesome.

The only workaround is right now to use an ssh tunnel to a machine with direct connectivity to the mainframes, which is not always posible (and is a very ugly kludge nevertheless).

[...]

PR #2151, opened 2024-05-23
Added simple proxy agent support using standard environment variables

phaumer:
**What It Does**

https://github.com/zowe/zowe-cli/issues/1305

I am proving a simple first increment for supporting a proxy http agent for Zowe CLI. It requires usage of the environment variables `HTTP_PROXY`, `http_proxy`, `HTTPS_PROXY`, `https_proxy`. If any of these env variables is set and depending if the Zowe session is configured for http or https it instantiates an appropriate http agent using the same libraries that VS Code uses: <https://github.com/TooTallNate/proxy-agents> 

If the z/OS system uses self-signed certificates then the proxy server must be configured to accept them.  If the proxy server itself is configured with self-signed certificates then the user needs to either import these certificates on their workstation or use the (not recommended) nodejs variable NODE_TLS_REJECT_UNAUTHORIZED=0. 

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

For the `next` branch I could envision to extension Zowe profiles with new proxy properties so users can specify their proxy server in their team config files. We can also consider using another third party library to read Windows and MacOS system settings for proxies.


Crosswind:
@phaumer Thank you for providing this. I am trying to test this but don't seem to have a lot of success. I am currently in an environment that has direct connection to the target system which makes the test a little more complicated. I am essentially trying to break the connection by sending the request through a non-existing proxy server first. 

I have a proxy server running on localhost:8888 (using https://github.com/tinyproxy/tinyproxy). I can easily verify that it works by using cURL.
Now, to test the Zowe CLI I am setting my environment variables to localhost:8889 (nothing is running on that port). I am expecting Zowe CLI to complain as it shouldn't be able to connect. Unfortunately I don't get any connection issues and the CLI just works.

```
> env | grep proxy
http_proxy=http://localhost:8889
https_proxy=http://localhost:8889
> curl -x localhost:8889 https://google.com  
curl: (7) Failed to connect to localhost port 8889 after 0 ms: Couldn't connect to server
> zowe zosmf check status --reject-unauthorized false | head
(node:16242) [DEP0051] DeprecationWarning: The `util.isNullOrUndefined` API is deprecated. Please use `arg === null || arg === undefined` instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
(node:16242) [DEP0044] DeprecationWarning: The `util.isArray` API is deprecated. Please use `Array.isArray()` instead.
(node:16242) [DEP0056] DeprecationWarning: The `util.isString` API is deprecated.  Please use `typeof arg === "string"` instead.
(node:16242) [DEP0052] DeprecationWarning: The `util.isNumber` API is deprecated. Please use `typeof arg === "number"` instead.
The user z01437 successfully connected to z/OSMF on 'S0W1.DAL-EBIS.IHOST.COM'.
zosmf_port:         10443
zosmf_saf_realm:    SAFRealm
zos_version:        04.27.00
zosmf_full_version: 27.0
api_version:        1
```

I haven't had the chance to dive deeper into the actual code to try and understand what may be missing. I am quite certain that my environment setup is correct. Any suggestions or things that I could test?

phaumer:
@Crosswind thanks. There should be an INFO level message in the imperative.log when the environment variable was found. I have to also test with other proxy servers. I was using Squid.

Crosswind:
Apologies, that was a mess-up on my side. It's working now. 
I get an error message if I try to use an incorrect proxy server and I can successfully connect if I go through my `tinyproxy` server.
I could only test the connection to a TLS-secured instance so I could only verify `HTTPS_PROXY` and `https_proxy`.

Regarding mention of self-signed certificates: I am using the cli option `--reject-unauthorized false` so I didn't have to specifically configure my proxy server. It should end up doing the same thing but it was much simpler and less intrusive like setting `NODE_TLS_REJECT_UNAUTHORIZED=0`.

For completeness: Does it make sense to also add `NO_PROXY` variable support? This is more or less for those people that need to use a proxy server for certain things but their `zowe cli` should not go through one.

phaumer:
@Crosswind thanks. I quickly added reading no_proxy. Have to do more testing for it, but let me know if that matches your expectations.

packages/imperative/src/rest/src/client/Proxy.ts
```diff
@@ -0,0 +1,94 @@
...
+import { URL } from 'url';
+import { Agent } from 'https';
+import { HttpProxyAgent } from 'http-proxy-agent';
+import { HttpsProxyAgent } from 'https-proxy-agent';
+
+import { HTTP_PROTOCOL_CHOICES, HTTP_PROTOCOL, HTTPS_PROTOCOL } from '../session/SessConstants';
+import { ISession } from '../session/doc/ISession';
+
+export class Proxy {
+
```
gejohnston, line 22: When adding a new family of functions in a class, we try to have a block comment for each function describing its purpose, parameters, and return value. Perhaps when you have finalized the details of the implementation, you could add such comments within the new Proxy class.

gejohnston (changes requested):
Thanks for devoting the effort to add a useful feature to the CLI SDK.

I have little expertise on handling proxies, and do not have an environment setup to test the modifications. However, the logic in the modifications make sense, and I see no adverse effect on existing functionality.

I left one comment that is more about convention than about programming logic. Otherwise, I am ready to approve this PR.

packages/imperative/src/rest/src/client/Proxy.ts
phaumer: @gejohnston thanks for the review and sorry for the delay. I now added a commit with the jsdoc comments following the pattern I saw other modules use in the folder.

packages/imperative/src/rest/src/client/Proxy.ts
```diff
@@ -0,0 +1,178 @@
...
+    /**
+     * Parses environment variables valid for https requests.
+     * @private
+     * @static
+     * @returns `string` if valid variable is found or undefined.
+     * @memberof Proxy
+     */
+    private static getHttpsEnvVariables(): string | undefined {
+        return env.HTTPS_PROXY ?? env.https_proxy ?? this.getHttpEnvVariables() ?? undefined;
```
traeok, line 136: Since `this.getHttpEnvVariables` falls back to `undefined`, the `?? undefined` can be removed here as well

packages/imperative/src/rest/src/client/Proxy.ts
```diff
@@ -0,0 +1,178 @@
...
+     * Parses environment variables valid for no proxy exceptions.
+     * @private
+     * @static
+     * @returns `string[]` of all hostnames found in the comma separated list
+     *          in lowercase without white spaces.
+     * @memberof Proxy
+     */
+    private static getNoProxyEnvVariables(): string[] | undefined {
+        const noProxyValue = env.NO_PROXY ?? env.no_proxy ?? undefined;
+        if (!noProxyValue) {
```
traeok, line 149: The suggestion about `?? undefined` also applies here

traeok (approved):
LGTM! Thanks for the enhancement @phaumer @rudyflores

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added support for proxy servers using a proxy http agent. Supports the usage of the environment variables HTTP_PROXY, http_proxy, HTTPS_PROXY, https_proxy. If any of these env variables is set and depending how the Zowe session is configured for http or https it instantiates an appropriate http agent. If the z/OS system uses self-signed certificates then the proxy server must be configured to accept them. If the proxy server itself is configured with self-signed certificates then the user needs to either import these certificates on their workstation, use rejectUnauthorized in their Zowe profile, or use the (not recommended) nodejs variable NODE_TLS_REJECT_UNAUTHORIZED=0. Zowe also looks for the environment variables, NO_PROXY, no_proxy. These work with a simple comma separated list of hostnames that need to match with the hostname of the Zowe profile.
```
t1m0thyj, line 7: Thanks for documenting the new behavior thoroughly 👍 

Could we make this long paragraph easier to read by splitting it up into bullet points that explain each environment variable (HTTP_PROXY, HTTPS_PROXY, NO_PROXY)?

Will defer to @anaxceron on this 😋

t1m0thyj (approved):
LGTM, thanks @phaumer and @rudyflores!

awharn (approved):
Tested this with a Squid forward proxy, and it appears to work perfectly. Thank you for the contribution! 😃

packages/imperative/CHANGELOG.md
anaxceron: Hi @phaumer -- I did some light editing on the text. Please let me know if we can go ahead w/ the content below for the changelog:

Enhancement: Added support for proxy servers using a proxy http agent. Supports the usage of the environment variables `HTTP_PROXY`, `HTTPS_PROXY` (not case sensitive).
- If any of these environment variables is set and depending how the Zowe session is configured for http or https, the REST client instantiates an appropriate http agent. 
- If the z/OS system uses self-signed certificates then the proxy server must be configured to accept them.
- If the proxy server itself is configured with self-signed certificates then the user needs to either import these certificates on their workstation, use `rejectUnauthorized` in their Zowe profile, or use the (not recommended) nodejs variable `NODE_TLS_REJECT_UNAUTHORIZED=0`.
- Zowe also looks for the environment variable `NO_PROXY`. These work with a simple comma separated list of hostnames that need to match with the hostname of the Zowe profile.

rudyflores: I went ahead and updated the changelog witth the suggestion above, thanks @anaxceron !

gejohnston (approved):

anaxceron (approved):

t1m0thyj (approved):

awharn:
@phaumer @rudyflores The Zowe 2.17.0 system demo is coming up on July 23rd. Would either of you be interested in doing a demo of this enhancement?

PR #2159, opened 2024-05-28
Remove unnecessary exports

awharn:
**What It Does**

Removes a set of exports that are not referenced outside of the related packages
Deprecates some functions/classes that are identical and appear twice in the codebase

This PR is presently in a draft state to get team opinions on additional things that can be removed. All of the items removed here did not appear outside of their respective packages. However, GitHub search only works against **V2 LTS**.

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**


t1m0thyj (approved):
LGTM, thanks @awharn!

packages/imperative/src/cmd/index.ts
```diff
@@ -13,8 +13,8 @@ export * from "./src/doc/profiles/definition/ICommandProfileSchema";
 export * from "./src/doc/profiles/definition/ICommandProfileProperty";
 export * from "./src/doc/profiles/definition/ICommandProfileTypeConfiguration";
 export * from "./src/doc/handler/ICommandHandler";
-export * from "./src/doc/handler/ICommandHandlerResponseChecker";
-export * from "./src/doc/handler/ICommandHandlerResponseValidator";
+// export * from "./src/doc/handler/ICommandHandlerResponseChecker";
+// export * from "./src/doc/handler/ICommandHandlerResponseValidator";
```
traeok, line 17: Quick question: I'm guessing we plan to remove these comments before merging?
Either way, no request for changes here; I just wanted to get an idea of what we plan to do with these line comments.

packages/imperative/src/operations/src/Operation.ts
```diff
@@ -28,6 +28,9 @@ export type IOperationUndoCompleted =
 export type IOperationResultReady<T> =
     (output: T, operationResults: IOperationResult<any>) => void;
 
+/**
+ * @deprecated
+ */
 export abstract class Operation<T> implements ITaskWithStatus {
```
traeok, line 34: Do we have a planned replacement for the `Operation` class going forward, or do we plan on removing it entirely? I'm not sure who is all using this as it is an abstract class, but we might want to elaborate on the long-term intentions in the deprecation message.

traeok (approved):
LGTM, thanks for cleaning up these exports and reviewing the deprecated functions 😋 
I just have a couple questions about the comments that were added in this PR.

packages/imperative/src/rest/index.ts
```diff
@@ -17,11 +17,11 @@ export * from "./src/client/doc/IRestClientError";
 export * from "./src/client/doc/IRestClientResponse";
 export * from "./src/client/doc/IRestOptions";
 export * from "./src/client/Headers";
-export * from "./src/client/AbstractRestClient";
-export * from "./src/client/CompressionUtils";
+// export * from "./src/client/AbstractRestClient";
```
gejohnston, line 20: While none of our apps directly import AbstractRestClient, we recently discovered that a member of the external Zowe community derives their own class by inheriting directly from AbstractRestClient. Until we implement a means for consumers to specify credential order-of-precedence, I think we will need to keep AbstractRestClient available to consumers.

gejohnston (changes requested):
These changes go a long way towards diffferentiating between those functions that we just happen to use and those functions that consumers are supposed to see.

I found one item which I do not think that we should hide from consumers. Check out my specific comment and see if you agree.

packages/imperative/src/operations/src/Operation.ts
awharn: From what I saw, nobody ever used this at all.

packages/imperative/src/cmd/index.ts
awharn: Potentially. I was thinking maybe we would leave them for awhile, and remove them later. After we get a couple releases into V3, and we haven't received any requests from consumers to add any of the removed content back to the exports.

packages/imperative/src/operations/src/Operation.ts
traeok: That's fair - since its still being exported, do you think we should err on the side of caution and add something short such as "No replacement planned, will be removed in v4"? I don't think that many devs would use this class either, but I think it would be helpful to provide context in case a private repo or company plug-in is leveraging this class without our knowledge.

awharn: I was planning on having that in the changelog, unless you think it needs to be in the typedoc too.

gejohnston (approved):

t1m0thyj (approved):

packages/zosfiles/src/index.ts
```diff
@@ -42,8 +42,4 @@ export * from "./methods/upload";
 export * from "./methods/utilities";
 
 export * from "./utils/ZosFilesAttributes";
-export * from "./utils/ZosFilesUtils";
-export * from "./doc/IOptionsFullResponse";
-export * from "./doc/IRestClientResponse";
-export * from "./doc/IOptions";
-export * from "./doc/types/ZosmfRestClientProperties";
```
zFernand0, line 49: Curious why these were removed instead of commented out like the rest? 😅 
Should we remove instead of comment-out?

zFernand0 (approved):
LGTM! 😋

packages/zosfiles/src/index.ts
awharn: Those were all duplicates. They all appear earlier in the file, and were exported twice.

packages/imperative/src/cmd/index.ts
awharn: We can create an issue with the file paths of all the files that have comments in them. The only reason I didn't want to remove them entirely is because it becomes simpler to restore them if a request does come in.

zFernand0: sounds good 😋 
- https://github.com/zowe/zowe-cli/issues/2168

packages/core/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Zowe core SDK package will be documented in this file.
 
+## Recent Changes
+
+- LTS Breaking: Modified the core SDK. [#2083](https://github.com/zowe/zowe-cli/issues/2083)
```
anaxceron, line 7: In call changelog references, can we say _why_ we're modifying the core SDK? For performance, for tech currency?

anaxceron (approved):
Thanks @awharn -- made one comment

packages/core/CHANGELOG.md
awharn: The only modification made was to deprecate the IHeaderContent interface to remove duplication between the Core SDK and Imperative, which is captured on the next line. Would it be preferred to remove this line and only have the deprecation line?

packages/imperative/CHANGELOG.md
```diff
@@ -5,6 +5,47 @@ All notable changes to the Imperative package will be documented in this file.
 ## `8.0.0-next.202406061600`
 
 - BugFix: Updated `braces` dependency for technical currency. [#2158](https://github.com/zowe/zowe-cli/pull/2158)
+- LTS Breaking: Modified the @zowe/imperative SDK [#2083](https://github.com/zowe/zowe-cli/issues/2083)
+  - Removed the following exported classes:
```
zFernand0, line 9: should we move this under a recent-changes header?

PR #2173, opened 2024-06-07
Fix error when downloading spool files with default encoding

t1m0thyj:
**What It Does**

Fixes z/OSMF error when `DownloadJobs.downloadSpoolContentCommon` is called and `encoding` property is undefined.

Also corrects the behavior of `Get.dataSet` and `Get.USSFile` methods so that they return an empty buffer instead of null for empty files.

**How to Test**

Run the following system test files and ensure that all tests pass:
* `DownloadJobs.system.test.ts`
* `cli.dir.upload.dtu.system.test.ts`
* `cli.files.upload.stds.system.test.ts`

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/zosfiles/CHANGELOG.md
+++ b/packages/zosfiles/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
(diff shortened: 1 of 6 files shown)


zFernand0 (approved):
LGTM! 😋 
Thanks for the quick fix! 😋

PR #2179, opened 2024-06-19
Verify credential manager access [v3]

### LINKED ISSUE #2111: Function to validate that credential manager will load successfully after `ProfilesInfo` constructor called

In Zowe Explorer, we utilize the `ProfilesInfo` API for loading our default and custom credential managers. Our issue is that when calling the constructor we expect the credential manager to be loaded during initialization, but instead the credential manager is loaded until `readProfilesFromDisk` is called. 

For this issue, I would like for the `ProfilesInfo` API to provide a way to check early whether or not the credential manager will fail when loading, so that way we won't rely on the `readProfilesFromDisk` function that can be called in multiple locations and can be harder to trace.

My current alternative is to call `await profileInfo.readProfilesFromDisk()` after the constructor in a `try/catch` that will catch error and handle accordingly, although not great since it uses the entire part of the `readProfilesFromDisk` code if successful, it currently is the only way to catch this error early, which is not conventional.

Imperative code screenshots:

image

image

image

awharn:
**What It Does**

Adds the `ProfileInfo.profileManagerWillLoad` function that returns if the credential manager will load without loading profiles.
Resolves #2111 in v3

**How to Test**

Run test suites

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
-- Enhancement: Added `ProfileInfo.profileManagerWillLoad` function to verify the credential manager will load before attempting to use it. [#2111](https://github.com/zowe/zowe-cli/issues/2111)
+- Enhancement: Added `ProfileInfo.profileManagerWillLoad` function to verify the credential manager can load. [#2111](https://github.com/zowe/zowe-cli/issues/2111)
 
 ## `8.0.0-next.202406111958`
 
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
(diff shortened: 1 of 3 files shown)


t1m0thyj (approved):
LGTM, thanks @awharn!

traeok (approved):

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added `ProfileInfo.profileManagerWillLoad` function to verify the credential manager will load before attempting to use it. [#2111](https://github.com/zowe/zowe-cli/issues/2111)
```
anaxceron, line 7: Needs some clarification: What is "it" in this sentence? And what is the thing that is attempting to use "it"?

awharn: I changed the wording, but in this case, we are just adding a function to verify the credential manager can load before calling another function that will attempt to use the credential manager. This is beneficial for things such as Zowe Explorer, which needs to know if there is going to be a problem loading credentials prior to actually attempting to access credentials from the credential vault.

PR #2189, opened 2024-07-02
Plugin registry field

jace-roell:
**What It Does**
- Resolves bug that resulted in each plugin to have identical public registries regardless of actual installation location/reference
-  Resolves bug that resulted in every plugin to have the same registry as the first if multiple were installed in the same command

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog


packages/imperative/src/imperative/__tests__/plugins/utilities/NpmFunctions.unit.test.ts
```diff
@@ -15,6 +15,8 @@
 import * as pacote from "pacote";
 import * as npmFunctions from "../../../src/plugins/utilities/NpmFunctions";
 import { PMFConstants } from "../../../src/plugins/utilities/PMFConstants";
+import { ExecUtils } from "../../../../utilities";
+import { IO } from "../../../../io";
```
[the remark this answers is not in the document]
jace-roell: resolved

packages/imperative/src/imperative/src/plugins/cmd/install/install.handler.ts
```diff
@@ -71,112 +74,218 @@ export default class InstallHandler implements ICommandHandler {
...
+                        "package..."
+                    )} is as well. ` + `They are mutually exclusive.`,
             });
         } else {
             try {
-                let installRegistry: any;
-
-                // Get the registry to install to
-                if (typeof params.arguments.registry === "undefined") {
-                    installRegistry = getRegistry().replace("\n",  "");
```
zFernand0, line 87: To fix the last few tests, we need to bring some of this logic/checks back 😋

zFernand0 (changes requested):
left a small comment with the area where we may need to make changes in order to fix the last few tests 😋

packages/imperative/src/imperative/src/plugins/cmd/install/install.handler.ts
```diff
@@ -71,112 +74,218 @@
...
-                            if (!packageInfo.registry) {
-                                packageInfo.registry = installRegistry;
+                            if (!packageInfo.location) {
+                                packageInfo.location = installRegistry;
+                            }
+
+                            let isDirTest: boolean;
+
+                            try{
+                                isDirTest = IO.isDir(packageInfo.package)
```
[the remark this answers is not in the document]
jace-roell: fixed

packages/imperative/CHANGELOG.md
```diff
@@ -2,9 +2,15 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Resolved bug that resulted in each plugin to have identical public registries regardless of actual installation location/reference
+- BugFix: Resolved bug that resulted in every plugin to have the same registry as the first if multiple were installed in the same command
```
anaxceron, line 8: - We should clarify what we mean by "the first" and "multiple": " .. to have the same registry as the first plug-in (?) if multiple plug-ins (?) were ... "
- Plug-in should be hyphenated, per the Zowe doc style
- Sentence should end w/ a period
- Also add an issue or PR link to the changelog item

packages/imperative/CHANGELOG.md
```diff
@@ -2,9 +2,15 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Resolved bug that resulted in each plugin to have identical public registries regardless of actual installation location/reference
```
anaxceron, line 7: - Plug-in is hyphenated per the Zowe doc style
- Should end w/ a period
- Please include an issue or PR link to this changelog item

anaxceron (changes requested):

adam-wolfe:
I am seeing issues with the location being undefined for packages installed from npmjs:

image

anaxceron (approved):
Looks great, thanks @jace-roell

packages/imperative/src/imperative/__tests__/plugins/cmd/install/install.handler.unit.test.ts
```diff
@@ -379,5 +381,90 @@ describe("Plugin Management Facility install handler", () => {
...
+
+        try{
+            await handler.process(params);
+        }
+        catch(e){
+            expect(e).toBeUndefined();
+        }
+
+        expect(mocks.install).toHaveBeenCalledWith("@private/sample1","privateRegistryUrl");
+    });
```
t1m0thyj, line 416: This test seems redundant with the one above for public scope, unless the install handler takes a different code path based on whether the scope is public or private.

packages/imperative/src/imperative/src/plugins/cmd/install/install.handler.ts
```diff
@@ -71,112 +74,218 @@ export default class InstallHandler implements ICommandHandler {
...
 
-                    // write the json file when done if not the plugin json file
-                } else {
-                    for (const packageString of params.arguments.plugin) {
-                        params.response.console.log("\n_______________________________________________________________");
-                        const pluginName = await install(`${packageString}`, installRegistry);
-                        params.response.console.log("Installed plugin name = '" + pluginName + "'");
-                        params.response.console.log(runValidatePlugin(pluginName));
+                        if (plugin.startsWith("@")) {
+                            installRegistry = getScopeRegistry(
```
t1m0thyj, line 226: It seems like we are calling `getScopedRegistry` twice for scoped plug-ins, first in the loop above to add the location to `plugins.json`, and then again here. Since `getScopedRegistry` runs an npm cli command, it may take a long time to run in some environments. Would it make sense to cache the results of `getScopedRegistry` from earlier and reuse them here?

t1m0thyj (commented):
Looks pretty good, thanks @jace-roell!

Left a few comments that are optional, will approve after testing the changes 🙂

packages/imperative/src/imperative/__tests__/plugins/cmd/install/install.handler.unit.test.ts
jace-roell: resolved

packages/imperative/src/imperative/src/plugins/cmd/install/install.handler.ts
jace-roell: resolved

t1m0thyj (commented):
After updating from Zowe V2 to this branch, when running `zowe plugins list` I see `location: undefined` for all pre-existing plug-ins.

Discussed with @zFernand0:
> we could read the existing plugins.json file and if it has the `registry` property, we can print that value as `location` in the `zowe plugins list` command (with a small warning)

zFernand0 (commented):
As Timothy mentioned, we would want to make sure that these changes are backwards compatible with V2 😋 

The functionality looks great though 😋

traeok:
Running into the same issue with `registry: undefined` for pre-existing plug-ins.
Aside from that functionality LGTM, will approve once addressed 😋

traeok (approved):
I'm no longer seeing an `undefined` registry for pre-existing plugins. Thanks for addressing that 👍 

---

Now that I am seeing the intended behavior for pre-existing plugins, I am wondering if it would be more readable to display a single error at the bottom of the output, and adding a "symbol/character" at the end of each applicable registry?

For example, this is how it looks now with a couple plugins:
image

Users with several Zowe plugins might be caught off guard by numerous pre-existing plugin warnings that show after upgrading to v3.

Here's how that above idea could look in the command output:
image

This is only a suggestion though, not requesting changes here. I think it looks ready to merge as-is 😋
 
LGTM, thanks @jace-roell!

t1m0thyj (changes requested):
Several integration tests are failing in GitHub workflows. Could you revert changes to the snapshot files in this PR? I don't believe they should have changed at all, unless there are any related to `zowe plugins` commands.

I like the suggestion from @traeok to show a single warning at the end of output but agree it's not a required change. 😋

traeok (approved):
Latest changes LGTM, thanks Jace

t1m0thyj (approved):
LGTM, thanks @jace-roell!

zFernand0 (approved):
LGTM! 😋

PR #2214, opened 2024-08-01
[v3] Ssh keypassphrase prompt

jace-roell:
**What It Does**

Resolved bug that resulted in user not being prompted for a keyPassphrase if in the secure array of ssh profile. 
SshBaseHandler will now prompt user up to 3 times to enter the correct keyPassphrase in the case that the stored value is incorrect or no value is stored

I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Resolved bug that resulted in user not being prompted for a keyPassphrase if in the secure array of ssh profile. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
+- Enhancement: SshBaseHandler will now prompt user up to 3 times to enter the correct keyPassphrase in the case that the stored value is incorrect or no value is stored. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
+
 ## `8.0.0-next.202407262216`
 
 - Update: See `5.26.1` for details
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Resolved bug that resulted in user not being prompted for a keyPassphrase if in the secure array of ssh profile. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
+- Enhancement: SshBaseHandler will now prompt user up to 3 times to enter the correct keyPassphrase in the case that the stored value is incorrect or no value is stored. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
+
 ## `8.0.0-next.202407181255`
 
 - BugFix: Resolved bug that resulted in each plug-in to have identical public registries regardless of actual installation location/reference. [#2189](https://github.com/zowe/zowe-cli/pull/2189)
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,11 +2,6 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
-## Recent Changes
-
-- BugFix: Resolved bug that resulted in user not being prompted for a keyPassphrase if in the secure array of ssh profile. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
-- Enhancement: SshBaseHandler will now prompt user up to 3 times to enter the correct keyPassphrase in the case that the stored value is incorrect or no value is stored. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
-
 ## `8.0.0-next.202407262216`
 
 - Update: See `5.26.1` for details
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
-- BugFix: Resolved bug that resulted in user not being prompted for a keyPassphrase if in the secure array of ssh profile. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
+- BugFix: Resolved bug that resulted in user not being prompted for a key passphrase if it is located in the secure creditial array of the ssh profile. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
 
 ## `8.0.0-next.202407262216`
 
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Resolved bug that resulted in user not being prompted for a keyPassphrase if in the secure array of ssh profile. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
+
 ## `8.0.0-next.202407262216`
 
 - Update: See `5.26.1` for details
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
-- BugFix: Resolved bug that resulted in user not being prompted for a key passphrase if it is located in the secure creditial array of the ssh profile. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
+- BugFix: Resolved bug that resulted in user not being prompted for a key passphrase if it is located in the secure credential array of the ssh profile. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
 
 ## `8.0.0-next.202407262216`
 
--- a/packages/imperative/src/rest/__tests__/session/ConnectionPropsForSessCfg.unit.test.ts
+++ b/packages/imperative/src/rest/__tests__/session/ConnectionPropsForSessCfg.unit.test.ts
@@ -1690,7 +1690,7 @@ describe("ConnectionPropsForSessCfg tests", () => {
                     parms: parms as any,
                 }
             );
-        expect(ConnectionPropsForSessCfg.secureSessCfgProps).toContain("keyPassphrase");
+        expect((ConnectionPropsForSessCfg as any).secureSessCfgProps).toContain("keyPassphrase");
     });
     describe("getValuesBack private function", () => {
         // pretend that console.log works, but put data into a variable
--- a/packages/imperative/src/rest/__tests__/session/ConnectionPropsForSessCfg.unit.test.ts
+++ b/packages/imperative/src/rest/__tests__/session/ConnectionPropsForSessCfg.unit.test.ts
@@ -1,13 +1,13 @@
 /*
- * This program and the accompanying materials are made available under the terms of the
- * Eclipse Public License v2.0 which accompanies this distribution, and is available at
- * https://www.eclipse.org/legal/epl-v20.html
- *
- * SPDX-License-Identifier: EPL-2.0
- *
- * Copyright Contributors to the Zowe Project.
- *
- */
+* This program and the accompanying materials are made available under the terms of the
+* Eclipse Public License v2.0 which accompanies this distribution, and is available at
+* https://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
 
 jest.mock("../../../logger/src/LoggerUtils");
 
@@ -1679,7 +1679,7 @@ describe("ConnectionPropsForSessCfg tests", () => {
                 },
             },
         };
-        const sessCfgWithConnProps: extendedSession =
+        const sessCfgWithConnProps: ISshSession =
             await ConnectionPropsForSessCfg.addPropsOrPrompt<ISshSession>(
                 initialSessCfg,
                 args,
--- a/packages/imperative/src/rest/__tests__/session/ConnectionPropsForSessCfg.unit.test.ts
+++ b/packages/imperative/src/rest/__tests__/session/ConnectionPropsForSessCfg.unit.test.ts
@@ -1,13 +1,13 @@
 /*
-* This program and the accompanying materials are made available under the terms of the
-* Eclipse Public License v2.0 which accompanies this distribution, and is available at
-* https://www.eclipse.org/legal/epl-v20.html
-*
-* SPDX-License-Identifier: EPL-2.0
-*
-* Copyright Contributors to the Zowe Project.
-*
-*/
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
 
 jest.mock("../../../logger/src/LoggerUtils");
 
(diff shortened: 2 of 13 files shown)


packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Resolved bug that resulted in user not being prompted for a keyPassphrase if in the secure array of ssh profile. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
+- Enhancement: SshBaseHandler will now prompt user up to 3 times to enter the correct keyPassphrase in the case that the stored value is incorrect or no value is stored. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
```
anaxceron, line 8: Can we explain what SShBaseHandler is? Is it a property, a class, a manufacturer?: "Enhancement: The **property/schema/cat/whatever** now prompts the user up to three times to ..."

packages/zosuss/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Zowe z/OS USS SDK package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Resolved bug that resulted in user not being prompted for a keyPassphrase if in the secure array of ssh profile. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
+- Enhancement: SshBaseHandler will now prompt user up to 3 times to enter the correct keyPassphrase in the case that the stored value is incorrect or no value is stored. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
```
anaxceron, line 8: See my comment above

packages/imperative/CHANGELOG.md
t1m0thyj: Do we need this enhancement in the Imperative changelog? Since it's specific to `SshBaseHandler`, I think it might make sense to include only in the z/OS USS changelog.

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -127,13 +127,24 @@ export class ConnectionPropsForSessCfg {
                     if (cmdArgs[argName] != null) { (sessCfgToUse as any)[override.propertyName] = cmdArgs[argName]; }
                     for (const prop of override.propertiesOverridden) {
                         // Make sure we do not prompt for the overridden property.
-                        if (!doNotPromptForValues.includes(prop)) { doNotPromptForValues.push(prop); }
+                        if (!doNotPromptForValues.includes(prop as any)) { doNotPromptForValues.push(prop as any); }
```
t1m0thyj, line 130: If possible we should avoid using `as any` here. Tested locally and after removing `as any` was getting errors about type `string | number | symbol` not matching type `string`.

I think this can be fixed by restricting the type to `(keyof SessCfgType & string)[]` in 4 places (inspired by [StackOverflow](https://stackoverflow.com/a/54784203/5504760) 😋):
* `promptForValues` in this method
* `doNotPromptForValues` in this method
* `propertiesOverridden` on the `IOverridePromptConnProps` interface
* `propsToPromptFor` on the `IOverridePromptConnProps` interface

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -161,20 +172,31 @@ export class ConnectionPropsForSessCfg {
         this.loadSecureSessCfgProps(connOptsToUse.parms, promptForValues);
 
         if (connOptsToUse.getValuesBack == null && connOptsToUse.doPrompting) {
-            connOptsToUse.getValuesBack = this.getValuesBack(connOptsToUse);
+            connOptsToUse.getValuesBack = this.getValuesBack(connOptsToUse as any);
```
t1m0thyj, line 175: We might be able to remove the `as any` here by updating the method definition of `getValuesBack` to use a type generic like the following:
```typescript
private static getValuesBack<SessCfgType extends ISession=ISession>(connOpts: IOptionsForAddConnProps<SessCfgType>):
```

packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -58,22 +59,108 @@ export abstract class SshBaseHandler implements ICommandHandler {
...
+                                    propertyOverrides: sshSessCfgOverride,
+                                    supportedAuthTypes: [
+                                        SessConstants.AUTH_TYPE_BASIC,
+                                    ],
+                                    propsToPromptFor: [
+                                        {
+                                            name: "keyPassphrase",
+                                            isGivenValueValid: (givenValue: {
+                                                [key: string]: any;
+                                            }) => {
```
t1m0thyj, line 119: For the `isGivenValueValid` callback, curious why the argument needs to be a key-value object? Could it be passed as a string, and then referenced inside the function as `givenValue` instead of `givenValue.keyPassphrase`?

packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -58,22 +59,108 @@ export abstract class SshBaseHandler implements ICommandHandler {
...
+                                                let saveKP: boolean = true;
+                                                const result = utils.parseKey(
+                                                    fs.readFileSync(
+                                                        sshSessCfgWithCreds[
+                                                            "privateKey"
+                                                        ]
```
t1m0thyj, line 125: Please use dot notation for referencing properties when possible 😋 
```suggestion
                                                        sshSessCfgWithCreds.privateKey
```

packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -58,22 +59,108 @@ export abstract class SshBaseHandler implements ICommandHandler {
...
+                        success = true;
+                    } catch (retryError) {
+                        this.console.log(
+                            "\n" +
+                                `Key passphrase authentication failed! (${
+                                    attempt + 1
+                                }/${maxAttempts})` +
+                                "\n"
+                        );
+                        attempt++;
```
t1m0thyj, line 154: The `attempt` variable could be incremented directly inside the string interpolation expression:
```suggestion
                        this.console.log(
                            "\n" +
                                `Key passphrase authentication failed! (${
                                    ++attempt
                                }/${maxAttempts})` +
                                "\n"
                        );
```

Not necessarily suggesting to change this, may want to keep the original syntax if you prefer incrementing on a separate line for readability 😋

t1m0thyj (dismissed):
Looks pretty good, thanks @jace-roell! Left a few comments

anaxceron (approved):
Looks good @jace-roell, thank you!

packages/imperative/src/rest/src/session/doc/IOverridePromptConnProps.ts
```diff
@@ -34,5 +34,11 @@ export interface IOverridePromptConnProps {
      * Prompting logic is only in place for host, port, user, and password, but cert, certKey, tokenType, and tokenValue may also need
      * to be overridden.
      */
-    propertiesOverridden: (keyof ISession)[];
+    propertiesOverridden: (keyof SessCfgType & string)[];
+
+    /**
+     * Allows for passing for additional properties to prompt for.
+     * Utilized in the case of a incorrect/not stored key passphrase.
```
zFernand0, line 41:
```suggestion
     * Allows passing additional properties for which to prompt.
     * Used in cases of an incorrect or missing key passphrase.
```

packages/imperative/src/rest/src/session/doc/IOptionsForAddConnProps.ts
```diff
@@ -51,7 +50,19 @@ export interface IOptionsForAddConnProps {
      * Specifies a list of authentication properties, and what they should override.
      * If one of these properties is available on the session, do not prompt for the other property.
      */
-    propertyOverrides?: IOverridePromptConnProps[];
+    propertyOverrides?: IOverridePromptConnProps<SessCfgType>[];
+
+     /**
+     * Allows for passing for additional properties to prompt for.
+     * Utilized in the case of a incorrect/not stored key passphrase. s
+     */
```
zFernand0, line 58:
```suggestion
      * Allows passing additional properties for which to prompt.
      * Used in cases of an incorrect or missing key passphrase.
      */
```

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -134,6 +134,17 @@ export class ConnectionPropsForSessCfg {
             }
         }
 
+        // Set default values on propsToPromptFor
+        if(connOpts.propsToPromptFor?.length > 0)
+        {
+            connOpts.propsToPromptFor.forEach(obj => {
```
zFernand0, line 140: My apologies for the confusion.
During the Zowe CLI standup (Aug 5th) I mentioned that this could be ommitted for coverage purposes.

I was wrong. For some reason I was under the impression that the sonar cloud reports where complaining about the SshBaseHandler statements not being covered.

I think it would be good to add a couple of tests for this 🙏🏽

zFernand0 (dismissed):
LGTM! 😋 

Left a small request for changes in the wording of the TSDoc.
Also, I'm sorry about the confusion on the tests discussion during the CLI standup this morning. We probably want to add a small test to make sure the `ConnectionPropsForSessCfg` handles the new functionality properly even though it's optional.

packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -58,22 +59,103 @@ export abstract class SshBaseHandler implements ICommandHandler {
...
+        } catch (e) {
+            this.console.log("Initial key passphrase authentication failed!" + "\n");
+            if (
+                e.message.includes("but no passphrase given") ||
+                e.message.includes("bad passphrase?")
+            ) {
+                const maxAttempts = 3;
+                let attempt = 0;
+                let success = false;
+                while (attempt < maxAttempts && !success) {
```
traeok, line 102: Not suggesting any changes for this pull request, but I could see parts of this logic being valuable for "prompting w/ `x` attempts" in other areas of the CLI 😋

Maybe at some point and if the demand is there, this code could be generalized and pulled into a utility class/function so it can be leveraged for different features.

packages/imperative/src/rest/src/session/doc/IOptionsForAddConnProps.ts
```diff
@@ -51,7 +50,19 @@ export interface IOptionsForAddConnProps {
...
+     * Allows passing additional properties for which to prompt.
+     * Used in cases of an incorrect or missing key passphrase.
+     */
+    propsToPromptFor?:
+    {
+        name: keyof SessCfgType & string,
+        secure?: boolean,
+        description?: string,
+        isGivenValueValid?: (givenValue: string) => boolean
+    }[];
```
traeok, line 65: Could this be defined as an interface/type, or do we feel that defining it as one would be redundant?
If we only intend on using the typed object within this interface, then we could probably leave it as-is.

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -338,7 +360,7 @@ export class ConnectionPropsForSessCfg {
      * NOTE(Kelosky): redundant from LoggerUtils.SECURE_PROMPT_OPTIONS - leaving
      * for future date to consolidate
      */
-    private static secureSessCfgProps: Set<string> = new Set(["user", "password", "tokenValue", "passphrase"]);
+    public static secureSessCfgProps: Set<string> = new Set(["user", "password", "tokenValue", "passphrase"]);
```
traeok, line 363: Quick question: Was this a required change to access to the set of secure props outside of this class?
I'm not opposed to doing this, but figured I would double-check, as I only see references to `this.secureSessCfgProps` in this PR (which would imply that we could leave it as `private` or `protected`).

packages/imperative/src/rest/__tests__/session/ConnectionPropsForSessCfg.unit.test.ts
zFernand0, line 1: I'm very curious about your IDE settings. 
It seems that it has a limit of < 100 columns per line, or it enforces closures to be properly indented (as opposed to being in-line). 😋 

Not suggesting to change it back, but for those reviewing, you might want to review this PR while hiding whitespaces 😋

jace-roell: I believe I used a formatter on ConnectionPropsForSessCfg.unit.test.ts which might be what you are referring to.

gejohnston (approved):

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -307,7 +329,7 @@ export class ConnectionPropsForSessCfg {
             impLogger.debug("Using basic authentication");
             sessCfg.type = SessConstants.AUTH_TYPE_BASIC;
         }
-        ConnectionPropsForSessCfg.setTypeForTokenRequest(sessCfg, connOpts, cmdArgs.tokenType);
+        ConnectionPropsForSessCfg.setTypeForTokenRequest(sessCfg, connOpts as any, cmdArgs.tokenType);
```
zFernand0, line 332:
```suggestion
        ConnectionPropsForSessCfg.setTypeForTokenRequest<SessCfgType>(sessCfg, connOpts, cmdArgs.tokenType);
```

I think we can remove this `as any` here if we modify the signature of `setTypeForTokenRequest`

```typescript
    private static setTypeForTokenRequest<SessCfgType extends ISession=ISession>(
        sessCfg: SessCfgType,
        options: IOptionsForAddConnProps<SessCfgType>,
        tokenType: SessConstants.TOKEN_TYPE_CHOICES
    )
```

packages/imperative/src/rest/src/session/doc/IOptionsForAddConnProps.ts
zFernand0: Yeah, I do see some value in cleaning this up a little.
```suggestion
    IPropsToPromptFor<SessCfgType>[];
```

And then import...

```typescript
import { IPropsToPromptFor } from "./IPropsToPropmtFor";
```

Which could be similar to `IOverridePromptConnProps` but with the following contents...

```typescript
export interface IPropsToPromptFor <SessCfgType extends ISession=ISession> {
  name: keyof SessCfgType & string,
  secure?: boolean,
  description?: string,
  isGivenValueValid?: (givenValue: string) => boolean
}
```

zFernand0 (changes requested):
LGTM! 😋 
Left a couple of comments in case you would like to make a few optional modifications to remove the newly added `as any` and to clean up the interface as Trae suggested 😋

packages/imperative/src/rest/src/session/doc/IPropsToPromptFor.ts
```diff
@@ -0,0 +1,8 @@
+import { ISession } from "../../..";
```
zFernand0, line 1: two quick things on this file...

1. We are missing the license header
2. I think it might be good to import from the specific file (or at least some more specific location) to avoid possible circular references in the future 😋

jace-roell: resolved

traeok (approved):

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Resolved bug that resulted in user not being prompted for a keyPassphrase if in the secure array of ssh profile. [#1770](https://github.com/zowe/zowe-cli/issues/1770)
```
anaxceron, line 7: We should spell out what is in the secure array to make this sentence more clear

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -358,7 +380,7 @@ export class ConnectionPropsForSessCfg {
      * @param connOpts Options for adding connection properties
      * @returns Name-value pairs of connection properties
      */
-    private static getValuesBack(connOpts: IOptionsForAddConnProps): (properties: string[]) => Promise<{ [key: string]: any }> {
+    private static getValuesBack<SessCfgType extends ISession=ISession>(connOpts: IOptionsForAddConnProps<SessCfgType>): (properties: string[]) => Promise<{ [key: string]: any }> {
```
zFernand0, line 383: Sorry for being picky, and this is not really a required change, but this line gets flagged by the linter as a warning because it's longer than 150 characters 😋

zFernand0 (approved):
LGTM! 😋

t1m0thyj (commented):
Changes LGTM, thanks @jace-roell! Will approve after changelogs have been updated and I have done manual testing 🙂

packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -58,22 +59,103 @@ export abstract class SshBaseHandler implements ICommandHandler {
...
+            );
         this.mSession = new SshSession(sshSessCfgWithCreds);
 
         this.mArguments = commandParameters.arguments;
-        await this.processCmd(commandParameters);
+
+        try {
+            await this.processCmd(commandParameters);
+        } catch (e) {
+            this.console.log("Initial key passphrase authentication failed!" + "\n");
```
t1m0thyj, line 94: Should we move this `log` statement inside the `if` block, since there are other kinds of errors that may occur?

packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -58,22 +59,103 @@ export abstract class SshBaseHandler implements ICommandHandler {
...
+                                "\n"
+                        );
+                        if (attempt >= maxAttempts) {
+                            throw new Error(
+                                "Maximum retry attempts reached. Authentication failed."
+                            );
+                        }
+                    }
+                }
+            }
```
t1m0thyj, line 157: In case the error is a different type (not related to bad passphrase), we should throw the original error message to aid in troubleshooting:
```suggestion
            } else {
                throw e;
            }
```

PR #2217, opened 2024-08-06
Use Fewer Deprecated Functions in Master

### LINKED ISSUE #2191: Remove deprecated methods in V2
When running Zowe CLI in Node 22, most commands display `DeprecationWarning` multiple times.

I would expect no deprecation warnings to be shown.

Implementation:
* We should remove deprecated utils functions as we did in V3:
    https://github.com/zowe/zowe-cli/pull/2169
* The ESLint `deprecation/deprecation` rule can be used to catch all usages of deprecated functions.
* We may want to leave references to `process.mainModule` as long as it does not cause deprecation warnings.

awharn:
**What It Does**

Resolves #2191
Reduces the use of deprecated functions in preparation to support Node 22

**How to Test**

Install, build, run commands to verify new build does not trigger deprecation warnings on Node 22

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


awharn:
Please wait to merge - system tests are in progress.

__tests__/__packages__/cli-test-utils/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI test utils package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Refactored code to reduce the use of deprecated functions to prepare for Node 22 support. [#2191](https://github.com/zowe/zowe-cli/issues/2191)
```
anaxceron, line 7: Can we explain why we need Node 22 support?

awharn:
System Tests are complete and pass.

__tests__/__packages__/cli-test-utils/CHANGELOG.md
awharn: I don't quite know how to explain it in the changelog. The CLI supports running on all Node LTS versions that are actively supported by the NodeJS project maintainers, and Node 22 will reach that point in October. That commitment is documented in the Zowe CLI system requirements. This PR sets the ground work to honor that commitment.

anaxceron: How about: " ... upcoming NodeJS 22 support"

awharn: Sounds good. Changelogs updated. Thanks!

anaxceron (approved):
Looks good, thanks @awharn!

packages/provisioning/__tests__/tsconfig.json
zFernand0, line 1: Is there a way we could avoid having to maintain a separate tsconfig.json file in every package?

I'm thinking of techniques similar to how jest config can be combined.

Not really requesting changes, though! 😋

packages/imperative/tsconfig-tests.json
zFernand0, line 1: Should this be moved inside the __tests__ directory and be renamed to tsconfig.json
Just to be consistent with other packages. 😋

packages/imperative/src/utilities/src/TextUtils.ts
```diff
@@ -274,7 +274,7 @@ export class TextUtils {
      * @returns {string} - a formatted string with the variables inserted
      */
     public static formatMessage(message: string, ...values: any[]): string {
-        if (!isNullOrUndefined(values)) {
+        if (!(values == null)) {
```
zFernand0, line 277: There are multiple occasions (in this file and maybe others) where we do `!(a == null)` instead of `a != null` Curious if we could replace them with the shorthand version of `!= null` 😋

packages/cli/src/zosfiles/create/Create.options.ts
```diff
@@ -36,6 +36,7 @@ export const ZosFilesCreateExtraOptions: { [key: string]: ICommandOptionDefiniti
      * @type {ICommandOptionDefinition}
      * @deprecated
      */
+    // eslint-disable-next-line deprecation/deprecation
```
t1m0thyj, line 39: Not requesting to change anything, but it seems weird that `eslint-disable` comment is needed here, since we're just defining a deprecated variable not referencing it elsewhere. 🤔

packages/imperative/src/security/__tests__/DefaultCredentialManager.unit.test.ts
zFernand0, line 1: Thanks for adding the deprecation to each individual line.
I'm sure it was very tedious to do 😋

packages/imperative/src/profiles/src/abstract/AbstractProfileManager.ts
```diff
@@ -200,12 +200,12 @@ export abstract class AbstractProfileManager<T extends IProfileTypeConfiguration
         if (parms.loadCounter != null) {
             this.mLoadCounter = parms.loadCounter;
         }
-        this.mLogger = isNullOrUndefined(parms.logger) ? this.mLogger : parms.logger;
+        this.mLogger = parms.logger == null ? this.mLogger : parms.logger;
```
zFernand0, line 203: It is possible that this comment may apply to other files...

Curious if we should take advantage of other operators like `??` 😋

packages/imperative/src/profiles/src/abstract/AbstractProfileManager.ts
```diff
@@ -200,12 +200,12 @@ export abstract class AbstractProfileManager<T extends IProfileTypeConfiguration
...
         this.mProfileType = parms.type;
         this.mProfileRootDirectory = parms.profileRootDirectory;
         this.mProfileTypeConfigurations = parms.typeConfigurations;
         this.mProductDisplayName = parms.productDisplayName;
-        if (isNullOrUndefined(this.profileTypeConfigurations) || this.profileTypeConfigurations.length === 0) {
+        if (this.profileTypeConfigurations == null || this.profileTypeConfigurations.length === 0) {
```
zFernand0, line 208: Similar to the previous comment...

what about using `?.` 😋 

There are edge cases where `?.` may not be desired since it can technically change the behavior. For example, `if (a != null && !a.booleanValue)` is not the same as `if (!a?.booleanValue)` since the condition will evaluate to true if `A` is undefined 😋

packages/imperative/src/imperative/__tests__/Imperative.unit.test.ts
```diff
@@ -21,6 +21,7 @@ import * as yargs from "yargs";
 import { ImperativeError } from "../../error/src/ImperativeError";
 
 describe("Imperative", () => {
+    // eslint-disable-next-line deprecation/deprecation
     const mainModule = process.mainModule;
```
zFernand0, line 25: Curious if there is a way to configure the deprecation/deprecation eslint plug-in to ignore all uses of `process.mainModule` 😋

packages/imperative/package.json
```diff
@@ -30,9 +30,9 @@
   "scripts": {
     "build:packages": "tsc --pretty",
     "build": "npm run build:packages && npm run build:webHelp",
-    "lint": "eslint \"packages/**/*.ts\" \"**/__tests__/**/*.ts\"",
-    "lint:packages": "eslint \"packages/**/*.ts\" --ignore-pattern \"**/__tests__/**/*.ts\"",
-    "lint:tests": "eslint \"**/__tests__/**/*.ts\"",
+    "lint": "eslint \"src/**/*.ts\" \"**/__tests__/**/*.ts\"",
+    "lint:packages": "eslint \"src/**/*.ts\" --ignore-pattern \"**/__tests__/**/*.ts\"",
+    "lint:tests": "eslint \"./__tests__/**/*.ts\"",
```
t1m0thyj, line 35: Don't we need to keep the `**` to include all tests because there are some inside subdirectories such as `packages/imperative/logger/__tests__`?

packages/zosmf/src/ZosmfBaseHandler.ts
```diff
@@ -40,11 +40,13 @@ export abstract class ZosmfBaseHandler implements ICommandHandler {
...
      * Loaded z/OSMF profile if needed
+     * @deprecated
      */
     protected mZosmfProfile: IProfile;
 
     /**
      * Loaded z/OSMF profile with meta information
+     * @deprecated
      */
     protected mZosmfLoadedProfile: IProfileLoaded;
```
zFernand0, line 51: Are this newly deprecated properties?

I believe they still exist in the next branch 😋 

Should we create an issue to remove them/replace them over there?

packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -41,6 +41,7 @@ export abstract class SshBaseHandler implements ICommandHandler {
 
     /**
      * Loaded z/OS SSH profile if needed
+     * @deprecated
      */
     protected mSshProfile: IProfile;
```
zFernand0, line 46: FYI, compared to the previous comment, this one seems to be gone form the next branch already 😋

packages/imperative/src/utilities/src/TextUtils.ts
```diff
@@ -39,7 +39,7 @@ export class TextUtils {
     public static getRecommendedWidth(preferredWidth: number = TextUtils.DEFAULT_WRAP_WIDTH): number {
         const widthSafeGuard = 8; // prevent partial words from continuing over lines
         const yargs = require("yargs");
-        const maxWidth = !isNullOrUndefined(yargs.terminalWidth() && yargs.terminalWidth() > 0) ?
+        const maxWidth = !((yargs.terminalWidth() && yargs.terminalWidth() > 0) == null) ?
```
t1m0thyj, line 42: This accurately reflects the old logic, but I think the old logic may have been wrong because the boolean statement won't evaluate to `null`. Should we change it to the following instead to match the vNext branch?
```suggestion
        const maxWidth = (!(yargs.terminalWidth() == null) && yargs.terminalWidth() > 0) ?
```

zFernand0 (approved):
LGTM! 😋 

All of my comments are mostly minor things that can be done at a later time, or be forgotten since they should not impact performance or behavior 😋

t1m0thyj (changes requested):
Looks pretty good, thanks @awharn!

packages/provisioning/__tests__/tsconfig.json
awharn: We made the same change in Next, but no, I don't think there is anything we can do. ESLint expects a TSConfig everywhere it scans, and you need different TSConfigs between tests and source. We could try to have them all extend a common TSConfig, but that would need to be a separate effort in V3. I don't expect we will change any of these TSConfigs through the rest of V2.

packages/imperative/src/imperative/__tests__/Imperative.unit.test.ts
awharn: There is not. I checked.

packages/zosmf/src/ZosmfBaseHandler.ts
awharn: I only deprecated them because the code around them didn't actually do anything, and the process of setting them used deprecated functions.

packages/imperative/src/profiles/src/abstract/AbstractProfileManager.ts
awharn: I was just doing a find and replace. I was not evaluating any of the other logic. I would think if we want to change something like this, it might make more sense as another issue in order to reduce scope creep.

packages/imperative/src/profiles/src/abstract/AbstractProfileManager.ts
awharn: Same as previous comment.

packages/cli/src/zosfiles/create/Create.options.ts
awharn: Yep, it is weird. But it really doesn't like that we are defining a deprecated option.

packages/.madgerc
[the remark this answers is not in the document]
awharn: Doing this kind of got lint stuck in an infinite loop. So I am going to say no, til I can figure out why this is happening. Or we can just leave it in V2 and not worry about it.

t1m0thyj (approved):
LGTM, thanks @awharn!

gejohnston (approved):
Thanks for the painstaking effort.

packages/zosmf/src/ZosmfBaseHandler.ts
awharn: After in person discussion, we are planning to remove these in V3 since nothing sets the protected variables anymore.

PR #2218, opened 2024-08-06
Fix error using session with auth type "none"

### LINKED ISSUE #2219: `RestClient.buildOptions` doesn't support session with `AUTH_TYPE_NONE`
**Describe the bug**

Since CLI v7.24.1, issuing a request thru the Imperative REST client with `type = "none"` fails with an error:
`ImperativeError: No credentials for a BASIC or TOKEN type of session.`

Credentials should not be required when the session type is `AUTH_TYPE_NONE`, and the following code should work:
```javascript
const { RestClient, Session } = require("@zowe/imperative");
const mySession = new Session({
    hostname: "example.com",
    port: 443,
    type: "none"
});
const restClient = new RestClient(mySession);
console.log(restClient.buildOptions("/api/v1", "GET"));
```

**Describe your environment**

**Additional context**

t1m0thyj:
**What It Does**

Fixes #2219

**How to Test**

Try issuing a REST request with no credentials on the session object:
```javascript
const { RestClient, Session } = require("@zowe/imperative");
const mySession = new Session({
    hostname: "example.com",
    port: 443,
    type: "none"
});
const restClient = new RestClient(mySession);
restClient.request({ resource: "/", request: "GET" }).then(console.log);
```

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**


zFernand0 (approved):
The change makes sense to me conceptually speaking.
I may need some help understanding it a bit better. 🙏🏽 

LGTM though 😋

traeok (approved):
Changes look good to me, thanks Timothy for the quick fix

anaxceron (approved):
amazing!

zFernand0 (approved):
Had a bit more time to test.
Thanks @t1m0thyj for helping! 🙏🏽 

LGTM! 😋 

Tested in 7.24.0 and this PR and they work as expected (no errors thrown).

Tested with 7.24.1 - a couple of versions in between - 7.28.2, and the latest vNext version and we get the error 😋 

Also did some testing with varios Zowe bundles 
2.9 - 2.15 (versions before CLI-7.24.1) and they all worked 😋 
Then, 2.16  started to fail (as expected)

gejohnston (approved):
I agree with the implementation using AUTH_TYPE_NONE.  AUTH_TYPE_NONE does not fall into our order-of-precedence. Its presence is an indicator that the items in our order-of-precedence do not even apply to the action.

I suppose AUTH_TYPE_NONE could be tested earlier and we could skip many of the credential checks, but then we would still have to set the headers and flags at the end of the function. Placing the test for AUTH_TYPE_NONE on an existing test (as you have done) minimizes the changes that have to be made, and may even make the logic easier for a human to follow.

awharn (approved):

PR #2229, opened 2024-08-12
Create web help lockfile only when publishing package

t1m0thyj:
**What It Does**

Fixes the nightly CLI bundle workflow which has still been failing: https://github.com/zowe/zowe-cli-standalone-package/actions/workflows/zowe-cli-bundle-all.yaml

We can generate the web help lockfile in the Imperative `prepublishOnly` script rather than in the web-help `prepare` script. This should limit the script so it only runs just before the Imperative package is published, and matches the way that we generate the CLI shrinkwrap file which also uses a `prepublishOnly` script 😋 

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/package.json
+++ b/packages/imperative/package.json
@@ -41,7 +41,8 @@
     "typedoc": "typedoc --options ./typedoc.json ./src/",
     "typedocSpecifySrc": "typedoc --options ./typedoc.json",
     "prepack": "node ../../scripts/prepareLicenses.js",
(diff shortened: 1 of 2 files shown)


traeok (approved):
LGTM, thanks Timothy for the fix

awharn:
Thanks for the fix @t1m0thyj! 🚀

PR #2232, opened 2024-08-12
[v2] ssh keypassphrase port from v3

jace-roell:
**What It Does**
Resolved bug that resulted in user not being prompted for a key passphrase if it is located in the secure credential array of the ssh profile. (port from v3)

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
+++ b/packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
@@ -97,7 +97,7 @@ export class ConnectionPropsForSessCfg {
     public static async addPropsOrPrompt<SessCfgType extends ISession>(
         initialSessCfg: SessCfgType,
         cmdArgs: ICommandArguments,
-        connOpts: IOptionsForAddConnProps = {}
+        connOpts: IOptionsForAddConnProps <SessCfgType> = {}
     ): Promise<SessCfgType> {
         const impLogger = Logger.getImperativeLogger();
 
@@ -113,8 +113,8 @@ export class ConnectionPropsForSessCfg {
         );
 
         // This function will provide all the needed properties in one array
-        const promptForValues: (keyof ISession)[] = [];
-        const doNotPromptForValues: (keyof ISession)[] = [];
+        let promptForValues: (keyof SessCfgType & string)[] = [];
+        const doNotPromptForValues: (keyof SessCfgType & string)[] = [];
 
         /* Add the override properties to the session object.
          */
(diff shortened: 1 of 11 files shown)


packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -15,36 +15,29 @@ import {
...
     /**
      * The session creating from the command line arguments / profile
      */
     protected mSession: SshSession;
 
-    /**
-     * Loaded z/OS SSH profile if needed
-     * @deprecated
-     */
-    protected mSshProfile: IProfile;
```
t1m0thyj, line 46: In V2, we can't make the breaking change to remove `mSshProfile` so we should keep it (as deprecated) for now.

packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -65,26 +58,108 @@ export abstract class SshBaseHandler implements ICommandHandler {
      */
     public async process(commandParameters: IHandlerParameters) {
         this.mHandlerParams = commandParameters;
-        // Why is this here? NOTHING uses it, but I suppose an extender MIGHT be... -awharn
-        // eslint-disable-next-line deprecation/deprecation
-        this.mSshProfile = commandParameters.profiles.get("ssh", false);
```
t1m0thyj, line 70: Same comment as above - we should keep this for V2

t1m0thyj (changes requested):
Looks pretty good, thanks for porting the changes @jace-roell! Left a few comments

packages/zosuss/src/SshBaseHandler.ts
jace-roell: resolved

anaxceron (approved):

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -366,7 +386,7 @@ export class ConnectionPropsForSessCfg {
                 let answer;
                 while (answer === undefined) {
                     const hideText = profileSchema[value]?.secure || this.secureSessCfgProps.has(value);
-                    let promptText = `${this.promptTextForValues[value]} ${serviceDescription}`;
+                    let promptText = `${this.promptTextForValues[value] ?? `Enter your ${value} for`} ${serviceDescription}`;
```
traeok, line 389: SonarCloud is flagging this line for nested template literals. Thoughts on making these changes to resolve the issue?

```suggestion
                    const valuePrompt = this.promptTextForValues[value] ?? `Enter your ${value} for`;
                    let promptText = `${valuePrompt} ${serviceDescription}`;
```

traeok (commented):
LGTM, thanks Jace for porting this over - I left a suggestion about a SonarCloud issue.

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
jace-roell: resolved

traeok (approved):

t1m0thyj (commented):
According to Zowe docs, the order of precedence for SSH commands is SSH key, then user & password:
image

When testing this PR locally if I've stored values for user and password in my base profile, and have an SSH profile that defines privateKey and contains keyPassphrase in the secure array, then I get prompted for a value for keyPassphrase as expected. But if I remove user and password from my base profile and leave keyPassphrase, then issue an SSH command, I am prompted for user and password which should not be needed.

I see the same behavior in Zowe V3, so not sure if this is actually an issue, and if so whether we want to handle it separately outside of this PR? 😋

t1m0thyj (approved):
LGTM, thanks @jace-roell!

Sorry for the false alarm - after further discussion realized that the command behaves as expected, and my test environment was tainted by a global config file.

awharn (approved):

PR #2236, opened 2024-08-14
[v3] Update next for deprecation replacement

jace-roell:
**What It Does**
Modified showMsgWhenDeprecated() to allow an empty string as a parameter when no replacement is available for deprecated command to print an alternative message.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/packages/imperative/src/utilities/src/CliUtils.ts
+++ b/packages/imperative/src/utilities/src/CliUtils.ts
@@ -373,7 +373,7 @@ export class CliUtils {
         let message = "";
         if (deprecatedReplacement != null) {
             const noNewlineInText = deprecatedReplacement.replace(/\n/g, " ");
-            if(commandWarning) message += ("\n\nWarning: This " + commandType + " has been deprecated.\n");
+            if(commandWarning) message += "\n\nWarning: This " + commandType + " has been deprecated.\n";
             if (deprecatedReplacement === "") {
                 message += "Obsolete component. No replacement exists";
             } else {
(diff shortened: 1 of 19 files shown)


packages/imperative/src/utilities/src/CliUtils.ts
```diff
@@ -378,20 +392,14 @@ export class CliUtils {
...
-            }
-
+            const oldCmd = handlerParms.positionals.join(" ");
             // display the message
             handlerParms.response.console.error("\nWarning: The command '" + oldCmd + "' is deprecated.");
-            handlerParms.response.console.error("Recommended replacement: " +
-                handlerParms.definition.deprecatedReplacement);
+            // Use consolidated deprecated message logic
+            const deprecatedMessage = CliUtils.generateDeprecatedMessage(handlerParms.definition);
+            handlerParms.response.console.error(deprecatedMessage);
```
t1m0thyj, line 402: Previously the `deprecatedReplacement` was printed as-is, now with the consolidated logic it will have newlines removed: `cmdDefinition.deprecatedReplacement.replace(/\n/g, " ")`. Is this change intentional?

t1m0thyj: Previously `DefaultHelpGenerator` stripped out new lines when parsing the deprecation message, and `CommandProcessor` kept them. I like that the new behavior is consistent 👍 

https://github.com/zowe/zowe-cli/blob/e10af2e2eaeafcbe3f68762405746a8b801bc7b0/packages/imperative/src/utilities/src/CliUtils.ts#L376

Not sure which we prefer - stripping out new lines vs keeping them. It may not matter much since I don't recall seeing a deprecation message that contains new lines 😋

t1m0thyj (approved):
LGTM, thanks @jace-roell!

gejohnston (approved):

PR #2237, opened 2024-08-20
Fix Yargs parsing strings as numbers

### LINKED ISSUE #1881: Fix yargs parsing string arguments as numbers
TL;DR Yargs parses string arguments as numbers if they look like numbers, which may be unexpected behavior for methods that expect string values.

> After seeing the context of zowe/zowe-cli#1883, I think I understand why 🙂 It seems like the default behavior of the yargs parser is to parse any options that look like numbers as numbers ([source](https://github.com/yargs/yargs-parser#parse-numbers)). This is undesirable behavior for jobid which should be parsed as a string regardless of whether a user enters it in the format `JOB1234` or just `1234`. Is this correct?
> 
> The solution in this PR will fix the issue for this command and any command handlers that call `toBeDefinedAndNonBlank`. But I wonder if there are any other places where we may hit a similar issue if command handlers call other methods that are expecting a string and receive a number instead.
>
> I am ok with approving this PR as-is if we believe it is the best solution. Alternatively, the behavior of yargs could be overridden (see [here](https://github.com/yargs/yargs/blob/main/docs/api.md#parserConfiguration)) to only parse options of type `number` as numbers and never parse items of type `string`. I don't know if we would consider that a bug fix, or a potentially risky breaking change?

_Originally posted by @t1m0thyj in https://github.com/zowe/imperative/pull/857#pullrequestreview-1049752029_

awharn:
**What It Does**

Prevents parsing string arguments as numbers if the value provided only consists of numerics.
Resolves #1881

**How to Test**

Run the `zowe files download ds "FAKE.DS" --rto 50 --encoding 1047 --show-inputs-only` command and observe the response timeout is blue (number) and encoding is white (string).

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/src/cmd/src/yargs/YargsConfigurer.ts
+++ b/packages/imperative/src/cmd/src/yargs/YargsConfigurer.ts
@@ -40,6 +40,11 @@ export class YargsConfigurer {
     ) {
     }
 
+    public static yargsConfiguration: Record<string, boolean> = {
+        "parse-numbers": false,
+        "parse-positional-numbers": false
+    };
+
     public configure() {
 
         /**
@@ -50,6 +55,7 @@ export class YargsConfigurer {
         this.yargs.help(false);
         this.yargs.version(false);
         this.yargs.showHelpOnFail(false);
+        this.yargs.parserConfiguration(YargsConfigurer.yargsConfiguration);
         // finally, catch any undefined commands
         this.yargs.command({
             command: "*",
--- a/packages/imperative/src/cmd/src/yargs/YargsConfigurer.ts
+++ b/packages/imperative/src/cmd/src/yargs/YargsConfigurer.ts
@@ -40,7 +40,7 @@ export class YargsConfigurer {
     ) {
     }
 
-    public static readonly yargsConfiguration: Record<string, boolean> = {
+    public static readonly yargsConfiguration: Readonly<Record<string, boolean>> = {
         "parse-numbers": false,
         "parse-positional-numbers": false
     };
--- a/packages/imperative/src/cmd/src/yargs/YargsConfigurer.ts
+++ b/packages/imperative/src/cmd/src/yargs/YargsConfigurer.ts
@@ -40,7 +40,7 @@ export class YargsConfigurer {
     ) {
     }
 
-    public static yargsConfiguration: Record<string, boolean> = {
+    public static readonly yargsConfiguration: Record<string, boolean> = {
         "parse-numbers": false,
         "parse-positional-numbers": false
     };
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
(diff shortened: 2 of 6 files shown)


awharn:
The reason we took this approach was because of a regression observed if we passed numbers to Yargs as number typed options. If a user supplied an option that took a number and provided no value `i.e zowe files list ds "FAKE.DS" --rto --encoding 939`, the handler validation would completely ignore the specified numeric option with no value. 

That is because yargs returns `undefined` for number options specified with no value, but returns "" for string options specified with no value. Therefore, our validation can't detect the missing value when it is a number, since it is the same as if it wasn't specified at all.

This behavior would be a regression with no easy fix. In this implementation, we continue to treat numerics as a string until validation is done to confirm the option was not specified with an empty value (""), and the value provided is numeric. Once that validation is completed, the value of the argument is converted from string to number.

adam-wolfe:
Looks good, but is it safe for us to assume that all numeric values will be integers?

packages/imperative/src/cmd/src/yargs/YargsConfigurer.ts
```diff
@@ -40,6 +40,11 @@ export class YargsConfigurer {
     ) {
     }
 
+    public static yargsConfiguration: Record<string, boolean> = {
```
t1m0thyj, line 43: SonarCloud is suggesting to define this as `public static const` or `public static readonly`.

t1m0thyj (approved):
LGTM, thanks @awharn!

traeok (approved):
LGTM, thanks @awharn for the fix!

packages/imperative/src/cmd/src/yargs/YargsConfigurer.ts
traeok: +1 for `public static readonly` as `public static const` doesn't work in TypeScript 😋 

Although `readonly` will prevent us from changing the entire record, you can still technically update keys inside it. Since the parsing configuration is only defined once and isn't modified after initialization, we could wrap the type in `Readonly`, and TypeScript will emit an error when any changes are made to the record:

```suggestion
    public static yargsConfiguration: Readonly<Record<string, boolean>> = {
```

image

adam-wolfe (approved):
LGTM, thanks @awharn

gejohnston (approved):

awharn:
Found a small bug - the change was not converting the values assigned to the alternate option names (aliases and camel case) to numbers, which caused system tests to fail for creating data-sets. Regression tests added, and am now using the `CliUtils.setOptionValue` function to set all options, camel and kebab case options, aliases, and camel and kebab case aliases to the same (number) value.

Re-requested reviews from all reviewers.

gejohnston (approved):

traeok (approved):

t1m0thyj (approved):
LGTM, thanks @awharn for updating this to handle aliases too! 😋

zFernand0 (approved):
Changes LGTM! 😋

janan07 (approved):

PR #2240, opened 2024-08-22
[v3] Tso issue command

jace-roell:
**What It Does**
Deprecated `IssueTsoCommand()` function and replaced with `IssueTsoCmd()` for compatibility with z/OS version 2.4.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog


packages/zosmf/src/CheckStatus.ts
```diff
@@ -36,6 +36,10 @@ export class CheckStatus {
         return ZosmfRestClient.getExpectJSON(session, infoEndpoint);
     }
 
+    public static async isZosVersionGreaterThan(session: AbstractSession, version: string): Promise<boolean> {
```
traeok, line 39: Thanks for adding this! We might also be able to leverage this in Zowe Explorer for some pre-existing features 😀 

_Reviewer note:_ At first I was thinking that `version` could be more strongly typed - it would prevent anyone from calling this function with a non-existent or unsupported version, but it would also limit this function's usability to versions that the SDK supports, and would require updating it for each z/OSMF release. I thought of a type that would grab all the version numbers from the string, but due to the caveats above, feel free to disregard. I just figured I'd post it in case anyone found it interesting 😋 

```typescript
type ZosmfVersion = (typeof ZosmfConstants.VERSIONS)[keyof typeof ZosmfConstants.VERSIONS];
public static async isZosVersionGreaterThan(session: AbstractSession, version: ZosmfVersion): Promise<boolean>;
```

packages/zostso/src/doc/IIssueTsoCmdResponse.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface IIssueTsoCmdResponse {
+
+    cmdResponse: {message: string}[];
+
+    tsoPromptReceived: "Y" | "N",
+
+    servletKey?: string;
+
+    keywordDetected?: "Y" | "N",
+}
```
traeok, line 21: Not requesting changes here, just a suggestion: It may be helpful from an extender's perspective to provide typedoc for the members in this interface.

traeok (commented):
Looks good Jace, thanks for working on this enhancement! I left some suggestions about types and the logic in `issueTsoCmd`.

anaxceron (commented):
Made suggestions for minor edits on the changelog item. Hope you find it helpful.

traeok (approved):
Tested and this is working great - LGTM. Thanks Jace for the enhancement! 😁

anaxceron (approved):
Looks great -- thank you @jace-roell

packages/cli/src/zostso/issue/command/Command.definition.ts
```diff
@@ -37,7 +37,15 @@ export const CommandDefinition: ICommandDefinition = {
...
-            description: "Suppress console messages from start of address space."
+            description: "Suppress console messages from start of address space.",
+            defaultValue: true
+        },
+        {
+            name: "stateful",
+            aliases: ["sf"],
+            type: "boolean",
+            description: "Statefulness of address space created for TSO Command.",
+            defaultValue: false
```
zFernand0, line 48: This is the crucial thing that we should document.

Users can explicitly ask for the old behavior if they specify `--ssm false`.

Otherwise, we will try the new behavior (if available) and fallback to the old behavior if the API is not available.

Also, if `--ssm false`, the value provided for `--stateful` won't really matter.

@anaxceron,
Could you help us here?

packages/zosmf/src/CheckStatus.ts
```diff
@@ -36,6 +36,10 @@ export class CheckStatus {
...
+        return (await CheckStatus.getZosmfInfo(session)).zosmf_version >= version;
```
zFernand0, line 40: Not sure what will be more valuable here.

If we should compare versions (i.e. `>=`) or if we should return an enum-like value (`ZosmfConstants.Versions.V2R1`, ...)

For our purposes, this works just fine 😋

packages/zosmf/src/constants/Zosmf.constants.ts
```diff
@@ -67,6 +67,15 @@ export const ZosmfConstants: { [key: string]: any } = {
          * @type {string}
          */
         UNABLE_TO_VERIFY_LEAF_SIGNATURE: "UNABLE_TO_VERIFY_LEAF_SIGNATURE"
+    },
+
+    VERSIONS: {
```
zFernand0, line 72: I think it'd be ideal to add a link here:
```suggestion
    // https://www.ibm.com/docs/en/zos/2.5.0?topic=service-retrieve-zosmf-information
    VERSIONS: {
```

zFernand0 (commented):
Still missing a few files to review.
I'll do another round of comments possibly.
so far, LGTM! 😋

zFernand0 (approved):
Left a few small comments. However, the functionality and code-comments look great.

LGTM! 😋

packages/zostso/src/constants/ZosTso.constants.ts
```diff
@@ -20,7 +20,7 @@ export class TsoProfileConstants {
         aliases: ["a"],
         description: "Your z/OS TSO/E accounting information.",
         type: "string",
-        required: true,
+        required: false,
```
zFernand0, line 23: I'd like to do some more testing around this before merging 😋 
Otherwise, we could revert this line change and just move my comment to a new issue 😋 
That way we don't have to wait any longer 😅

t1m0thyj (commented):
Looks pretty good, thanks @jace-roell! Left a few comments, will review again after manually testing the changes 😋

packages/cli/src/zostso/issue/command/Command.definition.ts
```diff
@@ -34,10 +34,24 @@ export const CommandDefinition: ICommandDefinition = {
     ],
     options: ([
         {
+            // Old API behavior will be utilized upon specifying --ssm to be false,
+            // otherwise try new API and if it fails, fallback to old API.
+
+            // Specifying --ssm to be false makes the value of --stateful have no impact on
```
gejohnston, line 40: I think that there is a relationship between the 'suppress-startup-messages' and 'stateful' options within the zosmf APIs that I do not fully understand. None-the-less, from the perspective of someone using the command, it does not seem like 'suppress-startup-messages' should control 'stateful'. It seems like we should think about it as 'stateful' will cause us to override 'suppress-startup-messages' to the desired value.

I am very late to the review of this PR, so this choice may have already been fully considered, and it could be reasonable to not change any of the paradigm at this late time.

packages/zosmf/src/CheckStatus.ts
gejohnston, line 39: From the implementation, should this function be named IsZosVersionAtLeast instead of GreaterThan?

gejohnston (approved):
I left a couple of opinions. As long as you have already considered the related choices, I do not require changes.

zFernand0 (dismissed):
Adam brought up a good point that I believe we might want to consider before merging this PR.

When running with the `--stateful` option, we probably want to show them the Servlet Key as part of the command output (and not just for the `--rfj` use case)

For example:
```shell
zFernand0 : ~/gh/zowe/cli.next > zdev tso issue cmd "time" --stateful
Servlet Key: test1234-key567899
IKJ56650I TIME-02:00:32 PM. CPU-00:00:00 SERVICE-574 SESSION-00:00:29 SEPTEMBER 17,2024
READY 
```

Where the first line `Servlet Key: test1234-key56789` goes to `.stderr` and the text `Servlet Key:` is displayed in yellow. 😋

t1m0thyj (dismissed):

zFernand0 (approved):
LGTM! 😋

packages/zostso/src/IssueTso.ts
```diff
@@ -34,50 +130,35 @@ export class IssueTso {
...
-        TsoValidator.validateSession(session);
-        TsoValidator.validateIssueParams(commandParms);
-        TsoValidator.validateNotEmptyString(commandParms.command, noCommandInput.message);
-        return IssueTso.issueTsoCommand(session, commandParms.accountNumber, commandParms.command, commandParms.startParams);
+    public static async issueTsoCommandCommon(
+        session: AbstractSession,
+        commandParms: IIssueTsoParms
+    ): Promise<IIssueResponse> {
+        return await IssueTso.issueTsoCmd(session, commandParms.command, {
+            addressSpaceOptions: {
```
zFernand0, line 158: After some more thought, I'm curious if we should keep backwards compatibility in this common method as well.
Meaning, pass the `suppessStartupMessages: false`. 😅 
I do believe this is a less common function as I only saw it in forks of the CLI and in the kotlin-externals. 😅 

Not a hard requirement though.
I kind of want to see if someone complains about this one 😋

packages/zostso/src/IssueTso.ts
```diff
@@ -9,23 +9,122 @@
...
+                        .join("\n"),
+                };
+                return response;
+            } catch (e) {
+                if (!e.mMessage.includes("status 404")) throw e;
+                useNewApi = false;
+            }
+        }
+        // Deprecated API Behavior [former issueTsoCommand() behavior]
+        if (opts.addressSpaceOptions != null || !useNewApi) {
```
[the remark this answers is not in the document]
awharn: We may want to fix this CodeQL warning before this gets merged.

t1m0thyj (approved):
LGTM, thanks @jace-roell!

I'm able to run the following command without a TSO profile 🎉 
`zowe tso issue cmd "TIME" --account=bogusValue`

packages/zostso/src/IssueTso.ts
```diff
@@ -1,31 +1,127 @@
...
+                        .join("\n"),
+                };
+                return response;
+            } catch (e) {
+                if (!e.mMessage.includes("status 404")) throw e;
+                useNewApi = false;
+            }
+        }
+        // Deprecated API Behavior [former issueTsoCommand() behavior]
+        if (!useNewApi) {
```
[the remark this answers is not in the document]
awharn: I think I understand why it is warning about this now. There are only 4 situations for this code block:

1) useNewApi is true and the try executes successfully. It returns on line 68.
2) useNewApi is true and the try fails, but is not a 404. The error is thrown on line 70.
3) useNewApi is true, and the try fails, and the error is a 404. The code continues, and useNewApi is set to false.
4) useNewApi is false.

Situations 1 and 2 never get to line 75, and situations 3 and 4 both have useNewApi as false. Therefore any condition that actually hits the negative check is always true, and the check is not needed.

awharn (approved):

PR #2254, opened 2024-09-05
Add --ignore-not-found flag logic to zosfilesBase.handler and update command definitions and en.ts for delete operations

ATorrise:
**What it Does**
This PR implements the `--ignore-not-found` flag for delete operations across DS, USS, zFS, and VSAM by:
- Adding logic in the `ZosFilesBaseHandler` to handle `--ignore-not-found` and suppress errors when a file or data set does not exist (HTTP 404 or IDC3012I).
  - The IDC3012I code is an IBM z/OS error message that indicates a VSAM entry does not exist in the catalog
- Updating the command definitions to include the `--ignore-not-found` flag.
- Modifying `en.ts` to reflect the new flag in descriptions and examples.
 
image

**How to Test**
Verify that the `--ignore-not-found` flag works for all delete operations (DS, USS, zFS, VSAM) by attempting to delete things that don't exist. Delete non-existent files/data sets with and without `--ignore-not-found`.

1. Test commands with --for-sure and --ignore-not-found flag (-fi)

| Command                                                | Type  |
|--------------------------------------------------------|-------|
| `zowe zos-files delete data-set "ibmuser.cntl" -fi`     | DS    |
| `zowe zos-files delete uss-file "/a/ibmuser/testcases" -fi` | USS   |
| `zowe zos-files delete zos-file-system "HLQ.MYNEW.ZFS" -fi` | zFS   |
| `zowe zos-files delete data-set-vsam "ibmuser.AAA.**.FFF" -fi` | VSAM  |

2. Test commands with just the --for-sure flag (-f)

| Command                                                | Type  | Expected Error Message                                          |
|--------------------------------------------------------|-------|-----------------------------------------------------------------|
| `zowe zos-files delete data-set "ibmuser.cntl" -f`      | DS    | "Data set not cataloged"                 |
| `zowe zos-files delete uss-file "/a/ibmuser/testcases" -f` | USS   | "No such file or directory"             |
| `zowe zos-files delete zos-file-system "HLQ.MYNEW.ZFS" -f` | zFS   | "NOT FOUND"         |
| `zowe zos-files delete data-set-vsam "ibmuser.AAA.**.FFF" -f` | VSAM  | "NOT DELETED"       |

**Review Checklist**
I certify that I have:
- [x] I tested my changes.
- [x] I added/updated automated tests.
- [x] I updated the changelog.
- [x] I followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md).

**Additional Comments**
This update is for all delete operations except MDS (which does not currently support this flag)


dkelosky:
Perhaps it's too late to adjust this, but I wanted to make an observation... `--quite` seems like a pretty generic flag that might have applicability to other command groups besides `zowe files delete` operations.  Should it be reserved for a broader purpose?  

I also came across [this suggested "standards" reference](https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html) which links to description:

> ‘quiet’
Used in many programs to inhibit the usual output. Every program accepting ‘--quiet’ should accept ‘--silent’ as a synonym.

Alternatively, a more descriptive command option for this command group could be used `--ignore-not-found`?

packages/cli/__tests__/zosfiles/__unit__/delete/zfs/zfs.handler.unit.test.ts
```diff
@@ -17,38 +26,45 @@ describe("ZfsHandler", () => {
...
+        // await expect(handler.processWithSession(commandParameters, {} as any)).rejects.toThrow(ImperativeError);
+        // await expect(handler.processWithSession(commandParameters, {} as any)).rejects.toThrow(error);
+        // await expect(handler.processWithSession(commandParameters, {} as any)).resolves.toEqual({ success: true });
+        await expect(handler.process(commandParameters)).resolves.toBe(undefined);
+        // await expect(handler.process(commandParameters)).resolves.toEqual({ success: true });
     });
 
     it("should throw file not found error (404) when --quiet is not used (-f)", async () => {
-        deleteZfsSpy.mockImplementationOnce(() => {
+        deleteZfsSpy.mockImplementation(() => {
```
zFernand0, line 90: Hey @ATorrise,
I forgot to add the `mockImplementationOnce` back on both tests.
It might be the cause of other tests failing. Sorry about that 😢

t1m0thyj:
> Perhaps it's too late to adjust this, but I wanted to make an observation... `--quiet` seems like a pretty generic flag that might have applicability to other command groups besides `zowe files delete` operations. Should it be reserved for a broader purpose?
> 
> Alternatively, a more descriptive command option for this command group could be used ~`--ignore-not-found`?

I think this is a good point that `--quiet` or `--silent` are pretty generic and could be used for other purposes in the future such as controlling log level.

If it's not too late to change, I support the idea of `--ignore-not-found` which is more explicit and used by some other CLIs like [kube-cli](https://github.com/kubernetes/kubernetes/issues/10828) on their delete commands 😋

packages/cli/__tests__/zosfiles/__system__/delete/zfs/__snapshots__/cli.files.delete.zfs.system.test.ts.snap
zFernand0, line 1: Ideally we would not have so many snapshots 😋 
Not asking to change anything at this point since many other tests use them for these success messages 😓

packages/cli/__tests__/zosfiles/__system__/delete/ds/__snapshots__/cli.files.delete.ds.system.test.ts.snap
```diff
@@ -35,10 +32,7 @@ exports[`Delete Data Set Success scenarios should delete a partitioned data set
   \\"message\\": \\"\\",
   \\"stdout\\": \\"Data set deleted successfully.\\\\n\\",
   \\"stderr\\": \\"\\",
-  \\"data\\": {
-    \\"success\\": true,
-    \\"commandResponse\\": \\"Data set deleted successfully.\\"
```
zFernand0, line 40: is this possibly reflecting a breaking change? 

If someone deletes a dataset with `--rfj`, should they still get the `data.commandResponse` and `data.success` values back?

packages/cli/src/zosfiles/delete/zfs/zfs.definition.ts
```diff
@@ -47,12 +47,23 @@ export const ZfsDefinition: ICommandDefinition = {
             description: strings.OPTIONS.FOR_SURE,
             type: "boolean",
             required: true
+        },
+        {
+            name: "ignore-not-found",
+            aliases: ["i", "ignore-nf"],
```
zFernand0, line 53: This comment applies to other option definitions.

To be consistent with our [rules for naming commands](https://github.com/zowe/zowe-cli/blob/master/docs/CommandFormatStandards.md#naming-commands) (and options), I think that instead of `ignore-nf`, it should be `inf`
We can keep the `i` if we want people to use `-if` when deleting datasets and files 😋 

```suggestion
            aliases: ["i", "inf"],
```

packages/cli/src/zosfiles/ZosFilesBase.handler.ts
```diff
@@ -39,33 +39,30 @@ export abstract class ZosFilesBaseHandler implements ICommandHandler {
...
-        if (!response.success && response.commandResponse) {
-            throw new ImperativeError({
-                msg: response.errorMessage || response.commandResponse
-            });
+        try {
+            const response = await this.processWithSession(commandParameters, session);
+            if (response.commandResponse) {
+                commandParameters.response.console.log(response.commandResponse);
+            }
+        } catch (error) {
```
zFernand0, line 56: To avoid the breaking change with `--rfj`, I think we should do the following:
```suggestion
            commandParameters.response.data.setObj(response);
        } catch (error) {
```

zFernand0 (changes requested):
LGTM! 😋 
left a small change-request to avoid a small breaking change on people processing out `--rfj`

packages/cli/__tests__/zosfiles/__system__/delete/ds/__snapshots__/cli.files.delete.ds.system.test.ts.snap
ATorrise: is it necessary to keep this in the test? I just removed  it because it didn't seem necessary since im also checking for an error code @zFernand0

zFernand0: After adding `commandParameters.response.data.setObj(response);` back, you may need to run system tests again.
And this `data: { }` object should be filled again 😋

packages/cli/src/zosfiles/ZosFilesBase.handler.ts
```diff
@@ -39,33 +39,31 @@ export abstract class ZosFilesBaseHandler implements ICommandHandler {
...
+            commandParameters.response.data.setObj(response);
+        } catch (error) {
+            if (commandParameters.arguments.ignoreNotFound && (error.errorCode === '404' || error.toString().includes("IDC3012I "))) {
+                // The IDC3012I code is an IBM z/OS error message that indicates that the
+                // requested dataset or VSAM entry does not exist in the catalog
+                commandParameters.response.data.setObj({ success: true });
+            } else {
+                throw new ImperativeError({
+                    msg: error.mMessage
+                });
```
zFernand0, line 65: This is fine! 🔥 🐶 🔥 

What about the processWithSession implementations that do not throw an error, but instead return "gracefully" with `success: false` and a proper response on the commandResponse object?

Even though the ZosFilesBaseHandler is [not exported to end users/developers](https://github.com/zowe/zowe-cli/blob/next/packages/cli/src/index.ts), we still need to maintain the behavior for our own APIs.

Here are the ones that I could find quickly:
- List.dataSetsMatchingPattern
- Download.allMembers
- Upload.pathToDataSet

I believe we still need to check for the response.success value and throw and error (in the `try` block)

zFernand0 (changes requested):
Thanks for addressing a few comments so quickly 😋 
I believe that running the whole suite of system tests might be beneficial as there is a slight change that should be caught by other seemingly unrelated system tests. 😓

packages/cli/src/zosfiles/ZosFilesBase.handler.ts
t1m0thyj: Changes to this file will affect every command in the `zowe zos-files` group. If the `IDC3012I` error is specific to VSAMs, would it make sense to handle it in `delete/vsam/Vsam.handler.ts` instead?

packages/cli/src/zosfiles/ZosFilesBase.handler.ts
```diff
@@ -39,33 +39,31 @@ export abstract class ZosFilesBaseHandler implements ICommandHandler {
...
-
-        // Ensure error gets thrown if request was unsuccessful.
-        // Sometimes it is useful to delay throwing an error until the end of the handler is
-        // reached, for example the upload API needs to return an API response even when it fails.
-        if (!response.success && response.commandResponse) {
-            throw new ImperativeError({
-                msg: response.errorMessage || response.commandResponse
-            });
+        try {
+            const response = await this.processWithSession(commandParameters, session);
```
zFernand0, line 52: I believe a few unit tests might be failing because of the `response.success === false` possibility.

ATorrise: my unit tests arent failing, im confused about what you mean

t1m0thyj: @ATorrise Tests are failing in GH workflows, here's an example run: https://github.com/zowe/zowe-cli/actions/runs/10944131154/job/30385249824

zFernand0 (commented):
waiting to review 😋

ATorrise:
zos files system tests are passing

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+- Enhancement: Add --ignore-not-found flag logic to zosfilesBase.handler and update command definitions and en.ts for delete operations. [#2254](https://github.com/zowe/zowe-cli/pull/2254)
```
zFernand0, line 6:
```suggestion
- Enhancement: Add `--ignore-not-found` flag logic to zosfilesBase.handler and update command definitions and en.ts for delete operations. [#2254](https://github.com/zowe/zowe-cli/pull/2254)
```

packages/cli/__tests__/zosfiles/__unit__/upload/ftds/__snapshots__/FileToDataSet.handler.unit.test.ts.snap
```diff
@@ -1,18 +1,6 @@
...
-      "to": "testing",
-    },
-  ],
-  "commandResponse": "uploaded",
-  "success": false,
-}
-`;
+exports[`Upload file-to-data-set handler process method should display error when upload file to data set 1`] = `undefined`;
 
 exports[`Upload file-to-data-set handler process method should display error when upload file to data set 2`] = `""`;
```
zFernand0, line 5: not something to resolve on this PR, but I'm curious if we can avoid the use of snapshots for empty strings or undefined/null values 😋

packages/cli/src/zosfiles/ZosFilesBase.handler.ts
```diff
@@ -39,33 +39,36 @@ export abstract class ZosFilesBaseHandler implements ICommandHandler {
...
         );
         const sessCfgWithCreds = await ConnectionPropsForSessCfg.addPropsOrPrompt<ISession>(
-            sessCfg, commandParameters.arguments, {parms: commandParameters}
+            sessCfg, commandParameters.arguments, { parms: commandParameters }
         );
-
         const session = new Session(sessCfgWithCreds);
-        const response = await this.processWithSession(commandParameters, session);
-
         commandParameters.response.progress.endBar(); // end any progress bars
```
zFernand0, line 50: should this be moved below the call to `processWithSession`? 
if we don't, some progress bars may terminate before they are supposed to 😅

zFernand0 (dismissed):
LGTM! 😋 

left a few small comments/questions, but nothing that should stop the PR from being merged if it gets enough approvals 🥳

t1m0thyj (changes requested):
Looks pretty good, thanks @ATorrise!

Will approve once this comment is addressed: https://github.com/zowe/zowe-cli/pull/2254#discussion_r1803211668

packages/cli/__tests__/zosfiles/__system__/delete/vsam/cli.files.delete.vsam.system.test.ts
```diff
@@ -120,14 +120,26 @@ describe("Delete VSAM Data Set", () => {
         });
 
         it("should delete a VSAM data set that has a retention period", async () => {
-            let response = runCliScript(__dirname + "/__scripts__/command/command_invoke_ams_define_for_statement.sh",
-                TEST_ENVIRONMENT, [dsname, volume]);
-            response = runCliScript(__dirname + "/__scripts__/command/command_delete_vsam_data_set.sh",
+            const response = runCliScript(__dirname + "/__scripts__/command/command_delete_vsam_data_set.sh",
```
awharn, line 123: The portion of this test that creates the VSAM to delete was removed, and this test does not pass.

awharn (dismissed):
Found one issue that should be corrected before merge.

packages/cli/src/zosfiles/delete/vsam/Vsam.handler.ts
```diff
@@ -18,10 +18,22 @@ import { ZosFilesBaseHandler } from "../../ZosFilesBase.handler";
...
+            });
+
+            return response;
+        } catch (error: any) {
+            if (commandParameters.arguments.ignoreNotFound &&
+                (error.errorCode === '404' || error.toString().includes("IDC3012I"))) {
+                return { success: true, commandResponse: "VSAM dataset not found but this is ignored" };
+            }
+            throw new ImperativeError({
+                msg: error.mMessage || "An unexpected error occurred."
```
awharn, line 35: Please add the original error to causeErrors.

packages/cli/src/zosfiles/ZosFilesBase.handler.ts
```diff
@@ -39,34 +39,35 @@ export abstract class ZosFilesBaseHandler implements ICommandHandler {
...
+            if (response.commandResponse) {
+                commandParameters.response.console.log(response.commandResponse);
+            }
+            commandParameters.response.data.setObj(response);
+        } catch (error) {
+            if (commandParameters.arguments.ignoreNotFound && error.errorCode === '404') {
+                commandParameters.response.data.setObj({ success: true });
+            } else {
+                throw new ImperativeError({
+                    msg: error.mMessage
```
awharn, line 66: Please add the original error to causeErrors.

packages/cli/src/zosfiles/ZosFilesBase.handler.ts
```diff
@@ -39,34 +39,35 @@ export abstract class ZosFilesBaseHandler implements ICommandHandler {
...
+                });
+            }
+            if (response.commandResponse) {
+                commandParameters.response.console.log(response.commandResponse);
+            }
+            commandParameters.response.data.setObj(response);
+        } catch (error) {
+            if (commandParameters.arguments.ignoreNotFound && error.errorCode === '404') {
+                commandParameters.response.data.setObj({ success: true });
+            } else {
```
awharn, line 64: We should check if an ImperativeError is thrown and just rethrow it. Otherwise we might have an ImperativeError in an ImperativeError.

packages/cli/src/zosfiles/delete/vsam/Vsam.handler.ts
awharn: Actually, we should probably check if it is an Imperative Error and if so, rethrow it.

packages/cli/CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 # Change Log
-
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+- Enhancement: Add `--ignore-not-found` flag logic to zosfilesBase.handler and update command definitions and en.ts for delete operations. [#2254](https://github.com/zowe/zowe-cli/pull/2254)
```
anaxceron, line 5: Suggesting to change this to something like: "Add the `--ignore-not-found` flag to avoid error messages potentially interrupting a script for automated batch processes. The flag bypasses warning prompts to confirm delete actions."

anaxceron (commented):
Suggesting some edits to the changelog.

packages/cli/src/zosfiles/ZosFilesBase.handler.ts
```diff
@@ -39,34 +39,35 @@ export abstract class ZosFilesBaseHandler implements ICommandHandler {
...
+                throw new ImperativeError({
+                    msg: error.mMessage
+                });
+            }
         }
+        commandParameters.response.progress.endBar();
```
awharn, line 70: I believe all of this code should be how it was prior to the changes, with a try/catch added around it and the special handling for a 404 with ignoreNotFound. Because of the order we are running this now, the daemon is never getting data back because we are not setting data, and the daemon is waiting for data that never comes. This causes the daemon to loop infinitely and freeze, which causes the tests to fail due to timing out.

packages/cli/src/zosfiles/ZosFilesBase.handler.ts
ATorrise: i guess i didnt really know what to do here because response isnt accessible here because of the catch block so i cant really just like revert back @awharn

packages/cli/src/zosfiles/ZosFilesBase.handler.ts
```diff
@@ -39,33 +39,39 @@ export abstract class ZosFilesBaseHandler implements ICommandHandler {
...
+            if (!response.success && response.commandResponse) {
+                throw new ImperativeError({
+                    msg: response.errorMessage || response.commandResponse
+                });
+            }
+            if (response.commandResponse) {
+                commandParameters.response.console.log(response.commandResponse);
+            }
+            commandParameters.response.data.setObj(response);
+            commandParameters.response.progress.endBar();
```
t1m0thyj, line 62: The integration tests are still failing because the order of these statements has been switched around. Restoring the original order should fix them 😋 
1. `commandParameters.response.progress.endBar();`
2. `if (response.commandResponse) ...`
3. `commandParameters.response.data.setObj(response);`
4. `if (!response.success && response.commandResponse) ...`

anaxceron (approved):
Good changelog, thank you @ATorrise!

zFernand0 (approved):
LGTM! 😋

packages/cli/__tests__/zosfiles/__system__/delete/uss/cli.files.delete.file.system.test.ts
```diff
@@ -123,13 +123,22 @@ describe("Delete File", () => {
...
+
         it("should delete a file with response timeout", async () => {
             const response = runCliScript(__dirname + "/__scripts__/command/command_delete_file.sh",
                 TEST_ENVIRONMENT, [ussname, "--for-sure", "--responseTimeout 5"]);
             expect(response.stderr.toString()).toBe("");
             expect(response.status).toBe(0);
             expect(response.stdout.toString()).toMatchSnapshot();
         });
+
+        it("should delete a file with --ignore-not-found flag", async () => {
```
t1m0thyj, line 135: Should this test have a second step to match the related test for data sets?
`// Now delete the USS file, verify no output in ignore-not-found mode`

packages/cli/__tests__/zosfiles/__system__/delete/vsam/cli.files.delete.vsam.system.test.ts
```diff
@@ -128,6 +128,20 @@ describe("Delete VSAM Data Set", () => {
...
+
+        it("should successfully delete a VSAM data set with --ignore-not-found flag", async () => {
+            // create vsam
+            const createResponse = runCliScript(__dirname + "/__scripts__/command/command_invoke_ams_define_statement.sh",
+                TEST_ENVIRONMENT, [dsname, volume]);
+            expect(createResponse.status).toBe(0);
+
+            // now delete
+            const deleteResponse = runCliScript(__dirname + "/__scripts__/command/command_delete_vsam_data_set.sh",
+                TEST_ENVIRONMENT, [dsname, "--for-sure", "--ignore-not-found"]);
```
t1m0thyj, line 140: Same question as for USS above - should we attempt the deletion a 2nd time and make sure it still works with `--inf` flag?

packages/cli/__tests__/zosfiles/__system__/delete/zfs/cli.files.delete.zfs.system.test.ts
```diff
@@ -114,6 +115,20 @@ describe("Delete z/OS File System", () => {
...
         });
+
+        it("should delete a ZFS with --ignore-not-found flag", async () => {
+            // first create zfs
+            const createResponse = runCliScript(__dirname + "/__scripts__/command/command_create_zfs.sh",
+                TEST_ENVIRONMENT, [fsname, volume]);
+            expect(createResponse.status).toBe(0);
+
+            const deleteResponse = runCliScript(__dirname + "/__scripts__/command/command_delete_zfs.sh",
+                TEST_ENVIRONMENT, [fsname, "--for-sure", "--ignore-not-found"]);
```
t1m0thyj, line 126: Same question as for USS above - should we attempt the deletion a 2nd time and make sure it still works with `--inf` flag?

packages/cli/__tests__/zosfiles/__unit__/create/zfs/zfs.handler.unit.test.ts
```diff
@@ -107,32 +107,35 @@ describe("Create z/OS file system handler", () => {
...
+                console: {
+                    log: jest.fn((logArgs) => {
+                        logMessage += "\n" + logArgs;
+                    }),
+                },
+                progress: {
+                    endBar: jest.fn(), // Mocking progress.endBar here
+                },
+            },
+        };
```
t1m0thyj, line 135: Not a request for changes, just a note that in the future it may be helpful to use `mockHandlerParameters` method from the cli-test-utils package to reduce the amount of code needed to mock the handler parameters object 😋

packages/cli/__tests__/zosfiles/__unit__/mount/fs/fs.handler.unit.test.ts
t1m0thyj, line 1: Curious why changes were needed in this file as well as the one above (`DataSetMatching.handler.unit.test.ts`), since it doesn't seem related to the rest of the PR. It's probably ok since these changes are to tests.

packages/cli/src/zosfiles/ZosFilesBase.handler.ts
```diff
@@ -39,33 +39,39 @@ export abstract class ZosFilesBaseHandler implements ICommandHandler {
...
-        if (response.commandResponse) {
-            commandParameters.response.console.log(response.commandResponse);
-        }
-
-        // Return as an object when using --response-format-json
-        commandParameters.response.data.setObj(response);
-
-        // Ensure error gets thrown if request was unsuccessful.
-        // Sometimes it is useful to delay throwing an error until the end of the handler is
-        // reached, for example the upload API needs to return an API response even when it fails.
```
t1m0thyj, line 64: Not a request for changes but please be careful about removing comments - some other ones in this file probably aren't very meaningful like "print out the response" so I understand removing them, but more verbose comments like this one may contain valuable information.

packages/cli/src/zosfiles/delete/vsam/Vsam.handler.ts
```diff
@@ -18,10 +18,26 @@ import { ZosFilesBaseHandler } from "../../ZosFilesBase.handler";
...
+        } catch (error: any) {
+            if (commandParameters.arguments.ignoreNotFound &&
+                (error.errorCode === '404' || error.toString().includes("IDC3012I"))) {
+                return { success: true, commandResponse: "VSAM dataset not found but this is ignored" };
+            }
+            if (error instanceof ImperativeError){
+                throw error;
+            }
+            throw new ImperativeError({
+                msg: error.mMessage || "An unexpected error occurred.",
```
t1m0thyj, line 38: The `mMessage` property is mostly likely undefined if the error is not an instance of `ImperativeError`. Should we do the following to match the base handler?
```suggestion
                msg: error.message,
```

packages/zosfiles/CHANGELOG.md
t1m0thyj, line 1: Please remove changelog update from this file since no user-facing changes were made to the zos-files SDK, only a formatting change

t1m0thyj (commented):
Looks pretty good, thanks @ATorrise! Left a few comments

anaxceron (approved):

ATorrise:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=zowe_zowe-cli&pullRequest=2254) **Quality Gate failed**
> Failed conditions  [4.2% Duplication on New Code](https://sonarcloud.io/component_measures?id=zowe_zowe-cli&pullRequest=2254&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=zowe_zowe-cli&pullRequest=2254)

@zFernand0 @t1m0thyj not sure what to do about this?

zFernand0:
> > ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=zowe_zowe-cli&pullRequest=2254) **Quality Gate failed**
> > Failed conditions  [4.2% Duplication on New Code](https://sonarcloud.io/component_measures?id=zowe_zowe-cli&pullRequest=2254&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> > [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=zowe_zowe-cli&pullRequest=2254)
> 
> `@zFernand0` @t1m0thyj not sure what to do about this?

I wouldn't worry too much about the duplication.
of course there could be some of it removed, but it might make the code a bit harder to follow.

If we do plan to reduce duplication, I guess I'd prefer to do it on a major release boundary, as part of our cleanup efforts 😋

packages/cli/CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 # Change Log
-
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+- Enhancement: Add the --ignore-not-found flag to avoid FNF error messages when deleting files so scripts are not interupted during automated batch processing. The flag bypasses warning prompts to confirm delete actions. [#2254](https://github.com/zowe/zowe-cli/pull/2254)
```
t1m0thyj, line 5: I'm not sure that FNF is a commonly known acronym, should we spell it out?

t1m0thyj (approved):
LGTM, thanks @ATorrise for the new feature and thorough system tests!

PR #2256, opened 2024-09-06
Add write permission to PRs

awharn:
**What It Does**

Added missing write permission to PRs for the projects automation.

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/.github/workflows/update-project.yml
+++ b/.github/workflows/update-project.yml
@@ -12,6 +12,9 @@ env:
   PR_STATUS_DRAFT: 'In Progress'
   PR_STATUS_READY: 'Review/QA'
 
+permissions:
+  pull-requests: write
+
 jobs:
   update-project:
     name: Move project item


.github/workflows/update-project.yml
```diff
@@ -12,6 +12,9 @@ env:
   PR_STATUS_DRAFT: 'In Progress'
   PR_STATUS_READY: 'Review/QA'
 
+permissions:
+  pull-requests: write
```
zFernand0, line 16: would it be good to add the `repository-projects` here?
It may not matter much sinc eth eproject is at the org level (I believe) and that's why we need the zowe-robot-token to have the right permissions 😋

zFernand0 (approved):

gejohnston (approved):

.github/workflows/update-project.yml
awharn: Yeah, I was thinking the same thing. Since it is an org project, we are using the Zowe Robot token, so it shouldn't matter.

PR #2266, opened 2024-09-18
chore: Allow a latest update from v2-lts in the interim

zFernand0:
**What It Does**

This enables `@latest` releases from the zowe-v2-lts branch.

Ideally this PR is closed without being merged. :yum:


zFernand0:
PR not needed since we didn't have a critical 7.x.x release to get out before the Zowe V3 GA

PR #2277, opened 2024-09-26
Remove Secrets SDK requirement when Imperative is a bundled dep

t1m0thyj:
**What It Does**

Resolves #2276

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] added/updated automated tests
- [ ] updated the changelog
- [ ] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

**Additional Comments**

PR #2278, opened 2024-09-26
Update release.config.js to allow V3 patch release

t1m0thyj:
**What It Does**

Allow V3 patch releases so we can publish #2277

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/release.config.js
+++ b/release.config.js
@@ -8,7 +8,7 @@ module.exports = {
         },
         {
             name: "master",
-            level: "none",
+            level: "patch",
             channel: "zowe-v3-lts"
         }
     ],


traeok (approved):
LGTM, thanks Timothy

zFernand0 (approved):
TL;DR: LGTM! 😋 

First, thanks @t1m0thyj for the offline discussion.

Given the circumstances, I believe y'all have chosen the path that will get us out the door with little to no burden on end users and developers.

It would seem that we would to update our CLI bundle to an RC3.

And IMO, I believe we should let plug-in and VSCEs developers know about the removal of the runtime dependency, in case they want to update. However, I do not see them having to update as a requirement. The reason being that if their build/package processes already passed with 8.0.0, they should be ok to continue as-is. For those experiencing a runtime dependency problem, it could be resolved by upgrading to 8.0.1, theoretically speaking.

PR #2286, opened 2024-10-02
fix(ds): Fix invalid response from DS list API for archived data sets

### LINKED ISSUE #2285: DS list API may return invalid response when data sets archived with CA Disk
When data sets are archived with CA Disk and have volume=ARCIVE, then z/OSMF APIs may fail to list them with attributes=true.

The `List.dataSetsMatchingPattern` API works around this by falling back to listing attributes for each data set individually:
https://github.com/zowe/zowe-cli/blob/22b8a00f9545dea972d7a4afd03ad1e8d6beb09a/packages/zosfiles/src/methods/list/List.ts#L376-L381

This may fail if the names of any archived data sets begin with the names of any non-archived data sets. For example: `A.B.C` (not archived) and `A.B.C.D` (archived). When z/OSMF lists attributes for `A.B.C`, rather than listing attributes for only 1 data set it parses the pattern as `A.B.C*` which matches an archived data set that it will choke on.

t1m0thyj:
**What It Does**

Fixes #2285 

**How to Test**

Find a ds pattern whose results include data sets with volume=ARCIVE but the first ds in the list is not migrated.

Notice that listing with attributes fails:
`zowe zos-files list data-set <dsPattern> --attributes`

After adding `maxLength=1`, it succeeds:
`zowe zos-files list data-set <dsPattern> --attributes --max-length 1`

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/zosfiles/CHANGELOG.md
+++ b/packages/zosfiles/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Zowe z/OS files SDK package will be documented in thi
 
 ## Recent Changes
 
-- BugFix: Fixed an issue with `List.dataSetsMatchingPattern` method where migrated data sets could break fetching attributes for other data sets. [#2285](https://github.com/zowe/zowe-cli/issues/2285)
+- BugFix: Fixed an issue with the `List.dataSetsMatchingPattern` method where migrated data sets could break fetching attributes for other data sets. [#2285](https://github.com/zowe/zowe-cli/issues/2285)
 
 ## `8.0.0`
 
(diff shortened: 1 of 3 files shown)


zFernand0 (approved):

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Fixed an issue with `List.dataSetsMatchingPattern` method where migrated data sets could break fetching attributes for other data sets. [#2285](https://github.com/zowe/zowe-cli/issues/2285)
```
anaxceron, line 7: super minor, but can we add "the" before `List.dataSets....`?

anaxceron (approved):
left a minor comment on the changelog

gejohnston (approved):
Nice job tracking down such an obscure scenario.

PR #2291, opened 2024-10-07
Fix Coverity issues in V3

awharn:
**What It Does**

Fixes issues found by Coverity

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
+++ b/packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
@@ -384,7 +384,7 @@ export class ConnectionPropsForSessCfg {
              * which do not create a mock for the connOpts.parms.response.console.log property.
              * In the real world, that property always exists for this CLI-only path of logic.
              */
-            if (promptForValues.length > 0 && connOpts?.parms?.response?.console?.log) {
+            if (promptForValues.length > 0 && connOpts.parms?.response.console.log) {
                 // We need to prompt for some values. Determine why we need to prompt.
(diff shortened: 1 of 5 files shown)


packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -399,14 +399,14 @@ export class ConnectionPropsForSessCfg {
 
                 reasonForPrompts += "Therefore, you will be asked for the connection properties " +
                     "that are required to complete your command.\n";
-                connOpts.parms.response.console.log(TextUtils.wordWrap(
+                connOpts?.parms?.response?.console?.log(TextUtils.wordWrap(
```
t1m0thyj, line 402: I'm not sure all these question marks are necessary, but they shouldn't hurt anything either 😋 

```suggestion
                connOpts?.parms.response.console.log(TextUtils.wordWrap(
```

t1m0thyj (approved):
LGTM, thanks @awharn!

gejohnston (approved):

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
awharn: Parms is optional on connOpts, so we will need to at least keep that one. We do the same check on Line 387, so we would need to remove that check as well.

zFernand0 (approved):

traeok (approved):

PR #2297, opened 2024-10-08
fix: require.main.filename check in `DefaultCredentialManager`

traeok:
**What It Does**

In some environments (namely ESM and the Node.js REPL), the `main` property is not always available within the `require` object, so the `if` check for `require.main.filename != null` results in a TypeError.

**How to Test**

This one is a bit difficult to test, but this should work as an example:

- Make an empty folder and set up a basic NPM project using `npm init`
- Build a TGZ for `@zowe/imperative` using `npm install && npm run build && npm pack` in the `packages/imperative` folder. Install the generated tarball in this new NPM project alongside `@zowe/secrets-for-zowe-sdk` as normal dependencies:
  - `npm install zowe-imperative-8.1.0.tgz @zowe/secrets-for-zowe-sdk`
- Open the Node.js REPL within this folder: `node -i`
- Import `DefaultCredentialManager` within the REPL and create a new instance of it. Then attempt to initialize it. Initialization should be successful with no errors thrown:

  ```javascript
  > const { DefaultCredentialManager } = await import("@zowe/imperative")
  > const credMgr = new DefaultCredentialManager()
  > await credMgr.initialize()
   ```

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/packages/imperative/src/security/src/DefaultCredentialManager.ts
+++ b/packages/imperative/src/security/src/DefaultCredentialManager.ts
@@ -120,7 +120,9 @@ export class DefaultCredentialManager extends AbstractCredentialManager {
             // our calling CLI. Since our caller must supply keytar, we search for keytar
             // within our caller's path.
             const requireOpts: any = {};
-            requireOpts.paths = [require.main?.filename, ...require.resolve.paths("@zowe/secrets-for-zowe-sdk")].filter(Boolean);
+            if (require.main?.filename != null) {
+                requireOpts.paths = [require.main.filename, ...require.resolve.paths("@zowe/secrets-for-zowe-sdk")];
+            }
             // use helper function for require.resolve so it can be mocked in jest tests
             const keytarPath = require.resolve("@zowe/secrets-for-zowe-sdk", requireOpts);
             Logger.getImperativeLogger().debug("Loading Keytar module from", keytarPath);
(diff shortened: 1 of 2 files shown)


anaxceron (approved):

awharn (approved):

packages/imperative/src/security/src/DefaultCredentialManager.ts
```diff
@@ -120,9 +120,7 @@ export class DefaultCredentialManager extends AbstractCredentialManager {
             // our calling CLI. Since our caller must supply keytar, we search for keytar
             // within our caller's path.
             const requireOpts: any = {};
-            if (require.main.filename != null) {
-                requireOpts.paths = [require.main.filename, ...require.resolve.paths("@zowe/secrets-for-zowe-sdk")];
-            }
+            requireOpts.paths = [require.main?.filename, ...require.resolve.paths("@zowe/secrets-for-zowe-sdk")].filter(Boolean);
```
t1m0thyj, line 123: Technically this changes the original behavior so that we always pass in a list of paths to `require.resolve`, but I don't think this change should be impactful 😋

t1m0thyj (approved):
LGTM, thanks @traeok!

packages/imperative/src/security/src/DefaultCredentialManager.ts
traeok: This is a good point - I don't want to make any assumptions, and its possible that the default module resolution is special for certain projects. I've updated this to add optional chaining to the `if` check so that we only apply the paths if a filename is provided in `require.main` 😋

Issue #1966, opened 2023-11-15
Enable APIML token renewal before expiration 


gejohnston:
**Is your feature or enhancement request related to a problem or limitation? Please describe**

When an APIML token expires, a REST request fails and the user is directed to login again.

**Describe your enhancement idea**

Detect that the token is due to (or has already) expired, and automatically perform an APIML login request. Then continue the user's original request.

**Describe alternatives you've considered**

- Create APIs to detect token expiration and perform re-authentication.
    - The same (or similar) APIs should be made available to both the CLI and to ZE extenders.

**Provide any additional context**

balhar-jakub:
Just a note there is API available via API ML for this type of exchange - https://docs.zowe.org/stable/user-guide/api-mediation/api-mediation-jwt-token-refresh/ 

It's not enabled by default, but I wanted to let you know, in case this is something you want to use, but it may be better to use the Personal Access Tokens for the authentication against the API Mediation Layer.

zFernand0:
Here is a bit more information about the APIs we could write in a non-breaking enhancement to the Imperative framework.

Thanks to the APIML Token structure, we could split it into 3 portions.
1. Token identification information: type (JWT), algorithm (RS256), and kID (???)
    ```json
    {
      "kid": "some_generated_token_id",
      "typ": "JWT",
      "alg": "RS256"
    }
    ```
2. Personal identification information: type, user, SAF Groups, issuer, expiration, iat (issued at time?), and more...
    ```json
    {
      "token_type": "Bearer",
      "sub": "USERNAME",
      "upn": "USERNAME",
      "groups": ["IZUUSER","ZWEADMIN", "MORE_GROUPS"],
      "realm": "SAFRealm",
      "iss": "zOSMF",
      "exp": 1700278625,
      "iat": 1700248625
    }
    ```
3. Some very long generated hash/string/token (no point in base64 decoding it 😋 ) 

---

The API may look something like this.
```typescript
export class CliUtils {
  // ...
  public static getTokenExpirationTime(apimlToken: string): string {
    // This is just pseudo-code : )
    const tokenDetails = JSON.parse(base64Decode(apimlToken.split(".")[1]));
    return tokenDetails.exp;
  }
```

adam-wolfe:
I think this is still something we should consider, even if only to avoid passing an expired token to the API ML.

PR #2298, opened 2024-10-08
feat(ProfileInfo): JWT token expiration detection

traeok:
**What It Does**

Fernando mentioned that extenders would benefit from having access to the JWT token expiration feature from Zowe Explorer, so I figured it made the most sense to put it the logic for it in `ProfileInfo`. That way, anyone using Imperative to access/manage profiles can leverage this functionality.

**How to Test**

Once merged, this can be tested in action after this PR is updated: https://github.com/zowe/zowe-explorer-vscode/pull/3174

Here's a manual test to show the behavior of the function itself:
- Checkout this branch and run `npm run build && npm pack` from the `packages/imperative` folder
- Copy the `zowe-imperative-8.1.0.tgz` file from this folder and place it into a new folder outside of the repo
- `cd` to the empty folder and run `zowe config init --no-prompt && npm init -y && npm install zowe-imperative-8.1.0.tgz @zowe/secrets-for-zowe-sdk` (you don't need secrets SDK, but its easier than getting around cred manager setup in ProfileInfo)
- Copy this config and replace the contents of `zowe.config.json`:

```jsonc
{
    "$schema": "./zowe.schema.json",
    "profiles": {
        "zosmf": {
            "type": "zosmf",
            "properties": {
                "port": 1337
            }
        },
        "project_base": {
            "type": "base",
            "properties": {
                "host": "localhost",
                // this token expired in 2001
                // if you want to test an unexpired token, this one expires in 2128: 123.eyJleHAiOjUwMDAwMDAwMDB9.456
                "tokenValue": "123.eyJleHAiOjEwMDAwMDAwMDB9.456"
            }
        }
    },
    "defaults": {
        "zosmf": "zosmf",
        "base": "project_base"
    }
}
```

- Save the following code snippet as `script.mjs` within this test folder and run it using `node script.mjs`

```javascript
const { ProfileInfo } = await import("@zowe/imperative");
const profInfo = new ProfileInfo("zowe");
await profInfo.readProfilesFromDisk({ projectDir: process.cwd() });
console.log("token expired:", profInfo.hasTokenExpiredForProfile("zosmf"));
```

The script should output `token expired: true`.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


traeok:
I don't have permissions on Zowe CLI to update the new issue reported by SonarCloud, but it is a false positive as `argValue` will almost always be type `string`, and even in cases where it isn't a string, all exceptions will be handled in the try/catch block.

awharn (approved):

gejohnston (approved):

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -179,6 +179,48 @@ export class ProfileInfo {
...
+                // The Date constructor expects a timestamp in milliseconds.
+                const msPerSec = 1000;
+                const expireDate = new Date(payloadJson.exp * msPerSec);
+                return expireDate < new Date();
+            }
+        } catch (err) {
+            return false;
+        }
+
+        return false;
```
t1m0thyj, line 221: Thanks for adding this to the ProfileInfo API! As the ProfileInfo API is used only by extenders like ZE, I'm curious if we'd want to move this part of the method to make it usable by CLI itself?

Perhaps could have a static `isTokenExpired` method on the `ConfigUtils` or `RestStandaloneUtils` classes, that only checks if the token is expired and doesn't handle loading it from config.

t1m0thyj (approved):
LGTM, thanks @traeok!

packages/imperative/src/config/src/ProfileInfo.ts
traeok: Good call - I added a function to `ConfigUtils` called `hasTokenExpired` and updated the `ProfileInfo` function to use it.
Also added an entry for the new function in the changelog 👍

zFernand0 (approved):
LGTM! 😋 

looking at the code reminded me of this issue:
- https://github.com/zowe/zowe-cli/issues/1966#issuecomment-1822748769

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added a new SDK method (`ConfigUtils.hasTokenExpired`) that allows developers to verify whether a given JSON web token has expired. [#2298](https://github.com/zowe/zowe-cli/pull/2298)
```
anaxceron, line 7: I think it'd be better to turn this into an action statement: Use the new SDK method `bbb` to check whether a given JSON web token has expired.
A note: Do you really use the token to verify expiration, as you originally wrote, or to check whether the token has expired?

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
...
+- Enhancement: Added a new SDK method (`ProfileInfo.hasTokenExpiredForProfile`) that allows developers to verify whether a JSON web token has expired for a given profile. [#2298](https://github.com/zowe/zowe-cli/pull/2298)
```
anaxceron, line 8: As the previous comment, better to turn this into an action: Use the new SDK method `ccc` to check whether a given JSON web token has expired for a given/specified profile.

anaxceron (changes requested):
Requesting some edits to the changelog

anaxceron (approved):
Fantastic changelog @traeok!

PR #2300, opened 2024-10-09
Add search callbacks to files SDK

awharn:
**What It Does**

Adds the ability to provide a function to the Data Sets search function, which is called if present and provided a list of data sets that are about to be searched. This allows extenders to prompt users asking if they want to continue a search, or automatically cancel a search if too many items are about to be searched. If not supplied, the search SDK keeps current behavior.

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/zosfiles/CHANGELOG.md
+++ b/packages/zosfiles/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Zowe z/OS files SDK package will be documented in thi
 
 ## Recent Changes
 
-- Enhancement: Added an optional `continueSearch` function to `ISearchOptions` to allow for user prompting and data set search cancellation after the list of data sets to be searched is retrieved. [#2300](https://github.com/zowe/zowe-cli/pull/2300)
+- Enhancement: Added an optional `continueSearch` function to the `ISearchOptions` interface. After a data set listing is completed, the new function is called with the list of data sets about to be searched. This allows the extender or end users to continue with the search or cancel it. [#2300](https://github.com/zowe/zowe-cli/pull/2300)
 
 ## `8.1.1`
 
--- a/packages/zosfiles/CHANGELOG.md
+++ b/packages/zosfiles/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added an optional `continueSearch` function to `ISearchOptions` to allow for user prompting and data set search cancellation after the list of data sets to be searched is retrieved. []()
+
 ## `8.1.1`
 
 - BugFix: Updated peer dependencies to `^8.0.0`, dropping support for versions tagged `next`. [#2287](https://github.com/zowe/zowe-cli/pull/2287)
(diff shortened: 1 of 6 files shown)


packages/zosfiles/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added an optional `continueSearch` function to `ISearchOptions` to allow for user prompting and data set search cancellation after the list of data sets to be searched is retrieved. [#2300](https://github.com/zowe/zowe-cli/pull/2300)
```
anaxceron, line 7: I suggest rewording to something like: Added an optional `continueSearch` function to the `ISearchOptions` prompt/function/whatever. After an initial data set search has been executed, the new function can be set to prompt users to continue the existing search or cancel it.

anaxceron (commented):
Suggested an edit to the changelog item.

packages/zosfiles/CHANGELOG.md
awharn: I made a change to try and follow that suggestion.

t1m0thyj (approved):
LGTM, thanks @awharn!

packages/zosfiles/src/methods/search/Search.ts
traeok, line 1: Thanks for adding comments here for clarification about the use of `cloneDeep` 😋

traeok (approved):
LGTM, thanks Andrew for the enhancement!

packages/zosfiles/src/methods/search/Search.ts
```diff
@@ -100,14 +102,33 @@ export class Search {
...
             } catch (err) {
                 failedDatasets.push(pds);
             }
         }
 
+        // Call the callback if it exists.
+        let resp: boolean = true;
+        if (searchOptions.continueSearch) {
+            // Call with a clone of the search items queue. Extenders should not be modifying the original.
+            resp = await searchOptions.continueSearch(cloneDeep(searchItemsQueue));
```
gejohnston, line 117: Since the callback is someone else's code (which might crash), should we put a try-catch safety net around the callback? Certainly a stack trace will identify if the failure occurred in the callback, but the failure could be perceived as a failure in our app/API.

I know we have numerous other callbacks, and maybe they do not check for crashes, but maybe we should start? This is not a hard request for change. If you consider the idea and make a different choice, that is ok.

gejohnston (approved):
I made one comment for your consideration, but I still approve.

zFernand0 (commented):
LGTM! 😋 

I kind of agree with Gene's comment here that we should probably catch errors coming from the callback 😋

packages/zosfiles/src/methods/search/Search.ts
awharn: How would we want to handle a crash? A graceful search termination, or just re-throwing it?

awharn: Spoke with many members of the team. We decided to let the code throw any failed callback, and expect anyone using the search APIs to handle their own errors. If a callback does throw, the error should point directly to the callback function.

PR #2301, opened 2024-10-09
all needed changes to meet AC besides tests

ATorrise:
**What It Does**

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/src/config/src/Config.ts
+++ b/packages/imperative/src/config/src/Config.ts
@@ -179,7 +179,8 @@ export class Config {
                 exists: false,
                 properties: Config.empty(),
                 global: layer === Layers.GlobalUser || layer === Layers.GlobalConfig,
-                user: layer === Layers.ProjectUser || layer === Layers.GlobalUser
+                user: layer === Layers.ProjectUser || layer === Layers.GlobalUser,
+                ignoreErrors: process.argv.includes("--help") || process.argv.includes("--version")
             });
         }
 
--- a/packages/imperative/src/config/src/api/ConfigLayers.ts
+++ b/packages/imperative/src/config/src/api/ConfigLayers.ts
@@ -18,6 +18,8 @@ import { ConfigConstants } from "../ConfigConstants";
 import { IConfigLayer } from "../doc/IConfigLayer";
 import { ConfigApi } from "./ConfigApi";
 import { IConfig } from "../doc/IConfig";
+import { TextUtils } from "../../../utilities";
+import { CommandResponse } from "../../../cmd/src/response/CommandResponse";
 
 /**
  * API Class for manipulating config layers.
@@ -31,7 +33,7 @@ export class ConfigLayers extends ConfigApi {
      * @param opts The user and global flags that indicate which of the four
      *             config files (aka layers) is to be read.
      */
-    public read(opts?: { user: boolean; global: boolean }) {
+    public read(opts?: { user: boolean; global: boolean; ignoreErrors: boolean}) {
         // Attempt to populate the layer
         const layer = opts ? this.mConfig.findLayer(opts.user, opts.global) : this.mConfig.layerActive();
         if (fs.existsSync(layer.path)) {
@@ -49,11 +51,19 @@ export class ConfigLayers extends ConfigApi {
                 layer.properties = JSONC.parse(fileContents.toString()) as any;
                 layer.exists = true;
             } catch (e) {
-                throw new ImperativeError({
-                    msg: `Error parsing JSON in the file '${layer.path}'.\n` +
-                        `Please check this configuration file for errors.\nError details: ${e.message}\nLine ${e.line}, Column ${e.column}`,
-                    suppressDump: true
-                });
+                const msg = `Error parsing JSON in the file '${layer.path}'.\n` +
+                `Please check this configuration file for errors.\nError details: ${e.message}\nLine ${e.line}, Column ${e.column}`
+                if (!opts.ignoreErrors){
+                    throw new ImperativeError({
+                        msg:msg,
+                        suppressDump: true
+                    });
+               } else {
+                    const cmdResp: CommandResponse = new CommandResponse({
+                        responseFormat: "default"
+                    });
+                    cmdResp.console.log(TextUtils.chalk.red(msg));
+               }
             }
             this.mConfig.api.secure.loadCached(opts);
         } else if (layer.exists) {
--- a/packages/imperative/src/config/src/doc/IConfigLayer.ts
+++ b/packages/imperative/src/config/src/doc/IConfigLayer.ts
@@ -17,4 +17,5 @@ export interface IConfigLayer {
     properties: IConfig;
     global: boolean;
     user: boolean;
+    ignoreErrors: boolean;
 }


packages/imperative/src/config/src/Config.ts
```diff
@@ -179,7 +179,8 @@ export class Config {
                 exists: false,
                 properties: Config.empty(),
                 global: layer === Layers.GlobalUser || layer === Layers.GlobalConfig,
-                user: layer === Layers.ProjectUser || layer === Layers.GlobalUser
+                user: layer === Layers.ProjectUser || layer === Layers.GlobalUser,
+                ignoreErrors: process.argv.includes("--help") || process.argv.includes("--version")
```
t1m0thyj, line 183: Should we also handle the case of the `zowe` command being run without any arguments since that prints help?

zFernand0: If the plan is to only support the `zowe` command being executed by itself, then I think adding an OR-condition might work.
```
  ||  process.argv[process.argv.length-1] === require.resolve('@zowe/cli')
```
  
---

If we want to handle eevry other command (like `zowe config` or `zowe jobs sumbmit`...
Wouldn't we have to check for whether there is no handler for them?

ATorrise: i added `||  process.argv[process.argv.length-1] === require.resolve('@zowe/cli')` on ln 184, lmk if I should do something else @zFernand0 @t1m0thyj

pujal0909:
test comment

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+- BugFix: Enabling commands with either the `--help` or `--version` flags to still display their information despite any config file issues [#2301](https://github.com/zowe/zowe-cli/pull/2301)
```
zFernand0, line 6:
```suggestion

- BugFix: Enabled commands with either the `--help` or `--version` flags to still display their information despite any config file issues [#2301](https://github.com/zowe/zowe-cli/pull/2301)
```

packages/imperative/src/config/src/Config.ts
```diff
@@ -179,7 +179,9 @@ export class Config {
...
```
zFernand0, line 183: This works for full option names.
should we have this working with aliases as well?
`-V` and `-h` 😋

packages/imperative/src/config/src/doc/IConfigLayer.ts
```diff
@@ -17,4 +17,5 @@ export interface IConfigLayer {
     properties: IConfig;
     global: boolean;
     user: boolean;
+    ignoreErrors: boolean;
```
zFernand0, line 20: Should we make this optional?

zowe-cli
zFernand0, line 1: Should we remove this submodule?

packages/imperative/src/config/src/api/ConfigLayers.ts
```diff
@@ -49,11 +51,19 @@ export class ConfigLayers extends ConfigApi {
...
                 layer.exists = true;
             } catch (e) {
-                throw new ImperativeError({
-                    msg: `Error parsing JSON in the file '${layer.path}'.\n` +
-                        `Please check this configuration file for errors.\nError details: ${e.message}\nLine ${e.line}, Column ${e.column}`,
-                    suppressDump: true
-                });
+                const msg = `Error parsing JSON in the file '${layer.path}'.\n` +
+                `Please check this configuration file for errors.\nError details: ${e.message}\nLine ${e.line}, Column ${e.column}`;
+                if (!opts.ignoreErrors){
```
zFernand0, line 56: Curious if we should  provide a default value in the function signature (to `false` based on this implementation) 😋

packages/imperative/src/config/src/api/ConfigLayers.ts
```diff
@@ -31,7 +33,7 @@ export class ConfigLayers extends ConfigApi {
      * @param opts The user and global flags that indicate which of the four
      *             config files (aka layers) is to be read.
      */
-    public read(opts?: { user: boolean; global: boolean }) {
+    public read(opts?: { user: boolean; global: boolean; ignoreErrors: boolean}) {
```
zFernand0, line 36: This may be out of scope, but should we use either an existing interface (or create a new one)?

If creating a new one... should we simplify the IConfigLayer and extend from the one we create here? 😋

zFernand0 (commented):
A few tests failing, but changes LGTM! 😋 

Curious if we agreed on having the error at the bottom of the output of towards the top.
Just seeing it in practice, someone could miss the error if they are viewing the help text 😋

packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
```diff
@@ -191,6 +191,61 @@ const FAKE_HELP_GENERATOR: IHelpGenerator = {
 
 const ENV_VAR_PREFIX: string = "UNIT_TEST";
 
+describe("Command Processor with --help and --version flags", () => {
```
t1m0thyj, line 194: Typically we have only one `describe` statement at top level of a unit test file.

Could these tests be moved inside the `describe("Command Processor", ...` block?

Also please fix the failing unit tests 😋

packages/imperative/src/config/src/Config.ts
t1m0thyj: We could reference constants defined in `packages/imperative/src/constants/src/Constants.ts` - `HELP_OPTION`, `HELP_OPTION_ALIAS`, etc.

packages/imperative/src/config/src/Config.ts
```diff
@@ -179,7 +179,9 @@ export class Config {
...
+                ||  process.argv[process.argv.length-1] === require.resolve('@zowe/cli')
```
t1m0thyj, line 184: I don't think we should have `process.argv` checks hardcoded in the `Config.ts` file because it may be used by extenders like Zowe Explorer and we don't want to introduce unexpected side effects based on the CLI args passed to a 3rd party app that uses our SDKs.

Could the `ignoreErrors` flag be added as a property on the `IConfigOpts` interface, and passed in to the `Config.load` method from `Imperative.init`?

t1m0thyj (dismissed):

packages/imperative/src/config/src/Config.ts
ATorrise: okay i think i made this change @t1m0thyj , let me know if i messed up

packages/imperative/src/config/src/api/ConfigLayers.ts
ATorrise: rather than setting default here what about `ignoreErrors: ignoreErrors || false` on line 182 in `packages/imperative/src/config/src/Config.ts`

packages/imperative/src/config/src/api/ConfigLayers.ts
ATorrise: i instead modified packages\imperative\src\config\src\Config.ts to set ignoreErrors to false if theres no existing ignoreErrors. now that im thinking about that retroactively, we are always going to have ignoreErrors set now and im not sure if that's wanted behavior

packages/imperative/src/config/src/api/ConfigLayers.ts
ATorrise: now that im thinking about that retroactively, this will cause ignoreErrors to always have a value and im not sure if that's wanted behavior given it was defined to be optional

zFernand0:
Added the `no-chagelog` label since the zosfiles sdk change was just to address a lint warning
- `The changelog was not changed in this pull request for packages/zosfiles/CHANGELOG.md.`

packages/cli/__tests__/zosfiles/__unit__/upload/ftds/FileToDataSet.handler.unit.test.ts
```diff
@@ -329,8 +329,9 @@ describe("Upload file-to-data-set handler", () => {
                     statusMessage: "Uploading to data set"
                 }
             });
-            expect(jsonObj).toMatchSnapshot();
-            expect(apiMessage).toMatchSnapshot();
+
+            expect(apiMessage).toBe("");
+
```
t1m0thyj, line 334: While I support removing snapshots when it makes sense, I think it's also important not to reduce the quality of our tests when doing so. Should we keep an `expect` statement in this test to validate the contents of the JSON object?

packages/imperative/src/config/src/Config.ts
```diff
@@ -117,14 +117,6 @@ export class Config {
         secure: ConfigSecure
     };
 
-    // _______________________________________________________________________
-    /**
-     * Constructor for Config class. Don't use this directly. Await `Config.load` instead.
-     * @param opts Options to control how Config class behaves
-     * @private
-     */
-    private constructor(public opts?: IConfigOpts) { }
```
t1m0thyj, line 126: I believe we need to keep this constructor, as removing the definition of `public opts` could be a breaking change.

packages/imperative/src/config/src/api/ConfigLayers.ts
```diff
@@ -49,11 +51,19 @@ export class ConfigLayers extends ConfigApi {
...
+                if (!opts?.ignoreErrors){
+                    throw new ImperativeError({
+                        msg:msg,
+                        suppressDump: true
+                    });
+                } else {
+                    const cmdResp: CommandResponse = new CommandResponse({
+                        responseFormat: "default"
+                    });
+                    cmdResp.console.log(TextUtils.chalk.red(msg));
```
t1m0thyj, line 65: Should we use `Logger.getAppLogger()` here instead? I'm concerned that the `CommandResponse` option may not work well for extenders like Zowe Explorer running outside the context of a console.

packages/imperative/src/constants/src/Constants.ts
```diff
@@ -59,6 +59,8 @@ export class Constants {
     public static readonly JSON_OPTION_ALIAS = "rfj";
     public static readonly HELP_OPTION = "help";
     public static readonly HELP_OPTION_ALIAS = "h";
+    public static readonly VERSION_OPTION: "version";
+    public static readonly VERSION_OPTION_ALIAS: "v";
```
t1m0thyj, line 63: The alias for version is a capital `V` which is case sensitive
```suggestion
    public static readonly VERSION_OPTION_ALIAS: "V";
```

packages/imperative/src/imperative/src/config/cmd/report-env/EnvItems.ts
```diff
@@ -68,8 +68,8 @@ function formatLogLevelMsg(logTypeName: string) {
 export const probTests: IProbTest[] = [
     {
         itemId: ItemId.NODEJS_VER,
-        probExpr: "semver.satisfies('{val}', '<18.x || 19.x || >=21.x')",
-        probMsg: "Only Node.js versions 18 and 20 are supported."
+        probExpr: "semver.satisfies('{val}', '<18.x || 19.x || 21.x || >=23.x')",
```
t1m0thyj, line 71: Thanks for updating this semver expression, I was also noticing this test fail locally 😍

t1m0thyj (changes requested):
Looks pretty good, thanks @ATorrise - made a few small requests

packages/cli/__tests__/zosfiles/__unit__/upload/ftds/FileToDataSet.handler.unit.test.ts
t1m0thyj: Please disregard this comment - after discussing with @zFernand0 it makes sense to remove this snapshot because the JSON object comes from a mock earlier in the test 😋

packages/imperative/src/config/src/Config.ts
ATorrise: I think to continue moving this pr forward, I'll just put it back and we can keep this as an open discussion point for later, possibly during a standup

packages/imperative/src/config/__tests__/Config.api.unit.test.ts
```diff
@@ -264,6 +265,65 @@ describe("Config API tests", () => {
...
+                const redSpy = jest.fn().mockImplementation((str: string) => str);
+                jest.spyOn(TextUtils, "chalk", "get").mockReturnValueOnce({red: redSpy});
+                let caughtError;
+                try {
+                    config.api.layers.read({user: false, global: false, ignoreErrors: true});
+                } catch(err) {
+                    caughtError = err;
+                }
+                expect(existsSpy).toHaveBeenCalledTimes(5); // Once for each config layer and one more time for read
+                expect(readFileSpy).toHaveBeenCalledTimes(1);
```
ATorrise, line 321: @t1m0thyj @zFernand0 im not sure why this case requires 6 calls instead of 5 like the rest. making sure this is OK with you guys

t1m0thyj: I think the extra call to `existsSync` may be happening because the logger isn't actually being mocked. See my other comment for why.

packages/imperative/src/config/__tests__/Config.api.unit.test.ts
```diff
@@ -264,6 +265,72 @@ describe("Config API tests", () => {
...
+
+            it("shouldnt error when ignoreErrors is true", async () => {
+                const config = await Config.load(MY_APP);
+                const existsSpy = jest.spyOn(fs, "existsSync").mockReturnValueOnce(true);
+                const readFileSpy = jest.spyOn(fs, "readFileSync");
+                const secureLoadSpy = jest.spyOn(config.api.secure, "loadCached");
+                const jsoncParseSpy = jest.spyOn(JSONC, "parse").mockImplementationOnce(() => { throw "failed to parse"; });
+
+                let logMsg: string = "Nothing logged";
+                jest.spyOn(Logger, "getImperativeLogger").mockImplementation(() => {
```
t1m0thyj, line 314: We should mock the app logger here since that's what the Config class is using:
```suggestion
                jest.spyOn(Logger, "getAppLogger").mockImplementation(() => {
```

packages/imperative/src/config/__tests__/Config.api.unit.test.ts
```diff
@@ -264,6 +265,72 @@ describe("Config API tests", () => {
...
+                    config.api.layers.read({user: false, global: false, ignoreErrors: true});
+                } catch(err) {
+                    caughtError = err;
+                }
+                expect(existsSpy).toHaveBeenCalledTimes(6); // Once for each config layer and one more time for read
+                expect(readFileSpy).toHaveBeenCalledTimes(1);
+                expect(secureLoadSpy).toHaveBeenCalledTimes(1);
+                expect(jsoncParseSpy).toHaveBeenCalledTimes(1);
+                expect(caughtError).toBeUndefined();
+                expect(logMsg).toContain("Nothing logged");
```
t1m0thyj, line 332: Once the logger mock is fixed, then `logMsg` should change to show the message that would be printed:
```suggestion
                expect(logMsg).toContain("Please check this configuration file for errors");
```

t1m0thyj (changes requested):
Thanks for addressing feedback, will approve once the unit test is fixed 😋

packages/imperative/src/config/__tests__/Config.api.unit.test.ts
ATorrise: oh whoops sorry and thanks for explaining

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Enabled commands with either the `--help` or `--version` flags to still display their information despite any config file issues [#2301](https://github.com/zowe/zowe-cli/pull/2301)
```
anaxceron, line 7: - Please write out "configuration."
- Sentence needs to end w/ a period.

anaxceron (approved):
Left some notes for minor edits to the changelog entry

t1m0thyj (commented):
When testing the PR I was still seeing a fatal error, which I fixed in https://github.com/zowe/zowe-cli/pull/2301/commits/71b032a5031462b8a5502ee3fc757a4c5e9c795e 🙂 

Please add an integration test that runs `zowe --help` and `zowe --version` with a bad config to verify the error is non-fatal.

t1m0thyj (approved):

packages/imperative/src/config/src/Config.ts
```diff
@@ -147,10 +147,10 @@ export class Config {
         opts = opts || {};
 
         // Create the basic empty configuration
-        const myNewConfig = new Config(opts);
+        const myNewConfig = new Config();
         myNewConfig.mApp = app;
         myNewConfig.mActive = { user: false, global: false };
-        myNewConfig.mVault = opts.vault;
+        myNewConfig.mVault = opts?.vault;
```
awharn, line 153: Since opts is defined on L147, is the `?` necessary? Just curious, since Coverity tends to be rather strict.

traeok (approved):
Thanks Amber for the fix! LGTM

packages/imperative/src/config/src/Config.ts
zFernand0: Makes sense to remove... the other option would've been to remove the assignment a few lines above 😋

awharn (approved):

Issue #2312, opened 2024-10-18
`ConvertV1Profiles` API does not support custom `ZOWE_CLI_HOME`


t1m0thyj:
The team config API is initialized with the default home directory which does not respect environment variables:
https://github.com/zowe/zowe-cli/blob/a1461ca9800cca692f8b3645c9509cfcc028a692/packages/imperative/src/config/src/ConvertV1Profiles.ts#L141-L146

Issue #2311, opened 2024-10-18
`ProfileInfo.onlyV1ProfilesExist` may be wrong when called by extenders


t1m0thyj:
The `onlyV1ProfilesExist` method checks if `ImperativeConfig.instance.config?.exists` to determine if V2 profiles exist.

When an extender using the ProfileInfo API calls this method, `ImperativeConfig.instance.config` is undefined and `profInfo.getTeamConfig().exists` should be checked instead.

Related to https://github.com/zowe/zowe-explorer-vscode/pull/3256

PR #2313, opened 2024-10-18
Fix Config API bugs that affect extenders converting V1 profiles

### LINKED ISSUE #2284: `ProfileInfo.profileManagerWillLoad` fails if `readProfilesFromDisk` was not yet called
**Describe the bug**

`ProfileInfo.profileManagerWillLoad` was added to check if the credential manager exists before loading profiles (using `readProfilesFromDisk`). However, if you haven't called `readProfilesFromDisk` yet and you call `ProfileInfo.profileManagerWillLoad`, it will throw an error. 

**Expected:** `ProfileInfo.profileManagerWillLoad` returns `true` or `false` regardless of whether profiles have been read from disk.
**Actual:** `ProfileInfo.profileManagerWillLoad` throws an error if profiles weren't yet read from disk.

t1m0thyj:
**What It Does**

Fixes #2284, #2311, #2312
- Refactored `ProfileInfo.profileManagerWillLoad` so that it works without requiring `readFromConfigDisk` to be called first.
- Fixed `ProfileInfo.onlyV1ProfilesExist` so that it works outside of Zowe CLI when `ImperativeConfig.instance.config` is undefined. This required changing it from a static method to an instance method.
- Fixed `ConvertV1Profiles` API detecting V1 profiles in `process.env.ZOWE_CLI_HOME`, but not creating V2 config there.

**How to Test**

1. Example of testing both an invalid and a good credential manager:
```javascript
const { ProfileCredentials, ProfileInfo } = require("@zowe/imperative");
(async () => {
    const profInfoBad = new ProfileInfo("zowe", {
        credMgrOverride: ProfileCredentials.defaultCredMgrWithKeytar(() => { throw new Error("bad cred mgr"); }),
    });
    console.log(await profInfoBad.profileManagerWillLoad());
    const profInfoGood = new ProfileInfo("zowe", {
        credMgrOverride: ProfileCredentials.defaultCredMgrWithKeytar(() => require("@zowe/secrets-for-zowe-sdk").keyring),
    });
    console.log(await profInfoGood.profileManagerWillLoad());
})();
```
2. (a) Should print `false` if both V1 profiles and V2 config are present, (b) Should ignore presence of meta files:
```javascript
const { ProfileInfo } = require("@zowe/imperative");
(async () => {
    const profInfo = new ProfileInfo("zowe");
    await profInfo.readProfilesFromDisk();
    console.log(profInfo.onlyV1ProfilesExist);
})();
```
3. Convert V1 profiles with a custom `ZOWE_CLI_HOME` dir:
```shell
export ZOWE_CLI_HOME=/tmp/.zowe
zowe config convert-profiles
```

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [x] updated the changelog

**Additional Comments**


packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -772,14 +777,27 @@ export class ProfileInfo {
      * detected the existence of old-school V1 profiles. We will not work with the
      * V1 profiles. This function can let you tell a user that they are incorrectly
      * trying to use V1 profiles.
-     *
+     * @deprecated Use non-static method instead to detect V2 profiles
      * @returns True - Means there is *NO* team config *AND* we detected that a V1 profile exists.
      *          False otherwise.
```
traeok, line 782: I agree with the new deprecation message, but since it no longer checks for existence of v2 profiles, should we update the `@returns` info in typedoc?

```suggestion
     * @returns {boolean} `true` if a V1 profile exists, and `false` otherwise.
```

traeok (changes requested):
This looks great, thanks for the fixes Timothy! I have a small request about updating typedoc for clarity, other than that the changes LGTM

packages/imperative/src/config/src/ProfileInfo.ts
gejohnston: I think that the original static method always tested for the existence of a team config. That test occurs in configUtils.onlyV1ProfilesExist. I thought that the reason for the new member function was to accommodate some imperative initialization actions which would not occur if the static method is called before the ProfileInfo class is instantiated.

With my level of understanding, it does raise the question of why we need to make an additional test for the existence of a team config in the new member function?

traeok: Ah, I see, thanks for clarifying on that Gene. That's actually why I made the request for changes as I noticed that the new method was doing a check for team config, so I assumed that the `ConfigUtils.onlyV1ProfilesExist` function did not. 

Since it does check for the team config in `ImperativeConfig.instance`, my request has changed to a follow-up question for my own clarification: is the team config check in the new method necessary?

anaxceron (approved):

packages/imperative/src/config/src/ProfileInfo.ts
t1m0thyj: > Could we somehow assign ImperativeConfig.instance.config from Zowe Explorer and eliminate the need for the first condition?

Yes, that approach would also work although I felt it might be a bit hacky, and I wouldn't expect other SDK users to know they need to set `ImperativeConfig.instance.config` in order for the getter to work.

Good catch on the Typedoc message, I'll update it 😋

gejohnston: Here is the way that I read this logic: The expression "ImperativeConfig.instance.config?.exists" in ConfigUtils can be false if there is no config or if there **IS** a config, but ImperativeConfig has not been fully initialized. The "extra" test in ProfileInfo avoids getting a false negative result in ConfigUtils, by ensuring that ConfigUtils is only called if there really is no config file. Please correct me if I have misinterpreted. It took me a long time re-reading this logic to draw my conclusion. The logic is not intuitively obvious, but I get that it is quite necessary.

traeok (approved):

PR #2319, opened 2024-10-23
[v3] Zos files attributes

jace-roell:
**What It Does**
Added optional `--attributes` flag to `zowe zos-files upload file-to-uss` to allow passing a .zosattributes file path for upload encoding format.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog


packages/cli/src/zosfiles/upload/ftu/FileToUSS.handler.ts
```diff
@@ -1,42 +1,93 @@
...
 
         const formatMessage = TextUtils.prettyJson(response.apiResponse);
         commandParameters.response.console.log(formatMessage);
         return response;
     }
+    private buildFilesMap(commandParameters: IHandlerParameters) {
+        let filesMap: IUploadMap = null;
+
+        // checking if binary-files or ascii-files are used, and update filesMap argument
+        if (commandParameters.arguments.binaryFiles) {
```
awharn, line 75: FileToUSS has no binaryFiles or asciiFiles option. Those are exclusive to DirToUSS.

packages/cli/src/zosfiles/upload/ftu/FileToUSS.handler.ts
```diff
@@ -1,39 +1,66 @@
...
+    IHandlerParameters,
+    ITaskWithStatus,
+    TaskStage,
+    TextUtils,
+} from "@zowe/imperative";
 import { Upload, IZosFilesResponse } from "@zowe/zos-files-for-zowe-sdk";
 import { ZosFilesBaseHandler } from "../../ZosFilesBase.handler";
+import { IUploadOptions } from "@zowe/zos-files-for-zowe-sdk";
+import { ZosFilesAttributes } from "@zowe/zos-files-for-zowe-sdk";
+import { IUploadMap } from "@zowe/zos-files-for-zowe-sdk";
```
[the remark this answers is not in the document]
jace-roell: fixed

adam-wolfe:
This seems to be working as expected. I'm wondering if we should also add this option for `zowe files dl uss-file` as well. That could be a separate pull request, though.

packages/zosfiles/__tests__/__system__/methods/upload/testfiles/encodingCheck.txt
ATorrise, line 1: i like the simplicity lol👍

packages/zosfiles/src/methods/upload/Upload.ts
```diff
@@ -854,8 +854,8 @@ export class Upload {
         };
     }
 
-    private static async uploadFile(localPath: string, ussPath: string,
-        session: AbstractSession, options: IUploadOptions) {
+    public static async uploadFile(session: AbstractSession, localPath: string, ussPath: string,
```
ATorrise, line 857: is it possible for this change of session in the parameter order to disrupt anything else in the codebase? @t1m0thyj  (gonna wait for response before approving pr @jace-roell)

awharn (approved):

ATorrise (changes requested):

packages/zosfiles/src/methods/upload/Upload.ts
t1m0thyj: This is a good catch, because we do need to be careful about changing the order of parameters.

In this case it should be OK because the method was private before, so no one was able to use it until now that we've changed it to be public 😋

packages/zosfiles/__tests__/__system__/methods/upload/testfiles/encodingCheck.txt
zFernand0, line 1: to make the test a bit more interesting... could we include these characters?

` ^ [ ] $ £ € `

zFernand0 (changes requested):
left a few comments that may help for testing and such 😋

packages/cli/src/zosfiles/upload/ftu/FileToUSS.handler.ts
```diff
@@ -1,39 +1,63 @@
...
-*/
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
```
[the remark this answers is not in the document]
zFernand0: Not saying that this needs to be updated as part of this PR 😋

PR #2320, opened 2024-10-23
Enhancing GetJobs.getJob 

pujal0909:
**What It Does**
Adds "exec-started" and "exec-ended" to IJob return data from GetJobs.getJob

**How to Test**
Download branch and test a command that uses GetJobs.getJob in the background - set a breakpoint and ensure that the job being returned has the expected timestamps. 

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x]   updated the changelog

**Additional Comments**
Other exec properties were included in the zosmf response from GetJobs. Should we include these other properties starred in the image? @t1m0thyj @zFernand0 @gejohnston @awharn 
Screenshot 2024-10-23 at 2 20 51 PM


packages/zosjobs/src/GetJobs.ts
```diff
@@ -153,6 +153,7 @@ export class GetJobs {
         Logger.getAppLogger().trace("GetJobs.getJobsCommon()");
         ImperativeExpect.toNotBeNullOrUndefined(session, "Required session must be defined");
         let query = JobsConstants.QUERY_ID;
+        params.execData = true; // always returning start and end time data
```
gejohnston, line 156: If a user explicitly turns off the execData flag ( for example --no-exec-data), I think that this statement will overwrite the user's choice. Maybe we need to check whether execData does not have a value of false, before attempting to return the new exec properties. Maybe we should set some local variable to reflect that determination and use that variable in later decisions in this function?

Admittedly, this seems like an unlikely user action, but it appears to be currently possible for a user to ask for no execData.

gejohnston (changes requested):

packages/zosjobs/src/GetJobs.ts
pujal0909: I resolved this. Let me know if it's good.

packages/zosjobs/src/GetJobs.ts
```diff
@@ -153,7 +153,7 @@ export class GetJobs {
         Logger.getAppLogger().trace("GetJobs.getJobsCommon()");
         ImperativeExpect.toNotBeNullOrUndefined(session, "Required session must be defined");
         let query = JobsConstants.QUERY_ID;
-        // params.execData = true; // always returning start and end time data
+        let includeExecData = params?.execData != false;
```
zFernand0: Should we use `!==` here? (`params?.execData !== false`)

Another alternative to default to `true` if the value is not specified is to use `??`
`parms?.execData ?? true`

packages/imperative/__tests__/src/packages/imperative/plugins/test_plugins/override_plugin/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "override-plugin",
+  "name": "@zowe/secrets-for-kubernetes-for-zowe-cli",
```
zFernand0, line 2: no objections on this, but not sure we need to modify the name of this test plugin 😋

zFernand0 (commented):
I like your observation! 🙏 
We should probably add the rest of the exec-data parameters to the iJob interface.

Ideally we would figure out a way to prevent changes in so many places 😢

zFernand0 (approved):
FWIW, this PR LGTM 😋 

Should we add a changelog entry to the CLI package since this will also affect the default behavior of the CLI? 😅 

Here is a quick before and after 🥳 

### Before
```
zFernand0 : ~/gh/zowe/cli > zowe jobs list jobs --prefix "izu*" --owner "*" --rfj | jq '.data[0]'
{
  "owner": "IZUSVR",
  "phase": 20,
  "subsystem": "JES2",
  "phase-name": "Job is on the hard copy queue",
  "job-correlator": "S0003689USI220MEDFD7F395.......:",
  "type": "STC",
  "url": "https://lpar.dev:1234/zosmf/restjobs/jobs/S0003689USI220MEDFD7F395.......%3A",
  "jobid": "STC03689",
  "class": "STC",
  "files-url": "https://lpar.dev:1234/zosmf/restjobs/jobs/S0003689USI220MEDFD7F395.......%3A/files",
  "jobname": "IZUSVR1",
  "status": "OUTPUT",
  "retcode": "SYS FAIL"
}
```

### After
```
zFernand0 : ~/gh/zowe/cli > zdev jobs list jobs --prefix "izu*" --owner "*" --rfj | jq '.data[0]'
{
  "owner": "IZUSVR",
  "phase": 20,
  "exec-member": "DE20",
  "subsystem": "JES2",
  "phase-name": "Job is on the hard copy queue",
  "job-correlator": "S0003689USI220MEDFD7F395.......:",
  "type": "STC",
  "url": "https://lpar.dev:1234/zosmf/restjobs/jobs/S0003689USI220MEDFD7F395.......%3A",
  "jobid": "STC03689",
  "exec-system": "DE20",
  "exec-submitted": "2024-10-14T08:25:30.470Z",
  "class": "STC",
  "files-url": "https://lpar.dev:1234/zosmf/restjobs/jobs/S0003689USI220MEDFD7F395.......%3A/files",
  "jobname": "IZUSVR1",
  "status": "OUTPUT",
  "retcode": "SYS FAIL",
  "exec-started": "2024-10-14T08:25:32.010Z"
}
```

packages/imperative/__tests__/src/packages/imperative/plugins/test_plugins/override_plugin/package.json
awharn: Historically, I think one of the tests changes it, and if something goes wrong, it gets left over as an artifact.

adam-wolfe:
I might be doing something wrong, but with getJobs() I do not see the new exec data, but with getJob() I do. I.e., I would expect to see the same output in the code below, but only getJob() gets the new fields:
```
    var jobsList = await GetJobs.getJobs(session)
    console.log(jobsList[0])

    var job = await GetJobs.getJob(session, jobsList[0].jobid)
    console.log(job)
```
Is this to be expected?

gejohnston (changes requested):
Your changes satisfy my earlier comments. However, Adam's comment seems like something that should be investigated and resolved before you merge your pull request. Once you address Adam's question, I am ready to approve.

packages/imperative/__tests__/src/packages/imperative/plugins/suites/sample-plugin.json
t1m0thyj, line 1: LGTM, thanks @pujal0909!

Just one request before I approve, please remove this file - I believe it may have been left behind from running tests 🙂

gejohnston (approved):

packages/zosjobs/src/GetJobs.ts
```diff
@@ -180,12 +181,6 @@ export class GetJobs {
                 }
                 query += JobsConstants.QUERY_JOBID + encodeURIComponent(params.jobid);
             }
-            if (params.execData) {
-                if (RestClient.hasQueryString(query)) {
-                    query += JobsConstants.COMBO_ID;
-                }
-                query += JobsConstants.EXEC_DATA;
-            }
```
traeok, line 188: Can we add this block back and remove the `if` block from lines 192-197? I think we'll want to keep the default behavior of only fetching exec data when the parameter is passed.

packages/zosjobs/src/doc/response/IJob.ts
```diff
@@ -34,6 +34,41 @@ export interface IJob {
...
+     */
+    "exec-submitted": string;
+
+    /**
+     * exec-system
+     * @type {String}
+     * @memberof IJob
+     */
+    "exec-system": string;
+
```
awharn, line 71: New required parameters were added to the IJob object, causing delete system tests to fail because these items must now be defined in order to delete a job. Should they be optional?

awharn (changes requested):

traeok (approved):

zFernand0:
> > ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=zowe_zowe-cli&pullRequest=2320) **Quality Gate failed**
> > Failed conditions  [25.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=zowe_zowe-cli&pullRequest=2320&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> > [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=zowe_zowe-cli&pullRequest=2320)
> 
> Can we ignore this SonarCloud Quality gate caught error? This file already contained heavily duplicated test data that supports test cases. `@t1m0thyj @zFernand0 @awharn @gejohnston`

I don't disagree with tackling this in a future issue.
The zos-jobs tests can definitely use some refactoring to reduce overall code duplication 😋

zFernand0 (approved):
LGTM! 😋 
Thanks for addressing all PR feedback 🥳 
Merging as soon as checks are done! 🥳

PR #2330, opened 2024-10-29
Add optional value to ISession for passed proxy values

JillieBeanSim:
**What It Does**
introduces an optional object in ISession for external proxy values to be sent by an extender like Zowe Explorer that would override the environment variable if set.

part one of ZE issue https://github.com/zowe/zowe-explorer-vscode/issues/3010

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**


gejohnston (approved):
I have no objections to this PR. In a response comment within this PR, could you explain the customer scenarios that this change will improve - just to better my understanding of how this change helps end users.

JillieBeanSim:
Hey @gejohnston thanks, this addition will help extenders of the cli in VS Code and/or possibly other environments.  An extender that is also a VS Code extension, ie. Zowe Explorer, may want to use the VS Code proxy settings instead of setting environment variables that are usually recognized by command line interfaces. if VSC proxy support is turned on, Zowe Explorer and it's extenders could allow users to set these values in the GUI interface and have them recognized with there rest calls.
Screenshot 2024-11-01 at 10 38 29 AM

awharn (approved):

traeok (approved):
LGTM, thanks Billie for adding support for proxies!

zFernand0:
This should be on public NPM after this workflow is done:
- https://github.com/zowe/zowe-cli-standalone-package/actions/runs/11687950797

PR #2339, opened 2024-10-31
Fix system tests for mac

awharn:
**What It Does**

Fixes various system test issues to allow the tests to run better on high performance Mac computers and alternate systems.
This includes:

- Skipping tests that rely on CIM for job cancellation
- Appending missing shebang lines
- Making shell scripts user executable
- Fixing DSN name too long errors
- Adding sleeps to scripts that require more time to pass between commands
- Fixing incorrect logic in shell scripts that used `until`
- Using commands that are intercompatible with MacOS (due to grep -P not existing on Mac)
- Fixing Logs handler incorrectly using client time when no time is specified
- Fixing an HTTP 1.1 socket reuse error with Node 20 and above
- Fixing the wrong port tests to also handle an ETIMEDOUT error due to firewalls dropping traffic instead of refusing it.

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


anaxceron (approved):

t1m0thyj (approved):
LGTM, thanks @awharn for testing the CLI on more environments and finding and fixing a few bugs along the way! 😋

PR #2352, opened 2024-11-05
Fix system tests 2

awharn:
**What It Does**

- Adds missing shebang lines to system test scripts.
- Alters `IO.createDirSync` logic to do a recursive create, and not throw an error if a race condition creates a directory after the `existsSync` check, but before the `mkdir` call.
- Fixes a bug in deleteJobs where the response was not passed along if `modifyVersion` wasn't specified, even though it was available.

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


packages/imperative/src/operations/src/Operation.ts
```diff
@@ -475,11 +474,7 @@ export abstract class Operation<T> implements ITaskWithStatus {
         for (let x = 0; x < order.length; x++) {
             this.log.info("Cleaning file: " + this.fileToUndo[x]);
             try {
-                if (fs.statSync(order[x]).isDirectory()) {
-                    removeSync(order[x]);
-                } else {
-                    fs.unlinkSync(order[x]);
-                }
+                fs.rmSync(order[x]);
```
t1m0thyj, line 477: To keep the same behavior as before, should we pass the options `{ recursive: true, force: true }` in case the item is a directory?

I'm not sure if this matters much since I believe the `Operations` class is dead code that we may plan to remove in V4.

t1m0thyj (approved):
LGTM, thanks @awharn for these fixes and removing as many references to `fs-extra` as possible! Left one question

packages/imperative/src/operations/src/Operation.ts
awharn: Sure, will make that change.

__tests__/__packages__/cli-test-utils/src/TestUtils.ts
```diff
@@ -61,6 +61,7 @@ export function runCliScript(scriptPath: string, testEnvironment: ITestEnvironme
             env: childEnv,
             encoding: "buffer"
         });
+        // eslint-disable-next-line @typescript-eslint/no-magic-numbers
```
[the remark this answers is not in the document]
awharn: No problem! :)

traeok (approved):
LGTM, thanks Andrew for fixing up the system tests and for the optimizations/bug fixes! 😁

anaxceron (approved):
Good changelog, thanks @awharn!

packages/imperative/src/operations/src/Operation.ts
```diff
@@ -475,11 +474,7 @@ export abstract class Operation<T> implements ITaskWithStatus {
         for (let x = 0; x < order.length; x++) {
             this.log.info("Cleaning file: " + this.fileToUndo[x]);
             try {
-                if (fs.statSync(order[x]).isDirectory()) {
-                    removeSync(order[x]);
-                } else {
-                    fs.unlinkSync(order[x]);
-                }
+                fs.rmSync(order[x], {recursive: true, force: true});
```
ATorrise, line 477: cool that you did extra work to simplify and clean this up! 👍 will try to follow your example in my prs :)

ATorrise (approved):

Issue #2175, opened 2024-06-09
[zowe files dl am] Add the ability to select members on a pattern or a range


FALLAI-Denis:
**Describe your enhancement idea**

The `zowe zos-files download all-members` command (`zowe files dl am` for short) allows to download all members of a PDS.

In many cases, member names meet a site-specific naming plan, and it would be desirable to be able to download based on their name:
* by applying a template based on ISPF standards: use of characters * and %,
* on a range of names with a start/from and an end/to names.

**Describe alternatives you've considered**

This would be the equivalent of the` zowe zos-files download data-sets-matching`, (`zowe files dl dsm`) command, applied to PDS members selection.

So maybe create a new command` zowe zos-files download all-members-matching`, (`zowe files dl amm`), with similar capacities:
* multiple selection expressions
* possibility of exclusion(s)

**Provide any additional context**

The z/OSMF API has `start` and `pattern` options to list the members of a PDS: [List the members of a z/OS data set](https://www.ibm.com/docs/en/zos/3.1.0?topic=interface-list-members-zos-data-set)

Thanks.

FALLAI-Denis:
👍

PR #2359, opened 2024-11-11
Select members enhancement

pujal0909:
**What It Does**
Created a new command zowe zos-files download all-members-matching that allows the user to input patterns to download specific members of a PDS (similar to zos-files download data-sets-matching).
https://github.com/zowe/zowe-cli/issues/2175

**How to Test**
Run the command zowe zos-files download all-members-matching with specific data set and patterns to download specific members.
Ex. zowe zos-files download amm "test-pds" "mem*"

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


packages/zosfiles/src/methods/download/Download.ts
```diff
@@ -298,7 +297,7 @@ export class Download {
 
             return {
                 success: true,
-                commandResponse: util.format(ZosFilesMessages.datasetDownloadedWithDestination.message, baseDir),
+                commandResponse: util.format(ZosFilesMessages.memberDownloadedWithDestination.message, baseDir),
```
awharn, line 300: This change seems to be making a handful of system tests fail, in particular the zos-files package's `partitioned data set - all members` system tests. Please update the tests.

awharn (changes requested):

packages/zosfiles/src/methods/download/Download.ts
pujal0909: That specific method is only downloading members not data sets. Originally, the message returned by it was "Data sets downloaded successfully", which I changed it to return "Member(s) downloaded successfully" since the method is downloading members.

ENCO#EDPG899423.ZOSFILE.DOWNLOAD.A63815.ENCO#ED
gejohnston, line 1: This weird empty file will added as part of this PR. Is this file an accident or part of a test for a weird file?

pujal0909: I think it was part of a system test. I deleted it.

packages/zosfiles/src/methods/download/Download.ts
gejohnston: Thanks for the clarification. That does seem like the right message for the situation. The change still raises the question of whether it would be considered a breaking change. Perhaps the team should discuss in a parking lot if the message should be improved, or remain the same because it would be considered breaking change?

adam-wolfe:
Functionally, this seems to work as expected. Is it possible to change the description for `pattern` in the help text. Right now it is:

```
   pattern               (string)

      The pattern or patterns to match data sets against. Also known as 'DSLEVEL'. The following special
      sequences can be used in the pattern:
      %: matches any single character
      *: matches any number of characters within a data set name qualifier (e.g. "ibmuser.j*.old" matches
      "ibmuser.jcl.old" but not "ibmuser.jcl.very.old")
      **: matches any number of characters within any number of data set name qualifiers (e.g. "ibmuser.**.old"
      matches both "ibmuser.jcl.old" and "ibmuser.jcl.very.old")
      However, the pattern cannot begin with any of these sequences. You can specify multiple patterns
      separated by commas, for example "ibmuser.**.cntl,ibmuser.**.jcl"
```

However, since we are filtering members, we don't really care about how these wildcards affect dsn qualifiers. If we can modify this description, I would change it to something like:

```
   pattern               (string)

      The pattern or patterns to match data sets against. Also known as 'DSLEVEL'. The following special
      sequences can be used in the pattern:
      %: matches any single character
      *: matches any number of characters within a member name
      You can specify multiple patterns separated by commas, for example ""
```
I'm not sure what we would use for member names for the multiple patterns example, but the command does work with multiple patterns.

adam-wolfe:
The changes to the help text look good. I would also remove `Also known as 'DSLEVEL'.` from the description for `pattern`, since we are filtering members, not data sets.

gejohnston (approved):

packages/cli/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ All notable changes to the Zowe CLI package will be documented in this file.
 
 ## `8.7.0`
 
+-Enhancement: Added new command zowe zos-files download all-members-matching, (zowe files dl amm), to download members matching specified pattern(s). The success message for the Download.allMembers API was changed from originally "Data set downloaded successfully" to "Member(s) downloaded successfully." The change also alters the commandResponse when using the --rfj flag. [#2359](https://github.com/zowe/zowe-cli/pull/2359)
+
+## Recent Changes
```
zFernand0, line 15: For new changelog items, I believe we want to move them towards the tope of the file. 
So, above the `## 8.8.0`  you want to have the `## Recent Changes` followed by the enhancement/bugFix entry 😋 
(Same way the changelog in zosfiles was modified) 😋

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added a `List.membersMatchingPattern` method to download all members that match a DSLEVEL pattern.[#2359](https://github.com/zowe/zowe-cli/pull/2359)
 ## `8.8.3`
```
zFernand0, line 8: nitpicky comment: 
- Can we have a new line in between? 😋

packages/zosfiles/src/methods/download/doc/IDownloadOptions.ts
```diff
@@ -103,4 +108,12 @@ export interface IDownloadOptions extends Omit<IDownloadSingleOptions, "stream">
      * Specifies whether hidden files whose names begin with a dot should be downloaded.
      */
     includeHidden?: boolean;
+
+    /**
+     * An optional pattern for restricting the response list
+     */
+    pattern?: string;
+
+    memberPatternResponse?: any;
```
zFernand0, line 117: Can we had a short description for this new optional property in the interface? 😋

zFernand0 (approved):
LGTM! 😋 

left a couple of comments, but nothing that should impact the overall behavior 🙏 

Great job! 🥳 

---

```
> zdev files download amm "sys1.parmlib" "ady*,b*" --extension ".md"
14 members(s) were found matching pattern.ls

Member(s) downloaded successfully.
Destination: sys1/parmlib
```
image

packages/cli/src/zosfiles/-strings-/en.ts
```diff
@@ -348,6 +348,23 @@ export default {
                     EX2: `Download the members of the data set "ibmuser.cntl" in text mode to the directory "jcl/"`
                 }
             },
+            ALL_MEMBERS_MATCHING: {
+                SUMMARY: "Download all members from a pds",
```
traeok, line 352: Thoughts on capitalizing "pds" here so it matches the other references to PDS in this file?

packages/zosfiles/src/constants/ZosFiles.messages.ts
```diff
@@ -407,6 +432,14 @@ export const ZosFilesMessages: { [key: string]: IMessageDefinition } = {
         message: "There are no data sets that match the provided pattern(s)."
     },
 
+    /**
+     * Message indicating that no members xsremain to be downloaded after the excluded ones were filtered out.
```
traeok, line 436:
```suggestion
     * Message indicating that no members remain to be downloaded after the excluded ones were filtered out.
```

traeok (commented):
Tested and functionality LGTM, thanks Pujal for the enhancement! I noticed a typo in the typedoc and had a suggestion for one of the strings in z/OS Files SDK.

traeok (approved):
Thanks for quickly addressing my feedback! Looks great 👍

PR #2372, opened 2024-11-19
Auth order spike

gejohnston:
**What It Does**

This PR contains a design document proposing how Zowe clients could enable users to specify their own preferred order of authentication.

This PR also contains a proof-of-concept prototype of a new AuthOrder class. When local code is modified to call AuthOrder it successfully performs the basic required functionality. 
- Since our real Zowe code has not been modified to call AuthOrder, the AuthOrder.ts file should be safe to merge into master.
- It can provide a good starting point for final implementation.

Certainly, the proposed design document should be the primary focus of reviews. Several secondary items about the document to consider are:

- I purposely chose to place this design document in the Zowe GitHub repository.
    - It is reasonable to allow interested, external people see the design of our open source app before it is implemented. 
    - Even after implementation, we can point people to such a document as an easy way to explain how the app is intended to operate.
    - We can consider whether or not we want our future designs to follow this publicly-available approach.

- It will be easier to read the document from the repo than from the file diffs.
    https://github.com/zowe/zowe-cli/blob/auth-order-spike/docs/Design_for_selecting_auth_type.md

- Alternatively, if you download the file and view it in the VSCode "Markdown Preview" window, you get a little splash of color which make some things easier to identify.

The following implementation stories were created as a result of this research spike:

[US1003134 CLI: Use the AuthOrder class in Zowe CLI/API](https://rally1.rallydev.com/#/227079578160/custom/730711030003?detail=%2Fhierarchicalrequirement%2F815981236335%2Fdetails)

[US1003140 CLI: Display authentication type from commands](https://rally1.rallydev.com/#/227079578160d/custom/54814896294?detail=%2Fuserstory%2F815983240577%2Fdetailshttps%3A%2F%2Frally1.rallydev.com%2F%23%2F227079578160d%2Fcustom%2F54814896294)

[US1003152 CLI: Include authOrder in zowe.schema.json](https://rally1.rallydev.com/#/227079578160d/custom/54814896294?Iteration=%2Fiteration%2F812764058331&Release=u&cpoid=227079578160&detail=%2Fhierarchicalrequirement%2F815983256969%2Fdetails&rankScope=%2Fiteration%2F812764058331&typeDef=25194920216)

[US1003153 CLI: Use authOrder for SSH connections](https://rally1.rallydev.com/#/227079578160d/custom/54814896294?Iteration=%2Fiteration%2F812764058331&Release=u&cpoid=227079578160&detail=%2Fhierarchicalrequirement%2F815984466197%2Fdetails&rankScope=%2Fiteration%2F812764058331&typeDef=25194920216)

[US1003155 ZE: Use the AuthOrder class in Zowe Explorer](https://rally1.rallydev.com/#/227079578160d/custom/54814896294?Iteration=%2Fiteration%2F812764058331&Release=u&cpoid=227079578160&detail=%2Fhierarchicalrequirement%2F815985428011%2Fdetails&rankScope=%2Fiteration%2F812764058331&typeDef=25194920216)

**Review Checklist**
I certify that I have:
- [x] Completed analysis and/or prototype of proposed software
- [x] Decided that no changelog update is required
- [x] Created stories for future implementation


traeok (approved):
I know that the tests are failing, but I approve of the design document, hence my approval. Very comprehensive and covered any questions that I had throughout my review. Thanks Gene for writing this proposal.

Given the context of where the changes would need to be made and following up on our previous in-person discussion, we may be able to leverage the `Config` class for these new functions. `ProfileInfo` has access to the loaded, merged config instance, so we can leverage this in Zowe Explorer to get access to the authentication order for a given profile. It seems that `ImperativeConfig.instance.config` might also be valid at the time a CLI command is run, so it could be used to access the authentication order as part of building the session.

zFernand0 (commented):
Hey Gene, 
Thanks for holding a meeting to discuss these concepts 🙏 

After moving things into a `/prototypes/imperative/src/rest/src/session/AuthOrder.ts` I think I'm fine to approve this PR 🙏

prototypes/Readme.md
```diff
@@ -0,0 +1,7 @@
...
+
+During experiments, you might place a new class into a particular package to confirm that an idea is feasible. However, that code may not be complete enough to be built and packaged into the product. Lint rules might fail on this new code. You may not have any tests yet, so code coverage verification may fail.
+
+You do not want this new code to cause Zowe build pipelines to fail, but you do not want to lose the useful work that has been completed. Placing your source file(s) under the **prototypes** directory is a way to keep your valuable experiment and not break the Zowe build process.
+
+Place your source file into a  path under the **prototypes** directory that mirrors the real directory path . Personally, I like to create a hard link from the file in my real file path to an identical file path under the **prototypes** directory. That way I can import, compile, and debug in the real directory until I complete experiments. Before committing, I can just delete the file in the real directory path and the remaining hard-linked file in the **prototypes** directory path will have all of the latest changes.
```
zFernand0, line 7: I know this is all internal docs, which is why this comment isn't a hard requirement.

Should we make this portion a bit more generic?
I like the suggestion, so maybe we could say something like...

```suggestion
Place your source file into a path under the **prototypes** directory that mirrors the real directory path. 

**Tip:** Create a hard link from the file to an identical file path under the **prototypes** directory. That way the import, compile, and debug operations work as expected. 

**Note:** Delete the file before committing. The remaining hard-linked file in the **prototypes** directory path will have all of the latest changes.
```

prototypes/packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
zFernand0, line 1: I believe this is just a copy of the original file, but with the addition of `AuthOrder.putTopAuthInSession` in the right locations. correct?

- This isn't something we need to worry about just now, but what if two different prototypes have changes on the same file? 😋
  - We can worry about this when the time comes 😋

zFernand0 (approved):
LGTM! 😋 

left some minor suggestions/questions, but nothing that should prevent this PR from being merged.

prototypes/Readme.md
gejohnston: Generalizing this advice is a good idea. I committed changes very similar to your suggestion.

prototypes/packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
gejohnston: The changes to both ConnectionPropsForSessCfg.ts and DataSet.handler.ts simply allow someone to run a quick test of the AuthOrder.ts class.

If we have changes to the same file for two different prototypes, I suggest that we could add an identifier to the the prototype's file name, like DataSet.handler_forFeature2.ts

Issue #2349, opened 2024-11-04
Give users the ability to create a copy of a pds


adam-wolfe:
**Is your feature or enhancement request related to a problem or limitation? Please describe**

For a user to create a copy of a PDS using Zowe CLI, they would have to allocate a new PDS, and then copy all of the members from the original PDS to the new PDS.

**Describe your enhancement idea**

Provide a command (or modify the existing copy command) that would create a copy of a PDS (i.e., do an `allocate like` and copy over all of the members).

**Describe alternatives you've considered**

**Provide any additional context**

It is not clear from the `zowe files copy ds` help text that PDSs cannot be copied. Trying to copy a PDS results in a semi-cryptic error message about the new data set not being catalogued.

PR #2393, opened 2024-12-20
copy pds into new target data set enhancement

pujal0909:
**What It Does**

The enhancement creates a new PDS and copies the source members when a target PDS is inputted that does not exist. The response informs the user a new PDS has been created.
https://github.com/zowe/zowe-cli/issues/2349
**How to Test**

Use the copy command with a target PDS that does not currently exist
zowe zos-files copy ds "existing-pds" "new-pds"

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


packages/zosfiles/__tests__/__system__/methods/copy/Copy.system.test.ts
```diff
@@ -61,6 +61,14 @@ describe("Copy", () => {
...
         });
         describe("Success Scenarios", () => {
+            afterEach(async () => {
+                try {
+                    await Delete.dataSet(REAL_SESSION, fromDataSetName);
+                    await Delete.dataSet(REAL_SESSION, toDataSetName);
+                } catch (err) {
+                    Imperative.console.info(`Error: ${inspect(err)}`);
+                }
+            });
```
t1m0thyj, line 71: Is this change necessary? This `afterEach` block seems redundant, since `before`/`after` blocks are inherited and there is already an `afterEach` block on line 55 that deletes these same data sets.

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -57,14 +57,20 @@ export class Copy {
         ImperativeExpect.toBeDefinedAndNonBlank(options["from-dataset"].dsn, "fromDataSetName");
         ImperativeExpect.toBeDefinedAndNonBlank(toDataSetName, "toDataSetName");
 
+        const targetDataSetExists = await this.dataSetExists(session, toDataSetName);
+
+        let newDataSet = false;
+        if(!targetDataSetExists) {
+            newDataSet = true;
```
t1m0thyj, line 64: When possible `const` is preferred over `let` 😋 This could be simplified to:
```suggestion
        const newDataSet = !targetDataSetExists;
        if (newDataSet) {
```

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -139,10 +161,7 @@ export class Copy {
      */
 
     public static async copyPDS (
-        session: AbstractSession,
-        fromPds: string,
-        toPds: string
-    ): Promise<IZosFilesResponse> {
+        session: AbstractSession, fromPds: string, toPds: string, newDataSet: boolean): Promise<IZosFilesResponse> {
```
t1m0thyj, line 164: Since `copyPDS` is a public method we can't make breaking changes to the function signature. If we add a parameter like `newDataSet` it needs to be optional:
```suggestion
        session: AbstractSession, fromPds: string, toPds: string, newDataSet = false): Promise<IZosFilesResponse> {
```

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -166,7 +185,9 @@ export class Copy {
             fs.rmSync(downloadDir, {recursive: true});
             return {
                 success:true,
-                commandResponse: ZosFilesMessages.datasetCopiedSuccessfully.message
+                commandResponse: newDataSet
+                    ? ZosFilesMessages.newDataSetCreated.message + ` - "${toPds}"`
```
t1m0thyj, line 189: We use the `ZosFilesMessages` class to define message strings in a common place so that they can be easily translated (in the future because we don't support i18n yet).

Please define the message with a `%s` placeholder so that the entire message is translatable. Then you can use the `util.format` function to replace the placeholder with a real value.
```typescript
    newDataSetCreated: {
        message: `New data set created - "%s"`
    },
```
```suggestion
                    ? util.format(ZosFilesMessages.newDataSetCreated.message, toPds)
```

t1m0thyj (changes requested):
Looks pretty good, thanks @pujal0909! Left a few comments

packages/zosfiles/__tests__/__system__/methods/copy/Copy.system.test.ts
pujal0909: Somehow that was added there by mistake. Thanks for catching that!

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -123,6 +129,22 @@ export class Copy {
         }
     }
 
+    /**
+     * Function that checks if the data set exists
+    **/
+    public static async dataSetExists(
```
t1m0thyj, line 135: If this method isn't used outside of this file, we may want to make it private for now.

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -123,6 +129,22 @@ export class Copy {
...
+        session: AbstractSession,
+        dataSetName: string
+    ): Promise<boolean> {
+        let dsnameIndex;
+        const dataSetList = await List.dataSet(session, dataSetName, {attributes: true, start: dataSetName});
```
t1m0thyj, line 140: Are attributes needed or if we just need names can we list without them?

t1m0thyj (commented):
Tested the functionality and it LGTM 🙂

@pujal0909 Please add a system test for the case where target data set does not exist.

Question for @adam-wolfe and @anaxceron - should we update the help text for the `zowe copy ds` command to explain that if the target data set does not exist, it will be created?

adam-wolfe:
> Tested the functionality and it LGTM 🙂
> 
> @pujal0909 Please add a system test for the case where target data set does not exist.
> 
> Question for @adam-wolfe and @anaxceron - should we update the help text for the `zowe copy ds` command to explain that if the target data set does not exist, it will be created?

I think so. I would change change this line: 
image

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,8 +2,11 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
-## `8.10.3`
 
+## Recent Changes
+- Enhancement: The `Copy.dataset` function now creates a new data set if the inputted target data set does not exist. [#2349](https://github.com/zowe/zowe-cli/issues/2349)
```
anaxceron, line 7: Can we use another word besides "inputted"? Like "entered" or "selected" or "included"?

anaxceron (commented):
Left a minor comment re: a changelog.

traeok (commented):
Functionality LGTM, thanks Pujal! I had couple small requests for changes and a follow-up question...

---

In the event where a user tries to copy to the same data set, can we return early from the command? Currently, if I try copying a PDS from source `A.B.C` to the same destination (`A.B.C`), the command copies each member, even though the source and destination are identical:

image

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,8 +2,11 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
-## `8.10.3`
 
+## Recent Changes
+- Enhancement: The `Copy.dataset` method now creates a new data set if the inputted target data set does not exist. [#2386](https://github.com/zowe/zowe-cli/pull/2386)
```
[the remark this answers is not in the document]
traeok: Please disregard, this comment was from earlier this morning and I see that this is now resolved 😅

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -166,7 +192,9 @@ export class Copy {
             fs.rmSync(downloadDir, {recursive: true});
             return {
                 success:true,
-                commandResponse: ZosFilesMessages.datasetCopiedSuccessfully.message
+                commandResponse: newDataSet
+                    ? util.format(ZosFilesMessages.newDataSetCreated.message, toPds)
```
awharn, line 196: Would it make sense to instead have the calling function change the command response message if it created a new data set? I just ask because this function isn't creating the new data set itself.

t1m0thyj (approved):
LGTM, thanks @pujal0909!

anaxceron (approved):

traeok (approved):

PR #2394, opened 2024-12-21
fix(secrets): Reduce keychain unlock prompts on MacOS

t1m0thyj:
**What It Does**

Reduces the number of keychain unlock prompts on MacOS for simultaneous access to secrets by multiple instances of the same application.

As a security feature, MacOS always re-prompts for keychain access after a secret has been modified by another process. This is why after running `zowe config secure` which updates credentials using `node`, you get a prompt for Zowe Explorer (the VS Code extension host process) to access credentials.

Suppose there are 3 VS Code instances open: A, B, and C. Currently when a vault change event is triggered, all 3 instances attempt to access the keychain simultaneously and Mac users will see 3 keychain prompts (ouch).

By adding a mutex associated with the calling process, we can make B & C wait to access credentials until A has finished accessing them. Now Mac users will see just 1 keychain prompt (0 would be ideal but 1 is way better than 3).

**How to Test**

This PR can only be tested on MacOS since the issue is OS-specific:
1. Download the `keyring.node` artifact and copy it into your ZE prebuilds folder.
2. Dequarantine the file with the command `xattr -d com.apple.quarantine *.node`
3. Close all VS Code windows and re-open multiple of them.
4. Run the `zowe config secure` command and press Enter repeatedly to skip all prompts.
5. Notice that only 1 prompt is shown for keychain access.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


traeok (approved):
Changes make sense to me, thanks for the fix @t1m0thyj - pretty cool that `fmutex` leverages the scope of the guard to auto-release the lock 😋 

As discussed offline, failing tests are related to `glibc` segfault with `ava` - once `workerThreads: false` is added, the stages will pass. Thanks for looking into this as well!

packages/secrets/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe Secrets SDK package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Reduced number of keychain unlock prompts on MacOS for simultaneous access to secrets by the same process.
```
gejohnston, line 7: The fix really addresses more than one process. I suggest the following wording:
... by multiple instances of the same application.

gejohnston (changes requested):
A small suggestion about the changelog. Otherwise ready to approve.

gejohnston (approved):

jace-roell (approved):

packages/secrets/core/src/os/mac/mod.rs
```diff
@@ -22,6 +23,27 @@ impl From<Error> for KeyringError {
...
+fn keyring_mutex() -> Result<Guard, KeyringError> {
+    // MacOS shows keychain prompt after secret has been modified by another process. We use cross-process mutex to
+    // block keychain access if there are multiple concurrent keychain operations invoked by the same process. This
+    // prevents multiple instances of the same app (e.g. VS Code) from triggering several keychain prompts at once.
+    let exe_path = std::env::current_exe()
+        .unwrap()
+        .to_string_lossy()
+        .replace(std::path::MAIN_SEPARATOR, "_");
+    let lock_path = std::env::temp_dir()
+        .join(format!("keyring_{}.lock", exe_path));
```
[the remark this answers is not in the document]
t1m0thyj: Resolved in https://github.com/zowe/zowe-cli/pull/2394/commits/265f087632d093ad42db0ff2608baf55f1f25d71

packages/secrets/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe Secrets SDK package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Reduced number of keychain unlock prompts on MacOS for simultaneous access to secrets by multiple instances of the same application.
```
anaxceron, line 7: Please add an issue or PR link

anaxceron (commented):
We should add an issue or PR link to the changelog entry

PR #2411, opened 2025-01-15
Download record range flag

jace-roell:
**What It Does**
Added --recordRange to `zowe jobs download output` command

**How to Test**
Run  command 'zowe jobs download o "<jobid>" --rr "0-100"` against a valid job
Where 0 is the start of the spool and 100 is the end (zero-indexed)

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog


anaxceron (approved):

packages/zosjobs/__tests__/__system__/DownloadJobs.system.test.ts
```diff
@@ -289,8 +290,86 @@ describe("Download Jobs - System tests", () => {
...
+                {
+                    jobFile: jesJCLJobFile,
+                    omitJobidDirectory: false,
+                    outDir: outputDirectory
+                }
+            );
+            expect(IO.existsSync(expectedFile)).toEqual(true);
+            expect(IO.readFileSync(expectedFile).toString()).toBeDefined();
+            expect(IO.readFileSync(expectedFile).toString()).toContain("J E S 2  J O B  L O G");
+            expect(IO.readFileSync(expectedFile).toString()).not.toContain("0------ JES2 JOB STATISTICS ------");
```
gejohnston, line 317: In addition to testing for key entries within the spool, maybe you could test for the total number of lines that were retrieved. This would confirm that different ranges actually captured different amounts of data.

gejohnston (approved):
I suggested an extra confirmation for some of your tests.
Give some consideration to whether it would be helpful.
Either way I approve this PR.

traeok (approved):
Functionality LGTM, thanks Jace!

I'm wondering if we should support for this argument to the `zowe jobs view spool-file-by-id` command as well? CLI users will likely want the record range for a specific job spool rather than for all spool files. For example, let's say you want to get the record range "50-75" for the 3rd spool file in the job - when running the download command w/ record range, that range doesn't exist and returns an error:

image

Since we use `GetJobs.getSpoolContentById` instead of `DownloadJobs.downloadSpoolContentCommon` for that command though, it may involve some refactoring, and I don't want to add scope creep. Curious to get your thoughts on this @adam-wolfe, this could be addressed separately if we want to avoid increasing scope - for now I will approve as this behavior is valid when expecting a record range that doesn't exist for all spool files.

zFernand0 (changes requested):
Functionality LGTM! 😋 
---

Can we add an example of this new option?

---

Also...
Not sure if this is fixed on another PR already, but...
I'm getting these file changes (mostly license headers being updated being)

```
On branch download-record-range
Your branch is up to date with 'origin/download-record-range'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   packages/imperative/src/utilities/__tests__/DeferredPromise.unit.test.ts
        modified:   packages/imperative/src/utilities/src/DeferredPromise.ts
```

zFernand0 (approved):
LGTM! 😋 
---

Thanks for addressing the feedback so quickly 🙏 

Just waiting for CI to finish.
Then lets have a merge-party and combine a few PRs 🥳 

---

Be on the lookout for unit and integration test snapshots

zFernand0 (approved):
LGTM! 😋 
---

PR #2419, opened 2025-01-22
copy dsclp command resulting in spacu attribute always being TRACK

jace-roell:
**What It Does**

Fixed an issue in the Copy.dataSetCrossLPAR() function where the `spacu` attribute of the copied data set was always set to `TRK`, regardless of the source data set's attributes.

**How to Test**

Run `zowe files copy dsclp <sourceMember> <targetMember> --target-zosmf-p <profile>`
Where 
- `<profile>` is the name of the profile for a remote LPAR (can be the same LPAR as the one being tested against)
- `<sourceMember>` is a member of a data set where the dataset has a `spacu` attribute of `CYLINDERS` (default)
- `<targetMember>` is the target location of the copied dataset/member

Observe that without the fix, the `spacu` of the copied data set will always be "TRACKS"
Observe that with the fix, the `spacu` of the copied data set will reflect that of the source dataset.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog


packages/zosfiles/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to the Zowe z/OS files SDK package will be documented in thi
 
 ## Recent Changes
 
+- Fixed an issue in the Copy.dataSetCrossLPAR() function where the `spacu` attribute of the copied data set was always set to `TRK`, regardless of the source data set's attributes. [#2412](https://github.com/zowe/zowe-cli/issues/2412)
```
anaxceron, line 7: I'd put the function in backticks:  " ... in the `Copy.dataSetCrossLPAR()` function where ... "

anaxceron (approved):

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,8 +2,13 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
+- Fixed an issue in the `Copy.dataSetCrossLPAR()` function where the `spacu` attribute of the copied data set was always set to `TRK`, regardless of the source data set's attributes. [#2412](https://github.com/zowe/zowe-cli/issues/2412)
+
 ## `8.12.0`
 
+- Fixed an issue in the `Copy.dataSetCrossLPAR()` function where the `spacu` attribute of the copied data set was always set to `TRK`, regardless of the source data set's attributes. [#2412](https://github.com/zowe/zowe-cli/issues/2412)
```
zFernand0, line 11: Should this be removed from here?
```suggestion
```

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -272,7 +272,7 @@ export class Copy {
                 /*
                 * If the source is a PDS and no member was specified then abort the copy.
                 */
-                if((sourceDataSetObj.dsorg == "PO" || sourceDataSetObj.dsorg == "POE") && sourceMember == undefined){
+                if((sourceDataSetObj.dsorg == "PO" || sourceDataSetObj.dsorg == "PO-E") && sourceMember == undefined){
```
zFernand0, line 275: Should we change all of these to `startsWith("PO")` ? 😋 
Otherwise, I think we should be using `===` instead of `==`

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -419,21 +419,21 @@ export class Copy {
...
+     * @param {string} getValue - The ALC value from the Get() call.
+     * @returns {string} - The ALC value in the format used by the Create() call.
      */
-    private static convertAlcTozOSMF( zosmfValue: string): string {
-        /**
-         *  Create dataset only accepts tracks or cylinders as allocation units.
-         *  When the get() call retreives the dataset info, it will convert size
-         *  allocations of the other unit types in to tracks. So we will always
-         *  allocate the new target in tracks.
-        */
```
zFernand0, line 435: Could we leave this comment here?
I guess I wouldn't want to loose some of the investigation that led to this comment 😋

zFernand0 (changes requested):
Functionality LGTM! 😋 
---

Left a few comments.
Also, I'm seeing the deferredPromise license headers being updated. 😋 
Feel free to address it on just one PR 😋 
But I believe @pujal0909 's PR already does 😋

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -316,7 +316,7 @@ export class Copy {
                     targetDataSetObj = TargetDsList.apiResponse.items[dsnameIndex];
                     targetFound = true;
 
-                    if((targetDataSetObj.dsorg == "PO" || targetDataSetObj.dsorg == "POE") && targetMember == undefined)
+                    if((targetDataSetObj.dsorg == "PO" || targetDataSetObj.dsorg == "PO-E") && targetMember == undefined)
```
zFernand0, line 319: Should we use startsWith across the entire file?
There are 4 places left 😋

zFernand0 (dismissed):
left one small comment about startsWith...
just a few more places 😋

packages/zosfiles/src/methods/copy/Copy.ts
jace-roell: Nice catch! Resolved. Thank you! @zFernand0

zFernand0 (approved):
LGTM! 😋 
---

Thanks for addressing the feedback 🙏 

Let's combine a few PRs and just have one minor release by the end of the week. 🥳

PR #2420, opened 2025-01-28
Cross-LPAR copy fails when copying a LARGE format dataset

jace-roell:
**What It Does**
- Resolved an issue where LARGE format datasets could not be copied cross LPAR due to dsorg/dsntype validation in the `Create.dataSetValidateOptions()` SDK method
- Added `--dsntype` flag to create sequential data set CLI command

**How to Test**
`zowe files create ps <sourceDataSet> --dsntype LARGE`

Run the cross LPAR copy command on this newly created data set
`zowe files copy dsclp <sourceDataSet> <targetDataSet> --target-zosmf-p <profile>`

Where
`<profile>` is the name of the profile for a remote LPAR (can be the same LPAR as the one being tested against)
`<sourceDataSet>` is a data set which has a `dsorg` of `PS-L`
`<targetDataSet>` is the target location of the copied data set

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments - Possible future story**
- Extended format sequential data sets (PS-E) face the same issue for copying data sets cross LPAR, there is additional complexities due to PS-E being created via a dsntype of `EXTREQ` OR `EXTPREF` so it may be difficult to determine how the copied data set should be created based on only "PS-E" being returned. 
- LPARs have varying support for extended format data set which may be an issue when copying cross LPAR. 
- An addtional story may be required in the future if necessary to resolve this bug.

Results from Creating data sets and copying data sets cross LPAR with varying `dsntype`s and their resultant `extx` attribute value:

Created: `EXTREQ`
Copy: `EXTREQ`
Source DS extx: 1
Target DS extx: 1

Created: `EXTPREF`
Copy: `EXTPREF`
Source DS extx: 1
Target DS extx: 1

Created: `EXTPREF`
Copy: `EXTREQ`
Source DS extx: 1
Target DS extx: 3

Created: `EXTREQ`
Copy: `EXTPREF`
Source DS extx: 1
Target DS extx: 3


packages/cli/CHANGELOG.md
```diff
@@ -1,7 +1,12 @@
 # Change Log
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added `--data-set-type` flag to create sequential data set command to allow for creating extended and large formatted sequential data sets [#2141](https://github.com/zowe/zowe-cli/issues/2141)
```
anaxceron, line 6: - Add the particle: "Added **the** `bbbb` flag ... "
- Should "command" be plural?: "... sequential data set **commands** to allow ..."
- Your sentence needs to end w/ a period.

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: The `Create.dataSetValidateOptions()` function now handles creating data sets with a `dsorg` of `PS-L` by changing the `dsntype` to `LARGE` [#2141](https://github.com/zowe/zowe-cli/issues/2141)
```
anaxceron, line 7: We should explain what `dsorg` and `dsntype` are

anaxceron (commented):
Made comments re: changelogs

anaxceron (approved):

packages/zosfiles/src/methods/create/Create.ts
```diff
@@ -249,35 +249,28 @@ export class Create {
...
-                        // Validate non-zero if 'dsorg' is PO
-                        if (tempOptions.dirblk === 0 && tempOptions.dsorg === "PO") {
+                        // Validate non-zero if dsorg starts with "PO"
+                        if (tempOptions.dirblk === 0 && tempOptions.dsorg.startsWith("PO")) {
                             throw new ImperativeError({ msg: ZosFilesMessages.invalidPODsorgDirblkCombination.message });
                         }
-
                         break;
 
-                    case "dsntype": {
```
awharn, line 263: Be aware that removal of this code results in the breakage of system test `Create Partitioned Data Set › Expected failures › should fail creating a partitioned data set extended due to invalid data-set-type` in `packages/cli/__tests__/zosfiles/__system__/create/pds/cli.files.create.dataset.pds.system.test.ts`. This test will need to be rewritten or removed if we plan to remove this validation.

packages/zosfiles/src/methods/create/Create.ts
```diff
@@ -249,35 +249,28 @@ export class Create {
...
-                    // Key to create a PDSE.
-                        const type: string = tempOptions.dsntype.toUpperCase();
-                        const availableTypes = ["BASIC", "EXTPREF", "EXTREQ", "HFS", "LARGE", "PDS", "LIBRARY", "PIPE"];
```
gejohnston, line 266: Wouldn't the removal of the test of availableTypes remove all validation of the dsnType command line option? I could supply an arbitrary string (like 'quicksand') as a dsnType. Have we made a conscience decision to let z/OS do the validation and return the z/OS message if an error occurs? Could this be considered a breaking change (admittedly an edge-case breakage)?

gejohnston (commented):
If the concerns in my question have already been considered, then I will approve.

packages/cli/src/zosfiles/create/ps/Ps.definition.ts
```diff
@@ -53,6 +53,7 @@ export const PsDefinition: ICommandDefinition = {
         ZosFilesCreateOptions.mgntclass,
         ZosFilesCreateOptions.dataclass,
         ZosFilesCreateOptions.unit,
+        ZosFilesCreateOptions.dsntype,
```
zFernand0, line 56: should we add an example for this new option?

packages/zosfiles/src/methods/create/Create.ts
zFernand0: TL;DR:  yeah, we want this back 🙏 
---

This is my fault. 
I was confused on the extent of our validation for `dsntype` and how that impacted the DS creation process.
Since this list matches the items listed on z/OSMF, I believe there is no need for us to let the request fail on z/OS.

**_Source_**: https://www.ibm.com/docs/en/zos/3.1.0?topic=dp-subparameter-definition-8

zFernand0 (approved):
LGTM! 😋 
---

Left a small comment about an example 🙏 
and also, Gene's comment is very valid. My apologies on the confusion 😢

packages/cli/__tests__/zosfiles/__system__/create/pds/cli.files.create.dataset.pds.system.test.ts
```diff
@@ -199,7 +199,8 @@ describe("Create Partitioned Data Set", () => {
         it("should fail creating a partitioned data set extended due to invalid data-set-type", () => {
             const response = runCliScript(__dirname + "/__scripts__/command/command_create_pdse_fail_dsntype.sh",
                 TEST_ENVIRONMENT, [user]);
-            expect(response.stderr.toString()).toContain("Invalid zos-files create command 'dsntype' option: NONLIBRARY");
+            expect(response.stderr.toString()).toContain("Unable to perform this operation due to the following problem.");
+            expect(response.stderr.toString()).toContain("NONLIBRARY");
```
traeok, line 203: I see that the system test was changed after `dsntype` validation was modified - since we are keeping the validation as-is now, should the changes in this test be reverted?

packages/zosfiles/src/methods/create/Create.ts
```diff
@@ -249,35 +249,34 @@ export class Create {
                         break;
 
                     case "dirblk":
-                    // Validate non-zero if dsorg is PS
-                        if (tempOptions.dirblk !== 0 && tempOptions.dsorg === "PS") {
+                        // Validate non-zero if dsorg starts with "PS"
+                        if (tempOptions.dirblk !== 0 && tempOptions.dsorg.startsWith("PS")) {
```
traeok, line 253: Thanks for refactoring to use `startsWith` 😋

traeok (changes requested):
Thanks for working on this fix Jace! Functionality makes sense to me, but I have a couple requests

traeok (approved):
LGTM, thanks Jace for the fix!

gejohnston (approved):

zFernand0:
Apologies for going back and forth on the release labels. I noticed a comment about a possible breaking change in behavior in another PR

> we don't need to be concerned about a breaking change in this case - although the PR adding the --safe-replace option was merged last week it hasn't been published yet.

Got a bit confused, but I believe that even if we publish this as a minor, we can publish a patch soon if needed

Issue #2259, opened 2024-09-10
Display secure values in `zowe config list` when the property `ZOWE_SHOW_SECURE_ARGS` is set


Joe-Winchester:
**Is your feature or enhancement request related to a problem or limitation? Please describe**

There are times when it's necessary to view the secure properties (for example to check whether there is a stale password).
This can be done in sorts using `zowe ...-show-inputs-only` to create a dry run command with `ZOWE_SHOW_SECURE_ARGS` which echos the value for that single command.

**Describe your enhancement idea**

Enhance `ZOWE_SHOW_SECURE_ARGS` to cover the `zowe config list` command.
This would allow a user to do a full display of all of the zowe configuration properties AND to see the secure properties.
This would help to show any property values stored securely to be able to view their contents and make actions such as updating expired passwords, refactoring profiles that are sharing the same TSO ID to be under a single nested profile, and more ...

**Provide any additional context**

This was brought up as a suggestion on the System Z Enthusiasts Discord Community Zowe channel on September 6th by PJ Farley.
https://discord.com/channels/880322471608344597/1229969123040301148/1281671442571857993

pjfarleyiii:
I put a thumbs-up emoji on this enhancement request, and would upvote it as well if someone can tell me where and how to do that.

zFernand0:
This sounds like a great enhancement for V3 🥳 

As a workaround for V2 users, you could set the env variable temporarily, use the `show-inputs-only`, and specify each profile. For example:
- `ZOWE_SHOW_SECURE_ARGS=TRUE zowe zosmf check status --show-inputs-only --zosmf-p DEV.lparX.zosmf`
- `ZOWE_SHOW_SECURE_ARGS=TRUE zowe zosmf check status --show-inputs-only --zosmf-p QA.zosmf`
- `ZOWE_SHOW_SECURE_ARGS=TRUE zowe zosmf check status --show-inputs-only --zosmf-p prod_zosmf`
- `ZOWE_SHOW_SECURE_ARGS=TRUE zowe zosmf check status --show-inputs-only --base-p base`

adam-wolfe:
Reviving this issue and linking to #2430

PR #2424, opened 2025-02-03
Secure value censorship refactor

awharn:
**What It Does**

Refactors LoggerUtils into a new Censor class to globally use the same list of censored options, and intelligently add options to censor on the fly. Also removes duplication (and deprecates functions) to the old utilities. Points old utilities to use the new utility when possible.

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/src/censor/src/Censor.ts
+++ b/packages/imperative/src/censor/src/Censor.ts
@@ -27,10 +27,10 @@ export class Censor {
     *
     * NOTE(Harn): This list should be kept in sync with the base profile secure definitions and MUST be in camel case.
     */
-    private static readonly MAIN_CENSORED_OPTIONS = ["auth", "p", "pass", "password", "passphrase", "credentials",
+    private static readonly MAIN_CENSORED_OPTIONS = ["auth", "pw", "pass", "password", "passphrase", "credentials",
         "authentication", "basicAuth", "tv", "tokenValue", "certFilePassphrase"];
 
-    private static readonly MAIN_SECURE_PROMPT_OPTIONS = ["user", "password", "tokenValue", "passphrase"];
+    private static readonly MAIN_SECURE_PROMPT_OPTIONS = ["user", "password", "tokenValue", "passphrase", "keyPassphrase"];
 
     // The censor response.
     public static readonly CENSOR_RESPONSE = "****";
@@ -74,7 +74,22 @@ export class Censor {
      * @returns True - if the given property is to be treated as a special value; False - otherwise
      */
     public static isSpecialValue(prop: string): boolean {
-        for (const v of this.CENSORED_OPTIONS) {
+        let specialValues = this.SECURE_PROMPT_OPTIONS;
+        const getPropertyNames = (prop: ICommandProfileProperty): string[] => {
+            const ret: string[] = [];
+            ret.push(prop.optionDefinition?.name);
+            prop.optionDefinitions?.map(opDef => ret.push(opDef.name));
+            return ret;
+        };
+
+        for (const profile of this.profileSchemas) {
+            // eslint-disable-next-line unused-imports/no-unused-vars
+            for (const [_, prop] of Object.entries(profile.schema.properties)) {
+                if (prop.secure) specialValues = lodash.union(specialValues, getPropertyNames(prop));
+            }
+        }
+
+        for (const v of specialValues) {
             if (prop.endsWith(`.${v}`)) return true;
         }
         return false;
@@ -168,7 +183,7 @@ export class Censor {
                     for (const prof of profiles) {
                         // If the profile exists, append all of the secure props to the censored list
                         const profName = censorOpts.commandArguments?.[`${prof}-profile`];
-                        if (censorOpts.config.api.profiles.get(profName)) {
+                        if (profName && censorOpts.config.api.profiles.get(profName)) {
                             censorOpts.config.api.secure.securePropsForProfile(profName).forEach(prop => this.addCensoredOption(prop));
                         }
                     }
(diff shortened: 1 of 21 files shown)


packages/imperative/CHANGELOG.md
```diff
@@ -2,7 +2,19 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Fixed inconsistent behavior with the `ZOWE_SHOW_SECURE_ARGS` environment variable continuing to mask secure properties when it should not. [#2430](https://github.com/zowe/zowe-cli/issues/2430)
+- Enhancement: Added the `Censor` class, consolidating all sensitive data hiding logic into one class. [#2424](https://github.com/zowe/zowe-cli/pull/2424)
+- Enhancement: Added `showSecureArgs` to `EnvironmentalVariableSettings` to allow extenders to determine if they should mask secure values. [#2424](https://github.com/zowe/zowe-cli/pull/2424)
```
anaxceron, line 9: we should explain what `showSecureArgs` and `EnvironmentalVariableSettings` are. Are they classes, properties, cats?
Comment applies to Lines 11-14 as well.

anaxceron (commented):
Left some comments re: minor edits for changelog

anaxceron (approved):
thanks for the changelog update, @awharn, everything looks good

pjfarleyiii:
Question from a CLI user:  Will this resolve the non-printing of secure user and password values in the output from `zowe config list`?  And if yes, will a new release be posted to the npm registry so that ordinary users can get back the ability to see secure values when really needed?

adam-wolfe:
> Question from a CLI user: Will this resolve the non-printing of secure user and password values in the output from `zowe config list`? And if yes, will a new release be posted to the npm registry so that ordinary users can get back the ability to see secure values when really needed?

That was not the intention with this PR, which was mainly to clean up how we determine what should be considered to be a secure property.

However, it's something we should consider. We'll take another look at https://github.com/zowe/zowe-cli/issues/2259.

packages/imperative/src/censor/src/Censor.ts
```diff
@@ -0,0 +1,340 @@
...
+                        }
+                    }
+                } else {
+                    // We only have a configuration file, assume every property that is secured should be censored
+                    censorOpts.config.api.secure.findSecure(censorOpts.config.mProperties.profiles, "profiles").forEach(
+                        prop => secureOptions.add(prop.split(".").pop())
+                    );
+                }
+                secureOptions.forEach(prop => this.addCensoredOption(prop));
+            }
```
traeok, line 258: SonarCloud flagged the `setCensoredOptions` function for complexity. I was wondering if we could move this part of the function into a helper function to reduce complexity? I think it would make it a bit easier to follow the different branches of logic, while also resolving the SonarCloud issue as a result.

packages/imperative/src/censor/src/Censor.ts
```diff
@@ -0,0 +1,340 @@
...
+        const censoredList = new Set<string>();
+        for (const option of this.MAIN_SECURE_PROMPT_OPTIONS) {
+            censoredList.add(option);
+            censoredList.add(CliUtils.getOptionFormat(option).kebabCase);
+        }
+        return Array.from(censoredList);
+    }
+
+    // Set a censored options list that can be set and retrieved for each command.
+    private static censored_options: Set<string> = new Set(this.DEFAULT_CENSORED_OPTIONS);
```
traeok, line 65: Not anything major, but I noticed that this property is formatted as lowercase & snake case. Shortly after in this class, the `m` prefix and camel case is used to declare the remaining private class members. Please update this to follow the same naming scheme

packages/imperative/src/censor/src/Censor.ts
```diff
@@ -0,0 +1,340 @@
...
+            ret.push(prop.optionDefinition?.name);
+            prop.optionDefinitions?.map(opDef => ret.push(opDef.name));
+            return ret;
+        };
+
+        for (const profile of this.profileSchemas) {
+            // eslint-disable-next-line unused-imports/no-unused-vars
+            for (const [_, prop] of Object.entries(profile.schema.properties)) {
+                if (prop.secure) specialValues = lodash.union(specialValues, getPropertyNames(prop));
+            }
```
traeok, line 181: If we aren't using the key, we should iterate using `Object.values` - this would eliminate the need for the eslint comment above

packages/imperative/src/censor/src/Censor.ts
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        for (const v of specialValues) {
+            if (prop.endsWith(`.${v}`)) return true;
+        }
+        return false;
```
traeok, line 187: Thoughts on joining these two possible return values? For example:

```suggestion
        return specialValues.some((v) => prop.endsWith(`.${v}`));
```

traeok (commented):
Thanks for all your work on this Andrew! I see there was quite a bit of refactoring done to keep our behavior consistent - its much appreciated 🙏 I had a couple comments from the code review, but wanted to post those ahead of time before continuing with manual testing.

packages/imperative/src/censor/src/Censor.ts
awharn: We could do that, though the data would need to be passed through to the second helper function, and it would probably only be called in this function. I attempted to do some reduction of calling `addCensoredOption` in here, which might not be as easily possible if I break this up into multiple functions.

traeok (approved):
LGTM, thanks Andrew!

gejohnston (approved):

PR #2427, opened 2025-02-03
Identical member names - PDS enhancement 

pujal0909:
**What It Does**

When copying PDSs, if the source and target have identical member names, the default behavior now is to prompt the user to confirm before overwriting the members contents (when the safe-replace option is not set to true). If the user uses the `replace` flag, it bypasses the prompt. 

**How to Test**

1. Copy a source PDS into a target PDS, with both having members with the same names. 
2. Notice that a prompt appears where you can select 'y/N'
3. Now copy the same source PDS into the same target PDS with the `replace` option. 
4. Notice that the prompt is skipped

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


packages/cli/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to the Zowe CLI package will be documented in this file.
 
 ## Recent Changes
 
+- Enhancement: When copying PDSs that have identical member names, the user is now prompted to confirm before the operation occurs in case of data loss. [#2349] (https://github.com/zowe/zowe-cli/issues/2349)
```
anaxceron, line 7: - Let's write out the PDSs: partitioned data sets
- might be clearer to say: " ... is now prompted to confirm the operation in case of potential data loss."

packages/cli/__tests__/zosfiles/__unit__/copy/ds/Ds.handler.unit.test.ts
```diff
@@ -234,7 +237,79 @@ describe("DsHandler", () => {
...
+            }
+        };
+        const promptMock = jest.fn();
+        promptMock.mockResolvedValue("y");
+
+        const promptForDuplicates = (handler as any)["promptForLikeNamedMembers"]({ prompt: promptMock });
+        const result = await promptForDuplicates();
+
+        expect(promptMock).toHaveBeenCalledWith(
+            `The source and target data sets have identical member names. The contents of those members will be overwritten.` +
```
anaxceron, line 276: I think it would be clearer to say which members will be overwritten: "The contents of the target members will be ..."

packages/cli/__tests__/zosfiles/__unit__/copy/ds/Ds.handler.unit.test.ts
```diff
@@ -234,7 +237,79 @@ describe("DsHandler", () => {
...
+            }
+        };
+        const promptMock = jest.fn();
+        promptMock.mockResolvedValue("N");
+
+        const promptForDuplicates = (handler as any)["promptForLikeNamedMembers"]({ prompt: promptMock });
+        const result = await promptForDuplicates();
+
+        expect(promptMock).toHaveBeenCalledWith(
+            `The source and target data sets have identical member names. The contents of those members will be overwritten.` +
```
anaxceron, line 312: see my comment above

packages/cli/src/zosfiles/copy/ds/Ds.handler.ts
```diff
@@ -35,10 +36,20 @@ export default class DsHandler extends ZosFilesBaseHandler {
     private promptForSafeReplace(console: IHandlerResponseConsoleApi) {
         return async (targetDSN: string) => {
             const answer: string = await console.prompt(
-                `The dataset '${targetDSN}' exists on the target system. This copy will result in data loss.` +
+                `The dataset '${targetDSN}' exists on the target system. This copy may result in data loss.` +
```
anaxceron, line 39: "can" instead of "may"?

packages/cli/src/zosfiles/copy/ds/Ds.handler.ts
```diff
@@ -35,10 +36,20 @@ export default class DsHandler extends ZosFilesBaseHandler {
...
                 ` Are you sure you want to continue? [y/N]: `
             );
             return answer != null && (answer.toLowerCase() === "y" || answer.toLowerCase() === "yes");
         };
     }
+
+    private promptForLikeNamedMembers(console: IHandlerResponseConsoleApi) {
+        return async() => {
+            const answer: string = await console.prompt (
+                `The source and target data sets have identical member names. The contents of those members will be overwritten.` +
```
anaxceron, line 49: can we specify which members will be overwritten?

packages/zosfiles/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to the Zowe z/OS files SDK package will be documented in thi
 
 ## Recent Changes
 
+- Enhancement: When copying PDSs that have identical member names, the user is now prompted to confirm before the operation occurs in case of data loss. [#2349] (https://github.com/zowe/zowe-cli/issues/2349)
```
anaxceron, line 7: please see my previous changelog comment

anaxceron (commented):
Left some comments, please review.

anaxceron (approved):

packages/cli/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to the Zowe CLI package will be documented in this file.
 
 ## Recent Changes
 
+- Enhancement: When copying partitioned data sets that have identical member names, the user is now prompted to confirm before the operation occurs in case of potential data loss. [#2349] (https://github.com/zowe/zowe-cli/issues/2349)
```
traeok, line 7: I know that it takes some time for users to start leveraging new command functionality, but we already released a version that copies one PDS to another without prompting for data loss. This PR changes the default behavior for the copy command so that users are now prompted before the copy can continue.

While this work definitely enhances the behavior of the existing command (and I prefer the new behavior as well), I think we might want to label it as a bug fix to justify the breaking change.

packages/zosfiles/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to the Zowe z/OS files SDK package will be documented in thi
 
 ## Recent Changes
 
+- Enhancement: When copying partitioned data sets that have identical member names, the user is now prompted to confirm before the operation occurs in case of potential data loss. [#2349] (https://github.com/zowe/zowe-cli/issues/2349)
```
traeok, line 7: See above comment regarding labeling as a bug fix. We could rephrase this entry to justify the change:

```suggestion
- BugFix: Users were not warned when copying partitioned data sets with identical member names. Now, the user is prompted to confirm before continuing the copy operation to avoid potential data loss. [#2349] (https://github.com/zowe/zowe-cli/issues/2349)
```

packages/cli/CHANGELOG.md
t1m0thyj: I believe we don't need to be concerned about a breaking change in this case - although the PR adding the `--safe-replace` option was merged last week it hasn't been published yet.

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -170,6 +180,22 @@ export class Copy {
...
 
+    /**
+     * Function that checks if source and target data sets have identical member names
+    */
+    private static async hasIdenticalMemberNames (
+        session: AbstractSession,
+        fromPds: string,
+        toPds: string
+    ): Promise <boolean> {
+        const sourceResponse = await List.allMembers(session, fromPds);
```
zFernand0, line 191: I do like how clean this implementation looks! 🙏 
And I know this might be a bit picky, but in order to minimize the number of calls to the mainframe, could we cache this call? 
`await List.allMembers(session, fromPds);`

or perhaps return it as part of the function too?  😅 

That way, the `copyPDS doesn't have to make the same call to get the list of members from the `fromPds` 😋

packages/cli/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ All notable changes to the Zowe CLI package will be documented in this file.
 
 ## Recent Changes
 
-- Enhancement: When copying partitioned data sets that have identical member names, the user is now prompted to confirm before the operation occurs in case of potential data loss. [#2349] (https://github.com/zowe/zowe-cli/issues/2349)
+- Enhancement: Users were not warned when copying partitioned data sets with identical member names. Now, the user is prompted to confirm before continuing the copy operation to avoid potential data loss. [#2349] (https://github.com/zowe/zowe-cli/issues/2349)
```
zFernand0, line 7: Going back to Trae's comment, should we allso document this one here as a bugFix? 😋 

Thanks for making the behavior safer 🙏

zFernand0 (approved):
LGTM! 😋 
---

I did leave some comments for possible improvements, but I do like the code as-is and works as expected 🙏

traeok (approved):
Changes LGTM, thanks @pujal0909 for enhancing the copy PDS command!

packages/cli/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ All notable changes to the Zowe CLI package will be documented in this file.
 
 ## `8.13.0`
 
+## Recent Changes
+
+- BugFix: Users were not warned when copying partitioned data sets with identical member names. Now, the user is prompted to confirm before continuing the copy operation to avoid potential data loss. [#2349] (https://github.com/zowe/zowe-cli/issues/2349)
```
t1m0thyj, line 8: Please move to the top of the changelog above the `8.13.0` header.

packages/zosfiles/__tests__/__system__/methods/copy/Copy.system.test.ts
```diff
@@ -65,12 +65,21 @@ describe("Copy", () => {
...
+                    Imperative.console.info(`Error: ${inspect(err)}`);
+                }
+            });
             describe("Sequential > Sequential", () => {
                 beforeEach(async () => {
                     try {
                         await Create.dataSet(REAL_SESSION, CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL, fromDataSetName);
                         await Create.dataSet(REAL_SESSION, CreateDataSetTypeEnum.DATA_SET_SEQUENTIAL, toDataSetName);
                         await Upload.bufferToDataSet(REAL_SESSION, Buffer.from("1234"), fromDataSetName);
+                        // await Upload.bufferToDataSet(REAL_SESSION, Buffer.from("abc"), fromDataSetName);
```
t1m0thyj, line 82: Can this line be removed?

packages/zosfiles/src/constants/ZosFiles.messages.ts
```diff
@@ -198,6 +197,13 @@ export const ZosFilesMessages: { [key: string]: IMessageDefinition } = {
         message: "Member(s) downloaded successfully.\nDestination: %s"
     },
 
+    /**
+     * Message indicating that the members contents were truncated due to lrecl
+     * @type {IMessageDefinition}
+     */
+    membersContentTruncated: {
+        message: "Member(s)' contents were truncated due to insufficient record lines. You can view the list of members here: %s"
```
t1m0thyj, line 205: Since `lrecl` is Logical Record Length, should this say "insufficient record length"? I believe that would be clearer than "lines" since the issue is the max length of lines not the total number of lines?

t1m0thyj (changes requested):
Looks pretty good, thanks @pujal0909! Left a few comments

t1m0thyj (approved):
LGTM, thanks @pujal0909!

packages/zosfiles/__tests__/__system__/methods/copy/Copy.system.test.ts
```diff
@@ -19,12 +19,13 @@ import { ITestPropertiesSchema } from "../../../../../../__tests__/__src__/prope
 import { join } from "path";
 import { readFileSync } from "fs";
 import { ITestEnvironment } from "../../../../../../__tests__/__src__/environment/ITestEnvironment";
-import { tmpdir } from "os";
 import path = require("path");
 import * as fs from "fs";
+import { List, ZosFilesUtils } from "@zowe/zos-files-for-zowe-sdk";
```
awharn, line 24: List and ZosFilesUtils are already defined in the import statement in Line 12.

awharn (changes requested):
This PR looks good, but there is a duplicate import in the system test that needs to be addressed.

awharn (approved):
Looks good to me, thanks @pujal0909!

PR #2433, opened 2025-02-09
Record lines truncation of target PDS Enhancement

pujal0909:
**What It Does**

In the scenario when copying from a source PDS to a target PDS where the target does not have enough record lines for a member, initially, it would copy over that member, truncate it's record length, throw a truncation error and would not continue copying over the subsequent members. Now, it will continue copying over the subsequent members and inform the user that member(s)' contents were truncated and they can view the list of members affected in a local file.

**How to Test**

1. With the current functionality, copy a source PDS into a target PDS. The target PDS should have a lower record lines value than the source.
2. Notice that a member with a long record length will be truncated, and subsequent members will not continue to copy over.
3. Now with the updated functionality, copy the same source PDS into a target PDS with a lower record lines value than the source.
4. Notice that the user is informed of the truncation and is given a local file of members with this error. The subsequent members were copied over as well.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

Should this technically be a bug fix?
Should the Github security suggestion about using the tmp library be ignored?


packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -201,13 +226,36 @@
...
+                    }
+                    else {
+                        Logger.getAppLogger().error(error);
+                    }
+                    continue;
+                }
+            }
+            const truncatedMembersFile = path.join(tmpdir(), 'truncatedMembers.txt');
+            if(truncatedMembers.length > 0) {
+                fs.writeFileSync(truncatedMembersFile, truncatedMembers.join('\n'));
```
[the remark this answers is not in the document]
gejohnston: This AI suggestion implies that a temp file will have a generated name.
I prefer your approach in which one name is consistently used.

gejohnston (approved):

traeok:
If we already warn users that some members were truncated from the copy operation, what are your thoughts on listing the affected members as part of the CLI response? Otherwise, users are forced to locate & open the local file to see what members were truncated.

If there were more than, say, 10 members, we could print so many and then add `...and X more` at the end of the output.

This isn't a request for changes, but I wanted to propose this as an option for warning the user - this may also help with identifying the troublesome members when copying within a CI environment.

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -201,13 +226,42 @@ export class Copy {
...
+                const firstTenMembers = truncatedMembers.slice(0, 10);
+                fs.writeFileSync(truncatedMembersFile, truncatedMembers.join('\n'));
+                const numMembers = truncatedMembers.length - firstTenMembers.length;
+                return {
+                    success: true,
+                    commandResponse:
+                        ZosFilesMessages.datasetCopiedSuccessfully.message + " " +
+                        ZosFilesMessages.membersContentTruncated.message + "\n\n" +
+                        firstTenMembers.join('\n') +
+                        `\n... and ${numMembers} more` +
```
gejohnston, line 259: Could this logic result in the message "and 0 more" ?

gejohnston: This is definitely a judgement call. If I were to hit the edge-case of exactly 10, I would think it odd to be told that "0 more" occurred. Alternatively if someone were automating on the output, it might be useful to know that "x more" would always be part of the message. I lean to removing the message when it would display "0 more". Perhaps wait a bit to see if others offer opinions.

ATorrise: i agree that adding "...and X more" should only be included if X is greater than 1

ATorrise (changes requested):
agreeing with your discussion with Gene that your truncation response should only include an attached file if truncated items list is greater than 10.

traeok (dismissed):
This looks great so far, thanks for enhancing the truncation warning! I have a small request regarding an `fs.rmSync` call but aside from that the functionality LGTM

zFernand0 (approved):
LGTM! 😋 
---

I'm curious to know if it would be possible to add a progress bar.
Trying to copy a PDS with almost 200 members looks like the command hands 😋 

Also... I'm afraid that some errors are not being shown 🤔 
image

---

As a final test, I let it run and here is what happened:
image

Here you can see that the command result only mentions the members that it was able to copy. (24 out of 193) 
But it didn't mention the members that were not copied. likely due to not having enough space on the target PDF 🙏 

---

All of that said, I do believe that this PR does handle quite a bit of scenarios and we could tackle these edge cases in a separate issue 🙏

packages/cli/CHANGELOG.md
```diff
@@ -1,9 +1,9 @@
 # Change Log
 All notable changes to the Zowe CLI package will be documented in this file.
 
-
 ## Recent Changes
 
+- BugFix: When using the `copy` command, when a target partitioned data set has a smaller record length than a source partitioned data set, it no longer stops the operation for subsequent members. The user can now view the affected members in a local file. [#2349] (https://github.com/zowe/zowe-cli/issues/2349)
```
anaxceron, line 6: - For better readability, change the second "when" to "if": "When using the `copy` command, **if** a target partitioned data set ... ."
- Also, rephrase the sentence a bit: " ... a source partitioned data set, the operation for subsequent members **no longer stops**." OR: " ... a source partitioned data set, the operation for subsequent members **is no longer interrupted**."

packages/zosfiles/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
 ## Recent Changes
+
+- BugFix: When using the `copy` command, when a target partitioned data set has a smaller record length than a source partitioned data set, it no longer stops the operation for subsequent members. The user can now view the affected members in a local file. [#2349] (https://github.com/zowe/zowe-cli/issues/2349)
```
anaxceron, line 7: See my previous comment

anaxceron (commented):
Left a couple comments re: changelog.

zFernand0:
Found this issue while testing, but this also on the `master` branch
```shell
> zdev files copy ds DS.DOES.NOT.EXIST fernando.test.pds.small      [±truncate-lrecl-target ✓(✹)]
Cannot read properties of undefined (reading 'dsorg')

Response From Service
TypeError: Cannot read properties of undefined (reading 'dsorg')
```


I believe we will need to handle missing datasets in the `isPDS` function, and maybe fail a bit more gracefully 🙏 
image

packages/zosfiles/src/methods/copy/Copy.ts
```diff
@@ -226,14 +237,51 @@
...
+                    else {
+                        Logger.getAppLogger().error(error);
+                    }
+                    continue;
+                }
+            }
+            const truncatedMembersFile = path.join(tmpdir(), 'truncatedMembers.txt');
+            if(truncatedMembers.length > 0) {
+                const firstTenMembers = truncatedMembers.slice(0, 10);
+                fs.writeFileSync(truncatedMembersFile, truncatedMembers.join('\n'), {flag: 'w'});
```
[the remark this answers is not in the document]
zFernand0: We've already determined that we are fine with writing the list of truncated members in a temporary file : )

Ideally we would make sure the file permissions are 600

PR #2436, opened 2025-02-14
Header generation refactoring

ATorrise:
This refactoring effort centralizes and standardizes header generation across the Files SDK and CLI, consolidating all header logic into a single utility class—`ZosFilesHeaders`—and replacing duplicate header creation code across multiple methods.

The new implementation leverages a unified `headerMap` to manage header generation functions, ensuring that headers like responseTimeout, etag, and others are applied consistently in all ZosFiles REST calls. 

- Header Generation now:
  - Requires:
    - the request `options`
  - Optionally takes:
    - `dataLength`  parameter to add a Content-Length header when needed.
    - `ZosFilesContext` enum that guides which Content-Type headers are applied for certain methods. These enum values include:
      - `USS_MULTIPLE`
        - Forces a general content-type header ("Content-Type": "application/json") when multiple USS files.
      - `ZFS`& `LIST`
        - No content-type headers are applied.
      - `UPLOAD`
        - Applies the content-type header ("X-IBM-Data-Type": encoding ). And will pass `Content-Type: application/octet-stream` for binary data. 
      - `DOWNLOAD`
        - applies `Content-Type` header. 
      - No context (default)
        - applies the default content-type header: {"X-IBM-Data-Type": "text"}
  
  ```typescript
  ZosFilesHeaders.generateHeaders<T>({
    options,
    context,
    dataLength,
  }: {
    options: T;
    context?: ZosFilesContext;
    dataLength?: number | string;
  }): IHeaderContent[]
  ```
  
 
Because of this centralization, organization, and newly-provided, clear header expectations, any future header changes or additions require modifications in only one location. Creating headers can now be done with a single function call without much thought required in future development of the methods within ZosFiles SDK.


packages/cli/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Created a centralized header class to be used across the files SDK with `ZosFilesHeaders.ts`. [#2436](https://github.com/zowe/zowe-cli/pull/2436)
```
anaxceron, line 6: We should say what `ZosFilesHeaders.ts` is: " ... across the files SDK with the **file/database/function/whatever** `ZosFilesHeaders.ts`."

packages/cli/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
...
+
 ## `8.13.0`
 
 - Enhancement: Added the `--data-set-type` flag to create sequential data set command to allow for creating extended and large formatted sequential data sets. [#2141](https://github.com/zowe/zowe-cli/issues/2141)
```
anaxceron, line 10: is there a word missing here?: " ... to create sequential data set command ... "

packages/core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe core SDK package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added recursive optional header to `ZosmfHeaders`. [#2436](https://github.com/zowe/zowe-cli/pull/2436)
```
anaxceron, line 7: - Is there a word missing here?: "Added recursive optional header ... "
- We should explain what `ZosmfHeaders` is: " ... to the **file/profile/whatever** `ZosmfHeaders`."

anaxceron (commented):
Left some comments re: changelogs

gejohnston (approved):

awharn (changes requested):
I attempted to run a `zowe files list ds` command on this branch, but it fails with the following error:

```
Response From Service
TypeError [ERR_HTTP_INVALID_HEADER_VALUE]: Invalid value "undefined" for header "undefined"
```

packages/zosfiles/__tests__/__unit__/methods/copy/Copy.unit.test.ts
```diff
@@ -84,7 +84,7 @@ describe("Copy", () => {
                     expect(copyExpectStringSpy).toHaveBeenLastCalledWith(
                         dummySession,
                         expectedEndpoint,
-                        expectedHeaders,
+                        expect.arrayContaining(expectedHeaders),
```
t1m0thyj, line 87: We want to make sure that no extra headers are being defined, so I don't think we should use `expect.arrayContaining` for headers in our unit tests.

If we need a way to ignore the order of the headers, we could create a Jest helper function that expects the exact same set of items but allows any order.

t1m0thyj (changes requested):

packages/core/src/rest/ZosmfHeaders.ts
```diff
@@ -115,7 +122,7 @@ export class ZosmfHeaders {
      * @static
      * @memberof ZosmfHeaders
      */
-    public static readonly X_CSRF_ZOSMF_HEADER: object = { "X-CSRF-ZOSMF-HEADER": true }; // "the value does not matter"
+    public static readonly X_CSRF_ZOSMF_HEADER: object = { "X-CSRF-ZOSMF-HEADER": "true" }; // "the value does not matter"
```
t1m0thyj, line 125: I assume this change was made to match the `IHeaderContent` type which defines all properties as strings 🙂 - should we change the `object` type to `IHeaderContent` to reflect this?

packages/core/src/rest/ZosmfHeaders.ts
```diff
@@ -32,6 +32,13 @@ export class ZosmfHeaders {
     public static readonly X_IBM_INTRDR_LRECL = "X-IBM-Intrdr-Lrecl";
 
 
+    /**
+     * header indicating recursive = true
+     * @static
+     * @memberof ZosmfHeaders
+     */
+    public static readonly X_IBM_RECURSIVE = {"X-IBM-Option": "recursive"};
```
t1m0thyj, line 40: Similar to previous comment - should this object be strongly typed as `IHeaderContent`?

packages/core/src/rest/ZosmfHeaders.ts
```diff
@@ -166,6 +173,8 @@ export class ZosmfHeaders {
      */
     public static readonly TEXT_PLAIN: IHeaderContent = { "Content-Type": "text/plain" };
 
+    public static readonly APPLICATION_JSON: IHeaderContent = { "Content-Type": "application/json" };
```
t1m0thyj, line 176: Do we need to define this option here? It seems like a duplicate of `Headers.APPLICATION_JSON` defined in Imperative.

packages/zosfiles/src/doc/IOptions.ts
```diff
@@ -18,7 +18,7 @@ import { IZosFilesOptions } from "./IZosFilesOptions";
  * @interface IOptions
  */
 export interface IOptions extends IZosFilesOptions {
-    /**
+    /*
```
t1m0thyj, line 21: This change looks like it may have been accidental since JsDoc typically starts with `/**` 🙂

packages/zosfiles/__tests__/__unit__/methods/copy/Copy.unit.test.ts
t1m0thyj: Instead of `expect.arrayContaining`, we may able to use the `expect. toIncludeSameMembers` helper function provided by the [jest-extended](https://www.npmjs.com/package/jest-extended) package.

zFernand0 (commented):
Changes overall LGTM! 😋 

I do agree with most of the comments made by others, so I'll wait for those to be addressed before approving 🙏

packages/zosfiles/__tests__/__unit__/methods/copy/Copy.unit.test.ts
ATorrise: ive implemented the jest-extended package but i think i still need to keep the expect.arrayContaining because the first expect statment that includes arrayContaining is about making sure the spy was called with the right arguments. the next expect is more granular about the headers.

zFernand0:
I'm trying to keep track of the test in the hopes that it would give us some perspective.

Nested `if` approach:
[80bef2d](https://github.com/zowe/zowe-cli/pull/2436/commits/80bef2de66d7dd46c15504c17bdddb586b1d8bdc)
image

Cleaner `switch` approach:
[c9e76a5](https://github.com/zowe/zowe-cli/pull/2436/commits/c9e76a51ac42a529379a32310d9d8af449eafdd9)
image

---

Personally, `switch` > `if`  for complex stuff like this 😋

ATorrise:
momentarily moving this to draft status for a few hours to make a significant change based on sonar cloud's good suggestion

packages/zosfiles/src/utils/ZosFilesHeaders.ts
```diff
@@ -0,0 +1,268 @@
...
+    // ============================//
+
+    /**
+     * Initializes the header map with predefined header generation functions.
+     * To extend header generation, add new keys and functions here.
+     */
+    private static headerMap = new Map<string, <T>(options: T, context?: ZosFilesContext) => IHeaderContent | IHeaderContent[]>();
+    static initializeHeaderMap() {
+        this.headerMap.set("binary", (options) => (options as any).binary === true ? ZosmfHeaders.X_IBM_BINARY : undefined);
+        this.headerMap.set("record", (options) => (options as any).binary !== true ? ZosmfHeaders.X_IBM_RECORD : undefined);
```
awharn, line 44: The system tests are failing because many data sets are being uploaded with a record header. This line seems to be setting the record header if the data set is not being transferred in binary mode. I believe that instead it should only be setting the record header if the record option is true.

awharn (changes requested):

ATorrise:
closing this PR and opening a new PR without the enhancement that only provides a fix for the RTO header

PR #2443, opened 2025-02-22
Add regex support to Search.dataSets

### LINKED ISSUE #2432: Add regex support to Search.dataSets

Searching is limited to only searching for a simple string.

I would like to be able to use regex when searching datasets and members. The functionality exists in the z/OSMF API, however it is not implemented for datasets yet.

I have looked at the zowe-cli code and how I would implement this. There are two options:
1. Copy what `searchJobs` has, which means to make `searchString` into an optional parameter, then add a new optional parameter alongside it called `searchRegex`
2. Keep `searchString` a required option, but add an optional boolean called `isRegex` or similar, which when true would call the API with the `research` query instead of the `search` one.

I like option 1 more because it would be a standard implementation between jobs and datasets. However, unfortunately this would cause a breaking change (as `searchString` would have to be made optional), which I would prefer to avoid (to allow this feature to come sooner), so therefore option 2 would have to do. But please let me know what your thoughts on that are.

This feature enhancement would also enable this enhancement on Zowe Explorer (which is ultimately what I want):

https://github.com/zowe/zowe-explorer-vscode/issues/3234

JWaters02:
**What It Does**

Closes #2432 

Implements the regex option on the data sets search:
- On mainframe search, it queries on `?research`
- On local search, it just uses normal regex matchAll

In #2432 I pondered about implementation of this option without causing a breaking change:

> There are two options:
> 1. Copy what `searchJobs` has, which means to make `searchString` into an optional parameter, then add a new optional parameter alongside it called `searchRegex`
> 2. Keep `searchString` a required option, but add an optional boolean called `isRegex` or similar, which when true would call the API with the `research` query instead of the `search` one.
> 
> I like option 1 more because it would be a standard implementation between jobs and datasets. However, unfortunately this would cause a breaking change (as `searchString` would have to be made optional), which I would prefer to avoid (to allow this feature to come sooner), so therefore option 2 would have to do. But please let me know what your thoughts on that are.

No one let me know what their thoughts on that were, but I think it is pretty straight forward that option 2 is the right way to go. So what I implemented was a new, optional `--regex` option (please let me know if the name should be changed) that when specified sets the searchString into a regex one. 

Obviously, one thing that needs to be taken into account, is that cases is not sensitive by default, which if combined with a regex string, may produce unintended effects - but that is up to the user (and our the docs to point it out). For example, this regular expression specifies only upper case characters, but since the `--case-sensitive` option is not passed, it will match on the lower case characters anyway:

image

I had to refactor the logic in `Search.dataSets` and `Search.searchLocal` functions to take into account that the matched string length, for the text highlighting in output, would not be the same as the search string length for a regex search string. 

Here are some random examples of using it:

image

**How to Test**

Run `zowe files search data-sets <pattern> <search string> --regex` and mess around with it!

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


JWaters02:
The build CI is failing because the snapshot has changed, how do I commit that?

zFernand0:
> The build CI is failing because the snapshot has changed, how do I commit that?

You could run `npx jest packages/cli/__tests__/zosfiles/__integration__/search/ds/cli.files.search.ds.integration.test.ts -u` 😋 

Thenk just push the corresponding snapshots which should be properly updated 😋

zFernand0 (commented):
LGTM! 😋 
---

Thanks for implementing this feature. 🙏 
Happy to approve once the build errors are resolved (snapshots updated) 🥳

packages/zosfiles/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to the Zowe z/OS files SDK package will be documented in thi
 
 ## Recent Changes
 - BugFix: Users were not warned when copying partitioned data sets with identical member names. Now, the user is prompted to confirm before continuing the copy operation to avoid potential data loss. [#2349] (https://github.com/zowe/zowe-cli/issues/2349)
+- Enhancement: Add the ability to search data sets with regex patterns by passing `--regex` into the search command. [#2432](https://github.com/zowe/zowe-cli/issues/2432)
```
t1m0thyj, line 7: The changelog check is failing because we need to have release notes in both packages that were changed - `packages/cli/CHANGELOG.md` and this one.

I would suggest copying this release note to the CLI changelog, then rewording it here in the SDK changelog to mention the `regex` interface option instead of the `--regex` CLI arg 😋

packages/zosfiles/src/methods/search/Search.ts
```diff
@@ -24,7 +24,8 @@ import { ISearchResponse } from "./doc/ISearchResponse";
 // This interface isn't used outside of the private functions, so just keeping it here.
 interface IInternalSearchResponse {
     responses: ISearchItem[],
-    failures: string[]
+    failures: string[],
+    matchLengths?: number[]
```
t1m0thyj, line 28: Would it make sense to add `matchLength` as an optional property to the `ISearchMatchLocation` interface? If we store it on each individual search response, then this array wouldn't be necessary 🙂

t1m0thyj (changes requested):
Thanks @JWaters02 for your contribution! Left a few comments

I agree the 2nd option you picked is the best one, at least for now in V3 to avoid breaking changes

packages/zosfiles/src/methods/search/Search.ts
JWaters02: This is why I love reviews~~ I was trying to do this when I first worked on this PR, but for some reason couldn't get it quite right - but now you've suggested exactly what needed to be done, it all makes sense! Thanks 🥰

packages/zosfiles/CHANGELOG.md
JWaters02: Thanks for clarifying!

JWaters02:
Tests failed for the same snapshot reason, but I comitted the updated snapshot already, so not sure what the issue is now.

awharn:
> Tests failed for the same snapshot reason, but I comitted the updated snapshot already, so not sure what the issue is now.

A recent change was made on the `master` branch that changed how line wrapping works in the CLI help. It appears the snapshots were based on the old line wrapping, and committed just after the new wrapping was merged into this branch.

I believe that another snapshot update is required, with the newest version of the CLI built and installed.

JWaters02:
> Additionally, if feasible, it would be nice to have a system tests to confirm the functionality is working.

Hi Andrew, I tried to set up the default_properties.yaml with as much detail as I could provide for the zxplore system, but I think I am simply missing a lot of permissions on that system to be able to run the tests, as most of them are failing due to errors on the zosmf server. I also wouldn't want to burden the zxplore server so probably best for me to leave it.

I can't test on my work system, because I am only able to do development on my home PC, which does not have the VPN that can access the work mainframe. 

I agree that system tests would be good, but I am just not able to do it, so best if someone else did.

packages/zosfiles/__tests__/__unit__/methods/search/Search.unit.test.ts
zFernand0, line 1: looks like this file was reformatted 😋 

LGTM though 😅

zFernand0 (approved):
LGTM! 😋 
---

Not sure how to feel about the system test 😅 
But, I believe we can get this one for now and circle back later. 🙏

traeok (approved):
Tested and changes LGTM, thanks @JWaters02!

PR #2467, opened 2025-03-24
adding rto to missing places in files sdk

ATorrise:
**What It Does**
Ensures responseTimeout is respected if it is sent in as an option either through the cli or via extenders calling the files sdk

**How to Test**
set a very short timeout (minimum 5) when accessing any methods in Utilities through a command (Download/Upload) or direct SDK call
image

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


packages/zosfiles/CHANGELOG.md
```diff
@@ -2,15 +2,20 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Adding the `responseTimeout` option to missing areas in the ZosFiles SDK so that the option is allowable on all methods.
```
anaxceron, line 7: Please add an issue or pr link to this

anaxceron (commented):
Left a comment re: the changelog

ATorrise:
image
When the response does time out.. the original implementation sends no error message or warning indicating what has happened. Should this be changed? @awharn @zFernand0 @t1m0thyj

t1m0thyj:
> image When the response does time out.. the original implementation sends no error message or warning indicating what has happened. Should this be changed? @awharn @zFernand0 @t1m0thyj

As long as we're passing the header on all z/OS Files requests we should be good - it's outside of our control if z/OSMF doesn't throw a timeout error on the server side. It appears that for large uploads which take longer than `rto` seconds, it doesn't time out, but I was able to reproduce a timeout error by listing a huge number of data sets:
```
$ zowe files ls ds "S*"

Response From Service
rc:       4
reason:   1
category: 2
message:  receiveResponseHeader: timeout receiving response (>5 secs): TsoServerConnection
```

gejohnston:
Regarding: When the response does time out.. the original implementation sends no error message

Can the rto option be placed in a profile within zowe.config.json? If so, couldn't a command and response like the following occur and be very unclear to the user?

image

I am not sure what we could do about it. Maybe if there is no error and 0 successes, we display a note that a timeout value of xxx was used on this command?

ATorrise:
after team discussion, it's possible that there's a mistake in checking return code for rto on that command.. going to look into this and see how a more reasonable response can be returned in these instances

> Regarding: When the response does time out.. the original implementation sends no error message
> 
> Can the rto option be placed in a profile within zowe.config.json? If so, couldn't a command and response like the following occur and be very unclear to the user?
> 
> image
> 
> I am not sure what we could do about it. Maybe if there is no error and 0 successes, we display a note that a timeout value of xxx was used on this command?

t1m0thyj:
> Regarding: When the response does time out.. the original implementation sends no error message
> 
> Can the rto option be placed in a profile within zowe.config.json? If so, couldn't a command and response like the following occur and be very unclear to the user?
> 
> image
> 
> I am not sure what we could do about it. Maybe if there is no error and 0 successes, we display a note that a timeout value of xxx was used on this command?

@gejohnston @ATorrise I believe this command response is expected and not related to the `--rto` option. All the first-level children of the `packages` folder are directories, so there are no files to upload. The `zos-files upload dir-to-pds` doesn't search for files recursively because this wouldn't make sense in the context of uploading files to a PDS.

gejohnston (approved):

ATorrise:
> @gejohnston @ATorrise I believe this command response is expected and not related to the `--rto` option.....
Thank you for catching this

anaxceron (approved):

packages/zosfiles/src/methods/utilities/Utilities.ts
```diff
@@ -78,8 +86,12 @@ export class Utilities {
         if (codeset) {
             payload.codeset = codeset;
         }
-        await Utilities.putUSSPayload(session, ussFileName, payload);
 
+        if (responseTimeout != null && responseTimeout >= 5) {
+            await Utilities.putUSSPayload(session, ussFileName, payload, responseTimeout);
+        } else {
+            await Utilities.putUSSPayload(session, ussFileName, payload)
+        }
```
zFernand0, line 94: I don't think there would be any harm in just calling the function since you handled the responseTimeout in the putUSSPlayload method 🙏 

```suggestion
        await Utilities.putUSSPayload(session, ussFileName, payload, responseTimeout);
```

packages/zosfiles/src/methods/utilities/Utilities.ts
```diff
@@ -102,9 +114,14 @@ export class Utilities {
...
-    public static async isFileTagBinOrAscii(session: AbstractSession, USSFileName: string): Promise<boolean> {
+    public static async isFileTagBinOrAscii(session: AbstractSession, USSFileName: string, responseTimeout?: number): Promise<boolean> {
         const payload = {request:"chtag", action:"list"};
-        const response = await Utilities.putUSSPayload(session, USSFileName, payload);
+        let response;
+        if (responseTimeout != null && responseTimeout >= 5) {
+            response = await Utilities.putUSSPayload(session, USSFileName, payload, responseTimeout);
+        } else {
+            response = await Utilities.putUSSPayload(session, USSFileName, payload)
+        }
```
zFernand0, line 124: Same as previous comment 😋

packages/zosfiles/src/methods/utilities/Utilities.ts
```diff
@@ -122,9 +139,14 @@ export class Utilities {
...
-    public static async applyTaggedEncoding(session: AbstractSession, USSFileName: string, options: IOptions): Promise<void> {
+    public static async applyTaggedEncoding(session: AbstractSession, USSFileName: string, options: IOptions, responseTimeout?: number): Promise<void> {
         const payload = { request: "chtag", action: "list" };
-        const response = await Utilities.putUSSPayload(session, USSFileName, payload);
+        let response;
+        if (responseTimeout != null && responseTimeout >= 5) {
+            response = await Utilities.putUSSPayload(session, USSFileName, payload, responseTimeout);
+        } else {
+            response = await Utilities.putUSSPayload(session, USSFileName, payload)
+        }
```
zFernand0, line 149: Same as previous comment 😋

packages/zosfiles/src/methods/utilities/Utilities.ts
```diff
@@ -148,11 +170,16 @@ export class Utilities {
...
         ImperativeExpect.toNotBeNullOrUndefined(newFilePath, ZosFilesMessages.missingUSSFileName.message);
         const oldFilePath = USSFilePath.charAt(0) === "/" ? USSFilePath : "/" + USSFilePath;
         const payload = { request: "move", from: path.posix.normalize(oldFilePath) };
-        const response = await Utilities.putUSSPayload(session, newFilePath, payload);
+        let response;
+        if (responseTimeout != null && responseTimeout >= 5) {
+            response = await Utilities.putUSSPayload(session, newFilePath, payload, responseTimeout);
+        } else {
+            response = await Utilities.putUSSPayload(session, newFilePath, payload)
+        }
```
zFernand0, line 182: Same as previous comment 😋

zFernand0 (approved):
LGTM! 😋 
---

Left a small suggestion, but the code works as-is 🥳

zFernand0:
Hey @ATorrise,
Sorry for noticing this so late, but it would seem that there were some lint warnings on the PR 😢 
https://github.com/zowe/zowe-cli/actions/runs/14179963487/job/39723564041
Do you think it'd be possible to resolve these in a follow-up PR? 🙏

PR #2468, opened 2025-03-24
Added authentication type to command response for --show-inputs-only

pujal0909:
**What It Does**

Added a line to `--show-inputs-only` output that displays the authentication used for the command.

**How to Test**

Use a command with `--show-inputs-only` to view the authentication type.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/src/cmd/src/CommandProcessor.ts
+++ b/packages/imperative/src/cmd/src/CommandProcessor.ts
@@ -44,6 +44,8 @@ import { IDaemonContext } from "../../imperative/src/doc/IDaemonContext";
 import { IHandlerResponseApi } from "./doc/response/api/handler/IHandlerResponseApi";
 import { Censor } from "../../censor/src/Censor";
 import { EnvironmentalVariableSettings } from "../../imperative/src/env/EnvironmentalVariableSettings";
+import { AbstractSession } from "../../rest/src/session/AbstractSession";
+import { ConnectionPropsForSessCfg, ISession } from "../../rest";
 
 
 /**
@@ -78,7 +80,7 @@ interface IResolvedArgsResponse {
      * @type {string[]}
      * @memberof IResolvedArgsResponse
      */
-    locations?: string[]
+    locations?: string[];
 }
 
 /**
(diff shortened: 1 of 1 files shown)


anaxceron (approved):

packages/imperative/src/cmd/src/CommandProcessor.ts
```diff
@@ -782,6 +782,10 @@ export class CommandProcessor {
         const secureInputs: Set<string> = new Set([...configSecureProps]);
         let censored = false;
 
+        const sessCfg: ISession = {};
+        ConnectionPropsForSessCfg.resolveSessCfgProps(sessCfg, commandParameters.arguments);
+        showInputsOnly.commandValues["authentication type"] = sessCfg.type;
```
t1m0thyj, line 787: The `commandValues` section is for listing CLI arguments, so I don't know if it makes sense to add this there.

Could we add `authenticationType` as a new property at the top level (as sibling of `commandValues`)?

packages/imperative/src/cmd/src/CommandProcessor.ts
t1m0thyj, line 1: Please add a unit test for the new functionality to `CommandProcessor.unit.test.ts` 😋

t1m0thyj (commented):
Looks pretty good, thanks @pujal0909!

gejohnston:
Can you paste sample output into a comment on this PR conversation? I would like to see how this new output might fit with output added in the future for authentication order.

pujal0909:
Screenshot 2025-03-26 at 11 02 57 AM

gejohnston (approved):
Thanks for the extra information. Looks good to me.

traeok (approved):
Tested and LGTM, thanks Pujal!

t1m0thyj (approved):
LGTM, thanks @pujal0909!

zFernand0 (changes requested):
Changes LGTM! 😋 
---

I could be doing something wrong, however...

I do have one slight problem... for some reason it's identifying this as `basic` instead of `token` auth

image

Here is what my QA.apiml profile looks like
image

zFernand0 (approved):
LGTM! 😋 
---

Please disregard my previous comment, it was due to not having a token stored on the vault. 🙏 

Should we still consider that as token auth?

zFernand0:
Merging and labeling `release-current`. That way we can combine this with other PRs before the end of the week 🙏

PR #2482, opened 2025-04-10
fix(secrets): Resolve error accessing secrets on MacOS

t1m0thyj:
**What It Does**

Fixes an error where if the parent process path is long, then the Secrets SDK attempts to create a socket file that exceeds the max path length for IPC sockets (103 chars on MacOS).

The Adler32 algorithm is not immune to hash collision (as pointed out by Copilot). However I think it's a good balance between speed and uniqueness for our use case. The worst that can happen in the rare case of a hash collision is that users see multiple keyring unlock prompts instead of just one.

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/secrets/CHANGELOG.md
+++ b/packages/secrets/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe Secrets SDK package will be documented in this file.
 
+## Recent Changes
+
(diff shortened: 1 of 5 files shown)


traeok (approved):
While I can't test this at the moment (not on a Mac), these changes make sense to me. Thanks Timothy for the fix!

gejohnston (approved):

awharn (approved):

anaxceron (approved):

zFernand0 (approved):
LGTM! 😋

PR #2486, opened 2025-04-15
Make changes to support connection timeouts and hostname validation

awharn:
**What It Does**

Adds a connection timeout to the Abstract Rest Client.
Adds hostname validation to REST requests.
Fixes linter errors.
Makes imports more granular to prevent circular imports.

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added a connection timeout to the Imperative REST Client, with a default of 60 seconds. []()
+- Enhancement: Added the `SOCKET_CONNECT_TIMEOUT` environment variable to the `EnvironmentalVariableSettings` class to allow extenders to determine how long to wait for a socket connection before timing out. []()
+- BugFix: Added checks to `AbstractRestClient` to ensure the `hostname` parameter does not contain a protocol. []()
+
 ## `8.18.0`
 
 - BugFix: Ensured that the `ProfileCredentials` class evaluates all layers to determine if the credentials are secure. [#2460](https://github.com/zowe/zowe-cli/issues/2460)
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -4,9 +4,9 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
-- Enhancement: Added a connection timeout to the Imperative REST Client, with a default of 60 seconds. []()
-- Enhancement: Added the `SOCKET_CONNECT_TIMEOUT` environment variable to the `EnvironmentalVariableSettings` class to allow extenders to determine how long to wait for a socket connection before timing out. []()
-- BugFix: Added checks to `AbstractRestClient` to ensure the `hostname` parameter does not contain a protocol. []()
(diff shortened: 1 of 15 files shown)


ATorrise (approved):

packages/zosfiles/src/methods/utilities/Utilities.ts
```diff
@@ -20,6 +20,8 @@ import * as path from "path";
 import { ZosFilesUtils } from "../../utils/ZosFilesUtils";
 
 export class Utilities {
+    private static readonly minimumTimeout = 5;
```
ATorrise, line 23: should this timeout be added as a constant? in Constants.ts? only because a min timeout of 5 is used a lot of places

gejohnston (approved):

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added a connection timeout to the Imperative REST Client, with a default of 60 seconds. [#2486](https://github.com/zowe/zowe-cli/pull/2486)
+- Enhancement: Added the `SOCKET_CONNECT_TIMEOUT` environment variable to the `EnvironmentalVariableSettings` class to allow extenders to determine how long to wait for a socket connection before timing out. [#2486](https://github.com/zowe/zowe-cli/pull/2486)
```
t1m0thyj, line 8: Should we say `ZOWE_SOCKET_CONNECT_TIMEOUT` here? I know the prefix is implied, but users may not be aware of this.

t1m0thyj (approved):

packages/imperative/CHANGELOG.md
awharn: I will update it. I was under the impression that Imperative consumers are more extenders, but I agree that is more clear for CLI and VSCE users.

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
...
+- BugFix: Added checks to `AbstractRestClient` to ensure the `hostname` parameter does not contain a protocol. [#2486](https://github.com/zowe/zowe-cli/pull/2486)
```
anaxceron, line 9: We should describe what `AbstractRestClient` is: " Added checks to **the** `AbstractRestClient` **class/property/whatever** to

anaxceron (commented):
Left a comment re: the changelog

packages/zosfiles/src/methods/utilities/Utilities.ts
awharn: I only did this here because of the several uses in this one file. The only other locations where I could find `5` being used was in the `zosfiles` strings file, and as a parameter for responseTimeout in the `zosmf` package. The only places where the change could be placed to share that value across those three files is either `zosmf` or `core`, and a new constants object would need to be created to export it. Seeing as the property here is readonly, I plan to leave the implementation as-is until we find more constants that we would like to expose at a lower level.

anaxceron (approved):
Good changelog, thanks @awharn!

zFernand0 (approved):
LGTM! 😋 
---

Pardon the late review 😅 
Thanks for adding this as an ENV 🥳

zFernand0:
tagging release current to combine with the 2 other open PRs 🙏

Issue #1650, opened 2023-03-08
`zowe auth login apiml` creates additional base profiles


zFernand0:
_Originally posted by @plavjanik in https://github.com/zowe/zowe-cli/issues/1341_

When I want to get a new token using `zowe auth login apiml`, it does not work well for me:

```
$ zowe auth login apiml                                                                                                                                                                                                                                                                                         
Enter the user name for your service (will be hidden): 
Enter the password for your service (will be hidden): 
Do you want to store the host, port, and token on disk for use with future commands? If you answer Yes, the credentials will be saved to a base profile named 'base_apiml_apiml'. If you answer No, the token will be printed to the terminal and will not be stored on disk. [y/N]:

Login successful. The authentication token is stored in the 'base_apiml_apiml' base profile for future use. To revoke this token and remove it from your profile, review the 'zowe auth logout' command.
```

This creates a new base profile:
```
        "base_apiml": {
            "type": "base",
            "properties": {
                "host": "sys32.lvn.broadcom.net",
                "port": 60004,
                "tokenType": "apimlAuthenticationToken"
            },
            "secure": [
                "tokenValue"
            ]
        },
        "base_apiml_apiml": {
            "type": "base",
            "properties": {
                "host": "sys32.lvn.broadcom.net",
                "port": 60004,
                "tokenType": "apimlAuthenticationToken"
            },
            "secure": [
                "tokenValue"
            ]
        }
    },
    "defaults": {
        "zosmf": "zosmf",
        "tso": "tso",
        "ssh": "ssh",
        "base": "base_apiml_apiml"
    },
```

There are two base profiles, I just wanted to refresh my token for the existing profile. How do I do that? The command fails too:
```
+ zowe zos-files create data-set-classic USERID.TESTBASH.DATASET1
Command Error:
This operation requires authentication.

z/OSMF REST API Error:
Rest API failure with HTTP(S) status 401

Host:      sys32.lvn.broadcom.net
Port:      1443
Base Path: 
Resource:  /zosmf/restfiles/ds/USERID.TESTBASH.DATASET1
Request:   POST
Headers:   [{"Accept-Encoding":"gzip"},{"X-CSRF-ZOSMF-HEADER":true}]
Payload:   POST

Error Details:
Token is not valid or expired.

For CLI usage, see `zowe auth login apiml --help`
```

zFernand0:
Hey @plavjanik,
[...]

PR #2491, opened 2025-04-24
Update Zowe Client REST APIs to obey the choice of authentication specified by a user

### LINKED ISSUE #2181: zowe auth login apiml creates 2nd base profile
Steps:
0. run `zowe config init` and respond to prompts
0. run `zowe config auto-init` and respond to prompts
0. verify `zowe jobs list jobs` functions correctly after
0. run `zowe auth login apiml`

This gives a response:

```
~/dev/systems/apiml> zowe auth login apiml
Do you want to store the host, port, and token on disk for use with future commands? If you answer Yes, the credentials will be saved to a base profile named 'base_apiml'. If you answer No, the token will be printed to the terminal and will not be stored on disk. [y/N]: y

Login successful. The authentication token is stored in the 'base_apiml' base profile for future use. To revoke this token and remove it from your profile, review the 'zowe auth logout' command.
```

My original base profile was:

```
        "base": {
            "type": "base",
            "properties": {
                "host": "mainframe.net",
                "rejectUnauthorized": true,
                "port": 60154,
                "tokenType": "apimlAuthenticationToken"
            },
            "secure": [
                "user",
                "password",
                "tokenValue"
            ]
        },
```

This `auth` command added:

```
        "base_apiml": {
            "type": "base",
            "properties": {
                "host": "mainframe.net",
                "port": 60154,
                "tokenType": "apimlAuthenticationToken"
            },
            "secure": [
                "tokenValue"
            ]
        }
```

Shouldn't login have instead performed a login to my existing base profile?

gejohnston:
**What It Does**

A user of Zowe CLI can specify which type of credential should be used to connect to the mainframe from any profile.
Background, use cases, and the approach for implementation are described in the following design document.

[Order of authentication in Zowe Clients](https://github.com/zowe/zowe-cli/blob/auth-order/docs/Design_for_selecting_auth_type.md)

This document will be available until this PR is merged. Afterwards, the document will be available from the master branch of the Zowe CLI repository.

This PR closes #2181 #1650

**How to Test**

- Place every possible type of credential into your base profile (they can be valid or bogus credentials).
- Place an authOrder property into your base profile or into the profile to which you will connect.
    - The property can contain one or multiple choices of authentication type.
    - The property syntax is "authOrder": "basic, token, bearer, cert-pem, none"
- Run a command and observe that the first authentication is used.
    - Switch the order and observe that the new first authentication is used.
- If the underlying REST request fails, the available credentials, the specified authentication order, and the type of authentication used for the current request are displayed.
- If you specify the `--show-inputs-only` option on your command, you will see a similar set of information displayed without actually making the request.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


packages/cli/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Change Log
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Updated the `zowe config auto-init` command to obey the choice of authentication specified by a user.
```
anaxceron, line 6: We should add an issue/pr link.

packages/cli/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
...
+- BugFix: Updated the `zowe auth login apiml` command to place its retrieved token into the current base profile instead of creating a new base profile. [#2181](https://github.com/zowe/zowe-cli/issues/2181) & [#1650](https://github.com/zowe/zowe-cli/issues/1650)
+  - The update also places a directive to use token authentication for the default zosmf profile if that profile is configured to connect to API-ML.
```
anaxceron, line 8: I'd put zosm in backticks

packages/core/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe core SDK package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Updated the `Login.apimlLogin` API function to record that it is making a request for a token into the session associated with its REST request.
```
anaxceron, line 7: We should include a pr/issue link here

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Updated the Zowe Client REST APIs to obey the choice of authentication specified by a user.
```
anaxceron, line 7: We should include a pr/issue link here

anaxceron (commented):
Left some comments re: changelogs

anaxceron (approved):
Excellent changelogs!

traeok (commented):
Thanks Gene, this looks great and is working well from my testing thus far! I noticed that the `authOrder` property is not in the properties of the schema for a base profile (or z/OSMF profile). I think it would help to add a description for the property to bring awareness to how it can be configured.

But this had me wondering... if we still need to add `authOrder` to the schema, what are your thoughts on making the `authOrder` type an array in the configuration file? I think this would help enforce proper configuration of this property, as we can use schema validation to guide users when adding to the list or changing the order.

If we add `authOrder` to the schema with this configuration:
```json
"authOrder": {
  "type": "array",
  "items": {
    "type": "string",
    "enum": [
      "token",
      "basic",
      "bearer",
      "cert-pem",
      "none"
    ]
  },
  "uniqueItems": true
},
```

Then it warns the user whenever duplicate items are present in the list:
image

And it also helps spot unsupported methods while providing the list of valid options:
image

gejohnston:
Regarding authOrder in the schama and being an array:

The following items were stated in the design document.
____________________

Such a new item within the properties object will not currently be included in the zowe.schema.json file. An IntelliSense-like editor will not display **authOrder** as an option when a user starts to add a new property in a profile.

- We do not want to require plugins/extensions to add **authOrder** to their profiles. Asking extenders to add AuthOrder to their set of options (and thus the schema) in the middle of a release's lifecycle can be a burden for those contributors.
      
- As a result, we feel that losing IntelliSense is an acceptable compromise to avoid adding work to every extender. Users will lose a nice convenience, but the user's config will not be limited in any way at runtime.

____________________

I believe that the CLI only updates the schema on initial install, so it raises the question of how that schema update would be applied. We can revisit this position if we misconstrued the implications, and we want to further extend the time that we spend on this story to resolve schema related operations.

Regarding authOrder being an array: Orginally it was an array. We felt that without schema assistance it would be less error-prone for users specify a string like: "basic, token, cert-pem", than to enter an array like: ["basic", "token", "cert-pem"]. Further, if we ever did add authOrder to the schema, users could specify authOrder on the command line. Array syntax on the command line becomes even more difficult for users to enter correctly. So we changed it to a string. Again, if the team feels that customers are better served with an array, that aspect could be reimplemented.

I encourage other team members to offer opinions on these decisions.

traeok:
> Regarding authOrder being an array: Orginally it was an array. We felt that without schema assistance it would be less error-prone for users specify a string like: "basic, token, cert-pem", than to enter an array like: ["basic", "token", "cert-pem"]. Further, if we ever did add authOrder to the schema, users could specify authOrder on the command line. Array syntax on the command line becomes even more difficult for users to enter correctly. So we changed it to a string. Again, if the team feels that customers are better served with an array, that aspect could be reimplemented.
> 
> I encourage other team members to offer opinions on these decisions.

Sorry for the confusion Gene - I was looking at the design doc on master while testing and did not see the format was changed to a string on this branch. I do recall reading about why its not in the schema as part of the design doc, and I don't want to assume that all folks are relying on the schema to edit the config. It was more of an afterthought based on setting it up in my environment. I can see how it would be easier to just specify it as a string on the command line, so I don't want to suggest any last-minute changes just to appeal to folks that use the schema.

That said, I think its reasonable to just have the options documented somewhere (as they have been) if we don't want it in the schema. As you mentioned, it is easier to introduce errors defining an array when the schema is not providing suggestions to the user.

---

Not related, but one new behavior I noticed is that when `authOrder` is defined in a top-level layer of my config as `token,cert-pem`, I am prompted to enter a hostname when using `zowe auth apiml login`:

image

For reference, this is the current structure I have in place:
```jsonc
{
    "profiles": {
        "z": {
            "type": "base",
            "profiles": {
                "####-apiml": {
                    "type": "base",
                    // this is a group with different profiles (such as zosmf)
                    "profiles": { ... }
                }
            },
            "secure": ["user"],
            "properties": {
                "authOrder": "token,cert-pem"
            }
        }
    }
}
```

I'm noticing that Zowe CLI on `master` has an issue where it prompts me for a password with this structure - but this may largely be because I have `user` in the secure array of my top-level group (`z`), but no password saved. I would still expect the password prompt as that bug seems separate from this PR, but the hostname prompt feels like unexpected behavior.

If we don't intend to support an `authOrder` where `basic` and `bearer` are omitted, then I can retest with a valid value. I saw there was a note on what happens when token isn't specified in the design doc, but wasn't sure how it is intended to behave if those two are missing from the auth order.

gejohnston:
Regarding `auth login` behavior:

This is exactly the kind of thing that we want to find.

Our intention is for any one or more combinations of authentication types in the authOrder to be valid.

None of the changes made to `auth login` were intended to change what it prompts for. During my testing, I have not seen a difference in prompting behavior between the master and authOrder branches. You may have created a config condition that I have not. I will attempt to reproduce your error.

packages/imperative/src/rest/src/session/AuthOrder.ts
```diff
@@ -0,0 +1,621 @@
...
+
+        sessCfg._authCache.didUserSetAuthOrder = false;
+        if (cmdArgs.authOrder) {
+            if (typeof cmdArgs.authOrder === "string") {
+                if (cmdArgs.authOrder.length > 0) {
+                    // user supplied an authOrder
+                    // convert user's comma-separated string into an array of auth types, and remove whitespace
+                    const userAuthOrder = cmdArgs.authOrder.split(',');
+                    for (let nextUserAuth of userAuthOrder) {
+                        nextUserAuth = nextUserAuth.trim();
```
zFernand0, line 384: This is one thing that Sonar cloud pointed out.

It recommends not reassigning loop-counters, which makes sense for for-loops in principle. However, this is a `for-of` and unless we are adding more stuff to `userAuthOrder` inside this loop, I don't believe there will be an issue here.

To address this, we could in theory trim the variable before it gets assigned, which is basically remapping the entire array. 
```
for (let nextUserAuth of userAuthOrder.map((v) => v.trim())) {
```

---

Again, I don't think that's necessary, thus I'm not requesting this to be changed : )

packages/imperative/src/censor/src/Censor.ts
```diff
@@ -333,4 +333,63 @@ export class Censor {
...
+        const propsToBeCensored = [...Censor.SECURE_PROMPT_OPTIONS, ...Censor.DEFAULT_CENSORED_OPTIONS];
+
+        // create copy of sessObj so that we can replace values in a censored object
+        let censoredSessObj;
+        if (createCopy) {
+            try {
+                censoredSessObj = JSON.parse(JSON.stringify(sessObj));
+            } catch(_error) {
+                return "Invalid session object";
+            }
```
zFernand0, line 372: SonarCloud points out that we should probably make use of the error here.
And I'm inclined to agree on this one. 
I think that at least we should log something (if we have access to the logger class here)

zFernand0 (commented):
Changes LGTM so far! 😋 
---

This is a preliminary review comment.
Mostly to address the SonarCloud issues found.

I was able to dismiss a false-positive regarding a `fakePassword` in a test file. 

Also, there was an issue with cognitive complexity on, but we tend to accept those. If we find that we spend too much time debugging a particular function/class we can always open a tech-debt issue to try and refactor it 🙏

packages/cli/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Change Log
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Updated the `zowe auth login apiml` command to place its retrieved token into the current base profile instead of creating a new base profile. [#2181](https://github.com/zowe/zowe-cli/issues/2181) & [#1650](https://github.com/zowe/zowe-cli/issues/1650)
+  - The update also places a directive to use token authentication for the default `zosmf` profile if that profile is configured to connect to API-ML.
```
zFernand0, line 7: I know it'd clear to most people but, I'm curious if we can be a bit more explicit here...

> the current base profile

I do believe this means the `default` base profile. Not to be confused with any `type: base` profiles in the nesting hierarchy of a zosmf profile (e.g. myBase.anotherBase.zosmf vs myDefaultBase).

packages/imperative/__tests__/__integration__/imperative/__tests__/__integration__/cli/config/import/cli.imperative-test-cli.config.import.integration.subtest.ts
```diff
@@ -219,7 +219,9 @@ describe("imperative-test-cli config import", () => {
 
             expect(response.status).toEqual(1);
             expect(response.stderr.toString()).toContain("Missing Positional Argument");
-            expect(response.stderr.toString()).toMatchSnapshot();
+            expect(response.stderr.toString()).toContain("Syntax Error:");
+            expect(response.stderr.toString()).toContain("Missing Positional Argument: location");
+            expect(response.stderr.toString()).toContain("Argument Description: File path or URL to import from.");
```
zFernand0, line 224: Thanks for removing snapshots in favor of more explicit strings 🙏

__tests__/__integration__/__snapshots__/imperative.secure.integration.test.ts.snap
```diff
@@ -1,21 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Imperative Secure Tests Imperative Test CLI Secure Tests imperative-test-cli config import failure scenarios should fail to import if location is not specified 1`] = `
-"
```
zFernand0, line 4: Was this always an outdated snapshot? 😕 
If so, thanks for cleaning it up 🙏

packages/cli/__tests__/auth/__system__/cli.auth.login.apiml.system.test.ts
```diff
@@ -144,7 +161,7 @@ describe("auth login/logout apiml create profile", () => {
        https://github.com/zowe/zowe-cli/issues/2005
     */
     // eslint-disable-next-line jest/no-disabled-tests
-    it.skip("TODO: After 2005 is fixed: should successfully issue the login command and create a team config", () => {
+    it.skip("TODO: After 2005 is fixed: should successfully login with password and create a team config", () => {
```
zFernand0, line 164: It seems that 2005 was closed on Jan 31, 2024 🤯 

Is this something we should not skip anymore?

packages/cli/__tests__/config/auto-init/__unit__/ApimlAutoInitHandler.unit.test.ts
```diff
@@ -340,6 +340,78 @@ describe("ApimlAutoInitHandler", () => {
...
+                ISession: iSessObj
+            },
+            {
+                arguments: {
+                    $0: "fake",
+                    _: ["fake"]
+                }
+            }
+        );
+        expect(iSessObj._authCache.authTypeToRequestToken).toBe(undefined);
```
zFernand0, line 412: Nothing to do here... 😅 
Just mind-blown due to the amount of setup just to expect this 🤯 
I guess that's how you got 99% coverage 🥳 
Thank you! 🙏

packages/core/__tests__/auth/__system__/Logout.system.test.ts
```diff
@@ -17,6 +17,9 @@ import { Login } from "../../../src/auth/Login";
 import { Logout } from "../../../src/auth/Logout";
 import { ZosmfRestClient } from "../../../src/rest/ZosmfRestClient";
 
+// import non-exported modules
+const AuthOrder = jest.requireActual("../../../../imperative/lib/rest/src/session/AuthOrder").AuthOrder;
```
zFernand0, line 21: This question is just for my own understanding...
Do we need to explicitly import `AuthOrder` this way? 

I do see it being exported as part of the `packages/imperative/src/rest/index.ts`

packages/core/__tests__/rest/__unit__/ZosmfRestClient.unit.test.ts
```diff
@@ -42,12 +42,19 @@ describe("ZosmfRestClient tests", () => {
...
+            const fakeSess = new Session({
                 hostname: "dummy",
                 type: SessConstants.AUTH_TYPE_BASIC,
                 user: "fakeUser",
                 password: "fakePass"
-            }));
+            });
+
+            // ensure that some available creds are cached
+            (AuthOrder as any).cacheCredsAndAuthOrder(fakeSess.ISession,
```
zFernand0, line 53: Same as previous comment(s) (regarding the need for `as any`)
```suggestion
            AuthOrder.cacheCredsAndAuthOrder(fakeSess.ISession,
```

I won't make this `"Same as previous comment"` in every location, but note that this may apply multiple times in various test files 🙏

packages/core/src/auth/Login.ts
```diff
@@ -33,6 +33,7 @@ export class Login {
         ImperativeExpect.toMatchRegExp(session.ISession.tokenType, "^apimlAuthenticationToken.*",
             `Token type (${session.ISession.tokenType}) for API ML token login must start with 'apimlAuthenticationToken'.`);
 
+        AuthOrder.makingRequestForToken(session.ISession);
```
zFernand0, line 36: Dumb question: (and this might be answered if I just continue reading the code) 
Is this something that extenders/plug-in devs need to do if they handle their own tokens?

As a general comment, could you help me understand how extenders may need to accommodate in order to tell Imperative that the extension/plug-in is trying to perform a token-request.

zFernand0 (commented):
Still halfway through the code. 😅 
Sorry for all the questions 🙏

traeok (approved):
Thanks Gene, this is working great and LGTM. I feel that this implementation especially helps us pivot away from the hardcoded precedence that's in Zowe Explorer, and other extenders / SDK consumers could also benefit from this. I've been using the `auth-order` branch since I last reviewed and haven't run into any issues while running various Zowe CLI commands. I did shift back to my daily config to work around #2495, but I find that my current config works well with this new property.

Fernando did raise a good question about how the `AuthOrder.makingRequestForToken` property could/should behave from an extender's point of view for their own token types. From what I've gathered when reviewing the code, this function doesn't necessarily do anything that's specific to a single token type, but rather prepares the internal session properties (`_authCache`, `authTypeOrder`, and `authTypeToRequestToken`) so that the correct credentials are applied and removed as part of the other `AuthOrder` functions. That said, we can always enhance the function in the future to allow passing custom criteria as part of these functions (if the demand is there of course).

I also agree with Fernando's request to log the error in the catch block (with unused `_error` variable) if possible, but don't have any separate requests so I'm going to approve this :yum:

packages/imperative/src/rest/src/session/AuthOrder.ts
traeok: I'm thinking that we can keep it as-is, especially because we do not use the array after the loop and the array is `string[]` (meaning that modifying each element won't affect any references to an object/class instance/etc)

packages/imperative/src/censor/src/Censor.ts
gejohnston: The are two wrinkles involved in this item. First, Logger depends on Censor. So, we cannot log a message from Cursor without introducing a circular dependency. Second, no matter what garbage I place into the sessObj during a call to JSON.parse(JSON.stringify(sessObj)), I cannot get it to throw an error. The catch block is really just a safety net that I cannot confirm will ever be used. But, I also cannot prove that it never will be called.

To satisfy the SonarCloud complaint, within the catch block, I placed the caught error into the string returned by the replaceValsInSess function. I will push this change soon.

packages/imperative/src/rest/src/session/AuthOrder.ts
gejohnston: Since I had to make another change to respond to other comments, I will just add the map operation.

packages/cli/CHANGELOG.md
gejohnston: It could be the default base profile, but it is not limited to the default base profile. A user could specify the --base-profile option on the login command and thus specify a base profile which is not the default. I used the term "current" in an attempt to generically refer to the base profile currently requested by the user. I am not opposed to changing this wording if someone can identify a simple phrase to encapsulate the possibilities available to the user.

packages/core/src/auth/Login.ts
gejohnston: The function AuthOrder.makingRequestForToken does not augment or reduce the capabilities that already within the Zowe client APIs. There has always been an object named connOpts, which contains a property named requestToken (with a value of true or false).

The connOpts object is passed into several different functions of the ConnectionPropsForSessCfg class. Callers set connOpts.requestToken before calling one of those functions. The connOpts object does not live for the duration from early initialization of the session to the actual REST request. The AuthOrder.makingRequestForToken simply ensures that the desire to make a token request is stored in the session, so that we know the caller's choice later during the process. AuthOrder.makingRequestForToken is called from ConnectionPropsForSessCfg when connOpts.requestToken === true.

The apimlLogin function (that you identified) is called from ApimlAuthHandler and ApimlAutoInitHandler (neither of which call any functions from ConnectionPropsForSessCfg). Because both ApimlAuthHandler and ApimlAutoInitHandler are attempting to get a token, we record that request-for-token within the apimlLogin function by calling the AuthOrder.makingRequestForToken function.

None of this provides extra token gathering capability for other extenders beyond what already existed in the zowe SDK. However, I believe that those extenders have no new limitations as a result of these AuthOrder modifications.

If we want to provide increased token-gathering abilities to extenders in the future, I think it would be best addressed in a new feature or story. We might need to make AuthOrder.makingRequestForToken public at that time to assist in that process.

packages/cli/__tests__/auth/__system__/cli.auth.login.apiml.system.test.ts
gejohnston: This was a good suggestion. Simply re-enabled the test, and it worked.

__tests__/__src__/environment/TestEnvironment.ts
```diff
@@ -103,15 +106,20 @@ export class TestEnvironment extends BaseTestEnvironment {
...
             user: SYSTEM_PROPS.zosmf.user,
             password: SYSTEM_PROPS.zosmf.password,
             hostname: SYSTEM_PROPS.zosmf.host,
             port: SYSTEM_PROPS.zosmf.port,
             type: "basic",
             rejectUnauthorized: SYSTEM_PROPS.zosmf.rejectUnauthorized,
             basePath: SYSTEM_PROPS.zosmf.basePath
-        });
+        };
+        (AuthOrder as any).cacheCredsAndAuthOrder(sessCfg, {});
```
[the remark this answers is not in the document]
gejohnston: I think that some of my recollection of events may be a bit off - this enhancement has been going on for a long time.

As you noticed, AuthOrder is in an index file, and cacheCredsAndAuthOrder is truly public. I think I needed to open things up a bit over the course of the implementation. I believe that I probably implemented some tests before things were opened up. I know that some of the tests could not originally do a normal import of AuthOrder.

As long as a test only needs to access cacheCredsAndAuthOrder, a normal import should work ok. I am revising such tests.

packages/cli/CHANGELOG.md
gejohnston: I changed the wording slightly.

packages/cli/__tests__/auth/__system__/cli.auth.login.apiml.system.test.ts
```diff
@@ -104,23 +102,42 @@ describe("auth login/logout apiml show token", () => {
         expect(response.status).toBe(0);
         const responseData = JSON.parse(response.stdout.toString()).data;
         expect(responseData.tokenType).toEqual("apimlAuthenticationToken");
-        expect(responseData.tokenValue).not.toEqual(token[token.length-3]); // Intermittent failure : (
-        expect(responseData.tokenValue.length).toEqual(token[token.length-3].length);
+        expect(responseData.tokenValue).toBeDefined();
+        expect(responseData.tokenValue.length).toBeGreaterThan(100);
```
t1m0thyj, line 106: If the intent of this test is to validate the token returned from the API ML, should we call `ConfigUtils.hasTokenExpired` to check that we have a non-expired JWT?

zFernand0 (approved):
LGTM! 😋 
---

Thanks for addressing all the questions and comments 🙏

packages/cli/__tests__/zosfiles/__unit__/ZosFilesBase.handler.unit.test.ts
```diff
@@ -23,6 +26,10 @@ describe("ZosFilesBaseHandler", () => {
             _commandParameters: IHandlerParameters,
             _session: AbstractSession
         ): Promise<IZosFilesResponse> {
+            // The authCache and authTypeOrder was added to the session by the super class.
+            // This is the only way that we can get them.
+            this.authCache = _session.ISession._authCache;
+            this.authTypeOrder = _session.ISession.authTypeOrder;
```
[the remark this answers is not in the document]
gejohnston: Seems reasonable.

packages/cli/__tests__/auth/__system__/cli.auth.login.apiml.system.test.ts
gejohnston: I am not sure why the original test implementation tested for length, but I can replace it with hasTokenExpired. That seems to better reflect the intent.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -232,26 +233,34 @@ export abstract class AbstractRestClient {
...
+        // The current best practice for consumers of any of these APIs is to let the
+        // default order maintain backward compatibility. If your app needs a credential
+        // that conflicts with credentials required by other profiles/services, you should
+        // instruct your end users to specify an 'authOrder' property in the profile
+        // related to your app within their zowe.config.json file.
+
+        if (this.constructor.name !== "RestClient" &&
+            this.constructor.name !== "ZosmfRestClient") {
+            // Maintain the historic behavior where other classes which extend from
+            // from AbstractRestClient have a default order with token at the top.
```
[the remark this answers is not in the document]
gejohnston: A new optional parameter with a default value seems like a cleaner, more intuitive solution.

t1m0thyj (changes requested):
Prior comments were addressed, requesting changes until password prompting issue is addressed

jace-roell (approved):

traeok (approved):
Thanks Gene! With the latest changes, the auth order still behaves as expected. Now, when I do a `zowe auth apiml login` with a nested profile (several layers), the token is properly stored on the provided base profile. Before, we saw a misleading prompt that asked to create a new base profile. In addition, I can now list data sets with Zowe CLI with token and username only - I am no longer prompted for a password. This is without introducing any changes to the config.

docs/Design_for_selecting_auth_type.md
```diff
@@ -143,7 +141,27 @@ We should add a new AUTH_TYPE_CHOICE of:
...
+
+While this kept the user/password and token separate, it required users to switch their base profile whenever the user wanted to use a profile that connected directly to a service versus using a profile that connected through API-ML. Alternatively the user could edit their configuration file to move user and password out of the existing base profile and then place a token into that existing base profile. All of these techniques were confusing, error-prone, and inconvenient.
+
+Because the new authOrder property enables a user to select which credentials should be used, every type of credential can co-exist on the same base profile. As a result, the `zowe auth login apiml` command now places the resulting token into the current default base profile (or into the base profile specified on the command line).
+
+The **login** command now also automatically places an **authOrder** property with a value of "token" into the default zosmf profile, so the token can be used immediately after logging into API-ML.
```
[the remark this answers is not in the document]
gejohnston: Made this change to the document.

t1m0thyj (approved):
LGTM, thanks for all your hard work on this @gejohnston!

zFernand0 (approved):
LGTM! 😋

PR #2496, opened 2025-05-07
refactor: use Winston instead of log4js & provide migration tools

traeok:
**What It Does**

- Replaces use of `log4js` with winston in `ConfigUtils.initImpUtils` to avoid clobbering downstream logging configurations in shared Node.js contexts (e.g. VS Code environments)
  - Winston uses a different design pattern by encapsulating individual loggers into Logger instances. This avoids collisions in global state, which is what log4js uses to store its logging configuration.
- Provides a `fromWinstonConfig` factory method on the `Logger` class to allow extenders to create new Logger instances using Winston configs
- Implements migration tools to translate log4js configurations to equivalent winston configs
- Adds a `log` function to the `Console` class to facilitate runtime re-routing of log messages to the correct console level 
- Adds more unit tests for the logger to verify expected behaviors
- Removed Node 18 from the build workflow as its now EOL

The logger still preserves identical behavior for Zowe CLI and when log4js is used (through `Logger.initLogger`) to prevent breaking changes. We can remove log4js support entirely when v4 comes around.

**How to Test**

1. Clone the extension sample at https://github.com/benjamin-t-santos/zowe-explorer-testing/ under the `log4js-issues` branch: `git clone https://github.com/benjamin-t-santos/zowe-explorer-testing/ --branch log4js-issues` 
2. Download [this ZIP](https://github.com/user-attachments/files/20207914/ZE-profileinfo-logger-changes.zip), extract the files inside and place the Zowe Explorer API `tgz` file in the extension sample root folder.
3. Update the `package.json` to point to the local Zowe Explorer API `tgz` built with the Imperative changes from this branch: `"@zowe/zowe-explorer-api": "file:<path-to-zowe-explorer-api-tgz>"`
4. Run `npm install && npm run compile`
5. After its built, install the Zowe Explorer VSIX from the ZIP in step 2. Once installed, restart VS Code if needed.
6. Once complete, click on the Run & Debug panel on the VSC Side Bar, and then click the Play button beside "Run Extension" at the top to start debugging the sample.
7. Open the Debug Console for the dev host of the extension sample. Observe that both debug messages through log4js are printed to the Debug Console. Before, one message made it to the console and the second one was redirected to imperative logger.
8. Now, check the extension logs for Zowe Explorer (in the ZE extension folder: `.vscode/extensions/zowe.vscode-extension-for-zowe-3.3.0-SNAPSHOT/logs/imperative.log`) and notice that no output is present from the sample extension (which is just using the default log4js logger). Also notice that the logs have identical format to pre-existing log4js output.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog


traeok:
Regarding the above CodeQL warnings: I'm inclined to dismiss these as I have not introduced any changes to the mentioned code (rather, these issues resurfaced after making changes in this file). 

Since the `Console` class is designed for logging data, I think its reasonable to allow "log injection" (newlines) as part of the message data.

traeok:
I don't have permission to resolve the SonarCloud issues highlighted on this PR, but can comment on them:

- **Member 'category?: string' is never reassigned; mark it as `readonly`.**
  - This could be considered a breaking change as any developer implementing a subclass of `Logger` could be modifying this property. I think we should keep this as-is until v4 to avoid breaking changes, and mark the SonarCloud issue as **accepted.**
- **'timestamp' will use Object's default stringification format ('[object Object]') when stringified.**
- **'message' will use Object's default stringification format ('[object Object]') when stringified.**
  - The `timestamp` and `message` properties are typed as `unknown` by the `logform` dependency (used by `winston`); I simply built the types of this function argument to match the expected values/types from `logform`.
  - I would consider these 2 SonarCloud issues as **false positives**

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Updated the `Logger` class to support the `winston` library, and introduced migration tools to switch from `log4js` to `winston`. For more information on how to migrate your logger instance to use the `winston` library, please refer to the ["Configuring logging" page](https://github.com/zowe/zowe-cli/wiki/Configuring-Logging) on the Zowe CLI wiki. [#2488](https://github.com/zowe/zowe-cli/issues/2488)
```
anaxceron, line 7:
```suggestion
- Enhancement: Updated the `Logger` class to support the `winston` library, and introduced migration tools to switch from `log4js` to `winston`. For more information on how to migrate your logger instance to use the `winston` library, refer to the ["Configuring logging" page](https://github.com/zowe/zowe-cli/wiki/Configuring-Logging) on the Zowe CLI wiki. [#2488](https://github.com/zowe/zowe-cli/issues/2488)
```

anaxceron (approved):
Looks good, @traeok. Left you one small edit.

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Updated the `Logger` class to support the `winston` library, and introduced migration tools to switch from `log4js` to `winston`. For more information on how to migrate your logger instance to use the `winston` library, refer to the ["Configuring logging" page](https://github.com/zowe/zowe-cli/wiki/Configuring-Logging) on the Zowe CLI wiki. [#2488](https://github.com/zowe/zowe-cli/issues/2488)
```
zFernand0, line 7: Thanks for documenting the migration approach from log4js to winston 🙏

packages/imperative/src/logger/src/log4jsToWinston.ts
```diff
@@ -0,0 +1,273 @@
...
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import { transports, format, LoggerOptions, addColors } from "winston";
+import * as os from "os";
+import * as dayjs from "dayjs";
+
+export const customLevels = {
```
zFernand0, line 16: I know some people may prefer `enum`s for memory efficiencies and stuff, but I think this is a clear way to access information about a specific log level 🙏 

No need to over-engineer things if we can achieve the same (or more) with just objects 😋

packages/imperative/src/logger/src/log4jsToWinston.ts
```diff
@@ -0,0 +1,273 @@
...
+ * Strips color codes (%[ and %]).
+ * @param pattern The log4js pattern string.
+ * @returns A Winston format object.
+ */
+export function translateLog4jsPattern(
+    pattern: string
+): ReturnType<typeof format.combine> {
+    // Mapping from log4js date format specifiers to dayjs format specifiers
+    // https://github.com/log4js-node/log4js-node/issues/1012#issuecomment-1017144708
+    const mapLog4jsToDayjsFormat = (log4jsFormat: string): string => {
```
zFernand0, line 53: Should we just export this function (so that we don't have to define it every time `translateLog4jsPattern` is called?

packages/imperative/src/logger/src/log4jsToWinston.ts
```diff
@@ -0,0 +1,273 @@
...
+                    }
+                    // Use dayjs to format the timestamp with the specific format for this token
+                    return dayjs(info.timestamp as string).format(
+                        formatString
+                    );
+                }
+            );
+
+            // Regex for non-date, non-color tokens handled previously
+            const otherTokenRegex = /(%p|%c|%m|%n|%h|%z|%%)/g;
```
zFernand0, line 113: Just for my own understanding... 
Isn't this terminal specific? 

For instance, there are tokens that instead of `%` may just escape the letter (e.g. `\u`). 
I might be confusing these tokens with something else though 😅

packages/imperative/src/logger/src/log4jsToWinston.ts
```diff
@@ -0,0 +1,273 @@
...
+): appender is { layout: { type: "pattern"; pattern: string } } {
+    return (
+        typeof appender === "object" &&
+        appender !== null &&
+        "layout" in appender &&
+        typeof (appender as any).layout === "object" &&
+        (appender as any).layout !== null &&
+        (appender as any).layout.type === "pattern" &&
+        typeof (appender as any).layout.pattern === "string"
+    );
```
zFernand0, line 157: I like the use of type guards here. Should we define it as a type?
i.e. `const type tAppenderPatter = { layout: { type: "pattern"; pattern: string } }`

Also, couldn't we simplify the function using optional-chaining (Elvis operator?)?

```typescript
const layout = appender?.layout;
if (layout != null) {
  return layout.type === "pattern" && typeof layout.pattern === "string" 
}
return false;
```

packages/imperative/src/logger/src/log4jsToWinston.ts
```diff
@@ -0,0 +1,273 @@
...
+export function isAppenderWithType(
+    appender: unknown,
+    type: string
+): appender is { type: string; filename?: string } {
+    return (
+        typeof appender === "object" &&
+        appender !== null &&
+        "type" in appender &&
+        (appender as any).type === type
+    );
```
zFernand0, line 272: Similar to previous comment 😋 

If we can, should we look into simplifying this function as well?

zFernand0 (approved):
LGTM! 😋 
---
Thanks for the testing steps! 🙏 
I left some comments, but nothing that requires changes to be made 😅 
Thanks a lot for keeping backwards compatibility when doing this! 🙏

packages/imperative/src/logger/src/log4jsToWinston.ts
traeok: These tokens are specific to log4js, this logic is mostly to help convert the old tokens to equivalent Winston formats 😋

packages/imperative/src/logger/src/log4jsToWinston.ts
traeok: Also something to consider: an enum transpiles to an object behind the scenes ^^

packages/imperative/src/logger/src/log4jsToWinston.ts
traeok: This was more of a C++ habit of mine where I like to use lambdas if I don't reuse the logic elsewhere 😋 
But we can definitely expose this inner function's logic in case an extender needs it.

Plan to address in an upcoming commit 👍

packages/imperative/src/logger/src/log4jsToWinston.ts
traeok: I agree that a type would be cleaner here, so I added one.
In regard to simplifying the function, each condition here feels necessary:

- `typeof appender === "object"` ensures that `appender` is an object
- `appender !== null` is required as it rules out the case where `typeof null === "object"`
- `"type" in appender` allows us to access the `type` property on `appender` without TypeScript errors

However, the last condition can be simplified to `appender.type === type`, so I'll adjust that as part of my next commit 😋

packages/imperative/src/logger/src/log4jsToWinston.ts
traeok: This won't work because `type` and `pattern` are not properties of `object`.
That said, I did clean up the logic in an upcoming commit and added a separate type guard for `layout is PatternLayout` (new type) for readability 😋

packages/imperative/src/logger/src/log4jsToWinston.ts
```diff
@@ -0,0 +1,289 @@
...
+    return (
+        typeof appender === "object" &&
+        appender !== null &&
+        "layout" in appender && isPatternLayout(appender.layout)
+    );
+}
+
+export const defaultPrintfFormat = 
+({ timestamp, level, message }: { timestamp: unknown, level: string, message: unknown }) =>
+    `[${timestamp}] [${level}] ${message}`;
```
zFernand0, line 176: SonarCloud has some good points...
is it possible for `timestamp` or `message` to be `typeof === "object"`s?
if so, the stringified version will just show `[object Object]` 😋  

**_Source:_** https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=2496&id=zowe_zowe-cli&open=AZasm3iCXjSg5CQ2xG3t

traeok: Commented on these above:

https://github.com/zowe/zowe-cli/pull/2496#issuecomment-2863272490

> 'timestamp' will use Object's default stringification format ('[object Object]') when stringified.
'message' will use Object's default stringification format ('[object Object]') when stringified.
The timestamp and message properties are typed as unknown by the logform dependency (used by winston); I simply built the types of this function argument to match the expected values/types from logform.
I would consider these 2 SonarCloud issues as false positives as the timestamp and message properties are properly stringified (as seen through testing)

packages/imperative/src/logger/src/Logger.ts
```diff
@@ -144,20 +148,126 @@ export class Logger {
...
+        }
+    }
 
-    constructor(private mJsLogger: log4js.Logger | Console, private category?: string) {
+    /**
+     * This flag is being used to monitor the logger configure status.
+     */
+    private initStatus: boolean;
 
+    constructor(private mJsLogger: log4js.Logger | winston.Logger | Console, private category?: string) {
```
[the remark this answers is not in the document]
zFernand0: 🤔  it's a private property, so, even if extending the class, people shouldn't be accessing the category directly 😅 
but it might be ogod to be cautious here, and just accept the Sonar Cloud issue 😋

packages/imperative/src/logger/src/log4jsToWinston.ts
zFernand0: Thanks for sharing! 🙏 
Changed the status as a false positive with a link to this conversation 🙏

packages/imperative/src/logger/src/Logger.ts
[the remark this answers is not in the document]
traeok: https://github.com/zowe/zowe-cli/issues/2505

packages/cli/__tests__/daemon/__unit__/DaemonDecider.unit.test.ts
```diff
@@ -240,12 +240,16 @@ describe("DaemonDecider tests", () => {
...
         let expectedCommChannel: string;
         if (process.platform === "win32") {
-            expectedCommChannel = `\\\\.\\pipe\\${os.userInfo().username}\\ZoweDaemon`;
+            expectedCommChannel = `\\\\.\\pipe\\${os.userInfo().username.toLocaleLowerCase()}\\ZoweDaemon`;
         } else {
             expectedCommChannel = path.join(os.homedir(), ".zowe/daemon/daemon.sock");
         }
 
-        expect((daemonDecider as any).mSocket).toEqual(expectedCommChannel);
+        if (process.platform === "win32") {
```
gejohnston, line 248: I do not see any difference between the "if" and "else" actions?

traeok: Ah thanks for spotting this Gene. Initially I had the `win32` branch checking for the lowercase channel path, but after updating the logic to use a common variable, we no longer need the if/else. I've addressed this in my upcoming commit 👍

PR #2507, opened 2025-05-24
Add extensionMap support to Download.allMembers

JWaters02:
**What It Does**

Enhancement: Added support for `IDownloadOptions.extensionMap` to the `Download.allMembers` method to allow for specifying the file extension for each downloaded member. If `extensionMap` isn't specified it just uses the `extension` option as before (which uses the default extension if that is also not specified).

**How to Test**

Try running systems tests in `Download.system.test.ts` that tests the `extensionMap` stuff and checking the download output.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**


anaxceron (commented):
Left a minor comment re: the changelog

packages/zosfiles/src/methods/download/Download.ts
```diff
@@ -259,6 +259,18 @@ export class Download {
...
                 const fileName = options.preserveOriginalLetterCase ? mem.member : mem.member.toLowerCase();
+
+                // Determine extension using extensionMap if provided
+                let extension = defaultExtension;
+                if (options.extensionMap != null) {
+                    // Use the member name as the key, lowercased if preserveOriginalLetterCase is false
+                    extension = options.extensionMap[fileName] ?? extension;
+                }
+                // Normalize the extension, remove leading periods
+                if (extension && extension.startsWith(".")) {
```
zFernand0, line 270: (nitpick) 😅 
Is the intention of this check to verify that extension is not an empty string `""`? 
If so, can we be a bit more explicit? (`extension !== "" && ...`)
I don't believe it's possible for extension to be null or undefined at this point 😋

zFernand0 (approved):
LGTM! 😋 
---
left a nitpicky comment about checking for empty string 😅 
not a hard requirement 😋

zFernand0 (approved):
LGTM! 😋 
---
All system tests pass except for one. However, the same test also fails on the master branch 😋 

We should be good to merge 😋

PR #2529, opened 2025-06-12
Add the ability to search an exact data set or PDS

awharn:
**What It Does**

Adds a `search-exact-name` option to the `zowe zos-files search data-set` command to allow users to specify they only wish to search one data-set.
Adds a `search-exact-name` option to the `ISearchOptions` interface for the same reason as above.

**How to Test**

Run the `zowe zos-files search data-set` command with the `search-exact-name` option when you have two data-sets, one of which having a high level qualifier that is a subset of the second. Specify the pattern to be the shorter of the two. (i.e. ZOWE.TEST.DS and ZOWE.TEST.DS.JCL, pattern ZOWE.TEST.DS).
Observe only one data set is searched.

**Review Checklist**
I certify that I have:
- [x] tested my changes (1125)
- [x] updated the changelog

**Additional Comments**


packages/cli/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added a `search-exact-name` option to `zowe zos-files search data-set` to search the contents of one data set or PDS [#2529](https://github.com/zowe/zowe-cli/pull/2529)
```
anaxceron: - We should explain what `zowe zos-files search data-set` is.
- Sentence needs to end w/ a period.

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added an optional `search-exact-name` option to `ISearchOptions` to search the contents of one data set or PDS [#2529](https://github.com/zowe/zowe-cli/pull/2529)
```
anaxceron: Same as previous comment.

packages/cli/__tests__/zosfiles/__unit__/search/ds/Datasets.handler.unit.test.ts
```diff
@@ -118,6 +118,100 @@ describe("Search Datasets handler", () => {
...
+                        })
+                    },
+                    progress: {
+                        startBar: jest.fn((_parms) => {
+                            // do nothing
+                        }),
+                        endBar: jest.fn(() => {
+                            // do nothing
+                        })
+                    }
```
zFernand0, line 187: we need to figure out a way (in a future PR) re: how we can reduce code duplication in tests 😋

packages/cli/__tests__/zosfiles/__unit__/search/ds/__snapshots__/Datasets.handler.unit.test.ts.snap
```diff
@@ -152,6 +152,33 @@ exports[`Search Datasets handler process method should search a data set if requ
...
+      ],
+      "member": "TESTMEM",
+    },
+  ],
+  "commandResponse": "Found \\"test\\" in 1 data sets and PDS members",
+  "success": true,
+}
+`;
+
+exports[`Search Datasets handler process method should search a data set if requested - with searchExactName 2`] = `""`;
```
zFernand0, line 175: Nitpick:
Seems like `apiMessage` is empty.
Instead of using snapshots for this, can we check right in the test itself?

packages/cli/__tests__/zosfiles/__unit__/search/ds/__snapshots__/Datasets.handler.unit.test.ts.snap
```diff
@@ -152,6 +152,33 @@ exports[`Search Datasets handler process method should search a data set if requ
...
+
+exports[`Search Datasets handler process method should search a data set if requested - with searchExactName 3`] = `
+"
+Found \\"test\\" in 1 data sets and PDS members"
```
zFernand0, line 179: Nitpick:
Kind of the same as the previous comment, since we kind of know the structure of the message, could we check it right in the test?

zFernand0 (approved):
LGTM! 😋 
--- 
Left some small comments but the feature LGTM as-is 😋

packages/cli/src/zosfiles/-strings-/en.ts
```diff
@@ -643,6 +643,7 @@ export default {
                 },
                 OPTIONS: {
                     CASESENSITIVE: "The search should be case sensitive",
+                    SEARCHEXACTNAME: "The search pattern should match the exact name of the data set to search.",
```
gejohnston: I found the description a bit difficult to follow. I suggest something like the following. It may be worthwhile to get advice from Ana.

"To be selected, a dataset must be an exact match to the search pattern."

gejohnston (changes requested):
I made a minor request about wording. Also, I almost always agree with Fernando's comments about testing objects for desired values within tests rather than using snapshots.

packages/cli/src/zosfiles/-strings-/en.ts
awharn: Perhaps something like "Treat the input pattern as an exact data set name."? Just trying to convey it will look for one, and only one, data set (or PDS).

gejohnston: That seems very concise and I think it conveys the meaning clearly. Thanks.

anaxceron: Offering two ideas:
- Search pattern and data set name must match (to get results).
- Search pattern must match the data set name (to get results).

awharn: I went with 'Search pattern must exactly match the data set name'. Technically, data sets that are searched without this option match the glob pattern (i.e. 'ZOWE.USER.*.DS' will match against 'ZOWE.USER.TEST.DS', but they are not an exact match).

t1m0thyj (approved):
LGTM, thanks @awharn!

pujal0909 (commented):
@awharn The feature looks good to me when a data set name is specified! But when I use a wildcard pattern like "test.pds*", and include the flag, it still returns a result (from test.pds1).

pujal0909 (approved):

PR #2538, opened 2025-06-19
Fixed invalid logonProc response for TSO command

pujal0909:
**What It Does**

When running a TSO command and an invalid logonProc is used, it now throws an accurate error.
https://github.com/zowe/zowe-explorer-vscode/issues/2528 (Part One)

**How to Test**

Link the CLI to ZE and run a TSO command on ZE with an invalid logon proc in a TSO profile.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

Do we think the error message should be more specific?


anaxceron (approved):

awharn (changes requested):
The added system test appears to fail if the fake procedure is not defined. A discussion on whether or not the system test is needed may be warranted.

packages/zostso/src/IssueTso.ts
```diff
@@ -110,6 +110,13 @@ export class IssueTso {
             });
         }
 
+        if(response.startResponse?.messages?.includes("IKJ56482I")) {
+            throw new ImperativeError({
+                msg: `Invalid logon procedure.`,
+                additionalDetails: response.startResponse?.messages
```
t1m0thyj: This is a nitpick, but sometimes our static code analysis tools are picky about the `?.` operator (optional chaining) 🙂  We know that `response.startResponse` is defined once we're inside the `if` block.
```suggestion
                additionalDetails: response.startResponse.messages
```

awharn:
> The added system test appears to fail if the fake procedure is not defined. A discussion on whether or not the system test is needed may be warranted.

Upon re-review, this is incorrect. An existing system test (`packages/cli/__tests__/zostso/__system__/issue/cli.zos-tso.issue.cmd.system.test.ts` test `zos-tso issue command › should honor the logon proc specified in the profile`) fails the expect on line 61.

Before it expected no error, but now it receives:
```
Invalid logon procedure.
Diagnostic Information
IKJ56482I THE PROCEDURE NAME F4K3PR0C HAS NOT BEEN DEFINED FOR USE
IKJ56429A REENTER -
```

t1m0thyj (approved):
LGTM, thanks @pujal0909 for updating the system test!

awharn (approved):

PR #2545, opened 2025-07-02
fix: pass PWSTR pointer by reference

traeok:
**What It Does**

Fixes an issue surfaced by a new version of the Rust compiler where a pointer variable was passed to the WinAPI function `FormatMessageW` by value instead of by reference, which is not considered sound in cases where the parameter is referred to by reference.

Now, we pass the parameter as a mutable reference to resolve the compiler error. The local variable `buffer` was updated to be mutable in order to support this change.

**How to Test**

Not really easy to test, just build the Secrets SDK and ensure that the build is successful.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

### DIFF
--- a/packages/secrets/src/keyring/Cross.toml
+++ b/packages/secrets/src/keyring/Cross.toml
@@ -11,6 +11,9 @@ passthrough = [
     "RUSTFLAGS"
 ]
 
+[target.aarch64-unknown-linux-gnu]
+image = "ghcr.io/cross-rs/aarch64-unknown-linux-gnu:main"
+
 [target.armv7-unknown-linux-gnueabihf]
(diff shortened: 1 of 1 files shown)


awharn:
Additional work:
- Move ARM64 builds to ARM64 runners
- Update Cross to support image toolchain options
- Fix ARM64 MUSL cross build

ATorrise (approved):
worked on my machine 😄

zFernand0 (approved):
Changes LGTM! 😋

t1m0thyj (approved):
LGTM, thanks @traeok for the quick fix and @awharn for fixing the cross-compilation!

PR #2556, opened 2025-07-10
fix: special character corruption on upload

jace-roell:
**What It Does**
Fixes a bug where uploaded special characters may be corrupted if they fall on chunk borders

**How to Test**

1. Create a file that contains many special characters and will be larger than the default chunk size. (32 kb)

ex: a file with this line 10,000 times. "ñøçüßΩ中日漢字éâêîôû¡¿☆☂＠＃＆％＋＝"

2. Upload the file
`zowe files upload ftu ~/special_char.txt /u/users/<user>/<dir>/<fileName> --encoding UTF-8`

3. Open the file via ZE (UTF-8)

4. Ctrl+F a line. If the instances found does not match the total lines then there is data corruption and you may look at these mismatched lines.

image
ex: this line has 10150 lines but only has 10144 instances of the expected line.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/zosfiles/__tests__/__system__/methods/upload/Upload.system.test.ts
+++ b/packages/zosfiles/__tests__/__system__/methods/upload/Upload.system.test.ts
@@ -1047,7 +1047,10 @@ describe("Upload USS file", () => {
 
             try {
                 fs.mkdirSync(path.dirname(testFilePath), { recursive: true });
-                fs.writeFileSync(testFilePath, content, "utf-8");
+                if(!fs.existsSync(testFilePath))
+                {
+                    fs.writeFileSync(testFilePath, content, "utf-8");
+                }
 
                 const expectedUniqueChars = new Set(stripNewLines(content));
 


packages/zosfiles/__tests__/__system__/methods/upload/testfiles/specialCharTestData.txt
zFernand0, line 1: is it possible to generate this file?

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Fixed a bug where uploaded special characters may be corrupted if they fall on chunk borders for streaming. [#2556](https://github.com/zowe/zowe-cli/issues/2555)
```
zFernand0: I would like to have Ana's opinion on this one 😋 

What do you think about any of these? 😋 
- BugFix: Resolved an issue where streaming uploads of special characters could result in data corruption at chunk boundaries
- BugFix: Resolved an issue where uploaded special characters may be corrupted if they fall at data transfer segments during streaming

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -223,6 +223,14 @@ export abstract class AbstractRestClient {
      */
     protected lastByteReceived: number = 0;
 
+    /**
+     * Last byte received when upload is being streamed
+     * @private
+     * @type {number}
+     * @memberof AbstractRestClient
+     */
+    protected lastByteReceivedUpload: number = 0;
```
traeok, line 232: Just curious, what happens when I try to make a subsequent request if this variable is not reset? Should we reset it or are we good to leave it as-is? 😋

packages/zosfiles/__tests__/__system__/methods/upload/Upload.system.test.ts
```diff
@@ -1023,6 +1024,35 @@ describe("Upload USS file", () => {
...
+            await Download.ussFile(
+                REAL_SESSION,
+                targetUssPath,
+                { encoding: "UTF-8", file: downloadedFilePath }
+            );
+
+            const downloadedContent = fs.readFileSync(downloadedFilePath, "utf-8");
+            const downloadedUniqueChars = new Set(stripNewLines(downloadedContent));
+
+            expect(downloadedUniqueChars).toStrictEqual(expectedUniqueChars);
```
traeok: (nit) we could just check if the contents are equal after stripping new lines, as there is logical overhead in the Set that causes each duplicate element to be stripped. Doing a `toStrictEqual` check between the two contents (before de-duplication) should have a similar effect w/ a negligible performance difference.

traeok (commented):
Thanks for working on this Jace - I had a question regarding the `lastByteReceivedUpload` variable, but the changes make sense to me. Plan to approve based on discussion about the class property.

traeok (approved):
Changes LGTM, thanks for quickly addressing feedback!

anaxceron (approved):

Issue #2560, opened 2025-07-15
Implement a way to see verbose logs in console


t1m0thyj:
**Is your feature or enhancement request related to a problem or limitation? Please describe**

If I want to view debug logs for a Zowe CLI command, currently I have to look in the `~/.zowe/logs` folder.

**Describe your enhancement idea**

Add a method (environment variable or CLI argument) to specify verbose logging.

**Describe alternatives you've considered**

**Provide any additional context**

PR #2562, opened 2025-07-17
Add verbose option to plugins install command

t1m0thyj:
**What It Does**

Adds `--verbose` option to the `zowe plugins install` command to make debugging easier

**How to Test**

Install a plug-in with custom install scripts like the DB2 plug-in: `zowe plugins install @zowe/db2-for-zowe-cli --verbose`

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (build 1214)

**Additional Comments**


packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
```diff
@@ -55,13 +58,17 @@ export function installPackages(npmPackage: string, npmArgs: INpmInstallArgs): s
...
-        execOutput = ExecUtils.spawnAndGetOutput(npmCmd, args, {
+        if (verbose && daemonStream == null) {
+            pipe[0] = pipe[1] = "inherit";
+        }
+
+        execOutput = (ExecUtils.spawnAndGetOutput(npmCmd, args, {
             cwd: PMFConstants.instance.PMF_ROOT,
             stdio: pipe
-        }).toString();
+        }) ?? "").toString();
```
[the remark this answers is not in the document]
zFernand0: I don't believe this is true 😅 

```shell
> $ node
Welcome to Node.js v22.15.1.
Type ".help" for more information.
> null.toString()
Uncaught TypeError: Cannot read properties of null (reading 'toString')
> undefined.toString()
Uncaught TypeError: Cannot read properties of undefined (reading 'toString')
```

packages/imperative/src/imperative/__tests__/plugins/utilities/NpmFunctions.unit.test.ts
```diff
@@ -164,4 +146,179 @@ describe("NpmFunctions", () => {
...
+            const spawnSpy = jest.spyOn(ExecUtils, "spawnAndGetOutput").mockReturnValue(stdoutBuffer);
+            
+            const result = npmFunctions.installPackages("samplePlugin", { 
+                prefix: "fakePrefix", 
+                registry: fakeRegistry 
+            });
+
+                         const calledArgs = spawnSpy.mock.calls[0]?.[1];
+             expect(calledArgs).not.toContain("--loglevel=debug");
+             expect(calledArgs).not.toContain("--foreground-scripts=true");
```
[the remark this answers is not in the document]
zFernand0: Makes sense to clean this up 😋

packages/imperative/src/imperative/src/plugins/cmd/install/install.handler.ts
```diff
@@ -148,7 +151,10 @@ export default class InstallHandler implements ICommandHandler {
                         params.response.console.log("\n_______________________________________________________________");
                         const registryInfo = NpmRegistryUtils.buildRegistryInfo(packageString, params.arguments.registry);
                         params.response.console.log("Location = " + registryInfo.location + "\n");
-                        const pluginName = await install(packageString, registryInfo);
+                        const pluginName = await install(packageString, registryInfo, false, params.arguments.verbose);
+                        if (params.arguments.verbose) {
+                            params.response.console.log("_______________________________________________________________");
+                        }
```
zFernand0, line 158: I'm guessing this is to make the output a bit cleaner 😋 
Feels a bit strange though 😅

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
```diff
@@ -43,9 +43,12 @@ export function findNpmOnPath(): string {
  * @return {string} command response
  *
  */
-export function installPackages(npmPackage: string, npmArgs: INpmInstallArgs): string {
+export function installPackages(npmPackage: string, npmArgs: INpmInstallArgs, verbose = false): string {
     const pipe: StdioOptions = ["pipe", "pipe", "pipe"];
     const args = ["install", npmPackage, "-g", "--legacy-peer-deps"];
+    if (verbose) {
+        args.push("--loglevel=debug", "--foreground-scripts=true");
```
zFernand0: For logLever...
- Should/Can we look at the app log level ENV var to determine how much output we should print?
- Should we also pass the `--loglevel=silly` flag if the user has specified `ZOWE_APP_LOG_LEVEL=TRACE`?
- Generally speaking, should we try to match our app_log_level to NPM's log level?
https://docs.npmjs.com/cli/v8/using-npm/logging#loglevel

That way people can control how much logs are printed out (when they pass the verbose flag)

packages/imperative/src/imperative/src/plugins/cmd/install/install.definition.ts
```diff
@@ -60,6 +60,10 @@ const loginDescription =
     "https://docs.npmjs.com/cli/login for NPM >= 9\n" +
     "https://docs.npmjs.com/cli/adduser for NPM < 9";
 
+const verboseDescription =
+    "Specifies that verbose output is printed for npm install.\n" +
+    "This may be useful for debugging errors during plugin installation.";
+
```
zFernand0, line 66: Since this is help text...
Could/Should Ana check this out too? 😋

t1m0thyj: @anaxceron Please review 😸

packages/imperative/src/imperative/src/plugins/cmd/install/install.handler.ts
t1m0thyj: Without the line added, it may be hard to tell the boundary between npm output and Zowe CLI output.

Happy to remove it if we think the line is unnecessary 😋

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
t1m0thyj: > Should we also pass the `--loglevel=silly` flag if the user has specified `ZOWE_APP_LOG_LEVEL=TRACE`?

I like this idea, and it seems like a good enough option for short-term - implemented in https://github.com/zowe/zowe-cli/pull/2562/commits/c39d19c05df956e8c35e6c685ed6db9fd0c81b7d

Matching our log level with npm seems outside the scope of this PR and could be addressed in https://github.com/zowe/zowe-cli/issues/2560.

anaxceron (approved):

packages/imperative/src/imperative/src/plugins/cmd/install/install.handler.ts
```diff
@@ -136,7 +136,10 @@ export default class InstallHandler implements ICommandHandler {
 
                             params.response.console.log("\n_______________________________________________________________");
                             params.response.console.log("Location = " + packageInfo.location + "\n");
-                            const pluginName = await install(packageArgument, registryInfo, true);
+                            const pluginName = await install(packageArgument, registryInfo, true, params.arguments.verbose);
+                            if (params.arguments.verbose) {
+                                params.response.console.log("_______________________________________________________________");
+                            }
                             params.response.console.log("Installed plugin name = '" + pluginName + "'");
                             params.response.console.log(runValidatePlugin(pluginName));
```
gejohnston, line 145: Do you have any idea of how we might ensure that this call to ValidatePlugin reports the same result as later running the `zowe plugins validate` command? We have seen a discrepancy in at least one case, and in that case, an old unsupported version of NodeJS was being used. 

I do not suggest any major effort that significantly increases the scope of this current enhancement. However, if something easy comes to mind, it might be good to add it while we are working in this area.

zFernand0 (approved):

packages/imperative/src/imperative/src/plugins/cmd/install/install.handler.ts
traeok: (UX nitpick) If I make my terminal narrow, the separator spans across multiple rows. We could use `process.stdout.columns` to determine how long to make the separator, e.g.:

```ts
if (params.arguments.verbose) {
    params.response.console.log("_".repeat((process.stdout.columns ?? 80) * 0.5));
}
```

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
```diff
@@ -43,9 +44,13 @@ export function findNpmOnPath(): string {
  * @return {string} command response
  *
  */
-export function installPackages(npmPackage: string, npmArgs: INpmInstallArgs): string {
+export function installPackages(npmPackage: string, npmArgs: INpmInstallArgs, verbose = false): string {
     const pipe: StdioOptions = ["pipe", "pipe", "pipe"];
     const args = ["install", npmPackage, "-g", "--legacy-peer-deps"];
+    if (verbose) {
+        const logLevel = Logger.getAppLogger().level === "TRACE" ? "silly" : "debug";
+        args.push(`--loglevel=${logLevel}`, "--foreground-scripts=true");
```
traeok: (nitpick) the `=true` part for foreground-scripts is redundant

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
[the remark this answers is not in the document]
traeok: I'm not sure if I agree about Copilot's suggestion being incorrect - but feel free to correct me if I'm missing something. Copilot is implying that the `ExecUtils.spawnAndGetOutput` function always returns a valid Buffer unless the function throws an error. Here's the logic for the function:

https://github.com/zowe/zowe-cli/blob/e60599d92a0497df50ba83a3affa783c90f5ae95/packages/imperative/src/utilities/src/ExecUtils.ts#L28-L43

The `cross-spawn` is simply a wrapper around `child_process.spawn` - in the Node examples, they use `.on` on `stdout` and `stderr` immediately after calling spawn, implying that neither would be `null` or `undefined`.
https://nodejs.org/api/child_process.html#child_processspawncommand-args-options

This means that `result.stdout` should always be a Buffer because you cannot define a Buffer with undefined or null, or the constructor throws an error. So if `result.stdout` is returned, the Buffer's `toString` method would work as expected. 

I can mark as resolved, but would like to keep this thread open in the meantime to discuss 😋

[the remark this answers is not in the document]
t1m0thyj: As long as `stdio` flags are set to "pipe", then process output gets returned from `spawn` and the return type should be never be nullish.

If `stdio` flags are set to "inherit", then process output gets redirected to the parent process and there is nothing to return.

Until this PR, I don't think we ever called our `spawnAndGetOutput` method with "inherit" mode. Technically we should enhance the return type to be a union of `Buffer | string | undefined`, but I'm concerned that would be a potential breaking change for extenders.

packages/imperative/src/imperative/src/plugins/cmd/install/install.handler.ts
t1m0thyj: Good idea, created an issue to track: #2567

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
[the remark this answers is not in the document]
t1m0thyj: Addressed in https://github.com/zowe/zowe-cli/pull/2562/commits/e55133066c7eaad1e1b19bcdc1d43650cfd138d4

packages/imperative/src/imperative/src/plugins/cmd/install/install.handler.ts
t1m0thyj: `runValidatePlugin` is the last step of the install handler, so I don't have much insight into this without doing more testing with the unsupported version of NodeJS.

packages/imperative/src/utilities/src/ExecUtils.ts
```diff
@@ -26,18 +26,34 @@ export class ExecUtils {
...
+    /**
+     * Spawn a process with arguments and throw an error if the process fails.
+     * Parameters are same as `child_process.spawnSync` (see Node.js docs).
+     * Output is inherited by the parent process instead of being returned.
+     * Use this method if you want the safe argument parsing of `spawnSync`
+     * combined with the smart output handling of `execSync`.
+     */
+    public static spawnWithInheritedStdio(command: string, args?: string[], options?: SpawnSyncOptions): void {
+        const result = spawn.sync(command, args, { ...options, stdio: "inherit" });
+        return this.handleSpawnResult<void>(result, `${command} ${args.join(" ")}`);
```
traeok: If `args` is `undefined`, then both functions fail with a `TypeError: Cannot read properties of undefined (reading 'args')`

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
```diff
@@ -55,13 +60,19 @@ export function installPackages(npmPackage: string, npmArgs: INpmInstallArgs): s
...
+                stdio: pipe
+            }).toString();
 
-        if(daemonStream != null) {
-            daemonStream.write(DaemonRequest.create({ stdout: execOutput}));
+            if (verbose && daemonStream != null) {
+                daemonStream.write(DaemonRequest.create({ stdout: execOutput }));
+            }
         }
     }
```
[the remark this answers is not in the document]
t1m0thyj: Silly Copilot, this code is fine and adding a comment doesn't change anything 😁

traeok (approved):
Thanks for addressing the recent feedback, LGTM

PR #2565, opened 2025-07-17
Fix links with special characters in the web help

### LINKED ISSUE #2308: Broken link in published webhelp for `zowe zos-files upload dir-to-uss`
**Describe the bug**

https://docs.zowe.org/stable/web_help/index.html?p=zowe_zos-files_upload_dir-to-uss contains a link that should point to https://git/-scm.com/docs/gitignore#_pattern_format, but is broken by a newline.

**Describe your environment**

**Additional context**

### LINKED ISSUE #2553: Broken link in Zowe CLi web-help
Hi Zowe CLI team,

We've noticed a broken link to our github repo in the Zowe CLI web-help due to `-` escaping:
https://docs.zowe.org/stable/web_help/?p=zowe_dbm-db2

Image

Could you please take a look?

Thanks,
Sergei.

awharn:
**What It Does**

Resolves #2308
Resolves #2553
Fixes links in the web help that contain special characters (i.e. hyphens)

**How to Test**

- Remove `~/.zowe/web-help`
- Run `zowe --help-web` and notice the entry for `zowe zos-files upload dir-to-uss` has a bad link to `https://git-scm.com`
- Remove `~/.zowe/web-help`
- Build and install this branch
- Run `zowe --help-web` and notice the entry for `zowe zos-files upload dir-to-uss` has a good link to `https://git-scm.com`

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/src/cmd/src/help/WebHelpGenerator.ts
+++ b/packages/imperative/src/cmd/src/help/WebHelpGenerator.ts
@@ -326,7 +326,7 @@ export class WebHelpGenerator {
         let htmlContent = "<h2>" + this.genBreadcrumb(rootCommandName, fullCommandName) + this.genPrintButton() + "</h2>\n";
 
         // Make regex check here to fix unintentional escaping in an HTTP(S) link
-        markdownContent = markdownContent.replaceAll(/https?:\/\/.+?(?=[ \n])/g, (match) => {
+        markdownContent = markdownContent.replaceAll(/https?:\/\/.+?(?=[ \n'"`\])}])/g, (match) => {
             return match.replaceAll("\\", "");
         });
 


gejohnston (approved):

packages/imperative/src/cmd/src/help/WebHelpGenerator.ts
```diff
@@ -324,6 +324,12 @@ export class WebHelpGenerator {
         }
 
         let htmlContent = "<h2>" + this.genBreadcrumb(rootCommandName, fullCommandName) + this.genPrintButton() + "</h2>\n";
+
+        // Make regex check here to fix unintentional escaping in an HTTP(S) link
+        markdownContent = markdownContent.replaceAll(/https?:\/\/.+?(?=[ \n])/g, (match) => {
```
traeok: - (nitpick) This regex is vulnerable to ReDoS with the string `http://` repeated several times
- URLs followed by punctuation or parentheses may still have the escaped hyphens, for example `Visit https://example-site.com.`, `"Go to https://example-site.com"`, or `(see https://example-site.com)`
- This is limited to handling fixing URLs with `http(s)` schemes only, but we may not intend to support other schemes - if that's the case please disregard this point

traeok (approved):
Changes make sense to me, I mentioned a couple possible edge cases in my comment but this PR solves the reported scenario so I'm going to approve 👍 Thanks Andrew

packages/imperative/src/cmd/src/help/WebHelpGenerator.ts
awharn: It was intended to only fix those particular cases, since both of our issues do involve that, and I don't think we want people to be putting links to things other then documentation in their documentation. As for the ReDoS concerns, that would only occur if someone were to repeat `http://` thousands of times in their command description or examples, which isn't much of a concern, but I also cannot think of a way to evaluate this in a better way.

awharn: As for the other cases, I will add breaks at single and double quotes, backticks, and closed brackets and parenthesis.

traeok: > I don't think we want people to be putting links to things other then documentation in their documentation

There's nothing stopping a developer from doing that 😛 but I agree, we should be cautious on what links are allowed

The ReDoS was definitely a nitpick, doesn't need to be addressed ^^

t1m0thyj (approved):
LGTM, thanks @awharn!

PR #2575, opened 2025-07-30
Fixed Explorer loading with this validity check on missing nested/base profiles

ATorrise:
**What It Does**
Any invalid base profile, nested or not, breaks the loading of Zowe Explorer and doesn't recognize the user's current team config, prompting them to create a new one. This PR creates a new validity check for previously-set base profiles so that ZE can still open and load valid profiles normally. 

image

**How to Test**
Run zowe config set defaults.base "base.invalid" and reload VS Code.
in zowe-explorer-vscode/package.json, add overrides so that you can connect the imperative changes to your ZE
image

Sorry for this but I don't think I can provide you guys with simple git URLs... the internet tells me that because the CLI is a monorepo, pnpm/npm will try to build the repo, which requires yarn.. which doesn't support subdir installs... and each override is for a subdir. If you can figure this out or im wrong, explain how to do it in comments!

You can use these (modify for your environment):
```
   "pnpm": {
     "overrides": {
       "@zowe/cli": "file:../zowe-cli/packages/cli",
       "@zowe/core-for-zowe-sdk": "file:../zowe-cli/packages/core",
       "@zowe/zosmf-for-zowe-sdk": "file:../zowe-cli/packages/zosmf",
       "@zowe/imperative": "file:../zowe-cli/packages/imperative",
       "@zowe/zos-files-for-zowe-sdk": "file:../zowe-cli/packages/zosfiles",
       "@zowe/zos-jobs-for-zowe-sdk": "file:../zowe-cli/packages/zosjobs",
       "@zowe/zos-console-for-zowe-sdk": "file:../zowe-cli/packages/zosconsole",
       "@zowe/zos-tso-for-zowe-sdk": "file:../zowe-cli/packages/zostso",
       "@zowe/zos-uss-for-zowe-sdk": "file:../zowe-cli/packages/zosuss"
     }
   }
   ```
pnpm install & ignore these peer errors
image

pnpm run build for explorer then use the Run and Debug tab in your explorer window to test that the newly populated window can load vscode with your valid profiles present in the explorer tab on this branch but not on main
image

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**
Fixes zowe/zowe-explorer-vscode#3372

### CODE REVIEW

File: packages/imperative/CHANGELOG.md [hunk 1]
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Added a validity check for base profiles when merging in ProfileInfo.mergeArgsForProfile. [#2575](https://github.com/zowe/zowe-cli/pull/2575)
```
anaxceron: Please describe what ProfileInfo.mergeArgsForProfile is, and use backticks: " .. when merging in method/command/property `ProfileInfo.mergeArgsForProfile`."

anaxceron (approved):

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -527,6 +527,32 @@ export class ProfileInfo {
...
 
+    /**
+     * Checks if a profile (including nested subprofiles) exists in the given profiles object.
+     * The profile name can use dot notation for nested profiles (e.g., "main.sub1.sub2").
+     *
+     * @param profileName - The dotted path to the profile (e.g., "main.sub1.sub2")
+     * @param profilesObj - The profiles object to search within
+     * @returns true if the profile exists, false otherwise
+     */
+    private static profileExists(profileName: string, profilesObj: any): boolean {
```
traeok: Can we come up with a more explicit type for `profilesObj`? Generally its discouraged to explicitly type a parameter as `any` unless we cannot determine the type at the time of the function call.

traeok (changes requested):
Thanks Amber, the changes make sense but I have a request for the new function.

packages/imperative/src/config/src/ProfileInfo.ts
ATorrise: 😅 yes sorry

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -601,6 +634,10 @@ export class ProfileInfo {
                     layerProperties = this.mLoadedConfig.findLayer(false, osLoc.global)?.properties;
                     realBaseProfileName = layerProperties?.defaults.base;
                 }
+                const profilesObj = layerProperties?.profiles;
+                if (realBaseProfileName && !ProfileInfo.profileExists(realBaseProfileName, profilesObj)) {
```
t1m0thyj: We already have an API to check if a profile exists - can we do the following instead of adding a new method and interface?
```suggestion
                if (realBaseProfileName && !this.mLoadedConfig.api.profiles.exists(realBaseProfileName)) {
```

ATorrise: sorry i didnt realize thank you for showing me this

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -53,6 +53,15 @@ import { IConfigLayer } from "..";
 import { Constants } from "../../constants";
 import { Censor } from "../../censor";
 
+/**
+ * Minimal interface for a profile node to support recursive nesting via the `profiles` property.
+ * Enables traversal of nested profiles using dot notation (ie: "base.sub1.sub2").
+ */
+interface IProfileNode {
+    profiles?: { [key: string]: IProfileNode };
+}
```
traeok: Let's reuse a pre-existing type that we can access here, for example (please see next comment for more context):

Image


```suggestion
type ProfileRecord = IConfig["profiles"];
```

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -527,6 +536,30 @@ export class ProfileInfo {
...
+
+        for (const level of levels) {
+            if ((current as Record<string, IProfileNode>)[level]) {
+                current = (current as Record<string, IProfileNode>)[level].profiles || (current as Record<string, IProfileNode>)[level];
+            } else {
+                return false;
+            }
+        }
+        return true;
+    }
```
traeok: Referring to my previous comment...

If you highlight over `layerProperties` in `ProfileInfo.mergeArgsForProfile`, you'll see that it it is of type `IConfig`:

Image

So, we can reuse the type of the `profiles` property in the `IConfig` interface here, with a couple code changes to avoid the repeated cast that's currently in place. Using the type recommended in the previous comment...

```suggestion
    private static profileExists(profileName: string, profilesObj: ProfileRecord): boolean {
        if (!profileName || !profilesObj) return false;

        const levels = profileName.split(".");
        let current: IConfigProfile | ProfileRecord = profilesObj;

        for (const level of levels) {
            const currentAsProfiles = current as ProfileRecord;
            if (currentAsProfiles[level]) {
                current = currentAsProfiles[level].profiles || currentAsProfiles[level];
            } else {
                return false;
            }
        }
        return true;
    }
```

t1m0thyj: I think this method and the corresponding interface can both be removed, since they're unused once my comment is addressed 😋

t1m0thyj (commented):
Integration tests are hanging for Node 22 + Windows, but that's not unique to this branch - they are failing on `master` as well.

Perhaps this is due to a change in the GitHub runner environment?
| Component | [Last Successful Run](https://github.com/zowe/zowe-cli/actions/runs/16580186453/job/46894230125) | [Newly Failing Run](https://github.com/zowe/zowe-cli/actions/runs/16807574452/job/47603990274) |
| --- | --- | --- |
| Runner Image | 20250720.1.0 | 20250727.1.0 |
| Node.js | 22.17.1 | 22.18.0 |
| Rust | 1.88.0 | 1.89.0 |

packages/imperative/src/config/src/ProfileInfo.ts
traeok: Ah, good point. Please disregard 👍

gejohnston (approved):

PR #2600, opened 2025-09-02
Make changes to support #2598

### LINKED ISSUE #2596: TSO SDK does not export `IIssueTsoCmdOpts` interface
The `issueTsoCommand` function has been deprecated in favor of `issueTsoCmd`, which uses the `IIssueTsoCmdOpts` interface. This interface should be exported from the TSO SDK package for extenders to use.

### LINKED ISSUE #2598: Setting log level to TRACE should not display credentials
See https://github.com/eclipse-che4z/che-che4z-lsp-for-hlasm/issues/305#issuecomment-3201069010. A common step when debugging ZE is to set log level to TRACE. This may result in B64-encoded credentials being logged. Although we warn about this in Zowe docs, it is still unexpected behavior that we should try to avoid.

awharn:
**What It Does**

Fixes logging basic credentials in the Authorization header in the AbstractRestClient
Fixes logging cookie based authentication in the AbstractRestClient
Fixes exports for the `IIssueTsoCmdOpts`, `IStartTsoAppParms`, and `ITsoAppCommunicationParms` interfaces
Fixes a daemon audit failure.
Fixes a secrets SDK audit failure.
Fixes #2598
Fixes #2596 

**How to Test**

Run a dataset list command in trace logger mode with an incorrect port number. Observe the Imperative logs.
Switch to this branch, and perform the same actions. Observe the issue is no longer present.

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (provide build number if applicable) - 1367 & 1368

**Additional Comments**

### DIFF
--- a/packages/secrets/src/keyring/Cargo.lock
+++ b/packages/secrets/src/keyring/Cargo.lock
@@ -3,10 +3,10 @@
 version = 4
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "adler2"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aho-corasick"
@@ -272,7 +272,7 @@ dependencies = [
 
 [[package]]
 name = "keyring"
-version = "1.0.0"
+version = "1.0.1"
 dependencies = [
  "napi",
  "napi-build",
@@ -492,9 +492,9 @@ checksum = "e5ea92a5b6195c6ef2a0295ea818b312502c6fc94dde986c5553242e18fd4ce2"
 
 [[package]]
 name = "secrets_core"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
- "adler",
+ "adler2",
  "cfg-if",
  "core-foundation",
  "core-foundation-sys",
--- a/packages/secrets/src/keyring/Cargo.toml
+++ b/packages/secrets/src/keyring/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 edition = "2021"
 name = "keyring"
-version = "1.0.0"
+version = "1.0.1"
 authors = ["Zowe Project"]
 license = "EPL-2.0"
 repository = "https://github.com/zowe/zowe-cli"


awharn:
Ignore the changelog failure - a linter error from another PR was fixed in this PR, but has no customer impact whatsoever.

packages/imperative/src/censor/src/Censor.ts
```diff
@@ -316,6 +326,49 @@ export class Censor {
...
+
+    /**
+     * Copy and censor over an object
+     * @param {Record<string, any>} data - the data to censor
+     * @returns {Record<string, any>} - the censored data
+     */
+    private static mCensorObject(data: Record<string, any>, secureValues: any[]): Record<string, any> {
+        const newData: Record<string, any> = {};
+        for (const [key, value] of Object.entries(data)) {
+            if (this.CENSORED_OPTIONS.includes(key) || secureValues.includes(value)) {
```
gejohnston, line 379: Trying to clarify my understanding: Is the intention to censor the value of a non-secure property if it happens to have the same value as another property which is defined as secure?

If not, then I misinterpreted, and that's fine.

If I am correct, then it could censor the values of some properties that are not marked as secure. I think this would also be ok since it guarantees the censoring of a known secure value regardless of what other properties happen to have that same value.

awharn: This does the latter, and will censor a non-secure property (or, in this case, value in an object) if it is known to be secure somewhere. This is because `host` can be considered a secure value, but may be under the key `hostname` in the session. This is also true for custom token headers (i.e. X-API-Token). Since any object can be passed in for censoring, we don't know if the key will necessarily exactly match the name of the property.

awharn:
Re-requested review due to the enhancement of allowing extenders to define additional headers to censor.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -646,7 +647,7 @@ export abstract class AbstractRestClient {
             return false;
         }
 
-        this.log.trace("Using cookie authentication with token %s", this.session.ISession.tokenValue);
+        this.log.trace("Using cookie authentication with token");
```
t1m0thyj: Thanks for the change, we definitely don't want to log token value here 😋 

(nitpick) Should we log token type instead for debugging purposes?
```suggestion
        this.log.trace("Using cookie authentication with token type %s", this.session.ISession.tokenType);
```

packages/imperative/src/censor/src/Censor.ts
```diff
@@ -55,6 +59,22 @@ export class Censor {
...
+        return Array.from(censoredList);
+    }
+
+    public static get CUSTOM_CENSORED_HEADERS(): string[] {
+        const censoredList = new Set<string>();
+        for (const header of this.INTERNAL_CUSTOM_CENSORED_HEADERS) {
+            censoredList.add(header);
+        }
+        return Array.from(censoredList);
+    }
```
traeok: I can see why this logic pattern is needed for `DEFAULT_CENSORED_OPTIONS` (in order to add the `kebabCase` format for the option), but it feels redundant here:

- Every time this getter is called, it has O(n) time complexity
- Uses a loop to copy array elements into the Set (no real de-duplication benefit here, since you already skip duplicates in the `addCensoredHeader` implementation on line 187)

Thoughts on spreading the array instead?


```suggestion
    public static get DEFAULT_CENSORED_HEADERS(): string[] {
        return [...this.MAIN_CENSORED_HEADERS];
    }

    public static get CUSTOM_CENSORED_HEADERS(): string[] {
        return [...this.INTERNAL_CUSTOM_CENSORED_HEADERS];
    }
```

This approach maintains the same behavior (doesn't mutate the original array), while being faster and avoiding memory allocations.

packages/imperative/src/censor/src/Censor.ts
```diff
@@ -162,6 +186,19 @@ export class Censor {
         }
     }
 
+    /**
+     * Add a header to censor
+     * @param {string} header - The name of the header to censor
+     */
+    public static addCensoredHeader(header: string) {
+        if (header != null) {
+            if (!this.INTERNAL_CUSTOM_CENSORED_HEADERS.includes(header)) {
```
traeok, line 187: see `includes` check mentioned in above comment

packages/imperative/src/censor/src/Censor.ts
awharn: Love the idea, thanks for the suggestion. You're absolutely right in this case. 😄

packages/imperative/src/rest/src/client/AbstractRestClient.ts
awharn: I am going to do a sensitive value check before we do, but sure.

t1m0thyj (approved):
LGTM, thanks @awharn!

traeok (approved):
Thanks again for the fix and for quickly addressing feedback 🙏

PR #2614, opened 2025-09-26
fix: Skip `\r` for data set and USS uploads for Windows clients 

traeok:
**What It Does**

Fixes a regression in the `AbstractRestClient` logic where newline normalization was only considered for downloads and not for uploads. Now, when uploading, the opposite logic is performed: carriage returns that come before new lines are stripped from the string/Buffer input before the file/data set is uploaded to the system.

**How to Test**

Pretty much impossible to test this unless you have a Windows machine or VM.

1. Make a text file in your Windows environment and type out the following contents (pressing Enter after the first Hello world to make a new line):
```
Hello world!
Hello world!
```
2. Upload these contents using `zowe zos-files upload file-to-uss`
3. Notice that on the target system, running `od -x <path-to-uss-file>` does not contain any `0D` (CR) sequences, only `15` (LF)
4. See the same behavior for `zowe zos-files upload file-to-data-set` when viewing the data set in ISPF with `hex on`

Here's a more-involved smoke test that just evaluates the behavior of IO.processNewlines on Windows. It should finish with RC 0 if all tests passed.

```ts
/**
 * Smoke test to verify CRLF normalization on chunk boundaries during upload on Windows.
 *
 * This test simulates the behavior of AbstractRestClient when normalizing newlines
 * during chunked uploads, specifically testing the edge case where CR appears at
 * the end of a chunk and LF appears at the start of the next chunk.
 *
 * Save this as `crlf-smoke-test.ts` in the `packages/imperative` folder and run this on Windows: `npx tsx crlf-smoke-test.ts`
 */
import { IO } from "./src/io/src/IO";
import * as os from "os";

interface TestCase {
    name: string;
    chunks: Buffer[];
    expected: string;
    description: string;
}

const testCases: TestCase[] = [
    {
        name: "CRLF split across chunk boundary",
        chunks: [
            Buffer.from("line1\r"),
            Buffer.from("\nline2")
        ],
        expected: "line1\nline2",
        description: "CR at end of chunk 1, LF at start of chunk 2 should become single LF"
    },
    {
        name: "Multiple CRLF across boundaries",
        chunks: [
            Buffer.from("line1\r"),
            Buffer.from("\nline2\r\n"),
            Buffer.from("line3\r"),
            Buffer.from("\n")
        ],
        expected: "line1\nline2\nline3\n",
        description: "Multiple CRLF sequences across chunk boundaries"
    },
    {
        name: "Standalone CR at chunk boundary",
        chunks: [
            Buffer.from("text\r"),
            Buffer.from("more")
        ],
        expected: "text\rmore",
        description: "CR not followed by LF should be preserved"
    },
    {
        name: "CR at end of stream",
        chunks: [
            Buffer.from("data\r")
        ],
        expected: "data\r",
        description: "Final CR should be preserved when no LF follows"
    },
    {
        name: "CRLF within single chunk",
        chunks: [
            Buffer.from("line1\r\nline2\r\nline3")
        ],
        expected: "line1\nline2\nline3",
        description: "CRLF within a chunk should be normalized to LF"
    },
    {
        name: "Mixed: CRLF in chunk and across boundary",
        chunks: [
            Buffer.from("a\r\nb\r"),
            Buffer.from("\nc\r\nd")
        ],
        expected: "a\nb\nc\nd",
        description: "Combination of CRLF within and across chunks"
    },
    {
        name: "Empty chunk after CR",
        chunks: [
            Buffer.from("test\r"),
            Buffer.from(""),
            Buffer.from("\nmore")
        ],
        expected: "test\nmore",
        description: "Empty chunk between CR and LF"
    }
];

function processChunks(chunks: Buffer[]): string {
    const result: Buffer[] = [];
    let lastByte: number | undefined = undefined;
    let heldByte: Buffer | undefined = undefined;

    for (let i = 0; i < chunks.length; i++) {
        let chunk = chunks[i];

        // If we held back a CR from the previous chunk, prepend it to current chunk
        if (heldByte != null) {
            chunk = Buffer.concat([heldByte, chunk]);
            heldByte = undefined;
        }

        // Process newlines (uploading = true)
        chunk = IO.processNewlines(chunk, lastByte, true);

        // If chunk ends with CR, hold it back until we see the next chunk
        if (chunk.byteLength > 0 && chunk[chunk.byteLength - 1] === 13) {
            heldByte = Buffer.from([chunk[chunk.byteLength - 1]]);
            chunk = chunk.subarray(0, chunk.byteLength - 1);
        }

        if (chunk.byteLength > 0) {
            result.push(chunk);
            lastByte = chunk[chunk.byteLength - 1];
        }
    }

    // Add held byte at the end if present
    if (heldByte != null) {
        result.push(heldByte);
    }

    return Buffer.concat(result).toString();
}

function runTests() {
    console.log("=".repeat(80));
    console.log("CRLF Chunk Boundary Normalization Smoke Test");
    console.log("=".repeat(80));
    console.log(`Platform: ${os.platform()}`);
    console.log(`Node version: ${process.version}`);
    console.log("");

    if (os.platform() !== "win32") {
        console.log("⚠️  WARNING: This test is designed for Windows (win32)");
        console.log("⚠️  On non-Windows platforms, processNewlines returns input unchanged");
        console.log("");
    }

    let passed = 0;
    let failed = 0;

    for (const testCase of testCases) {
        const result = processChunks(testCase.chunks);
        const success = result === testCase.expected;

        if (success) {
            console.log(`✓ PASS: ${testCase.name}`);
            passed++;
        } else {
            console.log(`✗ FAIL: ${testCase.name}`);
            console.log(`  Description: ${testCase.description}`);
            console.log(`  Expected: ${JSON.stringify(testCase.expected)}`);
            console.log(`  Got:      ${JSON.stringify(result)}`);
            console.log(`  Expected bytes: ${Buffer.from(testCase.expected).toJSON().data}`);
            console.log(`  Got bytes:      ${Buffer.from(result).toJSON().data}`);
            failed++;
        }
    }

    console.log("");
    console.log("=".repeat(80));
    console.log(`Results: ${passed} passed, ${failed} failed out of ${testCases.length} tests`);
    console.log("=".repeat(80));

    if (failed > 0) {
        process.exit(1);
    }
}

runTests();
```

**Review Checklist**
I certify that I have:
- [x] updated the changelog

### DIFF
--- a/packages/imperative/src/rest/src/client/AbstractRestClient.ts
+++ b/packages/imperative/src/rest/src/client/AbstractRestClient.ts
@@ -404,12 +404,12 @@ export abstract class AbstractRestClient {
                     this.log.debug("Writing data chunk of length %d from requestStream to clientRequest", data.byteLength);
 
                     // If we held back a CR from the previous chunk, prepend it to current chunk
-                    if (this.mNormalizeRequestNewlines && heldByte != null) {
-                        data = Buffer.concat([heldByte, data]);
-                        heldByte = undefined;
-                    }
-
                     if (this.mNormalizeRequestNewlines) {
+                        if (heldByte != null) {
+                            data = Buffer.concat([heldByte, data]);
+                            heldByte = undefined;
+                        }
+
                         this.log.debug("Normalizing new lines in request chunk to \\n");
                         data = IO.processNewlines(data, this.lastByteReceivedUpload, true);
 
@@ -434,8 +434,8 @@ export abstract class AbstractRestClient {
 
                     if (data.byteLength > 0) {
                         clientRequest.write(data);
+                        this.lastByteReceivedUpload = data[data.byteLength - 1];
                     }
-                    this.lastByteReceivedUpload = data.byteLength > 0 ? data[data.byteLength - 1] : this.lastByteReceivedUpload;
                 });
                 options.requestStream.on("error", (streamError: any) => {
                     this.log.error("Error encountered reading requestStream: " + streamError);


gejohnston (approved):

packages/imperative/src/io/src/IO.ts
```diff
@@ -274,6 +274,24 @@ export class IO {
             let prevByte = lastByte;
             for (let i = 0; i < original.length; i++) {
                 const currentByte = original[i];
+
+                if (isUploading) {
+                    // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+                    if (currentByte === 13) {
+                        // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+                        const nextByte = i + 1 < original.length ? original[i + 1] : undefined;
```
t1m0thyj, line 282: In order to properly handle chunk boundaries (where previous chunk ended with `\r` and current chunk begins with `\n`), should we check `prevByte` and `currentByte`, rather than `currentByte` and `nextByte`?

t1m0thyj (changes requested):
Thanks @traeok for the fix - tested on Windows and uploaded files look much better now 🎉  Have a concern about potential edge case

packages/imperative/src/io/src/IO.ts
traeok: Good point, I think the logic is a bit limited based on the existing design pattern for newline normalization in the SDKs - `processNewlines` does (mostly) stateless chunk processing and has no real knowledge of the previous chunk aside from the last byte (which is already uploaded in this case). But can look into refactoring AbstractRestClient and that function for a combined fix

traeok: Addressed in 901b568

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -399,13 +399,29 @@ export abstract class AbstractRestClient {
...
-                let heldByte: string;
+                let heldByte: Buffer;
                 options.requestStream.on("data", (data: Buffer) => {
                     this.log.debug("Writing data chunk of length %d from requestStream to clientRequest", data.byteLength);
+
+                    // If we held back a CR from the previous chunk, prepend it to current chunk
+                    if (this.mNormalizeRequestNewlines && heldByte != null) {
+                        data = Buffer.concat([heldByte, data]);
+                        heldByte = undefined;
+                    }
```
t1m0thyj: (nitpick) This could be consolidated into the `if` statement below:
```suggestion
                    if (this.mNormalizeRequestNewlines) {
                        if (heldByte != null) {
                            data = Buffer.concat([heldByte, data]);
                            heldByte = undefined;
                        }
                    ...
```

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -415,8 +431,11 @@ export abstract class AbstractRestClient {
                             this.mTask.percentComplete++;
                         }
                     }
-                    clientRequest.write(data);
-                    this.lastByteReceivedUpload = data[data.byteLength - 1];
+
+                    if (data.byteLength > 0) {
+                        clientRequest.write(data);
+                    }
+                    this.lastByteReceivedUpload = data.byteLength > 0 ? data[data.byteLength - 1] : this.lastByteReceivedUpload;
```
t1m0thyj: This could be consolidated into a single `if` statement:
```suggestion
                    if (data.byteLength > 0) {
                        clientRequest.write(data);
                        this.lastByteReceivedUpload = data[data.byteLength - 1];
                    }
```

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -399,13 +399,29 @@ export abstract class AbstractRestClient {
                 // if the user requested streaming write of data to the request,
                 // write the data chunk by chunk to the server
                 let bytesUploaded = 0;
-                let heldByte: string;
+                let heldByte: Buffer;
```
t1m0thyj, line 402: (nitpick) Changing from `string` to `Buffer` seems unnecessary but I don't see a problem with it 😋

PR #2618, opened 2025-10-05
Don't delete existing file if download fails

### LINKED ISSUE #2595: If downloading data set fails and file already exists in the path, it is deleted
**Describe the bug**

There is logic in the download data sets methods that if a download fails, it deletes the file at the same destination. 

I'm not really sure why this is the intended logic, because if a file already exists in the same destination (for example you are redownloading a file to get updates perhaps) and you get an error (let's say you've lost connection with the server), then the existing file will be deleted.

E.g.
https://github.com/zowe/zowe-cli/blob/2451323bcacdad7bc8e219a0809047b27680f04e/packages/zosfiles/src/methods/download/Download.ts#L168-L175

https://github.com/zowe/zowe-cli/blob/2451323bcacdad7bc8e219a0809047b27680f04e/packages/zosfiles/src/methods/download/Download.ts#L281-L289

If the API returns enough info to make it possible, then I think the intention is that it should delete partially downloaded files. That's what I'd expect anyway. 

**Describe your environment**

On a branch of Zowe Explorer main (3.3.0 SNAPSHOT) which is using zos-files 8.24.1

**Additional context**

This undesired behaviour actually happened to me mid testing so it's not just a pre-emptive thing from reading the code.

### LINKED ISSUE #2619: Various issues on Download.allDataSets
**Describe the bug**

Bugs:
1. Preserve original letter case option supplied does not preserve case (i.e. all caps) for directory names on download data sets matching/allDatasets command. The file names themselves are preserved though.
2. "4 data set(s) downloaded successfully to ./" `./` is technically correct I guess but it doesn't really make sense. It should just not show if --directory is not supplied to be consistent with all the other commands...
3. Sequential data set is not downloaded to the correct folder (if --directory is not supplied) and has wrong name. It should be next to the other PDSes but it is actually on the same level as the HLQ (see results).
4. When outputting member names ("Members:") in the dsorg status, is there a reason for the list being appended with a semi colon?

Image

Image

Image

Z02589.SPDS should be here under z02589 folder and be called just SDPS.txt

Image

**Describe your environment**

- Zowe CLI version installed (run command `zowe --version`): 8.27.1
- Installed plug-ins and their version numbers (run command `zowe plugins list --short`): N/A
- Node.js and NPM versions installed (run `node --version`, `npm --version`): v20.11.1, 10.8.3
- Environment variables in use: N/A
- Operating system and version: Win10 22H2
- Shell/terminal (bash, cmd, powershell, etc...): cmd

**Additional context**

I'm working on download API stuff anyway so I can probably take point on this (if you could assign me to the issue as well, thanks)

JWaters02:
**What It Does**

Fixes #2595 
Fixes #2619 (only issue 1 & 4)

**How to Test**

Run command to download a hekin chonker of a file (that takes a while to download) (successfully). Now run the command again but this time kill your network halfway through the download. See that the original file is no longer deleted.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/zosfiles/CHANGELOG.md
+++ b/packages/zosfiles/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to the Zowe z/OS files SDK package will be documented in thi
 
 - Enhancement: Added a download `destination` field to the API response for the `Download.ussFile` and `Download.dataSet` functions. [#2641](https://github.com/zowe/zowe-cli/pull/2641)
 - BugFix: Fixed an issue where if a file failed to download and an older version of it already existed in the destination, the old file was deleted. [#2618](https://github.com/zowe/zowe-cli/pull/2618)
-- BugFix: Fixed an issue where the preserve letter case option was not preserving the case in generated directory names in the `Download` `data-sets-matching` command. [#2618](https://github.com/zowe/zowe-cli/pull/2618)
+- BugFix: Fixed an issue where the preserve letter case option was not preserving the case in generated directory names in the `download data-sets-matching` command. [#2618](https://github.com/zowe/zowe-cli/pull/2618)
 
 ## `8.29.1`
 


packages/zosfiles/src/methods/download/Download.ts
```diff
@@ -271,18 +283,29 @@ export class Download {
                     extension = extension.replace(/^\.+/g, "");
                 }
 
+                const memberFilePath = posix.join(baseDir, fileName + IO.normalizeExtension(extension));
```
traeok, line 299: We should use the platform-specific implementation of path instead of `posix`. The current logic assumes that the path is in POSIX format, which may have unintended side effects for Windows clients.

JWaters02: I used it because I saw that other places in the code also use posix.path.join so should I also be changing those to path.join? except for the uss file bits

JWaters02: @traeok ?

traeok: I've addressed your requests on the other two PRs, checking this out now

traeok: I think its fine to keep this for now since we have this pattern elsewhere - I can file a separate issue to handle platform-specific paths

t1m0thyj (changes requested):
After merging changes from master branch, several unit tests are failing

traeok:
Blocking from merging until up-to-date and system tests are passing.
Updating the branch to be current with master should resolve most of the system tests that are currently failing on this PR.

JWaters02:
> I probably spent too much time trying to reproduce the original problems, only to realize that I needed to pass the overwrite flag. 😅 I can see that we do provide overwrite: true in the test. So for a next time, could we add those details to the issue? 🙏

I'm really sorry you wasted your time on this 😭 , please do reach out to me on slack or here if you are confused about anything as I am always monitoring. I will try to include more info next time.

> Also noticed some back and forth on item number 2 of https://github.com/zowe/zowe-cli/issues/2619, which was a bit confusing as well 😅

Apologies once more, I thought just updating the description of the PR was enough and that the overall PR file compare not showing "issue 2" related code not changing would be enough.

> Blocking from merging until up-to-date and system tests are passing.
Updating the branch to be current with master should resolve most of the system tests that are currently failing on this PR.

Hi, I don't work for a vendor so I have no capacity to run any system tests. How do I find out which tests are failing and how can I try to fix them without being able to test them myself? Thanks

traeok:
@JWaters02 no worries about running them on your end as I can run them internally. I tried running them against this PR but it needs to be updated with master to solve some of the known test failures. Once it's up to date with master I can re-run system tests and post the output here

packages/zosfiles/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ All notable changes to the Zowe z/OS files SDK package will be documented in thi
 ## Recent Changes
 
 - Enhancement: Added a download `destination` field to the API response for the `Download.ussFile` and `Download.dataSet` functions. [#2641](https://github.com/zowe/zowe-cli/pull/2641)
+- BugFix: Fixed an issue where if a file fails to download and one already exists in the destination, the old file is deleted. [#2618](https://github.com/zowe/zowe-cli/pull/2618)
```
```suggestion
- BugFix: Fixed an issue where if a file failed to download and an older version of it already existed in the destination, the old file was deleted. [#2618](https://github.com/zowe/zowe-cli/pull/2618)
```

packages/zosfiles/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ All notable changes to the Zowe z/OS files SDK package will be documented in thi
...
+- BugFix: Fixed an issue where the preserve letter case option was not preserving the case in generated directory names in the `Download` `data-sets-matching` command. [#2618](https://github.com/zowe/zowe-cli/pull/2618)
```
```suggestion
- BugFix: Fixed an issue where the preserve letter case option was not preserving the case in generated directory names in the `download data-sets-matching` command. [#2618](https://github.com/zowe/zowe-cli/pull/2618)
```

PR #2620, opened 2025-10-06
Fix overwrite option not being used in data set and uss file downloads

### LINKED ISSUE #2594: Download data set commands should support overwrite flag

I am implementing the download commands in Zowe Explorer with a user option to allow overwriting downloaded files if modified (for downloading a whole data set or all data sets or a uss directory). 

Additionally, for performance reasons, if you download a massive data set, let's say with 1k members, and it fails on file 900, instead of redownloading all of them, it should just download the last 100 (or just the files that failed if don't fail fast). But if user wants to always overwrite (in case any of the first 900 files are modified) then respect the option.

However I have found that all data set related commands do not support the existing overwrite option - only `ussDir` uses it. Existing downloaded files are always overwritten.

Ideally we would use the existing `overwrite` option.

However, this would be a breaking change because
- Current default behaviour: (overwrite is not used) download the file regardless of if it exists or not
- New default behaviour: (overwrite is false/not supplied) if the file already exists and has been modified, don't download the file

So actually we need to add a new option that has the opposite default of `overwrite`, let's say `skipExisting` or similar.

Then new default behaviour will be the same as before, but now if `skipExisting` is true, then check if file exists and if so, don't download the file.

Question for maintainers: If we are implementing a new option for all functions except `ussDir`, could we mark `overwrite` as deprecated, implement the new option in all places (including ussDir alongside `overwrite`) and then remove `overwrite` in next major release?

Just don't provide any overwrite option and warn that it will always overwrite except when downloading a uss directory?

OR

Implement overwrite functionality at the ZE level. 

This is not a good solution however as I would not be able to use the `downloadAllMembers` or `allDataSets` functions for their representive commands in ZE. 

Because I can check if the files are all there and overwrite is off, then don't do the download. But if one file is missing then the call to the `downloadAllMembers`/`allDataSets` functions would still redownload the entire directory with all the files overwritten. 

And I can't do both because of how the task is tracked so in that case I'd have to implement the method that downloads each indidivual file and recreate basically all the logic of the whole directory download but with overwriting supported.

JWaters02:
**What It Does**

Fixes #2594 

Added new interfaces `IDownloadAmResult` and `IDownloadAmResponse` to provide detailed results when downloading all members of a PDS.

Added new optional field `skippedExisting` to `IDownloadDsmResult`.

**How to Test**

Run zowe zos-files download all-members, all-members-matching, data-set, data-sets-matching and uss-file commands with and without overwrite option, deleting some files in the middle to test capabilities of each.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

There is one failing test, but I am stumped how to fix it...

### DIFF
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -3,7 +3,11 @@ All notable changes to the Zowe CLI package will be documented in this file.
 
 ## Recent Changes
 
-- Added support for `--overwrite` option to all Download methods. The default behavior is no longer to always overwrite existing files. [#2620](https://github.com/zowe/zowe-cli/pull/2620)
+- BugFix: Added support for `--overwrite` option to all Download methods. The default behavior is no longer to always overwrite existing files. [#2620](https://github.com/zowe/zowe-cli/pull/2620)
+
+## `8.29.0`
+
+- Enhancement: Added support for providing options to both default and custom credential managers. [#2601](https://github.com/zowe/zowe-cli/issues/2601) 
 
 ## `8.27.4`
 
--- a/.github/workflows/secrets-sdk.yml
+++ b/.github/workflows/secrets-sdk.yml
@@ -4,13 +4,10 @@ env:
   APP_NAME: keyring
   MACOSX_DEPLOYMENT_TARGET: 10.13
 on:
-  push:
-    paths:
-      - "packages/secrets/**"
-      - ".github/workflows/secrets-sdk.yml"
   pull_request:
     paths:
       - "packages/secrets/**"
+      - ".github/workflows/secrets-sdk.yml"
   workflow_dispatch:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -22,7 +19,6 @@ defaults:
 
 jobs:
   build:
-    if: (github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository) && !contains(github.event.head_commit.message, '[ci skip]')
     strategy:
       fail-fast: false
       matrix:
--- a/__tests__/__packages__/cli-test-utils/package.json
+++ b/__tests__/__packages__/cli-test-utils/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zowe/cli-test-utils",
-  "version": "8.28.0",
+  "version": "8.29.0",
   "description": "Test utilities package for Zowe CLI plug-ins",
   "author": "Zowe",
   "license": "EPL-2.0",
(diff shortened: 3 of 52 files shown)


JWaters02:
> @JWaters02 revisiting the error details, I think the issue may be a side effect of having a space in the path of your current working directory ("GitHub Projects"). Can you try renaming that folder or moving the repo to a path without a space to see if the issue goes away?

As yes good observation, you are right. I fixed this in #2627

t1m0thyj:
@JWaters02 Apologies for this, it should be re-opened now.

There wasn't a misclick - when a PR is merged, GitHub auto-closes issues/PRs that are mentioned with the "Fixes" or "Resolves" keyword. So the description of #2627 triggered it:
> Resolves https://github.com/zowe/zowe-cli/pull/2620#issuecomment-3394185671

JWaters02:
> @JWaters02 Apologies for this, it should be re-opened now.
> 
> There wasn't a misclick - when a PR is merged, GitHub auto-closes issues/PRs that are mentioned with the "Fixes" or "Resolves" keyword. So the description of #2627 triggered it:
> 
> > Resolves [#2620 (comment)](https://github.com/zowe/zowe-cli/pull/2620#issuecomment-3394185671)

Ahh okay I see thanks. I knew it did it for issues but had never seen it do it for PRs 😅

JWaters02:
> As for the second test, this seems like an issue specific to building the correct destination path on Linux and MacOS. This may be related to path normalization between platforms - you might want to double check that the platform-specific path import is used to ensure proper path generation, e.g. `import * as path from "path";` - I can provide more advice once I take a deeper look at the tests/logic.

@traeok not sure whats going wrong, mind taking another look again? I can't test on unix. apologies

packages/cli/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Added support for `--overwrite` option to all Download methods. The default behavior is no longer to always overwrite existing files. [#2620](https://github.com/zowe/zowe-cli/pull/2620)
```
traeok: (nitpick) missing `BugFix` prefix here

zFernand0 (commented):
Changes LGTM! 😋 
---
Still doing some testing!

Q: Since this is categorized as a bug... do we want to try and port these changes (including the new interfaces and option) back to V2?

---

UPDATE:
After discussion in standup, we will wait for someone to ask for this to be ported over to V2 🙏

packages/cli/CHANGELOG.md
zFernand0: resolved in c81f902

PR #2629, opened 2025-10-20
Feat: secure cert support

JillieBeanSim:
**What It Does**

macOS & Windows:
Core Features:
Certificate retrieval from macOS Keychain or Credential Manager via certAccount property
Profile validation recognizing certAccount/certKeyAccount
User prompting for certificate access authorization
Comprehensive error handling with fallback guidance

macOS Known Limitation:
Private key export blocked by macOS Security Framework (documented with workarounds)

Extra Documentation:
Certificate_Keychain_Limitations.md - Private key export limitations and 4 workaround options
Certificate_Access_Prompting.md - User prompting configuration guide

**How to Test**

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

Screenshot 2025-12-09 at 11 50 59 AM
Screenshot 2025-12-09 at 11 52 04 AM
if key not exportable:
Screenshot 2025-12-09 at 11 54 28 AM


packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -270,7 +279,7 @@ export class ConnectionPropsForSessCfg {
      *      ConnectionPropsForSessCfg.resolveSessCfgProps(sessCfg, cmdArgs);
      *      sessionToUse = new Session(sessCfg);
      */
-    public static resolveSessCfgProps<SessCfgType extends ISession>(
+    public static async resolveSessCfgProps<SessCfgType extends ISession>(
```
zFernand0: I know this is still in draft mode. But I believe this might constitute a breaking change 🤔

zFernand0 (commented):
LGTM! 😋 
---

Changes are looking good.
I left a comment about a possible breaking change 😢 

---

Regarding the integration test doc:
There isn't a ton of doc in terms of integration tests, but the `npm run pretest:integration` should get the test CLIs installed.
Do they pass locally for you?

JillieBeanSim:
Thanks @zFernand0 I was finally able to get them to run locally and duplicate failures shown in the pipeline, but only on a Windows git bash terminal.  When running on macOS and other Windows terminals the majority of the tests fail, in the other Windows terminals it's due to bash scripts.  The squad may want to look into javascript scripts as they should be able to run on multiple OSs and terminals. Also, I noticed the first step in the `npm run test:integration` script the pretest:integration is ran 😄 working on the fixes now.

packages/secrets/core/src/os/win.rs
```diff
@@ -349,3 +356,271 @@ pub fn find_credentials(
...
+
+    // For now, let's try to read the private key from the file system
+    // The P12 import should have made the key accessible, but extracting it through
+    // Windows CryptoAPI is complex. Let's try reading the .key file directly as a fallback
+    
+    // Try to read the private key from the original key file
+    let key_file_paths = [
+        "C:\\Users\\Administrator\\Desktop\\certs\\split\\BJSIMM.CLIENT.key",
+        "C:\\Users\\Administrator\\Desktop\\certs\\BJSIMM.CLIENT.key"
+    ];
```
traeok: This logic needs refactored so that the key file paths are not hardcoded paths (specific to your user)

packages/secrets/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe Secrets SDK package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added support for secure certificates in keyring for Windows and macOS. [#2325](https://github.com/zowe/zowe-cli/issues/2325)
```
traeok: We may want to mention here that Linux remains unsupported at this time. It is implied in the error messaging and the code but not explicitly documented, so developers might try to invoke the functions expecting them to work.

packages/secrets/core/src/os/mac/mod.rs
```diff
@@ -186,3 +186,198 @@ pub fn find_credentials(
...
+    // If certificate search failed, try password lookup with base64 decoding
+    // (for private keys stored as passwords)
+    let keychain = SecKeychain::default().unwrap();
+    match keychain.find_password(service.as_str(), account.as_str()) {
+        Ok((pw, _)) => {
+            let pw_str = String::from_utf8(pw.to_owned())?;
+            match base64::decode(&pw_str) {
+                Ok(bytes) => Ok(Some(bytes)),
+                Err(_) => Ok(None), // Not base64-encoded, treat as not found
+            }
```
traeok: https://github.com/marshallpierce/rust-base64/blob/13f4fe86e565b3a8ed9402d3b8b1bcf83ab9817c/src/decode.rs#L123-L130

`base64::decode` is deprecated, consider using the `base64::engine::general_purpose::STANDARD.decode` approach used in the Windows implementation

traeok (changes requested):
Thanks for working on this Billie, left a couple requests for my initial review. Going to manually test once the hardcoded path comment is resolved

docs/Certificate_Access_Prompting.md
```diff
@@ -0,0 +1,143 @@
...
+With `promptForCertAccess: true`, running a command like:
+```bash
+zowe zos-files list ds BJSIMM --zosmf-profile tvt
+```
+
+Will produce:
+```
+Zowe CLI would like to access certificate 'User BJSIMM on tvt4002' from your system keychain. Allow? [y/N]: y
+Zowe CLI would like to access private key 'User BJSIMM on tvt4002' from your system keychain. Allow? [y/N]: y
+```
```
traeok: While this prompting is nice, I am unable to reproduce on Windows. Using a certificate with certAccount set to "User traeok," I get prompted for my password instead of accessing the cert from the keychain. This happens whether I pass `cert-account` as a CLI option or setting `certAccount` under my z/OSMF profile properties.

Image

traeok: Image

Here's what my Certificate Manager looks like. Am I missing something?

JillieBeanSim: Hey @traeok when you imported your certificate make sure the key is exportable.  For it to prompt for password it seems the key wasn't exportable.  with the same cert files I have for other CLI cert options, here are the steps I followed to import the certificate on Windows
image

traeok: Thanks Billie, marking the key as exportable seemed to help. However I noticed that I'm still not prompted on Windows, it just lists data sets when providing `certAccount` in profile properties or `--cert-account` as the CLI parameter. Might be something OS or environment specific?

JillieBeanSim: what are you expecting to be prompted for?  I don't think Windows prompts users to access the credential manager like the macOS does even with the existing securely stored values.  I may need to update the md files to be more OS specific and rename the cert being referenced to a more generic name

traeok: I was expecting the "Zowe CLI would like to access certificate" Y/N prompt based off of what the doc says. I think it makes sense on Windows to not prompt though. I just noticed this while following that prompting doc as there aren't test steps in the PR, so I figured I'd start w/ the expected prompt behavior for testing.

JillieBeanSim: I pushed updates to the md files. let me know if anything else needs addressed in them please. thanks

JillieBeanSim: if the prompting that some OSs do for securely stored values is already documented we may just need to remove that md file and just keep the one that goes into the non-exportable key. let me know, thanks

packages/secrets/core/src/os/win.rs
```diff
@@ -349,3 +358,406 @@ pub fn find_credentials(
...
+    const CERT_SYSTEM_STORE_LOCAL_MACHINE: u32 = 0x00020000;
+    
+    // Try both Current User and Local Machine certificate stores  
+    let store_names = ["MY", "Root", "CA", "AddressBook"]; // AddressBook is "Other People" store
+    let store_locations = [
+        CERT_SYSTEM_STORE_CURRENT_USER,
+        CERT_SYSTEM_STORE_LOCAL_MACHINE,
+    ];
+
+    for &_store_location in &store_locations {
```
traeok, line 480: `_store_location` is unused here, is this expected? If so, we could remove the outer `for` loop. Otherwise the inner logic may need refactored to consider the different store locations.

packages/secrets/core/src/os/win.rs
```diff
@@ -349,3 +358,406 @@ pub fn find_credentials(
...
+                    CertEnumCertificatesInStore(h_store, std::ptr::null())
+                };
+            }
+
+            while !cert_context.is_null() {
+                // Try to get the private key for this certificate
+                if let Ok(Some(key_bytes)) = extract_private_key_from_cert(cert_context) {
+                    unsafe { CertCloseStore(h_store, 0) };
+                    return Ok(Some(key_bytes));
+                }
```
traeok, line 521: On Windows, if a specific match isn't found by subject, the logic falls back to enumerating all certificates. This makes it possible for an unrelated certificate to be returned just because it has a valid private key, which might not match the intended host. Is this intended behavior? I noticed with the macOS logic, if the specific account isn't found, it returns None.

Should we consider refining the certificate search or removing it entirely if it may return a mismatched certificate?

packages/secrets/core/src/os/unix.rs
```diff
@@ -219,3 +219,37 @@ pub fn find_credentials(
...
+/// Returns the certificate (decoded from base64) stored as the password for the given service/account.
+///
+/// - `service`: The service name that matches the certificate entry
+/// - `account`: The account/name for the certificate entry
+///
+/// Returns:
+/// - A `KeyringError` indicating that this feature is not supported on Linux/Unix
+///
+pub fn get_certificate(service: &String, account: &String, _optional: bool) -> Result<Option<Vec<u8>>, KeyringError> {
+    Err(KeyringError::Os(
```
traeok, line 233: Even though the API isn't implemented for Linux, I'm wondering if we should return `Ok(None)` when `optional` is `true` 🤔 

That way we can gracefully fallback if the request was optional under that OS. This is more from the perspective of secrets_core or the Secrets SDK being used directly by developers.

packages/secrets/core/src/os/mac/mod.rs
```diff
@@ -186,3 +187,219 @@ pub fn find_credentials(
...
+                            0, // flags - allow prompting
+                            std::ptr::null(),
+                            &mut exported_data
+                        );
+
+                        if export_status == 0 && !exported_data.is_null() {
+                            let data = CFData::wrap_under_create_rule(exported_data as *mut _);
+                            let mut bytes = Vec::new();
+                            bytes.extend_from_slice(data.bytes());
+                            return Ok(Some(bytes));
```
traeok, line 320: I noticed certificates are exported as PEM (expected), but private keys are exported as raw DER format. For Windows, however, the implementation exports both in PEM format. Should we match the export format for private keys with the Windows implementation?

packages/secrets/core/src/os/win.rs
```diff
@@ -349,3 +358,406 @@ pub fn find_credentials(
...
+        }
+    }
+
+    Ok(None)
+}
+
+/// Extract the private key from a certificate context and return it in PEM format
+fn extract_private_key_from_cert(cert_context: *const CERT_CONTEXT) -> Result<Option<Vec<u8>>, KeyringError> {
+    // Helper to wrap DER into PEM header/footer
+    fn der_to_pem(label: &str, der: &[u8]) -> Vec<u8> {
```
traeok, line 539: The `der_to_pem` helper could be reused/generalized to apply to both the Windows and macOS implementation (to have matching private key export formats)

docs/Certificate_Keychain_Limitations.md
```diff
@@ -0,0 +1,311 @@
...
+## Overview
+
+When using `certAccount` and `certKeyAccount` properties to authenticate with certificates stored in macOS Keychain, you may encounter an issue where the **certificate** is successfully retrieved but the **private key** cannot be exported. This document explains why this happens and provides several workarounds.
+
+## The Problem
+
+### Error Message
+
+```
+Private key export not supported - identity found but key cannot be exported for account 'User BJSIMM on tvt4002'
```
t1m0thyj, line 12: If we intend for these docs to be user-facing, can we use a generic username (and system name if applicable)?

```suggestion
Private key export not supported - identity found but key cannot be exported for account 'User IBMUSER on lpar1'
```

IMPLEMENTATION_STATUS.md
t1m0thyj, line 1: Thanks for including this file to explain the design and make review easier 🙂 I assume before the PR is merged, we'll want to either move it to the `docs` directory, or remove it if intended to be temporary?

packages/imperative/src/config/src/ProfileInfo.ts
```diff
@@ -522,7 +522,13 @@ export class ProfileInfo {
...
 
         // resolve the choices among various session config properties
-        ConnectionPropsForSessCfg.resolveSessCfgProps(sessCfg, cmdArgs, connOpts);
+        // Note: resolveSessCfgProps is async (may prompt or auto-store). To keep
+        // createSession synchronous (non-breaking API) we intentionally do not
+        // await it here. Callers that need the resolved/possibly-prompted
+        // properties should call the async APIs directly.
+        // Fire-and-forget the resolver so background work (prompts) can proceed
+        // without changing the synchronous contract.
+        void ConnectionPropsForSessCfg.resolveSessCfgProps(sessCfg, cmdArgs, connOpts);
```
t1m0thyj, line 531: Generally I think we prefer to avoid using `void` keyword on async functions where possible, but in this case I assume it's needed to avoid breaking changes as described in the comment?

Is it also correct that the only prompt that can happen here is a question asking the user to grant certificate access? So in worst case scenario if we don't wait for the method to resolve, then I assume we end up with a session that is missing certificate-related properties.

packages/imperative/src/rest/src/session/AuthOrder.ts
```diff
@@ -688,7 +731,191 @@ export class AuthOrder {
...
+            pushIf(certKeyAccountVal);
+            if (certAccountVal && !candidates.includes(certAccountVal)) pushIf(certAccountVal);
+        }
+
+        pushIf((sessCfg as any).profile);
+        pushIf((sessCfg as any).account);
+        return candidates;
+    }
+
+    private static writeCertToTempFile(sessCredName: string, certBuf: Buffer): string {
```
t1m0thyj, line 779: Writing the certs to a temp file makes me a bit uneasy, would prefer if we can avoid doing so. If we want to guarantee that certs are handled securely, then ideally we shouldn't store them on disk at all. And having to track these temp file paths in memory and dispose of the files later on adds extra overhead.

I assume the main reason for storing them in files was to avoid breaking changes, and fit the `ISession` type that expects `cert` and `certKey` to be file paths. Some alternative approaches that could work:
* Add optional properties to `ISession` for `certLoaded` and `certKeyLoaded` that contain the values of the certificates cached in memory.
  * **Pro:** As long as the properties are optional, they shouldn't cause a breaking change.
  * **Con:** It may be confusing to have duplicate `cert` properties on the session, but this could be alleviated by making `cert`/`certKey` and `certLoaded`/`certKeyLoaded` mutually exclusive.
* Change the types of `cert` and `certKey` properties on `ISession` from `string` to `string | Buffer` where a buffer contains loaded contents rather than a file path.
  * **Pro:** This approach may be familiar to developers - I believe it's similar to other packages like `zos-node-accessor` upload method.
  * **Con:** This may be considered too much of a breaking change, even if we keep the behavior for `string` case the same.

t1m0thyj (commented):
Thanks @JillieBeanSim for working on this! Left a few questions below. I plan to review further and do manual testing soon 😋

JillieBeanSim:
hey all, thanks for the reviews @t1m0thyj & @traeok. I have a request from customer to support der certs where the key wasn't set as exportable. So I will be looking into that and reworking the proposed solution.

Issue #2635, opened 2025-11-04
Incorrect auth type in session for extender profiles without authOrder after switching authentication


likhithanimma1:
**Describe the bug**

I switched the authentication method of my extender profile from basic to token-based. The team configuration file correctly reflects this change and shows the authentication as token-based.
However, my iSession still contains both user/password and tokenType details. I don’t have any authOrder configured.
When the session is retrieved through `imperative.ConnectionPropsForSessCfg.addPropsOrPrompt`, the primary authentication is incorrectly set to basic, and all my token-related details are placed under `_authCache: { availableCreds: [Object] }`. This causes issues when certain queries attempt to access token-related information.

**Expected and actual results**

When my configuration specifies token-based authentication for a profile, I expect the retrieved session to use the same primary authentication method as defined in the config, especially when no authOrder is set.

**Describe your environment**

- Zowe CLI version installed (run command `zowe --version`): v8.28.0
- Node.js and NPM versions installed (run `node --version`, `npm --version`): v22.18.0 , v10.9.3
- Operating system and version: mac

JTonda:
We noticed a related issue with the FTP vsce and we will investigate that first.  We will see if the modifications correct your problem.

Can you provide sample logic that reproduces this behavior from your extension?

likhithanimma1:
> We noticed a related issue with the FTP vsce and we will investigate that first. We will see if the modifications correct your problem.
> 
> Can you provide sample logic that reproduces this behavior from your extension?

After some investigation, we found that as an extender to Zowe Explorer, everything is working fine — we don’t see any issues since the auth order is being added even to the extender profiles. However, when we move over to the CLI, that’s where the issue occurs.
With the adoption of `authOrder` in CLI plugins that extend Zowe CLI, the issue is resolved. However, this introduces a breaking change for CLI extenders, as they are now required to adopt this change for their extensions to continue functioning as expected. To handle this, there should be a fallback mechanism to maintain the previous behavior if the extenders have not yet adopted the change.

gejohnston:
It was mentioned that the plugin may **NOT** be logging into APIML to get its token. The following authOrder assignment may work for such a plugin:

"authOrder": "bearer"

adam-wolfe:
@likhithanimma1, is this issue still a problem?

PR #2660, opened 2025-12-12
Add instructions when prompting for a token

gejohnston:
**What It Does**

When a command prompts for a token, display instructions on how to obtain a token.
Identified while investigating the issue [Incorrect auth type in session for extender profiles without authOrder after switching authentication 2635](https://github.com/zowe/zowe-cli/issues/2635)

**How to Test**

Login to your service. For example:
```
zowe auth login apiml
```

Logout of your service. For example:
```
zowe auth logut apiml
```

Run a command and see the new instructions. For example:
```
zowe zos-files list data-set "SYS1.PARMLIB*"

Some required connection properties have not been specified in your Zowe client configuration.
Therefore, you will be asked for the connection properties that are required to complete your command.

Your profile is configured to use a token for authentication. You can supply a token at the prompt.
Alternatively, terminate this command with Control-C and login to your service to obtain a token.
For example, the command 'zowe auth login apiml' will login to the API ML service. You can then
run your original command again, which will no longer prompt for a token.

Enter your tokenType for your service:
```

To see the latest token instructions as review feedback is incorporated, view the changes in this file: 
```
packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

This pull request is opened as a draft so that the review feedback about the displayed text can be incorporated before tests are written.

### DIFF
--- a/packages/imperative/__tests__/__integration__/imperative/__tests__/__integration__/cli/config/secure/cli.imperative-test-cli.config.secure.integration.subtest.ts
+++ b/packages/imperative/__tests__/__integration__/imperative/__tests__/__integration__/cli/config/secure/cli.imperative-test-cli.config.secure.integration.subtest.ts
@@ -218,4 +218,31 @@ describe("imperative-test-cli config secure", () => {
         // Check the securely stored JSON
         expect(securedValueJson).toEqual(expectedSecuredValueJson);
     });
+
+    it("should display instructions when a token must be secured", async () => {
+        runCliScript(__dirname + "/../init/__scripts__/init_config_prompt.sh", TEST_ENVIRONMENT.workingDir);
+        const baseProfile: IConfigProfile = {
+            type: "base",
+            properties: {
+                "host": "example.com",
+                "port": 443,
+                "user": "fakeUser",
+                "tokenType": "ExtenderTokenType"
+            },
+            secure: [
+                "tokenValue"
+            ]
+        };
+        runCliScript(__dirname + "/../set/__scripts__/set.sh", TEST_ENVIRONMENT.workingDir,
+            ["profiles", JSON.stringify({ project_base: baseProfile }), "--json"]);
+        const response = runCliScript(__dirname + "/__scripts__/secure_prompt.sh", TEST_ENVIRONMENT.workingDir, [], {
+            IMPERATIVE_TEST_CLI_SECURE_VALUES: "anotherFakeValue"
+        });
+
+        expect(response.stdout.toString()).toContain("A profile requires a token of type 'ExtenderTokenType' to be securely stored.");
+        expect(response.stdout.toString()).toContain("Supply a token at the prompt. Or press ENTER at the prompt and later log in to");
+        expect(response.stdout.toString()).toContain("your service for a token.");
+        expect(response.stderr.toString()).toEqual("");
+        expect(response.status).toEqual(0);
+    });
 });


packages/imperative/src/rest/__tests__/session/ConnectionPropsForSessCfg.unit.test.ts
```diff
@@ -2042,6 +2042,55 @@ describe("ConnectionPropsForSessCfg tests", () => {
...
+
+            /* Pretend that we do not have any V1 profiles.
+             * onlyV1ProfilesExist is a getter property, so mock the property.
+             */
+            Object.defineProperty(ConfigUtils, "onlyV1ProfilesExist", {
+                configurable: true,
+                get: jest.fn(() => {
+                    return false;
+                }),
+            });
```
traeok: (nitpick) Consider restoring this property back to its original value after the test passes - while the mocked property doesn't currently affect the results of other tests in the suite, restoring it prevents the mock from affecting future tests in this file.

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -405,10 +406,33 @@ export class ConnectionPropsForSessCfg {
...
 
             for (const propNm of promptForValues) {
                 const sessPropNm = AuthOrder.getPropNmFor(propNm, PropUse.IN_SESS);
                 const cfgPropNm = AuthOrder.getPropNmFor(propNm, PropUse.IN_CFG);
+                if ((propNm === "tokenType" || propNm === "tokenValue") && connOpts.parms?.response.console.log) {
+                    // Typically users will not know how to type in a token.
+                    // Add extra instructions for obtaining a token and only display them once.
+                    if (tokenInstructions.length === 0) {
+                        // instructions have not yet been displayed
+                        if (somePromptHasOccurred) {
```
traeok: (nitpick) `somePromptHasOccurred` is technically redundant since you can infer whether a prompt has occurred based on whether `answers` has at least one key. But, I think its fine to keep it for readability and as an explicit flag.

traeok (approved):
Thanks Gene, changes LGTM. I thought it was interesting that there is a message shown for the default z/OSMF profile when running `zowe auth login`, but it is a valid comment so I don't think it needs changed here. It just caught me off guard because I was logging into the "red" server (production instance) and saw a comment about the default z/OSMF profile from my "blue" server (profile does not use API ML). Just something that others might run into so I wanted to mention it for awareness.

Image

t1m0thyj (approved):

packages/imperative/src/rest/__tests__/session/ConnectionPropsForSessCfg.unit.test.ts
awharn: Alternatively, consider using `jest.replaceProperty` instead.

gejohnston: This was a problem that previously existed in these tests. My new test propagated the problem. I set a "default" state in the beforeEach section.

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
gejohnston: Since I made other changes, I removed somePromptHasOccurred in favor of "answers" to reduce clutter.

PR #1819, opened 2023-10-05
Fix Newline Normalization on Upload (V1)

awharn:
**What It Does**
Resolves #1815 in Zowe CLI V2

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] added/updated automated tests
- [x] updated the changelog
- [x] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

**Additional Comments**

PR #2665, opened 2026-01-02
fix(imperative): Remove pacote dep and replace with `npm pack`

t1m0thyj:
**What It Does**

Removes the large dependency `pacote` (~10MB including its transitive deps), which was used in Imperative to determine package name/version when a plug-in is installed from non-registry source (e.g. a TGZ or local folder).

This PR replaces `pacote` with the `npm pack` command in dry-run mode. The JSON output of this command has kept the same format for a long time (since [npm 7](https://docs.npmjs.com/cli/v7/commands/npm-pack)).

**How to Test**

Run `zowe plugins install <pkgSpec>` where `<pkgSpec>` is a TGZ file or path to local package.

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (`#1819`)

**Additional Comments**

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -12706,10 +12706,10 @@
         "@zowe/zos-uss-for-zowe-sdk": "8.29.8",
         "@zowe/zos-workflows-for-zowe-sdk": "8.29.8",
         "@zowe/zosmf-for-zowe-sdk": "8.29.8",
-        "find-process": "1.4.7",
-        "lodash": "4.17.21",
-        "minimatch": "9.0.5",
-        "tar": "7.5.3"
+        "find-process": "^1.4.7",
+        "lodash": "^4.17.21",
+        "minimatch": "^9.0.5",
+        "tar": "^7.5.4"
       },
       "bin": {
         "zowe": "lib/main.js"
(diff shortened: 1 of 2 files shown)


packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
```diff
@@ -95,15 +95,22 @@ export function installPackages(npmPackage: string, npmArgs: INpmInstallArgs, ve
  * Fetch name and version of NPM package that was installed
  * @param pkgSpec The package name as specified on NPM install
  */
-export async function getPackageInfo(pkgSpec: string): Promise<{ name: string, version: string }> {
+export function getPackageInfo(pkgSpec: string): { name: string, version: string, [key: string]: any } {
```
traeok: (nit) Consider using `unknown` instead of `any` when labeling types, so that other keys in the object are not used without safely casting them first.

traeok (approved):
I second Fernando's comments, I think this is a safe change and I'm a fan of how much it reduces our dependency tree.

Going to approve even though its in draft as the changes make sense to me. Thanks Timothy!

packages/cli/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Updated `tar` dependency to resolve technical currency. [#2665](https://github.com/zowe/zowe-cli/pull/2665)
```
ATorrise, line 6: 💯

PR #2672, opened 2026-01-26
awharn/enhancement/support_config_env_variables

awharn:
**What It Does**

Add support for environment variables in the team configuration files.
Update daemon dependencies to resolve audit failures.

**How to Test**

New functionality:
- Export/set an environment variable (i.e. `export TEST_VAR=<username>`)
- Use that variable in a config (i.e. `"user": "$TEST_VAR"`)
- Run a command that uses the property
- Observe the property is replaced with the environment variable

Advanced new functionality:
- Use that variable in a config in a complex variable (i.e. `"user": "${TEST_VAR}01"`)
- Run a command that uses the property
- Observe the text in the braces is replaced with the environment variable

Regression tests:
- Unset the environment variable
- Use that old variable in a config (i.e. `"user": "$OMEUSER"`)
- Run a command that uses the property
- Observe the property is treated as a string literal

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (provide build number if applicable) [1847]

**Additional Comments**

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
+- Enhancement: Added an optional `scrtData` object to the ISession interface for a Session object. The `scrtData` is transmitted as a new HTTP header named `Zowe-SCRT-client-feature` to the targeted REST service. [#2678](https://github.com/zowe/zowe-cli/pull/2678)
 - Enhancement: Updated config files to support environment variables in `$VARIABLE` and `${VARIABLE}` formats. [#2672](https://github.com/zowe/zowe-cli/pull/2672)
   - Properties that contain a valid environment variable cannot be set or changed with the `Config.set` function. They can only be updated in an editor.
   - Properties that map to an unset environment variable are treated as a literal string.
--- a/packages/imperative/src/rest/__tests__/client/AbstractRestClient.unit.test.ts
+++ b/packages/imperative/src/rest/__tests__/client/AbstractRestClient.unit.test.ts
@@ -1975,6 +1975,564 @@ describe("AbstractRestClient tests", () => {
             });
         });
 
+        describe("addScrtHeader", () => {
+            const restSession = new Session({
+                hostname: "FakeHostName",
+                scrtData: {
+                    "featureName": "FakeFeatureName",
+                    "productId": "fakeProductId",
+                    "productVersion": "fakeProductVersion"
+                }
+            });
+            const restReqOpts = {
+                "resource": "/NotZosmf/url",
+                "reqHeaders": [] as any
+            };
+            const privateRestClient = new RestClient(restSession) as any;
+            const impLogger = Logger.getImperativeLogger();
+            let impErrorLoggerSpy: jest.SpyInstance;
+
+            beforeEach(() => {
+                jest.clearAllMocks();
+                (restReqOpts as any).resource = "/NotZosmf/url";
+                (restReqOpts as any).reqHeaders = [];
+                jest.spyOn(Logger, "getImperativeLogger").mockReturnValue(impLogger);
+                impErrorLoggerSpy = jest.spyOn(impLogger, "error");
+            });
+
+            it("should NOT add a header for a zosmf request", () => {
+                (restReqOpts as any).resource = "/zosmf/url";
+                const impDebugLoggerSpy = jest.spyOn(impLogger, "debug");
+
+                privateRestClient.addScrtHeader(restReqOpts);
+
+                expect(restReqOpts.reqHeaders.length).toBe(0);
+                expect(impDebugLoggerSpy).toHaveBeenCalledWith(
+                    "addScrtHeader: SCRT headers are NOT sent to z/OSMF."
+                );
+            });
+
+            it("should log an error if it fails to create a header", () => {
+                const isScrtValidSpy = jest.spyOn(privateRestClient, "isScrtValid").mockReturnValue(false);
+                const formScrtHeaderValSpy = jest.spyOn(privateRestClient, "formScrtHeaderVal");
+
+                privateRestClient.addScrtHeader(restReqOpts);
+
+                expect(restReqOpts.reqHeaders.length).toBe(0);
+                expect(formScrtHeaderValSpy).toHaveBeenCalled();
+                expect(isScrtValidSpy).toHaveBeenCalled();
+                expect(impErrorLoggerSpy).toHaveBeenCalledWith(
+                    "formScrtHeaderVal: No SCRT header is created when SCRT data is invalid."
+                );
+            });
+
+            it("should add a header when when SCRT properties are available", () => {
+                privateRestClient["mSession"]["mISession"].scrtData = {
+                    featureName: "Fake Feature name",
+                    productId: 'FkProdId',
+                    productVersion: '12.34.56'
+                };
+                const formScrtHeaderValSpy = jest.spyOn(privateRestClient, "formScrtHeaderVal");
+
+                privateRestClient.addScrtHeader(restReqOpts);
+
+                expect(formScrtHeaderValSpy).toHaveBeenCalled();
+                expect(restReqOpts.reqHeaders.length).toBe(1);
+                expect(restReqOpts.reqHeaders[0]).toEqual(
+                    { "Zowe-SCRT-client-feature":   "featureName='Fake Feature name', " +
+                                                    "productId='FkProdId', productVersion='12.34.56'"
+                    }
+                );
+            });
+
+            it("should NOT included non SCRT properties in the header", () => {
+                privateRestClient["mSession"]["mISession"].scrtData = {
+                    featureName: "Valid SCRT property of Feature name",
+                    productId: 'ValPrdId',
+                    productVersion: '11.22.33',
+                    nonScrtProp1: "nonScrtVal1",
+                    nonScrtProp2: "nonScrtVal2",
+                    nonScrtProp3: "nonScrtVal3",
+                };
+                const formScrtHeaderValSpy = jest.spyOn(privateRestClient, "formScrtHeaderVal");
+
+                privateRestClient.addScrtHeader(restReqOpts);
+
+                expect(formScrtHeaderValSpy).toHaveBeenCalled();
+                expect(restReqOpts.reqHeaders.length).toBe(1);
+                expect(restReqOpts.reqHeaders[0]).toEqual(
+                    {
+                        "Zowe-SCRT-client-feature":
+                            "featureName='Valid SCRT property of Feature name', " +
+                            "productId='ValPrdId', productVersion='11.22.33'"
+                    }
+                );
+                expect(impErrorLoggerSpy).toHaveBeenCalledTimes(3);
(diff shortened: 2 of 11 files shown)


packages/imperative/src/config/__tests__/ConfigEnvironmentVariables.unit.test.ts
```diff
@@ -0,0 +1,418 @@
...
+            ConfigEnvironmentVariables.replaceEnvironmentVariablesInConfigLayer(configObject, fakeConfig as any, fakeLayer as any);
+            expect(fakeConfig.mEnvVarManaged?.length).toEqual(1);
+            expect(replaceEnvironmentVariablesInStringSpy).toHaveBeenCalledTimes(1);
+            expect(coercePropValueSpy).toHaveBeenCalledTimes(1);
+            expect(replaceEnvironmentVariablesInConfigLayerSpy).toHaveBeenCalledTimes(3);
+            expect(configObject.fake.properties.user).toEqual("notarealuser");
+            expect(fakeConfig.mEnvVarManaged?.[0]).toEqual({
+                global: false,
+                user: false,
+                propPath: ".fake.properties.user",
```
t1m0thyj: Discussed offline and this doesn't seem like a valid property path since it begins with a `.`. We may want to handle empty paths differently and throw an error for them?

awharn: I altered the behavior so that if the path provided is empty, the propPath will begin with the key on the provided object. I will push these changes as soon as integration tests pass locally.

anaxceron (approved):

packages/imperative/src/config/src/Config.ts
```diff
@@ -117,6 +119,13 @@ export class Config {
         secure: ConfigSecure
     };
 
+    /**
+     * @internal
+     * A list of config properties managed by environment variables.
+     * These should not be saved if the config changes.
+     */
+    public mEnvVarManaged: IConfigEnvVarManaged[] = [];
```
traeok, line 127: From the perspective of an extender: how can the extender find out which properties are managed by environment variables? Currently this list is internal to Imperative, so this means that extenders would have to introduce error handling around the use of `Config.set` to gracefully handle the properties that are managed by env. vars. This could be perceived as breaking since the changes in `Config.set` introduce a new case for an exception to be thrown.

traeok: Additional comment: It seems unusual to make `mEnvVarManaged` public but `internal`. Is this to work around an access issue internally? Would there be benefit in an internal getter and making the array private/protected?

packages/imperative/src/config/src/ConfigEnvironmentVariables.ts
```diff
@@ -0,0 +1,160 @@
...
+            let propPath: string;
+
+            if (path) { propPath = path + "." + key; }
+            else { propPath = key; }
+
+            if (typeof obj[key] === 'object' && !Array.isArray(obj[key]) && obj[key] != null) {
+                this.replaceEnvironmentVariablesInConfigLayer(obj[key], config, layer, propPath);
+            } else if (typeof obj[key] == 'string' && obj[key].includes("$") &&
+            this.findEnvironmentVariables(obj[key]).size > 0) {
+                const replacementValue = ConfigUtils.coercePropValue(
```
traeok, line 41: Should we consider providing the type information from the schema to this function, so the type isn't guessed for each prop?

I think this may have come up in the past around coercing properties: if you would have an env var `$MYPASSWORD="false"`, this would be coerced to a boolean `false`. If this behavior is expected, then please disregard this comment.

packages/imperative/src/config/src/ConfigEnvironmentVariables.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * @internal
+     * Replace environment variables in a config layer with their real values
+     * @param obj The config object to iterate over
+     * @param config The overall config class to modify
+     * @param layer The config layer being operated on
+     * @param path The current high level path to the properties being evaluated
+     */
+    public static replaceEnvironmentVariablesInConfigLayer(obj: any, config: Config, layer: IConfigLayer, path: string = "") {
```
traeok, line 30: There appears to be a limitation with properties meant to have array values, in that these properties are skipped instead of being handled as a list of strings. Is it expected for these to be unsupported? If so, should we document the limitation so both plug-in developers and end users are aware?

I know this is uncommon but someone might try this after the env. var support is released, so I think its good to be explicit.

packages/imperative/src/config/src/Config.ts
```diff
@@ -437,6 +466,12 @@ export class Config {
         opts = opts || {};
 
         const layer = this.layerActive();
+        this.mEnvVarManaged.forEach((value) => {
+            if (value.global == layer.global && value.user == layer.user && value.propPath == propertyPath) {
+                throw new ImperativeError({msg: `The property ${propertyPath} is managed by environment variables and cannot be set.`});
+            }
+        });
```
traeok, line 481: As mentioned in a previous comment, this introduces a new scenario in which `Config.set` can throw an error.
Could this be considered breaking?

I think would be considered less breaking if callers had a way of checking whether a property is managed by an environment variable. But as it stands this isn't possible unless you access the `mEnvVarManaged` array marked as internal.

traeok (commented):
Thanks for working on this Andrew, left a few suggestions and questions around the current implementation.

Here are a couple items I wanted to mention, but didn't really fit as a comment on specific code:

---
There are now two different mechanisms for environment variable support:

- New feature (this PR): `$VAR / ${VAR}` syntax in config files, resolved at load time, tracked via `mEnvVarManaged`
- Existing feature: `ZOWE_OPT_*` environment variables handled in `ProfileInfo.overrideWithEnv()`, reported with `ProfLocType.ENV`

These serve different purposes but the naming and behavior could be confusing:
- The new feature replaces values in the config at load time
- The existing feature overrides values after profile merging (higher precedence)

Consider documenting the distinction clearly, as this affects the order of precedence. Per the existing docs, env vars override service profiles.

---
We have a doc `Using Team Configuration.md` in the `docs` folder of the repo. Should this be updated to show the new environment variable syntax? The doc seems generally out of date, so if we don't add it here I think we should at least document this new behavior in Zowe Docs.

packages/imperative/src/config/src/Config.ts
awharn: I thought we decided as a team that we did want it to throw an error if we want to overwrite a property that is provided by an environment variable. We can make a getter that allows extenders to view the mEnvVarManaged array. We could also make the behavior opt-in, but then we will need to make a change in ZE to get the behavior we want.

packages/imperative/src/config/src/ConfigEnvironmentVariables.ts
awharn: Since the intention for this feature is to allow individuals in web based editors to store credentials securely using environment variables, I don't know that we need array support for the first iteration of this feature. However, I did not explicitly document that, so it can be added to the documentation.

packages/imperative/src/config/src/ConfigEnvironmentVariables.ts
awharn: I don't know that the Config is aware of the schema information. I was under the impression that only the ProfileInfo APIs were.

packages/imperative/src/config/src/Config.ts
awharn: The ConfigEnvironmentVariables class needs access to that list, since we wanted to keep that code in a separate file. The only way to allow access to that class, without also allowing access to everyone else, was to mark it as public and internal.

See my comment below about the breaking change and handling `Config.set`.

packages/imperative/src/config/src/Config.ts
traeok: I think we want to throw an error, but I don't think its clear to extenders as to what properties are managed by env vars. If they know the variable is managed with an env var, they can skip the `Config.set` call altogether. Does that clear up my suggestion?

packages/imperative/src/config/src/ConfigEnvironmentVariables.ts
traeok: Gotcha, if we are limited by what the Config class currently has access to, then this could be considered a separate issue.

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Updated config files to support environment variables in `$VARIABLE` and `${VARIABLE}` formats.
+  - Properties that contain a valid environment variable cannot be set or changed with `Config.set`. They can only be updated in an editor.
```
anaxceron: sorry, andrew, was re-reading this and thought we should explain what `Config.set` is: changed with the `Config.set` class/property/cat/hat/whatever

awharn: Updated.

anaxceron (approved):
Looks good @awharn, thank you!

packages/imperative/src/config/src/ConfigEnvironmentVariables.ts
traeok: Deferring this to a future enhancement

packages/imperative/src/config/src/Config.ts
traeok: Extenders now have access to the variables managed by the environment, resolving my above concern

packages/imperative/src/config/src/ConfigEnvironmentVariables.ts
traeok: Doc was added 😋 thanks

packages/imperative/src/config/src/ConfigEnvironmentVariables.ts
```diff
@@ -0,0 +1,160 @@
...
+        return potentialEnvironmentVariables;
+    }
+
+    /**
+     * Find and replace all known environment variables in a string
+     * @param candidate The string to replace environment variables in
+     * @returns {string} The string with the environment variables replaced with their values
+     */
+    private static replaceEnvironmentVariablesInString(candidate: string): string {
+        let modifiedString = candidate;
```
ATorrise, line 117: replaceEnvironmentVariablesInString is called multiple times in the forEach loop of replaceEnvironmentVariablesInConfigLayer, the regex search position seems to need reset before each call 

"Note that the [lastIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex) property will not be reset when searching a different string, it will start its search at its existing [lastIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex)."

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#finding_successive_matches 

        this.simpleEnvironmentVariableRegexGlobal.lastIndex = 0;
        this.complexEnvironmentVariableRegexGlobal.lastIndex = 0;

awharn: Should be fixed in 7291a19

PR #2678, opened 2026-02-04
Transmit SCRT via HTTP header

gejohnston:
**What It Does**

Added a new optional 'scrtData' object to the session.ISession interface. Extenders can place SCRT properties into the scrtData object. When scrtData is present in a session, a new HTTP header named `Zowe-SCRT-client-feature` is added to any REST request sent using that session. The intention is that the receiving REST service will record the SCRT data.

SCRT properties can also be supplied in an environment variable named `ZOWE_SCRT_CLIENT_FEATURE`. This environment variable is only intended to be used by orchestrator apps (like Ansible) which do not make direct HTTP requests of REST services. If SCRT data is specified in the environment variable, any SCRT data in the session is ignored so that multiple records of SCRT data is not produced.

- No SCRT header is added if the target service is `zosmf`. 
  - The target consumer is a client extender who is sending REST requests to an application-specific mainframe REST service.
- SCRT properties are validated for format and length.

**How to Test**

Since Zowe CLI and Zowe Explorer only target z/OSMF, and no SCRT header is created for z/OSMF, new functionality cannot be demonstrated by running those apps. Neither app should see any difference in behavior as a result of these SCRT changes.

Only by adding temporary logic to Zowe CLI or Zowe Explorer can you display a confirmation that an SCRT header is being created. Therefore, independent manual testing may not be justified.

- In a command handler (like ZosFilesBaseHandler in the CLI) manually add SCRT data to a session.
- In AbstractRestClient.addScrtHeader, replace the zosmfUrlRegEx to something that does **NOT** match a zosmf URL.
- In AbstractRestClient.request, before calling https.request, write an imperative.log error message which displays buildOptions.headers.
- Rebuild the app.
- Run an operation. In the example above, the operation might be a `zowe zos-files list data-set` command.
- You will see an imperative.log message containing the SCRT header, for example:
  
  ```json
  {
    "Authorization": "Basic NotRealBase64Data==",
    "Accept-Encoding": "gzip",
    "X-IBM-Max-Items": "0",
    "Zowe-SCRT-client-feature": "featureName='Session Feature Name', productId='SesPrdId', productVersion='11.22.33'",
    "X-CSRF-ZOSMF-HEADER": true
  }
  ```

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/src/rest/__tests__/client/AbstractRestClient.unit.test.ts
+++ b/packages/imperative/src/rest/__tests__/client/AbstractRestClient.unit.test.ts
@@ -2173,11 +2173,10 @@ describe("AbstractRestClient tests", () => {
                 });
                 const privateRestClient = new RestClient(restSession) as any;
 
-                const isScrtValidSpy = jest.spyOn(privateRestClient, "isScrtValid").mockReturnValue(true);
-
                 const impLogger = Logger.getImperativeLogger();
                 const impErrorLoggerSpy = jest.spyOn(impLogger, "error");
                 jest.spyOn(Logger, "getImperativeLogger").mockReturnValue(impLogger);
+                jest.spyOn(privateRestClient, "isScrtValid").mockReturnValue(true);
 
                 const scrtHeaderVal = privateRestClient.formScrtHeaderVal({});
 


packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added an optional scrtData object to ISession. The scrtData is transmitted as a new new HTTP header named `Zowe-SCRT-client-feature` to the targeted REST service. [#2678](https://github.com/zowe/zowe-cli/pull/2678)
```
anaxceron: - we should add backticks to scrtData
- we should explain what ISession is: the ISession hat/cat/property/file/whatever

prototypes/packages/imperative/src/utilities/src/Readme.md
```diff
@@ -0,0 +1,11 @@
+
+
+    AsyncMutex.ts 
+
+The AsyncMutex class is a copy of the same class from ZRS. At one point it was needed for a feature which was later refactored to not need AsyncMutex. If AsyncMutex is needed in Imperative in the future, this file could be moved into the utilities/src directory, where it could be used by both Imperative and ZRS. The copy in ZRS could then be removed to avoid code duplication.
```
anaxceron:
```suggestion
The `AsyncMutex` class is a copy of the same class from ZRS. At one point it was needed for a feature which was later refactored to not need `AsyncMutex`. If `AsyncMutex` is needed in Imperative in the future, this file could be moved into the `utilities/src` directory, where it could be used by both Imperative and ZRS. The copy in ZRS could then be removed to avoid code duplication.
```

prototypes/packages/imperative/src/utilities/src/Readme.md
```diff
@@ -0,0 +1,11 @@
...
+
+
+
+    isVsCodeApp_for_ProcessUtils.ts 
+
+A function named isVsCodeApp( ) was added to the ProcessUtils class in support of a feature which was later refactored to not need isVsCodeApp. The isVsCodeApp function was operational. If isVsCodeApp is needed in the future, the code from this file could be pasted into ProcessUtils.ts
```
anaxceron:
```suggestion
A function named `isVsCodeApp( )` was added to the `ProcessUtils` class in support of a feature which was later refactored to not need `isVsCodeApp`. The `isVsCodeApp` function was operational. If `isVsCodeApp` is needed in the future, the code from this file could be pasted into `ProcessUtils.ts`.
```

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added an optional `scrtData` object to the ISession interface for a Session object. The `scrtData` is transmitted as a new new HTTP header named `Zowe-SCRT-client-feature` to the targeted REST service. [#2678](https://github.com/zowe/zowe-cli/pull/2678)
```
traeok: There's an extra `"new"` in the changelog entry:


```suggestion
- Enhancement: Added an optional `scrtData` object to the ISession interface for a Session object. The `scrtData` is transmitted as a new HTTP header named `Zowe-SCRT-client-feature` to the targeted REST service. [#2678](https://github.com/zowe/zowe-cli/pull/2678)
```

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -657,6 +660,246 @@ export abstract class AbstractRestClient {
...
+     *      A string containing a properly formatted value for a Zowe-SCRT-client-feature header.
+     *      If no SCRT data has been set, null is returned.
+     */
+    private formScrtHeaderVal(scrtData: IScrtData): string | null {
+        const header = Object.entries(scrtData)
+            .map(([k, v]) => `${k}='${v}'`)
+            .join(", ");
+
+        return header;
+    }
```
traeok, line 794: This function will never return `null` since `Object.entries(...).map(...).join()` would return an empty string at minimum:

```node
> const o = {}
undefined
> Object.entries(o).map(([k, v]) => `${k}='${v}'`).join(", ")
''
```

Consider dropping `| null` from the return type or handling the edge case where `header` is an empty string:

```suggestion
    private formScrtHeaderVal(scrtData: IScrtData): string {
        const header = Object.entries(scrtData)
            .map(([k, v]) => `${k}='${v}'`)
            .join(", ");

        return header;
    }
```

Another thing to consider is that this function does not sanitize single quotes. Someone could pass a feature name w/ a single quote, causing the header to be malformed. We may not expect devs to pass a single quote in a feature name, but if the feature name does not have rules around single quotes, we might want to consider sanitizing them.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -657,6 +660,246 @@ export abstract class AbstractRestClient {
...
+
+    /**
+     * Validates the supplied SCRT data.
+     *
+     * @param {IScrtData} scrtData - The data items required for SCRT reporting
+     *
+     * @returns {boolean} True if scrtData contains valid data. False otherwise.
+     */
+    private  isScrtValid(scrtData: IScrtData): boolean {
```
traeok:
```suggestion
    private isScrtValid(scrtData: IScrtData): boolean {
```

traeok (commented):
Thanks Gene and Pujal for your work on this! Left a couple suggestions, still reviewing but code looks great so far 👍

packages/imperative/src/rest/src/client/AbstractRestClient.ts
gejohnston: The scrtData was validated before formScrtHeaderVal() was called, which would have prevented formScrtHeaderVal from being called if the ScrtData was empty, null, etc. Even so, our logic becomes much more clear by moving the call to isScrtValid() into formScrtHeaderVal. The formScrtHeaderVal function then has a real reason to return null when isScrtValid fails. Good suggestion.

I also altered logic to escape any un-escaped quotes in the SCRT property values. Another good thing to do.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -657,6 +660,276 @@ export abstract class AbstractRestClient {
...
+        const scrtEnvNm = "ZOWE_SCRT_CLIENT_FEATURE";
+        let scrtData: IScrtData = null;
+
+        const scrtStr = process.env[scrtEnvNm];
+        if (!TextUtils.hasNonBlankValue(scrtStr)) {
+            return null;
+        }
+
+        // parse featureName from the env var value
+        const enclosingQuote = "['|\"]";
```
ATorrise: const enclosingQuote = "['\"]";  -> change ['|"] to ['\"]
i think the intent is to decipher between single or double quotes but | is being seen as a literal character here rather than an operational character

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -657,6 +660,276 @@ export abstract class AbstractRestClient {
...
+        if (!this.isScrtValid(scrtData)) {
+            Logger.getImperativeLogger().error(
+                `${funName} No SCRT header is created when SCRT data is invalid.`
+            );
+            return null;
+        }
+
+        // escape any un-escaped quotes in SCRT properties
+        let scrtHeaderVal: string = "";
+        const unEscapedQuotesRegEx = /([^\\])(["'])/g;
```
ATorrise: the goal for this regex is to add a backslash before any unescaped quote? i think this regex needs to be changed :

### Before: `/([^\\])(["'])/g` 

| Input | Result |
|-------|--------|
| `it's`  | `it\'s` ✓ |
| `'hello` | `'hello` ✗ |
| `a"b` |  `a\"b` ✓ |

### Proposed After: `/(^|[^\\])(["'])/g`

| Input |  Result |
|-------|---------|
| `it's`  | `it\'s` ✓ |
| `'hello`  | `\'hello` ✓ |
| `a"b`  | `a\"b` ✓ |

### Edge case: `''` (two quotes in a row)

| Input | Regex  | Result |
|-------|---------|--------|
| `''` | Before |  `''` ✗ |
| `''` | After | `\''` ✓ (first escaped) |

packages/imperative/src/rest/src/client/AbstractRestClient.ts
gejohnston: It is weird that I left the bar | character in this RegEx, since the very next line has a very similar RegEx WITHOUT the bar. Thanks for catching this.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
gejohnston: It is really helpful that you considered edge cases that I overlooked. It made me revisit the approach. I think that the original complex RegEx (which was attempting to match an escape character \ before a quote) was not warranted. Programmers might put escape characters in their property values within their source code. However, I believe that by the time that this code reads a property, the escapes have already been removed from the data itself by the interpreter.

It appears that this code can match any quote and always place an escape before a quote. This approach will escape all of the edge cases: a quote at the beginning, multiple quotes in a row, as well as simple quotes in the middle of the text.

This code will then generate a string into a header with the quotes escaped. The server program that reads the text from the header will again see no escape chracters since the escapes will have already been removed as the header string is interpreted into data.

We will not get final proof of this approach until we integrate our client with a server.
Thanks again.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -657,6 +660,275 @@
...
+            Logger.getImperativeLogger().error(
+                `${funName} No SCRT header is created when SCRT data is invalid.`
+            );
+            return null;
+        }
+
+        // escape any single or double quotes in SCRT property values
+        let scrtHeaderVal: string = "";
+        for (const [scrtPropName, scrtPropVal] of Object.entries(scrtData)) {
+            const valToUse = scrtPropVal.replace(/(["'])/g, "\\$1");
```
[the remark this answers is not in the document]
gejohnston: Copilot made a further suggestion which i implemented.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -657,6 +660,277 @@ export abstract class AbstractRestClient {
...
+            );
+            return null;
+        }
+
+        // escape any escape characters and any single or double quotes in SCRT property values
+        let scrtHeaderVal: string = "";
+        for (const [scrtPropName, scrtPropVal] of Object.entries(scrtData)) {
+            const valToUse = String(scrtPropVal)
+                .replace(/\\/g, "\\\\")
+                .replace(/(["'])/g, "\\$1");
```
zFernand0: Some dumb questions around escaping characters:

- Should we trim the `scrtProfVal` before appending it to the header string?
- Are there any other values we should be careful with?
  - Specially for anything that may be invalid for a header value, i.e. CRLF, null characters, non-ascii.
- Should we consider URI encoding the `scrtPropVal`?
  - Will REST APIs handle special non-URI encoded characters (if we decide not to do so)?

prototypes/packages/imperative/src/utilities/src/AsyncMutex.ts
zFernand0, line 1: Thanks for keeping these around 🙏

packages/imperative/src/rest/src/client/AbstractRestClient.ts
gejohnston: I suppose that a product could have space at the beginning or end of a name, but I think it is more likely that they made a typo when they specified the value in their session. I think that using trim is good approach until a product team makes a case for such spaces. I will add this.

It also seems that restricting the values to printable ASCII characters is an appropriate safety net. I will add this. However, I think that such a test belongs in the isScrtValid function instead of here.

I think that URL-encoding the header value may create problems. I do not think it is necessary because the header value is not used as part of the URL. Some legitimate SCRT values may contain values that have to be escaped for URL encoding. This would also place a burden on the REST API SDK to URL-decode what is supposed to be a simple text string. Unless we encounter some unforeseen problem in the future, I do not feel that we should URL-encode the header value.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
gejohnston: Please re-review.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
gejohnston: Please re-review the latest modifcations.

zFernand0 (approved):
LGTM! 😋

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -657,6 +660,289 @@ export abstract class AbstractRestClient {
...
+        if (!this.isScrtValid(scrtData)) {
+            Logger.getImperativeLogger().error(
+                `${funName} No SCRT header is created when SCRT data is invalid.`
+            );
+            return null;
+        }
+
+        // escape any escape characters and any single or double quotes in SCRT property values
+        let scrtHeaderVal: string = "";
+        for (const [scrtPropName, scrtPropVal] of Object.entries(scrtData)) {
```
ATorrise, line 773:
```bash
        const validScrtProps = ["featureName", "productId", "productVersion"]; //only loop over SCRT headers?
        let scrtHeaderVal: string = "";
        for (const [scrtPropName, scrtPropVal] of Object.entries(scrtData)) {
            if (!validScrtProps.includes(scrtPropName)) {
                continue; // non-SCRT properties
            }
```

ATorrise: do we want to escape all headers in general or just scrt? since scrt seems to have a custom format that doesnt apply to `{ "Content-Type": "application/json" }` for example

gejohnston: I agree with your opinion that we should not escape headers other than SCRT headers. However, I think that the existing logic already achieves that goal.

- The function formScrtHeaderVal() only forms the value for an SCRT header, so it will not escape any other header.
- The for loop iterates over a set of SCRT properties. It does not iterate over values for any other type of header.
- The SCRT properties over which we are iterating have already been validated by the earlier call to the function isScrtValid(). The isScrtValid function already ensures that the only properties in the object are the same properties that you identified : “featureName", "productId", "productVersion".

I do not think that any changes are required, but if you feel that I have overlooked something, we can discuss it.

pujal0909 (approved):
This looks good to me. Thanks @gejohnston

PR #2686, opened 2026-03-01
Upload a file to directory without need to specify its name twice 

### LINKED ISSUE #1209: Upload files to directory
I would like to upload a file into a USS directory under the same name. If I specify a directory as a target of the UNIX command `cp` it works. But with Zowe CLI, it fails:

```
zowe zos-files upload ftu go.mod /a/plape03/md5sum/                                                 
Command Error:
z/OSMF REST API Error:
Rest API failure with HTTP(S) status 500
category: 6
rc:       8
reason:   91947108
message:  fopen() error
details: 
  - EDC5123I Is a directory. (errno2=0x05620064)


Error Details:
HTTP(S) error status "500" received.

Protocol:  https
Host:      ca32.lvn.broadcom.net
Port:      1443
Base Path: 
Resource:  /zosmf/restfiles/fs/a%2Fplape03%2Fmd5sum%2F
Request:   PUT
Headers:   [{"Content-Type":"text/plain"},{"Accept-Encoding":"gzip"},{"X-CSRF-ZOSMF-HEADER":true}]
Payload:   undefined
```

It would be more useful for me and probably other people if it can upload the file into the directory using its original filename in the same way as the `cp` command does.

I would like to have an option (e.g. `--make-directories`) that would create missing directories of the target path so I can copy a file from my computer to z/OS without thinking if the directories are already there.

likhithanimma1:
What It Does

Solves https://github.com/zowe/zowe-cli/issues/1209


packages/zosfiles/src/methods/upload/Upload.ts
zFernand0, line 1: The changes definitely address the problem as stated in the issue. 🙏 

---

Scenarios to consider for follow-up issue/PR

Should we consider creating directories recursively?
- `zowe files upload ftu package.json /u/fernando/missing-dir/missing-dir/missing-dir/`

What about when the path does not end in `/` and the user specifies `--md`?
- `zowe files upload ftu package.json /u/fernando/missing-dir-01 --md`
  - Uploads `package.json` to `/u/fernando/missing-dir-01/` with the name `package.json`
- `zowe files upload ftu package.json /u/fernando/missing-dir-02`
  - Uploads `package.json` to `/u/fernando/` with the name `missing-dir-02`

---

Here is the output of the scenarios above 🙏 

Image

zFernand0 (approved):
LGTM! 😋

packages/cli/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Enhanced `zowe zos-files upload file-to-uss` to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added a `--make-directories` option to automatically create missing target directories during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
```
anaxceron:
```suggestion
- BugFix: Enhanced the `zowe zos-files upload file-to-uss` command to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added the `--make-directories` option to automatically create missing target directories during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
```

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Enhanced `zowe zos-files upload file-to-uss` to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added a `--make-directories` option to automatically create missing target directories during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
```
```suggestion
- BugFix: Enhanced the `zowe zos-files upload file-to-uss` command to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added the `--make-directories` option to automatically create missing target directories during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
```

packages/zosfiles/src/methods/upload/Upload.ts
likhithanimma1: Sure Fernando. Thanks for the inputs. :)

packages/zosfiles/src/methods/upload/Upload.ts
```diff
@@ -596,6 +596,36 @@ export class Upload {
...
+        let isUssDirectory = ussname.endsWith('/');
+
+        if (!isUssDirectory) {
+            const isDirectoryExist = await this.isDirectoryExist(session, ussname);
+            if (isDirectoryExist) {
+                isUssDirectory = true;
+            } else if (options.makeDir) {
+                const hasExtension = path.posix.extname(ussname) !== '';
+                if (!hasExtension) {
+                    isUssDirectory = true;
```
traeok: USS files don't require extensions - so files like `Makefile`, `README`, or `/u/users/binary_file` would be treated as directories since they have no extension.

The same applies for some directories - a directory like `configs.d` could be misinterpreted as a file instead of a directory. Consider pursuing one of these three options:

- Requiring the path to end with / to be treated as a directory when --make-dir is used and the path doesn't exist (preferred)
- Adding an explicit --directory flag to disambiguate
- Or document this clearly so users know to add a trailing / for files w/o extension (could still be confusing to end-users)

packages/zosfiles/src/methods/upload/Upload.ts
```diff
@@ -596,6 +596,36 @@ export class Upload {
...
+                }
+            }
+        }
+
+        if (isUssDirectory) {
+            const baseName = path.basename(inputFile);
+            if (!ussname.endsWith(baseName)) {
+                ussname = path.posix.join(ussname, baseName);
+            }
```
traeok: Consider using a stricter check to avoid subtle edge cases around file names:


```suggestion
            if (path.posix.basename(ussname) !== baseName) {
                ussname = path.posix.join(ussname, baseName);
            }
```

packages/zosfiles/src/methods/upload/Upload.ts
```diff
@@ -596,6 +596,36 @@ export class Upload {
...
+        }
+
+        if (options.makeDir) {
+            const parentDir = path.posix.dirname(ussname);
+            const parentExists = await this.isDirectoryExist(session, parentDir);
+            if (!parentExists) {
+                await Create.uss(session, parentDir, "directory");
+            }
```
traeok, line 630: Do we intend to support creating deeply-nested directories for an upload? For example, if the user specifies a path like `/u/user/a/b/c/file.txt` and `a`, `b`, and `c` do NOT exist, this create call would fail.

Consider refactoring this so that directories are created recursively, or explicitly document that this only supports creating a single level of directories (parent directory).

packages/cli/CHANGELOG.md
```diff
@@ -1,13 +1,18 @@
...
+- BugFix: Enhanced the `zowe zos-files upload file-to-uss` command to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added the `--make-directories` option to automatically create missing target directories during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
+
 ## `8.30.3`
 
 - BugFix: Updated the `tar` dependency to resolve technical currency. [#2696](https://github.com/zowe/zowe-cli/pull/2696)
 
 ## `8.30.2`
 
 - BugFix: Updated the `minimatch` dependency to resolve technical currency. [#2685](https://github.com/zowe/zowe-cli/pull/2685)
+- BugFix: Enhanced `zowe zos-files upload file-to-uss` to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added a `--make-directories` option to automatically create missing target directories during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
```
anaxceron:
```suggestion
- BugFix: Enhanced the `zowe zos-files upload file-to-uss` command to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added the `--make-directories` option to automatically create missing target directories during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
```

traeok (dismissed):
Thanks for working on this Likhitha! I left a few requests but the implementation makes sense so far 👍 

I noticed there could be up to 3 round-trip requests before the upload goes out:

1. `isDirectoryExist(session, ussname)` when path doesn't end with `/`
2. `isDirectoryExist(session, parentDir)` when `makeDir` is true
3. `Create.uss(session, parentDir, "directory")` when `makeDir` is true and parent doesn't exist

This could have a significant impact on time to upload for bulk uploads or scripts. Consider allowing users to skip the directory check if they know that the target is a file.

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,9 +2,14 @@
...
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Enhanced the `zowe zos-files upload file-to-uss` command to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added the `--make-directories` option to automatically create missing target directories during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
+
 ## `8.30.2`
 
 - BugFix: Updated the `minimatch` dependency to resolve technical currency. [#2685](https://github.com/zowe/zowe-cli/pull/2685)
+- BugFix: Enhanced `zowe zos-files upload file-to-uss` to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added a `--make-directories` option to automatically create missing target directories during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
```
```suggestion
- BugFix: Enhanced the `zowe zos-files upload file-to-uss` command to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added the `--make-directories` option to automatically create missing target directories during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
```

packages/cli/CHANGELOG.md
zFernand0, line 1: Now that there is a new version of flatted in 8.31.1, I believe these entries needs moved under a new `## Recent Changes` header above ```##  `8.31.1` ```

zFernand0 (commented):
Changes LGTM! 😋 
---

I will wait for Trae's comments to be resolved since we would want to prevent slowness for bulk actions 🙏

anaxceron (approved):

packages/cli/src/zosfiles/upload/Upload.options.ts
```diff
@@ -146,5 +146,25 @@ export const UploadOptions: {[key: string]: ICommandOptionDefinition} = {
...
+        aliases: ["md"],
+        description: strings.MAKE_DIR,
+        type: "boolean"
+    },
+
+    /**
+     * The skip directory check option for upload
+     */
+    skipDirectoryCheck: {
+        name: "skipDirectoryCheck",
```
awharn:
```suggestion
        name: "skip-directory-check",
```

I believe this should be in kebab case, in line with the rest of the upload options.

packages/zosfiles/__tests__/__unit__/methods/upload/Upload.unit.test.ts
```diff
@@ -2606,4 +2455,275 @@ describe("z/OS Files - Upload", () => {
             });
         });
     });
-});
+
+    describe("fileToUssFile", () => {
```
awharn: This test suite seems to have been moved from its original location in the file. Is there a reason why?

awharn (changes requested):

packages/cli/src/zosfiles/upload/Upload.options.ts
likhithanimma1: Thanks @awharn . Resolved it.

packages/zosfiles/src/methods/upload/Upload.ts
likhithanimma1: Hey @traeok . Thankyou for the inputs. I did make changes in the code. Please let me know if any more are required.

packages/zosfiles/src/methods/upload/Upload.ts
```diff
@@ -596,6 +596,42 @@ export class Upload {
...
+        // Check if ussname is a directory (ends with '/') or if it exists as a directory on z/OS
+        const isUssDirectory = ussname.endsWith('/');
+
+        // If path ends with '/', it's a directory - append the filename
+        if (isUssDirectory) {
+            const baseName = path.basename(inputFile);
+            ussname = path.posix.join(ussname, baseName);
+        } else if (!skipDirCheck && !options.makeDir) {
+            try {
+                const existsAsDir = await this.isDirectoryExist(session, ussname);
```
awharn, line 611: `isDirectoryExist` is misleadingly named. It will return `true` for USS files that exist as well. Therefore, if you specify a path to an existing file (in an attempt to overwrite), this logic will treat it as a directory, which results in a server error.

It should be possible to enhance `isDirectoryExist` to verify that it is, indeed, a directory. Since it is a public API, it might break some logic in extenders, but it seems more like a bug fix. And a change like that shouldn't affect our logic in Upload.

awharn (changes requested):

packages/zosfiles/src/methods/upload/Upload.ts
likhithanimma: Thanks @awharn. I just modified isDirectoryExist. Can please review now.

traeok:
Running system tests now, will approve if passing 👍

traeok (approved):
System tests passed, changes LGTM. Thanks Likhitha!

awharn (approved):
The changes look great, thanks @likhithanimma1!

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Enhanced the `zowe zos-files upload file-to-uss` command to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added the `--make-directories` option to automatically create missing target directories during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
```
zFernand0: Should we add the `--skip-directory-check` here?

packages/zosfiles/src/methods/upload/Upload.ts
```diff
@@ -641,6 +677,36 @@ export class Upload {
...
+
+        await this.createDirectoriesRecursively(session, parentDir, skipCheck);
+
+        if (skipCheck) {
+            await Create.uss(session, dirPath, "directory").catch((err) => {
+                if (err?.mDetails?.msg?.includes("already exists")) {
+                    return;
+                }
+                throw err;
+            });
```
zFernand0, line 714: It gets a bit tricky when people combine both flags. 

For example, if I specify an existing file and then a directory I see two different errors

`zowe files upload ftu package.json /tmp/file/dir --md --sd`
Image

`zowe files upload ftu package.json /tmp/file/dir/ --md --sd`
Image

Can we provide a better error message in the second scenario?

zFernand0 (changes requested):
Changes MSTM! 😋 
---

I do have a few concerns about the new skip-directory-check flag 😅

packages/zosfiles/src/methods/upload/Upload.ts
likhithanimma1: Thanks for testing @zFernand0 . I am not sure why the second scenario fails. Can u please provide details of /tmp/file/dir/
Screenshot 2026-04-15 at 12 08 52 AM

zFernand0: oh, sorry, I should've been more explicit.

Assuming that the file `/tmp/file` already exists...
When you try to create a directory on top of it, that's the scenario where you get the error.

SSH (or OMVS): 
- `mkdir -p ~/tmp` 
- `touch ~/tmp/file`
- `mkdir ~/tmp/file/dir`

likhithanimma1: Thankyou @zFernand0 . I have pushed the changes please let me know if any.

zFernand0 (approved):
LGTM! 😋 
---

Thanks for addressing the PR comments 🙏 

Image

zFernand0 (changes requested):
Before merging, can we increase patch coverage?

It seems that with the latest error handling, we dropped below 80% 😢

likhithanimma1:
> Before merging, can we increase patch coverage?
> 
> It seems that with the latest error handling, we dropped below 80% 😢

Hey @zFernand0 . I have added coverage. Thankyou :) Didnot see it before. My apologies.

packages/zosfiles/__tests__/__unit__/methods/upload/Upload.unit.test.ts
zFernand0, line 1: Thanks for moving the Spy initialization to the beforeEach hook (and restore to afterEach).
It makes each individual test a little more reliable 🙏 

As part of a larger effort, we could do the same for other existing tests. but for now this is good as it is 😋

zFernand0 (approved):
LGTM! 😋

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Enhanced the `zowe zos-files upload file-to-uss` command to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added the `--make-directories` option to automatically create missing target directories and the `--skip-directory-check` option to skip checking if the target path is a directory during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
```
zFernand0: nitpick: This should be from the perspective of the SDK, meaning `makeDirectories` and `skipDirectoryCheck` 

Also, it should say `Enhancement:` instead of `BugFix:` 🙏

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Enhanced the `zowe zos-files upload file-to-uss` command to support uploading a file to a USS directory using its original filename when the target is a directory (similar to UNIX cp) and added the `--make-directories` option to automatically create missing target directories and the `--skip-directory-check` option to skip checking if the target path is a directory during upload.[#2686](https://github.com/zowe/zowe-cli/pull/2686)
```
zFernand0: This should say `Enhancement:` instead of `BugFix:` 🙏

zFernand0 (approved):
Thanks for addressing changelog comments 🙏

zFernand0:
Labeling `release-current` to combine with other PRs in a single minor version update 🙏

PR #2688, opened 2026-03-03
Feature: Support secure certificates

### LINKED ISSUE #2325: Secrets SDK: Add ability to load client certs from OS cert storage

Currently, Zowe CLI and Explorer for VS Code support using client certificates to authenticate to z/OSMF and the Zowe API ML. This implementation does not (as I understand it) support loading certs from the certificate storage facilities built into the user's operating system. 

Enhance the Secrets SDK to enable users to specify paths to certificates within the OS certificate store. I am not sure what this means for Linux users, but on MacOS it would require interacting with the keychain, and on Windows this would require interacting with the Windows certificate store.

The end goal would be for Zowe CLI and Zowe Explorer for VS Code users to be able to specify the path and name of the certificate to use for authenticating in their z/OSMF and API ML(nested/base) profiles.

JillieBeanSim:
**What It Does**

Add support for securely stored certificates in macOS and Windows environments and adopted the support for zosmf type profiles

**How to Test**

import your P12 cert into the macOS Keychain or Windows Certificate Store, you don't have to worry about having to set the key as exportable or anything. Do enter cert password when prompted during import. add CN name of cert for value of the new profile property "certAccount" and set authOrder as cert-pem, ie.

```
    "zosmfSecCerts": {
             "type": "zosmf",
                    "properties": {
                        "host": "host.com",
                        "rejectUnauthorized": false,
                        "port": 443,
                        "certAccount": "{CERT CN}",
                        "authOrder": "cert-pem"
                    },
                "secure": []
                },
``` 

expect previous behavior especially with macOS where you may be prompted by security for OS password when retrieving the cert for use.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**


packages/imperative/src/rest/src/client/NativeHttpsClient.ts
```diff
@@ -0,0 +1,155 @@
...
+
+export class NativeHttpsClient {
+    private static readonly ENV_FLAG_MACOS = "ZOWE_MACOS_NATIVE_HTTPS";
+    private static readonly ENV_FLAG_WINDOWS = "ZOWE_WINDOWS_NATIVE_HTTPS";
+
+    public static isEnabled(session: ISession): boolean {
+        return (process.platform === "darwin" || process.platform === "win32")
+            && session?.certAccount != null
+            && (this.isFlagEnabled() || (session as any)._useNativeHttpsForNonExportable === true);
+    }
```
zFernand0: Is this a Mac and Windows only feature?

Or should we consider `linux` (and other platforms like `openbsd`)?

_Source:_ https://nodejs.org/api/process.html#processplatform

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -193,14 +193,21 @@ export class ConnectionPropsForSessCfg {
...
+                    const hasCertAccount = sessCfgToUse._authCache?.availableCreds?.certAccount;
+
+                    // Only prompt for cert/certKey if certAccount is not available
+                    if (!hasCertAccount) {
+                        if (!sessCfgToUse._authCache?.availableCreds?.cert && !doNotPromptForValues.includes("cert")) {
+                            promptForValues.push("cert");
+                        }
+                        if (!sessCfgToUse._authCache?.availableCreds?.certKey && !doNotPromptForValues.includes("certKey")) {
+                            promptForValues.push("certKey");
+                        }
```
zFernand0, line 207: Since we are adding a new property specific to the cert-pem order of precedence, I believe we will need to update the docs 🙏 

_Link:_ https://docs.zowe.org/stable/user-guide/cli-authentication-methods#using-client-certificates

packages/core/src/constants/Core.constants.ts
zFernand0, line 1: Added @anaxceron as reviewer for her input regarding changelog entires and help text additions.

The help text can be viewed in this file or in any of the `.snap` files.

Here is a picture for ease of review/discussion 🙏 

Every other command will have this new option:
Image

And the `zowe config auto-init --help` will have this option:
Image

packages/imperative/__tests__/src/rest/client/KeychainAgent.unit.test.ts
zFernand0, line 1: Some lint warnings in this and the NativeHttpsClient.unit.test.ts files.

Image

packages/imperative/src/rest/__tests__/client/AbstractRestClient.unit.test.ts
```diff
@@ -1852,14 +1852,14 @@ describe("AbstractRestClient tests", () => {
                 const restOptions: any = {
                     headers: {}
                 };
-                const pemCertWasSet: boolean = restClient["setCertPemAuth"](restOptions);
+                const pemCertWasSet: boolean = await restClient["setCertPemAuth"](restOptions);
```
zFernand0: This made me question if we were introducing a breaking change, however the function is private `setCertPemAuth` (called by  `buildOptions` - which is also private). 

And buildOptions is cllaed in  the `request` function which already returned a promise. Meaning it's fair game to make the inner function `async`, as long as we continue to call the promise.resolve (or `.mResolve(...)`)

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -296,41 +297,70 @@ export abstract class AbstractRestClient {
...
-            this.addScrtHeader(options);
-
-            const buildOptions = this.buildOptions(options.resource, options.request, options.reqHeaders);
-
-            // Perform the actual http request
-            let clientRequest: http.ClientRequest;
-            if (this.session.ISession.protocol === SessConstants.HTTPS_PROTOCOL) {
-
-                clientRequest = https.request(buildOptions, this.requestHandler.bind(this));
+            (async () => {
```
zFernand0: this is what I was referring to in the previous comment 🙏 

Smart way to avoid a potential race-condition / breaking-change 🙏

zFernand0 (commented):
Changes MSTM! 😋 
---

I'm down to the last 30 files. 🙏 

For those reviewing...
- Exactly 100 files were `.snap` changes 😅 
- Consider reviewing the PR without whitespaces 😋 
    Image

packages/imperative/src/rest/src/client/NativeHttpsClient.ts
JillieBeanSim: yes macOS and Windows at the moment, previous discussions we said we would see if community requested linux as well

packages/imperative/__tests__/src/rest/client/KeychainAgent.unit.test.ts
JillieBeanSim: lint warning addressed in commit https://github.com/zowe/zowe-cli/pull/2688/commits/f18d92e75c57f100f5b27481310ae5275d7c3313

anaxceron (approved):
Looks good to me!

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -329,41 +330,70 @@ export abstract class AbstractRestClient {
...
-            ImperativeExpect.toBeDefinedAndNonBlank(options.request, "request");
-            ImperativeExpect.toBeEqual(options.requestStream != null && options.writeData != null, false,
-                "You cannot specify both writeData and writeStream");
-
-            // putTopAuthInSession was originally done in the RestClient constructor.
-            // As a safety net to ensure that no logic has placed a different cred in the
-            // session since then, we again place only the top cred in the session.
-            AuthOrder.putTopAuthInSession(this.session.ISession);
-
-            // form a header from scrtData and place the header into the options.reqHeaders
```
gejohnston, line 357: The call to the addScrtHeader function has been removed by this PR.  Probably due to the merging of PRs after this implementation was started. We need to keep this function call.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -376,139 +406,152 @@ export abstract class AbstractRestClient {
...
-                } else {
-                    reject(this.populateError({
-                        msg: "Failed to send an HTTP request.",
-                        causeErrors: errorResponse,
-                        source: "client"
-                    }));
-                }
-            });
+                    if (errorResponse.code === "ECONNRESET" && clientRequest.reusedSocket) {
+                        this.request(options).then((response: string) => {
```
awharn: This should be `this._request`, following some changes that enabled request queuing. If left in the current implementation, a new request will return to the back of the queue, while this request hangs waiting for it (thus decreasing the number of parallel requests by 1, which causes a deadlock if the max concurrent requests is 1).

awharn (dismissed):

gejohnston (changes requested):
I identified that a call to a function named addScrtHeader is being incorrectly removed by this PR. The deletion probably occurred when resolving a merge form master into this branch.

There have been a number of recent enhancements to the AbstractRestClient. The Zowe team should closely review that file for any other merge errors.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
JillieBeanSim: that can happen when the PR sits there for almost a month before being reviewed. I don't recall addressing conflicts in this file, mainly the CHANGELOG files are the one's I had conflicts in so my changes may have just overridden them on a merge from main.  
If many changes came into this file since I created the PR I may need help getting stuff back in correctly without breaking my work.

JillieBeanSim: the comment and line under have been added back in commit https://github.com/zowe/zowe-cli/pull/2688/commits/baaeca9b831dab9a6537fd5856a390d2a9dc9296

packages/imperative/src/rest/src/client/AbstractRestClient.ts
JillieBeanSim: changed from this.request to this._request in commit https://github.com/zowe/zowe-cli/pull/2688/commits/baaeca9b831dab9a6537fd5856a390d2a9dc9296

packages/imperative/src/rest/src/session/doc/ISession.ts
```diff
@@ -138,6 +138,16 @@ export interface ISession {
      */
     passphrase?: string;
 
+    /**
+     * The account name for a PKCS#12 certificate stored in the credential manager.
+     * When specified, the certificate and private key will be retrieved from the
+     * system keychain instead of from files.
+     * (macOS only - uses non-exportable keys from Keychain Access)
```
traeok: Is this property really exclusive to macOS or am I just confused by the wording in parentheses? Based on my initial review, it sounds like the feature also supports Windows through this property.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -1069,17 +1114,78 @@ export abstract class AbstractRestClient {
...
+                this.session.ISession.certAccount,
+                ImperativeConfig.instance.cliHome
+            );
+
+            if (!isExportable) {
+                // Key is non-exportable - automatically enable native HTTPS client
+                const platformName = process.platform === "darwin" ? "macOS" : "Windows";
+                this.log.info(`Certificate '${this.session.ISession.certAccount}' has a non-exportable private key. ` +
+                    `Automatically using ${platformName} native HTTPS client.`);
+                (this.session.ISession as any)._useNativeHttpsForNonExportable = true;
```
traeok: Could we add this `_useNativeHttpsForNonExportable` field on `ISession` (even if marked `@internal`) for developer awareness / type safety? The `as any` cast can be fragile and this logic could be unknowingly broken by future refactoring.

packages/secrets/core/src/os/win/cert_store.rs
```diff
@@ -0,0 +1,218 @@
...
+                }
+                
+                // For non-exportable keys, we cannot export the key data
+                // Return an informative error message
+                CertFreeCertificateContext(cert_context);
+                
+                Err(KeyringError::Library {
+                    name: "Windows Certificate Store".to_owned(),
+                    details: "Private key cannot be exported (likely non-exportable). Use create_identity_context and sign_with_identity for non-exportable keys.".to_owned(),
+                })
```
traeok: It seems like this function does not actually check whether a key is exportable, it just checks its size. Is this an accurate way of determining whether a key is exportable? If so, is there some supporting docstring that we could add as a comment in this function? 

It may not be necessary for the first phase of implementation, but we could try using `NCryptExportKey` to see if a key can be exported: https://learn.microsoft.com/en-us/windows/win32/api/ncrypt/nf-ncrypt-ncryptexportkey

In the meantime, if this is intentional design, consider returning early in `NativeHttpsClient.isKeyExportable` if the platform is `win32` to make this behavior explicit.

packages/secrets/core/src/os/win/cert_store.rs
```diff
@@ -0,0 +1,218 @@
...
+
+/// Helper function to encode a string as a null-terminated UTF-16 string
+pub(super) fn encode_utf16(s: &str) -> Vec<u16> {
+    let mut chars: Vec<u16> = s.encode_utf16().collect();
+    chars.push(0);
+    chars
+}
+
+/// Helper function to convert Windows error to string
+fn win32_error_as_string(error: WIN32_ERROR) -> String {
```
traeok: These functions are duplicates of the original implementations in `win/mod.rs` which goes against DRY principle. Can we make the functions public in `mod.rs` and then reuse them here?

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -329,41 +330,72 @@ export abstract class AbstractRestClient {
...
+                this.addScrtHeader(options);
+                const buildOptions = await this.buildOptions(options.resource, options.request, options.reqHeaders);
+
+                if (NativeHttpsClient.isEnabled(this.session.ISession)
+                    && this.session.ISession.protocol === SessConstants.HTTPS_PROTOCOL) {
+                    if (options.requestStream != null || options.responseStream != null) {
+                        throw new ImperativeError({
+                            msg: "macOS native HTTPS client does not yet support streaming requests/responses.",
+                            additionalDetails: "Disable ZOWE_MACOS_NATIVE_HTTPS or avoid streaming for this request.",
+                        });
```
traeok: Tracing through the helper function `NativeHttpsClient.isEnabled` and the surrounding logic, it looks like this error can be thrown for Windows users - but, the message mentions macOS. Can we revisit the message here to make it platform-agnostic?

packages/secrets/core/src/os/mac/nsurlsession_https.m
```diff
@@ -0,0 +1,271 @@
...
+
+- (void)URLSession:(NSURLSession *)session
+          dataTask:(NSURLSessionDataTask *)dataTask
+    didReceiveData:(NSData *)data {
+    [self.responseData appendData:data];
+}
+
+- (void)URLSession:(NSURLSession *)session
+              task:(NSURLSessionTask *)task
+didCompleteWithError:(NSError *)error {
```
traeok: It's been a while since I looked at Objective-C code, so sorry in advance if this is phrased poorly 😅 

I believe `didReceiveData` and `didCompleteWithError` are unreachable methods - when `dataTaskWithRequest:completionHandler` is used, NSURLSession won't call the delegate's `didReceiveData` or `didCompleteWithError` methods. Consider removing these two methods to avoid confusion during future maintenance.

See this page from Apple dev doc for more context: https://developer.apple.com/documentation/foundation/urlsession/datatask(with:completionhandler:)-e6xv#Discussion

traeok (dismissed):
Thanks for working on this Billie and sorry for the delayed review. I left a few suggestions and a couple requests, but the code looks good so far. Doing some manual testing but wanted to post these ahead of time.

packages/imperative/src/cmd/src/CommandProcessor.ts
```diff
@@ -515,12 +515,26 @@ export class CommandProcessor {
...
                 // prompt for any requested --options
                 if (this.definition.options != null && this.definition.options.length > 0) {
+                    const hasCertAccount = preparedArgs.certAccount != null || preparedArgs["cert-account"] != null;
                     for (const option of this.definition.options) {
                         // check if value provided
                         if (preparedArgs[option.name] != null) {
                             // string processing
                             if (typeof preparedArgs[option.name] === "string" &&
                                 preparedArgs[option.name].toUpperCase() === this.promptPhrase.toUpperCase()) {
+                                if (hasCertAccount && option.name === "password") {
```
gejohnston: When the AuthOrder module was created, the intent was to consolidate into that module all logic about what creds are used or prompted for. It is possible that AuthOrder already handles this situation without this new logic. That would have to be tested to confirm the right behavior. If AuthOrder's existing logic for a certificate does not already handle the situation, I think that it would be preferable to add logic to the AuthOrder module to keep all such logic consolidated in one place, instead of adding new logic here or in other places.

gejohnston (dismissed):

packages/imperative/src/cmd/src/CommandProcessor.ts
JillieBeanSim: Since this deals specifically with prompts and validating existence of an item and not prompting due to the existence of item this is where it belongs. We do not get and store this key/password value, it will be grabbed and passed when doing the request to support non-exportable keys

gejohnston: This prompt may not be the prompt that you were thinking of. With cert-pem as your authOrder (as you have in the description of this PR), the AuthOrder class should *PREVENT* any prompt for password like the prompt shown, below (in which the command is using basic auth, but has no password).
```
> zowe zos-files list data-set "SYS1.PARMLIB*"
Required connection properties are missing in your Zowe client configuration. Provide these properties at the prompts to complete your command.

Enter the password for your service (will be hidden):
```


I think that the logic in CommandProcessor handles the situation in which the user explcitly places an option on the command line, but does not want the value shown on the command line, so the user supplies the special value PROMPT* to force the user to be prompted for a value. The prompt text is different for this situation, see below:

```
> zowe zos-files list data-set "SYS1.PARMLIB*" --password PROMPT*
"password" Description: Mainframe (z/OSMF) password, which can be the same as your TSO password.
Please enter "password":
```

Granted, the user will be prompted in the case above, but that is due to an explicit action by the user. It makes no sense for the user to supply --password PROMPT* on the command line when using "certAccount", but that is what the user explicitly chose to do. I think that we should not remove the "password" option from preparedArgs in this code to avoid disrupting the "Available creds" described later. Note that AuthOrder rules will ignore even such an explicitly-provided password value during later credential prompting and during eventual connection to the mainframe, if "certAccount" is the top AuthOrder choice.

AuthOrder keeps the desired creds in the session, and keeps a list of all credential types that were available to the command. If we remove password from the list of available parms, it will not appear in the list of "Available creds" when the command fails to connect to the mainframe service. An example of the "Available creds" information is shown below.

```
-> zowe zos-files list data-set "SYS1.PARMLIB*"
Failed to send an HTTP request.

Response From Service
Error: connect ECONNREFUSED 11.22.33.44:123

Diagnostic Information
HTTP(S) client encountered an error. Request could not be initiated to host.
Review connection details (host, port) and ensure correctness.

Protocol:          https
Host:              Your.Test.LPAR.com
Port:              1234
Base Path:         ibmzosmf/api/v1
Resource:          /zosmf/restfiles/ds?dslevel=SYS1.PARMLIB*
Request:           GET
Headers:           [{"Accept-Encoding":"gzip"},{"X-IBM-Max-Items":"0"},{"X-CSRF-ZOSMF-HEADER":true}]
Payload:           undefined
Allow Unauth Cert: true
Available creds:   user,password,tokenType,tokenValue,cert,certKey,base64EncodedAuth
Your auth order:   basic,token,bearer,cert-pem
Auth type used:    basic
```

I have seen modifications for "certAccount" in AuthOrder (although I have not reviewed them in detail). Those changes may already prevent the "missing property value" type of prompt. If not, I am reasonably confident that it can be made to prevent such a prompt.

packages/secrets/core/src/os/mac/nsurlsession_https.m
JillieBeanSim: will remove the unreachable code since we use dataTaskWithRequest:completionHandler: which handles data in the completion block; going to retest though before committing the change on this one

traeok:
One more thing I wanted to mention: it seems that this PR proposes adopting a native HTTPS client instead of the `RestClient` implementation to facilitate requests involving certificate use. I am concerned that this deviates from the typical flow for HTTP requests though the Zowe SDKs and client ecosystem.

When the native HTTPS path is active, the request bypasses significant portions of `AbstractRestClient`: streaming support, proxy configuration, timeout handling, compression/decompression, progress tracking, ~~the SCRT header~~, keep-alive/connection reuse, and the error handling pipeline all live within the Node.js request flow. The native client reimplements a subset of this in Rust/Objective-C/WinHTTP, which means every future enhancement to `AbstractRestClient` would also need to be mirrored in the native clients to maintain parity.

Is there a way we can provide the required certificate data to the REST client in the Node.js layer? For example, could the native layer handle only the certificate/key retrieval and identity context operations (which it already does via `getCertificate`, `getPrivateKey`, `createIdentityContext`, `signWithIdentity`), and then feed that material into Node.js TLS - either through the existing `KeychainAgent` approach or a custom `tls.connect` with a `privateKeyEngine`/signing callback? That way, we keep all HTTP request logic in `AbstractRestClient` and only cross into native code for the keychain/certificate store operations that Node.js can't do on its own.

I understand that non-exportable private keys could be driving the implementation of the native HTTPS path; Node.js TLS requires the raw private key bytes, which the OS won't provide for non-exportable keys. But it might be worth investigating whether Node.js OpenSSL engine support or a custom secure context could bridge that gap, before we commit to a parallel HTTP client implementation that we'd need to maintain long-term.

Edit(4/6): Thought SCRT header was not being passed, but it is so I updated the above comment. The remaining features mentioned are still unsupported.

packages/imperative/src/cmd/src/CommandProcessor.ts
JillieBeanSim: those changes you mention for the certificate did not address the issue seen and why I added this, I kept getting prompted for password and added handling in the prompt method that drives the prompt to eliminate it if certAccount is used.

t1m0thyj:
> this is the only way to support non-exportable key with secure certificates, I tried multiple approaches and this native HTTPS approach was the only one that would work with non-exportable keys from secure storage.

I investigated a bit and think I've found a way to support non-exportable keys with only a custom agent, similar to how `https-proxy-agent` adds proxy support, without needing a custom/native HTTP client. Changes are in this branch and I've tested that it works on MacOS: https://github.com/JillieBeanSim/zowe-cli/compare/feat/secCert-2...zowe:zowe-cli:timothy/test-certs
We can discuss more next week 😋

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -1069,17 +1082,59 @@ export abstract class AbstractRestClient {
...
+            this.log.trace("Using certificate store authentication (certAccount)");
+
+            if (process.platform !== "darwin" && process.platform !== "win32") {
+                throw new ImperativeError({
+                    msg: "Certificate account authentication (certAccount) is only supported on macOS and Windows.",
+                });
+            }
+
+            // eslint-disable-next-line @typescript-eslint/no-var-requires
+            const { KeychainAgent } = require("./KeychainAgent");
```
awharn: Since the KeychainAgent is only used within the try/catch block, could the import be moved into the block as well? The Secrets SDK does fail to load if certain Visual Studio executables are not available on Windows systems, and if a user is using another Secrets plug-in, or has secure credentials disabled, this could be the first instance that causes an error to surface.

awharn (commented):
I noticed one thing in addition to my requested change - when I have an invalid hostname and provide cert-account, I get an error 502 back instead of an error 400/401. This is also true when the certificate is not valid for the system that we are connecting to. Will this different behavior mess with anything in Zowe Explorer?

packages/imperative/src/cmd/src/CommandProcessor.ts
gejohnston: @JillieBeanSim 
In a local copy of the source, I temporarily remove this block of code.
```
if (typeof preparedArgs[option.name] === "string" &&
    preparedArgs[option.name].toUpperCase() === this.promptPhrase.toUpperCase()) {
    /* Temporarily remove for a test **************************
    if (hasCertAccount && option.name === "password") {
        this.log.debug("Skipping prompt for option %s because certAccount is present in profile",
            option.name);
        preparedArgs[option.name] = undefined;
        const camelCase = CliUtils.getOptionFormat(option.name).camelCase;
        preparedArgs[camelCase] = undefined;
        if (option.aliases != null) {
            for (const alias of option.aliases) {
                preparedArgs[alias] = undefined;
            }
        }
        continue;
    }
    **************************************************/
    // prompt has been requested for an --option
    this.log.debug("Prompting for option %s which was requested by passing the value %s",
        option.name, this.promptPhrase);

```
I put every possible type of credential into my config. I then set the certAccount property in my config to a cert that resides in the CertMgr. Since that cert is not configured on the server, we expect that the REST request will fail, but I did not see any prompt for password. Note that the password is listed in the set of available creds in the output.
```
zowe zos-files list data-set "SYS1.PARMLIB*"
Failed to send an HTTP request.

Response From Service
Error: socket hang up

Diagnostic Information
HTTP(S) client encountered an error. Request could not be initiated to host.
Review connection details (host, port) and ensure correctness.

Protocol:          https
Host:              Your.QA.LPAR.com
Port:              1443
Base Path:
Resource:          /zosmf/restfiles/ds?dslevel=SYS1.PARMLIB*
Request:           GET
Headers:           [{"Accept-Encoding":"gzip"},{"X-IBM-Max-Items":"0"},{"X-CSRF-ZOSMF-HEADER":true}]
Payload:           undefined
Allow Unauth Cert: true
Available creds:   user,password,tokenType,tokenValue,cert,certKey,certAccount
Your auth order:   cert-pem
Auth type used:    cert-pem
```

It appears to me that the Zowe client infrastructure is handling credentials and prompting correctly, and that this block of logic is not necessary. Is it possible that you are testing with a plugin or extension that is doing the prompting? If so, perhaps the correction to undesired prompts should be made in that location?

packages/imperative/src/rest/src/client/AbstractRestClient.ts
traeok: > The Secrets SDK does fail to load if certain Visual Studio executables are not available on Windows systems

This is no longer true now that we ship a static C runtime, but the other 2 possibilities are valid concerns

JillieBeanSim: @awharn since this is new functionality all together ZE would have to adopt and handle when cert-account is present.
A 502 Bad Gateway error occurs when a server acting as a gateway or proxy receives an invalid response from an upstream server. This typically indicates a communication failure between servers, often due to issues like server overload, misconfigurations, or network problems.
Since values are invalid/misconfiguration, wouldn't this be correct? With the use of non-exportable keys we can't store and run any checks so it would be up to the Rest API or API ML to validate and return error if invalid.
I didn't make any changes around hostname, and not sure what checks are already in place around that field and value.

JillieBeanSim: I tested with my expired cert (direct connect with zosmf) and in the error it includes this which explains the issue:

Response From Service
TLS handshake failed with {mySystem}.com: Failure(Error { code: -9806, message: "connection closed via error" }) [errSSLClosedAbort: server aborted the handshake — client certificate or its chain was likely rejected; check that all intermediate CA certificates are present in the keychain]

did edit error here to remove the hostname

JillieBeanSim:
Hey all, I tried restarting the failed check and it failed again with call to docker failing.  Is it just a pipeline glitch or something?  latest commit didn't change anything in this package that is failing a check.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
awharn: I just wanted to make sure, because I know that ZE prompts for updated fields if authentication fails. Since this is an enhancement, I don't think this change is a big deal, I just wanted to make sure the behavior was known. And thanks for moving the import! 🚀

awharn:
> Hey all, I tried restarting the failed check and it failed again with call to docker failing. Is it just a pipeline glitch or something? latest commit didn't change anything in this package that is failing a check.

One of the Ubuntu package mirrors was down, which was breaking the 32 bit ARM builds because the libsecret-dev package could not be installed. I rebuilt it a couple more times, and it appears Ubuntu fixed the issue.

t1m0thyj (changes requested):
Thanks @JillieBeanSim for your work on this! ❤️ 

Will re-review once Gene's comment is addressed: https://github.com/zowe/zowe-cli/pull/2688#discussion_r3089383725

packages/imperative/src/cmd/src/CommandProcessor.ts
JillieBeanSim: addressed in commit https://github.com/zowe/zowe-cli/pull/2688/commits/b39d4feb58ee2f62731a35f7f588488983433c7d

__types__/hogan.d.ts
t1m0thyj, line 1: Curious why this file needed changes? Were there build failures related to the `hogan` dependency?

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -1069,17 +1069,58 @@ export abstract class AbstractRestClient {
...
+                // eslint-disable-next-line @typescript-eslint/no-var-requires
+                const { KeychainAgent } = require("./KeychainAgent");
+                const agent = new KeychainAgent(
+                    this.session.ISession.certAccount,
+                    ImperativeConfig.instance.cliHome,
+                    {
+                        rejectUnauthorized: this.session.ISession.rejectUnauthorized,
+                        ca: this.session.ISession.serverCertificate,
+                    }
+                );
```
traeok: A new `KeychainAgent` is created for every request, which spawns a new OS thread and performs a TLS handshake via Rust. For bulk operations this can be expensive, for example `O(n)` threads, handshakes, and socket files for operations like `zos-files download all-members` or `zos-jobs download`.

I'd recommend caching the agent on the session so we can reuse the agent across requests for that session. This is more performant and avoids unnecessary overhead of setting up the agent for each request.

packages/imperative/src/rest/src/client/KeychainAgent.ts
```diff
@@ -0,0 +1,155 @@
...
+                "createTlsPipe is not supported by the current version of @zowe/secrets-for-zowe-sdk. " +
+                "Please rebuild the native bindings."
+            );
+        }
+
+        const host = options.servername || options.host || "localhost";
+        const port = options.port || AbstractSession.DEFAULT_HTTPS_PORT;
+        const rejectUnauthorized = options.rejectUnauthorized ?? true;
+
+        const localPath: string = await this.keyring.createTlsPipe(host, port, this.certAccount, rejectUnauthorized);
```
traeok, line 131: Is it expected that we don't pass the CA to Rust when creating the TLS pipe? Users relying on a custom CA would silently fall back to OS trust, or fail verification w/o a way to recover.

Consider passing CA through the function or reject the cert at validation time w/ a clear error

packages/secrets/core/src/os/mac/tls_pipe.rs
```diff
@@ -0,0 +1,220 @@
...
+    use security_framework::trust::SecTrust;
+
+    let raw_identity = identity.as_concrete_TypeRef();
+    // CFRetain increments the refcount before we hand ownership to sf_identity via
+    // wrap_under_create_rule. When `identity` is dropped at the end of this function
+    // (decrement), sf_identity still holds a valid +1 reference in the spawned thread.
+    unsafe { CFRetain(raw_identity as *const _); }
+    let sf_identity = unsafe { SecIdentity::wrap_under_create_rule(raw_identity as *mut _) };
+    
+    thread::spawn(move || {
```
traeok, line 99: The socket file is never removed after the thread exits. Successful connections leave `/tmp/zowe_secrets_core_*.sock` behind indefinitely.

Consider adding `remove_file` calls to the exit paths (whether success or failure)

traeok (commented):
Left some comments and plan to do some manual testing after meetings. Thanks @t1m0thyj for refactoring the native layer to simplify request handling in Rust to the TLS handshake, few suggestions on those changes but most of the code LGTM

packages/secrets/core/src/os/mac/tls_pipe.rs
t1m0thyj: Resolved in https://github.com/zowe/zowe-cli/pull/2688/commits/0f2891768639934146eb3f7c7e40512778356a26

packages/imperative/src/rest/src/client/AbstractRestClient.ts
t1m0thyj: Resolved in https://github.com/zowe/zowe-cli/pull/2688/commits/eaa71cc1419a5ea04a619c8a8aec879b1f7c0b17

packages/imperative/src/rest/src/client/KeychainAgent.ts
t1m0thyj: Resolved in https://github.com/zowe/zowe-cli/pull/2688/commits/5946798b35b8354c75862ac60095c65a3e9b8732 - the `serverCertificate` property wasn't being assigned anywhere so removed it.

packages/imperative/src/rest/src/client/KeychainAgent.ts
```diff
@@ -0,0 +1,148 @@
...
+                const privateKeyDer: Buffer = await this.keyring.getPrivateKey(this.certAccount);
+                if (privateKeyDer != null) {
+                    return this.buildExportableTlsSocket(options, privateKeyDer);
+                }
+            } catch (err: any) {
+                if (!err.message?.includes("non-exportable")) {
+                    throw err;
+                }
+            }
+        }
```
t1m0thyj: Do we need to maintain separate support for exportable keys (`getCertificate`, `getPrivateKey`, etc)? Not opposed to keeping it if we see benefit, but wondering how much it benefits us since:
* Support is currently limited to MacOS
* The native agent/socket used for non-exportable keys could be reused for exportable ones

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -1069,17 +1069,44 @@ export abstract class AbstractRestClient {
...
+     * @returns {boolean} True if this function sets authentication options. False otherwise.
      * @memberof AbstractRestClient
      */
     private setCertPemAuth(restOptionsToSet: any): boolean {
         if (!(this.session.ISession.type === SessConstants.AUTH_TYPE_CERT_PEM)) {
             return false;
         }
+
+        // Handle certAccount (keychain/certificate store-based certificate) - macOS and Windows
+        if (this.session.ISession.certAccount != null) {
```
t1m0thyj, line 1087: If someone has both types of cert defined in the same profile, then `certAccount` will take precedence, and `certFile`/`certKeyFile` will be ignored.

I don't believe this is a breaking change - new behavior seems ok as long as its documented 👍

Should we consider displaying a warning to the user in the case that both types of cert are defined?

packages/imperative/src/rest/src/session/doc/ISession.ts
```diff
@@ -138,6 +138,16 @@ export interface ISession {
      */
     passphrase?: string;
 
+    /**
+     * The account name for a PKCS#12 certificate stored in the credential manager.
```
t1m0thyj: The certs aren't stored as PKCS#12 in the OS credential manager. Should we say something like the following:
```suggestion
     * The subject name for a client certificate stored in the credential manager.
```

t1m0thyj (commented):
Thanks @JillieBeanSim! I've tried to address feedback on the native agent code. Left a few minor comments, but I'm excited to get this in for Zowe 3.5 😋

__types__/hogan.d.ts
JillieBeanSim: yes this fixed build failures related to hogan dep

packages/imperative/src/rest/src/client/AbstractRestClient.ts
JillieBeanSim: Hey @t1m0thyj i believe having precedence on the more secured path is best practice.  Do we want a warning in the output or a warning log message?

packages/imperative/src/rest/src/client/KeychainAgent.ts
JillieBeanSim: I can test the diff scenarios and see, thanks

JillieBeanSim: yes no special handling needed for exportable vs non-exportable, pushed changes in commit https://github.com/zowe/zowe-cli/pull/2688/commits/4bfb2424b828240342221add4fac9b5c8484e52d

packages/imperative/src/rest/src/client/AbstractRestClient.ts
JillieBeanSim: @t1m0thyj actually looking at the code, especially for linux platform I wonder if we should reverse the precedence because if not darwin or win32 an error is thrown and certFile/certKey are ignored if all 3 in profile.

JillieBeanSim: updated with checks for certFile & certKey for linux with log message and move forward with alternate cert authentication

t1m0thyj (approved):

traeok (approved):
Tested after the latest round of changes and functionality seems to be working as expected w/ `certAccount` and a z/OSMF test profile. LGTM

t1m0thyj:
The `uuid` audit failure should be ok to ignore for now - it only applies to `@zowe/cli-test-utils` package. Updating to the fixed version is tricky due to some breaking changes, which I believe @awharn is looking into 😋

__types__/hogan.d.ts
zFernand0: I'm not going to claim to understand what's happening here.

But as a quick test, I reverted these changes, deleted node_modules, and install+built the project and everything worked just fine 🤔 

Is this because some upstream application expects us to bundle hogan?

My main concern is that if we do this, we will be on the hook for changing  this file every time hogan makes a change. And I don't think we should be doing that. 😓

packages/imperative/package.json
```diff
@@ -47,6 +47,7 @@
   "dependencies": {
     "@types/semver": "^7.5.0",
     "@types/yargs": "^17.0.32",
+    "agent-base": "^7.0.0",
```
zFernand0: It seems that `agent-base` 7.x is their last CommonJS version. (source: [agent-base changelog](https://github.com/TooTallNate/proxy-agents/blob/main/packages/agent-base/CHANGELOG.md))
Have we considered alternative packages that provide more active support for their CommonJS distribution?

Would there be any issues from the scanning utilities used by ISVs? I'm concerned that their 7.x distribution stream will fall behind rather quickly given that it was last published 10 months ago. So they may not be adjusting to the onslaught of attacks to the NPM ecosystem in the last month alone.

There is a positive aspect about this package (compared to other alternatives). It does not have any dependencies 🙏 

Consider going with the latest 7.x version 🙏 
```suggestion
    "agent-base": "^7.1.4",
```

packages/secrets/core/src/os/win/tls_pipe.rs
```diff
@@ -0,0 +1,234 @@
...
+        0x80090330 => " [SEC_E_DECRYPT_FAILURE: TLS record decryption failure]",
+        0x80090331 => " [SEC_E_ALGORITHM_MISMATCH: no cipher suites in common with the server]",
+        0x8009030E => " [SEC_E_NO_CREDENTIALS: no credentials found — \
+                        certificate or private key may not be in the correct store]",
+        0x800B0101 => " [CERT_E_EXPIRED: a certificate in the chain has expired]",
+        0x800B0109 => " [CERT_E_UNTRUSTEDROOT: certificate chain terminated in an untrusted root]",
+        0x800B010A => " [CERT_E_CHAINING: certificate chain could not be built to a trusted root — \
+                        intermediate CA certificates may be missing from the Windows cert store]",
+        0x800B0110 => " [CERT_E_WRONG_USAGE: certificate is not valid for client authentication]",
+        _ => "",
```
zFernand0, line 36: I love how thorough this error list is.

I do wonder if we plan to have automated tests for each and every of these?

If not, should we trim down the number to the subset we do test for? An perhaps leave the rest as a generic channel error?

packages/secrets/core/Cargo.lock
zFernand0, line 1: Are there any concerns with the number of added packages here?

I'd be curious to see what ISVs Scanning tools show here.

packages/secrets/core/src/os/mac/tls_pipe.rs
```diff
@@ -0,0 +1,239 @@
...
+        -9813 => " [errSSLNoRootCert: certificate chain not anchored in a trusted root CA]",
+        -9814 => " [errSSLCertExpired: a certificate in the chain has expired]",
+        -9815 => " [errSSLCertNotYetValid: a certificate in the chain is not yet valid]",
+        -9825 => " [errSSLPeerBadCert: server reported the client certificate is malformed]",
+        -9826 => " [errSSLPeerUnsupportedCert: server does not support the client certificate type]",
+        -9827 => " [errSSLPeerCertRevoked: server reported the client certificate as revoked]",
+        -9829 => " [errSSLPeerCertUnknown: server does not recognize the client certificate]",
+        -9831 => " [errSSLPeerUnknownCA: server cannot verify the certificate chain to a known CA — \
+                   intermediate CA certificates may be missing from the keychain]",
+        _ => "",
```
zFernand0, line 46: Similar to a previous comment...

Do we have (or plan to have) automated tests for each and every one of these error scenarios?

If not, should we consider having a subset of these and provide a more generic error message?

packages/secrets/src/keyring/Cargo.lock
zFernand0, line 1: Again, curious what scanning tools say about these new packages added here 🙏

__types__/hogan.d.ts
zFernand0: Requested change:
- Please undo these changes or share steps to reproduce the build error (and a screenshot).

packages/secrets/core/src/os/win/tests.rs
zFernand0, line 1: Should we have some basic tests for  MacOS (in`os/mac`)?

zFernand0 (commented):
LGTM! 😋 
---

It just works! 🥳 

Thanks for al the work on this.

I left some comments, but most of them can be handled in a PR at a later time.

---

My only request is to please reconsider the Hogan changes.

__types__/hogan.d.ts
JillieBeanSim: tested build and unit tests run with this reverted and looks good, the error I was experiencing may have been due to the native https which is no longer in the branch

packages/secrets/core/src/os/win/tests.rs
JillieBeanSim: addressed in commit https://github.com/zowe/zowe-cli/pull/2688/commits/dadce2adeb6eafa0e0fcacffe2e0e45e20952ccb

zFernand0 (approved):
LGTM! 😋

zFernand0:
Audit workflow will be fixed in a follow up PR 🙏

Issue #2687, opened 2026-03-03
[zosuss] Allow running ssh commands in exec mode for improved performance


t1m0thyj:
**Is your feature or enhancement request related to a problem or limitation? Please describe**

I wish I was able to issue USS (SSH) commands in exec mode, rather than being restricted to shell mode.

**Describe your enhancement idea**

Expose an option in the z/OS USS SDK and on the CLI command `zowe uss issue cmd` to run in exec mode.

The default behavior should remain shell mode for backward compatibility, since some commands may require a login shell.

**Describe alternatives you've considered**

Use the `ssh` CLI instead

**Provide any additional context**

https://www.npmjs.com/package/ssh2#Client-methods

PR #2689, opened 2026-03-04
Add exec mode for running ssh commands

likhithanimma1:
**What It Does**
Solves https://github.com/zowe/zowe-cli/issues/2687

**How to Test**

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (build number: 2010)

**Additional Comments**

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+
+## Recent Changes
+
+- BugFix: Fixed an issue on Windows where non-normalized drive letter casing caused configuration load failures. [#2699](https://github.com/zowe/zowe-cli/pull/2699)
+
 ## `8.30.4`
 
 - BugFix: Updated `flatted` dependency for technical currency. [#2700](https://github.com/zowe/zowe-cli/pull/2700)
--- a/packages/imperative/src/config/__tests__/Config.unit.test.ts
+++ b/packages/imperative/src/config/__tests__/Config.unit.test.ts
@@ -874,4 +874,45 @@ describe("Config tests", () => {
             expect(spyOnFsWriteFileSync).not.toHaveBeenCalled();
         });
     });
+
+    describe("reload path normalization", () => {
+        it("should normalize projectDir using realpathSync.native on Windows", async () => {
+            Object.defineProperty(process, 'platform', { value: 'win32', configurable: true });
+
+            const rawPath = "C:/users/test/../project";
+            const normalizedPath = "C:\\users\\project";
+            const realPathSpy = jest.spyOn(fs.realpathSync, "native").mockReturnValue(normalizedPath as any);
+
+            const config = new (Config as any)();
+            config.mApp = MY_APP;
+            config.mLayers = [];
+
+            jest.spyOn(ConfigLayers.prototype, "read").mockImplementation(() => {});
+            jest.spyOn(ConfigSecure.prototype, "load").mockResolvedValue(undefined);
+
+            await config.reload({ projectDir: rawPath });
+
+            expect(realPathSpy).toHaveBeenCalledWith(rawPath);
+            expect(config.mProjectDir).toBe(normalizedPath);
+
+            Object.defineProperty(process, 'platform', { value: os.platform(), configurable: true });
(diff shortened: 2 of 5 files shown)


packages/zosuss/src/Shell.ts
```diff
@@ -139,6 +144,106 @@ export class Shell {
...
+     * @param cwd - Working directory
+     * @param stdoutHandler - Handler for stdout data
+     * @returns Promise with exit code
+     */
+    public static async executeExecCwd(session: SshSession,
+        command: string,
+        cwd: string,
+        stdoutHandler: (data: string) => void
+    ): Promise<any> {
+        const cwdCommand = `cd ${cwd} && ${command}`;
```
traeok: Considering that this is new code and flagged by GitHub security, can we add quotes to prevent command injection?

```suggestion
        const cwdCommand = `cd "${cwd}" && ${command}`;
```

packages/zosuss/src/Shell.ts
```diff
@@ -139,6 +144,106 @@ export class Shell {
...
+                } else if (err.message.includes(this.connRefusedFlag)) {
+                    reject(new ImperativeError({
+                        msg: ZosUssMessages.connectionRefused.message + ":\n" + err.message
+                    }));
+                } else {
+                    reject(new ImperativeError({
+                        msg: ZosUssMessages.unexpected.message + ":\n" + err.message
+                    }));
+                }
+            });
```
traeok, line 180: This logic is an exact copy of the handler in `executeSsh` above, which adds code duplication. It also makes future changes in this area less convenient to maintain, as we would have to apply the same fix twice. To adhere to the DRY principle, can we extract this handler into a private static method so we can call it from both `executeSsh` and `executeExec`?

packages/zosuss/src/Shell.ts
```diff
@@ -139,6 +144,106 @@ export class Shell {
...
+    /**
+     * Execute a command using exec mode (no shell wrapping, faster)
+     * @param session - SSH session
+     * @param command - Command to execute
+     * @param stdoutHandler - Handler for stdout data
+     * @returns Promise with exit code
+     */
+    public static executeExec(session: SshSession,
+        command: string,
+        stdoutHandler: (data: string) => void): Promise<any> {
```
traeok: The return type `Promise<any>` feels a bit loose - I know its based off the existing pattern in `executeSsh`, but when we resolve the promise, we are returning `rc` (which is a number). With this in mind, I think it could be more strongly typed as `Promise<number>`.

packages/zosuss/src/Shell.ts
likhithanimma1: Addressed

packages/zosuss/src/Shell.ts
likhithanimma1: I have extracted it into seperate private static method :)

packages/zosuss/src/Shell.ts
likhithanimma1: Thankyou @traeok . I have added the handler level tests and also system tests. Can test the PR now.

packages/cli/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added `--exec` flag to `zosuss` for faster command execution. Exec mode bypasses shell initialization and profile loading, providing up performance improvement. Shell mode remains the default for backward compatibility.
```
traeok: Please add the issue link to the changelog entry:

```suggestion
- Enhancement: Added `--exec` flag to `zosuss` for faster command execution. Exec mode bypasses shell initialization and profile loading, providing a performance improvement. Shell mode remains the default for backward compatibility. [#2687](https://github.com/zowe/zowe-cli/issues/2687)
```

packages/cli/__tests__/zosuss/__unit__/issue/ssh/Ssh.handler.unit.test.ts
zFernand0, line 1: There are a few lint warnings in this file 😅 
If there are unused variables, please make sure they have `_`  in front 🙏 

Image

PR #2697, opened 2026-03-10
Add Queue Utility + Implement REST Request Queues

### LINKED ISSUE #2543: Implement SDK-wide support for request throttling / max concurrent requests

In some corporate environments, mainframe systems can get bogged down through multiple concurrent requests. We should offer the ability to throttle the amount of maximum concurrent requests made by Zowe SDKs to prevent denial of service.

Implement a method to throttle the amount of maximum concurrent requests issued by the Zowe SDKs. Expose the option so that it can be configured by API consumers/downstream developers.

The plug-in / extension / downstream consumer can implement throttling logic, but it feels like something that could be done as part of SDK request handling.

awharn:
**What It Does**

Adds a queue utility to provide a promise queue.
Implements the promise queue framework in the `AbstractRestClient` to make it easier for extenders to consume.
Fully implements queuing in the `ZosmfRestClient`.
Fixes #2543 

**How to Test**

[test-request-queue.ts](https://github.com/user-attachments/files/25883805/test-request-queue.ts)
Modify and run the test-request-queue script provided.

Before behavior - comment out line 29 - `ZosmfRestClient.setThrottlingOptions({ maxConcurrentRequests: 2 });`
After behavior - restore line 29

Observe the number of created address spaces with throttling options set is considerably lower when compared to the before behavior.

Note: This will create >2 address spaces, but 2 will only ever be in progress via z/OSMF at any given time. The additional ones will abend S222 due to the address space getting too large. This is a known z/OSMF behavior, and the address spaces are reused til they are exhausted. However, when the throttling is in place, fewer overall address spaces will be created, compared to when there is no request throttling.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

This *technically* introduces the following breaking changes for consumers of the z/OSMF REST Client:
- Default maximum concurrent requests reduced from infinity to 2^31-1 (~2.147 billion).
- Requests will time out if waiting in the queue for 2^31-1 (~2.147 billion) milliseconds (~24.855 days).

We do not expect users to encounter breaking behavior, given the extremes.


packages/imperative/src/utilities/src/queue/Queue.ts
```diff
@@ -0,0 +1,112 @@
...
+import { ImperativeError } from "../../../error";
+import type { IQueueItem } from "./doc/IQueueItem";
+import type { IQueuePool } from "./doc/IQueuePool";
+import type { IQueueThrottleOptions } from "./doc/IQueueThrottleOptions";
+
+export class Queue {
+    private readonly defaultQueue: string = "default";
+    private mQueueTimeout: number = Constants.MAX_SIGNED_32BIT_NUMBER;
+    private mMaxConcurrentRequests: number = Constants.MAX_SIGNED_32BIT_NUMBER;
+    private mQueue: IQueuePool = {};
```
t1m0thyj: (nit) Since the `IQueuePool` type is a fairly simple one-line type and seems to be used only once for this private variable, it may not need to be defined in a separate file:
```suggestion
    private mQueue: {[key: string]: IQueuePoolEntry} = {};
```

packages/core/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to the Zowe core SDK package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added a `requestQueue` to the `ZosmfRestClient` class. [#2543](https://github.com/zowe/zowe-cli/pull/2543)
```
anaxceron: We should describe what `requestQueue` is.

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added a `Queue` utility to create and manage the processing of `Promise` queues. [#2543](https://github.com/zowe/zowe-cli/pull/2543)
+- Enhancement: Added a `requestQueue` getter to `AbstractRestClient` to allow extenders to create and manage a queue of REST requests to help prevent server overload. [#2543](https://github.com/zowe/zowe-cli/pull/2543)
```
anaxceron: We should describe what `AbstractRestClient` is

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
+- Enhancement: Added handling for `requestQueue` to the `AbstractRestClient`'s `request` function, if the `requestQueue` is present. [#2543](https://github.com/zowe/zowe-cli/pull/2543)
```
anaxceron: We should describe what `requestQueue` and `AbstractRestClient` are

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -287,14 +288,47 @@ export abstract class AbstractRestClient {
     }
 
     /**
-     * Perform the actual http REST call with appropriate user input
+     * Getter for the request queue, if one exists.
+     * Must be implemented at the end Rest Client level to be static.
+     * @memberof AbstractRestClient
+     * @returns {Queue}
+     */
+    protected get requestQueue(): Queue {
```
traeok: Should we explicitly type this as `Queue | undefined` so the caller knows that the queue could come back as falsy?

packages/imperative/src/rest/src/client/AbstractRestClient.ts
```diff
@@ -287,14 +288,47 @@ export abstract class AbstractRestClient {
...
+        if (this.requestQueue) {
+            let requestPool: string = undefined;
+            if (this.session.ISession.hostname) {
+                requestPool = this.session.ISession.hostname;
+                if (this.session.ISession.port) { requestPool += ":" + this.session.ISession.port.toString(); }
+            }
+            return this.requestQueue.enqueue(this._request.bind(this, options), requestPool);
+        } else {
+            return this._request(options);
+        }
```
traeok: The `else` part of this `if-else` can be removed as the `if` block returns


```suggestion
        return this._request(options);
```

packages/core/src/rest/ZosmfRestClient.ts
```diff
@@ -24,6 +25,44 @@ import { ZosmfHeaders } from "./ZosmfHeaders";
...
+        }
+        else if (options) { ZosmfRestClient.mRequestQueue.setThrottlingOptions(options); }
+    }
+
+    /**
+     * Reimplementation of the request queue getter to get the static request queue.
+     * @type {Queue}
+     * @memberof ZosmfRestClient
+     */
+    protected get requestQueue(): Queue {
```
traeok, line 53: See comment about making this more explicit (`Queue | undefined`)

packages/imperative/src/utilities/src/queue/Queue.ts
```diff
@@ -0,0 +1,112 @@
...
+        // Exit if there are no more requests in the queue or we are at the maximum number of requests in progress.
+        if (!this.mQueue[queue].requestPool.length || this.mQueue[queue].inProgress >= this.mMaxConcurrentRequests) { return; }
+
+        // Get the next item to process, and increment the inProgress counter.
+        const execItem = this.mQueue[queue].requestPool.shift();
+        this.mQueue[queue].inProgress++;
+
+        try {
+            // Try to process the item. If it timed out in the queue, throw an error and pick up the next one.
+            if (execItem.timedOut) { throw new ImperativeError({msg: "Request timed out while in the request queue."}); }
```
traeok, line 95: If `execItem.timedOut` is true, the code throws without calling `clearTimeout`. The `setTimeout` callback has already fired so the timer reference leaks (Node.js returns a `Timeout` object w/ references as part of `setTimeout`). Even though the callback has fired, its good practice to always clear the timeout.

I'd recommend either calling `clearTimeout(execItem.timeout)` unconditionally before the `if` branch, or in the `finally` block of `dequeue`.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
traeok, line 427: If the connection is reset, we retry the request here. However, now that there is queuing in `request`, the retry attempt is added to the end of the queue. With concurrency set to 1, this would push the retry to the back of the queue rather than being retried immediately. As a result, other requests in this queue may run before that request gets the opportunity to retry.

Should a retry execute immediately without re-entering the queue? Or is this current behavior expected?

packages/core/src/rest/ZosmfRestClient.ts
```diff
@@ -24,6 +25,44 @@ import { ZosmfHeaders } from "./ZosmfHeaders";
...
+     * Create a request queue if one does not exist. Set the throttling options if provided and they do exist.
+     * @param { IQueueThrottleOptions } options - The throttling options to apply to the z/OSMF REST client
+     */
+    private static verifyRequestQueue(options: IQueueThrottleOptions = {}): void {
+        // Create the queue if it is missing. If it is not missing, and we have options, set those options on the queue.
+        if (!ZosmfRestClient.mRequestQueue) {
+            ZosmfRestClient.mRequestQueue = new Queue(options);
+        }
+        else if (options) { ZosmfRestClient.mRequestQueue.setThrottlingOptions(options); }
+    }
```
traeok, line 46: The `else if (options)` is always true when `options` is `{}`. Should this parameter's default be `undefined`, or could `options` be changed to be optional?

If we want to keep the default, consider changing the check to `Object.keys(options).length > 0`.

packages/imperative/src/utilities/src/queue/doc/IQueueItem.ts
```diff
@@ -0,0 +1,18 @@
...
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+export interface IQueueItem {
+    func: Function,
+    timeout: NodeJS.Timeout,
+    timedOut: boolean,
+    resolve: any
+    reject: any
```
traeok: `resolve` and `reject` are using explicit `any` for the types, can we narrow this a bit to match the `Promise` class expectations?

```suggestion
    resolve: (value: unknown) => void,
    reject: (reason?: unknown) => void
```

packages/imperative/src/utilities/src/queue/Queue.ts
```diff
@@ -0,0 +1,112 @@
...
+*
+*/
+
+import { Constants } from "../../../constants";
+import { ImperativeError } from "../../../error";
+import type { IQueueItem } from "./doc/IQueueItem";
+import type { IQueuePoolEntry } from "./doc/IQueuePoolEntry";
+import type { IQueueThrottleOptions } from "./doc/IQueueThrottleOptions";
+
+export class Queue {
```
traeok, line 18: One thing that stood out to me: there is no way to cancel or drain the queue of its requests. This may be acceptable for now, but it could come in handy down the line. No requests with this comment, just an observation.

packages/imperative/src/utilities/src/queue/Queue.ts
```diff
@@ -0,0 +1,112 @@
...
+    private mMaxConcurrentRequests: number = Constants.MAX_SIGNED_32BIT_NUMBER;
+    private mQueue: {[key: string]: IQueuePoolEntry} = {};
+
+    constructor(options?: IQueueThrottleOptions) {
+        if (options) { this.setThrottlingOptions(options); }
+        this.createQueue();
+    }
+
+    /**
+     * Changes and existing request queue's throttling options.
```
traeok:
```suggestion
     * Changes an existing request queue's throttling options.
```

packages/imperative/src/utilities/src/queue/index.ts
```diff
@@ -0,0 +1,13 @@
...
+* Eclipse Public License v2.0 which accompanies this distribution, and is available at
+* https://www.eclipse.org/legal/epl-v20.html
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+export * from "./Queue";
```
gejohnston, line 12: Do we intend to deliver the general queueing functionality to all SDK consumers at this time, or do we just need it as a capability within imperative? If we do not export it, we will be free to make changes without worrying about breaking changes. Either choice is legitimate. I am just trying to ensure that we have thought about it.

awharn: We could do either, but I did need to export it to set it up on the z/OSMF REST client.

packages/imperative/src/rest/src/client/AbstractRestClient.ts
awharn: This behavior is not completely expected. I will change it to call `_request` instead. It may cause one extra item to run, but I think this is better and more expected behavior.

packages/core/src/rest/ZosmfRestClient.ts
awharn: In this case, do we want to do that? verifyRequestQueue ensures a queue exists.

traeok: Discussed offline and determined that the type makes sense in the `AbstractRestClient`, but not in the z/OSMF client. Marking as resolved.

packages/imperative/src/utilities/src/queue/index.ts
awharn: I am almost certain we will need to export it in order for third party SDKs and rest clients (including ours) to consume it. Since we have to implement it in the ZosmfRestClient level due to making it static per REST client, our only options are to export it, or move the ZosmfRestClient into Imperative and re-export it from core.

packages/imperative/src/utilities/src/queue/Queue.ts
awharn: I agree with this observation. Down the road, we can add a function to remove something from the queue. The current implementation has the existing behavior, where once you initiate a REST request, you cannot cancel it. It just becomes more obvious in this case, since now we have a list of the requests, whereas before, they just disappeared into the aether and existed in the code as unresolved promises.

packages/imperative/src/utilities/src/queue/index.ts
gejohnston: Ok. thanks for thinking about it.

Issue #2695, opened 2026-03-09
`ProfileInfo.mergeArgsForProfile` sometimes misses secure base profile properties


traeok:
**Describe the bug**

Say that I have a service profile with an extender profile type, and I have it inheriting properties from my `project_base` profile in my project-level team configuration. Example of my `profiles` section of that team config:

```json
"extenderprofile": {
  "type": "extendertype",
  "properties": {
    "basePath": "extender/api/v1",
    "authOrder": "token,bearer"
  },
  "secure": []
},
"project_base": {
  "type": "base",
  "properties": {
    "host": "hostname",
    "port": 12345,
    "rejectUnauthorized": false,
    "tokenType": "apimlAuthenticationToken",
    "user": "trae"
  },
  "secure": ["tokenValue"]
}
```

The `defaults` section has `project_base` as the default base profile, and `extenderprofile` as the default for the extender's profile type.

When calling `ProfileInfo.mergeArgsForProfile` with the attributes returned by `ProfileInfo.getDefaultProfile`, `tokenValue` is not returned in `knownArgs` or `missingArgs` - even if `getSecureVals` is `true`.

**Expected and actual results**

**Expected:** `ProfileInfo.mergeArgsForProfile(profAttrs, { getSecureVals: true })` should return `tokenValue` in its `knownArgs` array if stored in the `project_base` profile.

```typescript
this.profileApi = new ProfileInfo("zowe", {
    credMgrOverride: ProfileCredentials.defaultCredMgrWithKeytar(() => require("@zowe/secrets-for-zowe-sdk").keyring),
});

// ... later on

await this.profileApi.readProfilesFromDisk({ projectDir });

const profAttrs = this.profileApi.getDefaultProfile(Constants.PROFILE_TYPE);
if (profAttrs == null) {
    return;
}
this.logger.debug("getDefaultProfile response: " + JSON.stringify(profAttrs));

const mergedArgs = this.profileApi.mergeArgsForProfile(profAttrs, { getSecureVals: true });
this.logger.debug("mergeArgsForProfile response: " + JSON.stringify(mergedArgs));
```

**Actual:** `ProfileInfo.mergeArgsForProfile` does not return `tokenValue` in `knownArgs` or `missingArgs` in this scenario.

<img width="744" height="434" alt="Image" src="https://github.com/user-attachments/assets/8bddba4c-9833-4504-b6cc-bab0c1b7b0dd" />

**Describe your environment**

WIP

[...]

PR #2699, opened 2026-03-12
Windows path normalization

jace-roell:
**What It Does**
- Fixed an issue on Windows where non-normalized drive letter casing caused configuration load failures.

**How to Test**

On a windows machine. run the test scenario described in https://github.com/zowe/zowe-cli/issues/2695

The tokenValue and other secure properties from the base profile should be properly populated.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/src/config/src/Config.ts
+++ b/packages/imperative/src/config/src/Config.ts
@@ -182,7 +182,7 @@ export class Config {
      * @throws An ImperativeError if the configuration does not load successfully
      */
     public async reload(opts?: IConfigOpts) {
-        // Normalize projectDir for Windows before loading the config
+        // Normalize projectDir and homeDir for Windows before loading the config
         if (process.platform === 'win32') {
             if(opts?.projectDir) {
                 opts.projectDir = fs.realpathSync.native(opts.projectDir);


packages/imperative/src/config/src/Config.ts
```diff
@@ -182,6 +182,16 @@ export class Config {
      * @throws An ImperativeError if the configuration does not load successfully
      */
     public async reload(opts?: IConfigOpts) {
+        // Normalize projectDir for Windows before loading the config
+        if (process.platform === 'win32') {
+            if(opts?.projectDir){
+                opts.projectDir = fs.realpathSync.native(opts.projectDir);
+            }
+            if(opts?.homeDir){
```
traeok: (nitpick) consider making the code formatting more consistent here - there's a space after the `if` on line 186 (before curly) but this is not the case for the `if` statements on 187 and 190

packages/imperative/src/config/__tests__/Config.unit.test.ts
```diff
@@ -874,4 +874,45 @@ describe("Config tests", () => {
             expect(spyOnFsWriteFileSync).not.toHaveBeenCalled();
         });
     });
+
+    describe("reload path normalization", () => {
+        it("should normalize projectDir using realpathSync.native on Windows", async () => {
+            Object.defineProperty(process, 'platform', { value: 'win32', configurable: true });
```
traeok, line 880: Consider using `jest.replaceProperty` here instead:


```suggestion
            jest.replaceProperty(process, 'platform' as any, 'win32');
```

This ensures that the `process.platform` value is cleaned up by Jest if this test fails from a future change (isolates test failures to a single suite). It also avoids the need to restore the value on line 898.

This pattern also occurs within the other tests in this suite.

traeok: Disregard this suggestion, we found that since `process.platform` is read-only, we cannot use the Jest function. `Object.defineProperty` is acceptable in this case.

packages/imperative/src/rest/src/session/ConnectionPropsForSessCfg.ts
```diff
@@ -171,13 +171,21 @@ export class ConnectionPropsForSessCfg {
...
+                !sessCfgToUse.authTypeOrder.includes(SessConstants.AUTH_TYPE_BASIC);
+
             switch (sessCfgToUse.authTypeOrder[0]) {
                 case SessConstants.AUTH_TYPE_BASIC:
                     if (!sessCfgToUse._authCache?.availableCreds?.user && !doNotPromptForValues.includes("user")) {
                         promptForValues.push("user");
                     }
-                    if (!sessCfgToUse._authCache?.availableCreds?.password && !doNotPromptForValues.includes("password")) {
-                        promptForValues.push("password");
+                    if (!passwordNotRequiredByAuthOrder) {
```
gejohnston: The implication is that if "basic" is not in a non-default authOrder, that we should not prompt for user and password. The new comment reinforces that idea. If so, shouldn't the test for passwordNotRequiredByAuthOrder encapsulate both the push of "password" and the push of "user"? Or am I missing some nuance?

jace-roell: After further offline talk, it seems we can remove the auth order check entirely. Thanks!

packages/imperative/src/imperative/src/auth/__tests__/BaseAuthHandler.config.unit.test.ts
```diff
@@ -47,6 +47,8 @@ describe("BaseAuthHandler config", () => {
     });
 
     beforeEach(() => {
+        jest.spyOn(fs, 'realpathSync').mockImplementation((path: any) => path);
+        (fs.realpathSync as any).native = jest.fn().mockImplementation((path: any) => path);
```
zFernand0, line 51: Apologies if this was already discussed...

Can we use `jest.spyOn` instead? 

`jest.spyOn(fs.readpathSync, "native")...`

packages/imperative/src/imperative/src/auth/__tests__/BaseAuthHandler.unit.test.ts
```diff
@@ -83,6 +84,8 @@ describe("BaseAuthHandler", () => {
                 }
             })
         });
+        jest.spyOn(fs, 'realpathSync').mockImplementation((path: any) => path);
+        (fs.realpathSync as any).native = jest.fn().mockImplementation((path: any) => path);
```
zFernand0, line 88: Same as previous comment...

Although rare, but I'm mostly worries about other tests that may rely on the actual implementation of `native` 😋

packages/imperative/src/imperative/src/auth/__tests__/BaseAuthHandler.config.unit.test.ts
jace-roell: I had tried that on a previous commit and attempting to access the function directly didn't seem to work. [1a293e1](https://github.com/zowe/zowe-cli/pull/2699/commits/1a293e1c1c9a9872eb51ec5dab7fcd0c48d0267e)

zFernand0: Thanks for trying it out 🙏

PR #2712, opened 2026-04-08
feat(config): add --merge and --dry-run flags to zowe config import

chipset:
## Summary

- **`--merge` (`-mg`)**: Merges properties from an imported config into the existing `zowe.config.json` instead of overwriting it. Existing values win on all conflicts — profiles are deep-merged, defaults use existing keys, plugins are appended without duplication, and `autoStore` is only set if not already defined.
- **`--dry-run` (`-dr`)**: Previews the resulting config without writing anything to disk. Works standalone (preview a full overwrite) or combined with `--merge` (preview a safe merge). Schema download is skipped during dry runs.
- Updated the "file already exists" skip message to advertise `--merge` and `--dry-run` alongside the existing `--overwrite` hint.
- Added `CHANGES.md` documenting the new behaviour, merge semantics, and a behaviour matrix.

## Motivation

`zowe config import` previously offered only two modes: skip (if the target file already existed) or overwrite (`--overwrite`). Users with an existing `zowe.config.json` had no safe way to adopt new defaults or profiles from a shared/team configuration without losing their own settings. This PR closes that gap.

## Behaviour matrix

| Flag(s) | File exists? | Result |
|---|---|---|
| _(none)_ | No | Write imported config as-is |
| _(none)_ | Yes | Skip (no change) |
| `--overwrite` | Yes | Overwrite existing file |
| `--merge` | No | Write imported config as-is |
| `--merge` | Yes | Deep-merge; **existing values win** |
| `--dry-run` | Any | Print result to console, no disk write |
| `--merge --dry-run` | Yes | Print merge preview, no disk write |
| `--merge --overwrite` | Any | Error (mutually exclusive) |

## Merge semantics (field-by-field)

| Field | Rule |
|---|---|
| `profiles` | Deep-merged via `lodash.mergeWith`; existing property values and arrays are preserved |
| `defaults` | Shallow spread — existing keys are never overwritten; only missing keys are added |
| `plugins` | Additive — new plugin names appended only if not already in the list |
| `autoStore` | Only set from imported config if the target does not already define it |

## Test plan

- [x] 10 new unit tests in `import.handler.unit.test.ts` covering all flag combinations, mutual exclusion, skip message text, dry-run output, and field-level merge correctness
- [x] 5 new integration tests in the CLI config import subtest file covering help output, `--dry-run`, `--merge --dry-run`, `--merge` on existing file, and `--merge` first-time import
- [x] New fixture `test.config.for.merge.json` with overlapping and new profiles/defaults used by integration tests
- [ ] Manual smoke test: `zowe config import <url> --merge --dry-run` against a live profile

## Files changed

| File | Change |
|---|---|
| `src/config/cmd/import/import.definition.ts` | Added `--merge` and `--dry-run` options and three new usage examples |
| `src/config/cmd/import/import.handler.ts` | Implemented safe merge logic, dry-run preview, mutual exclusion guard, updated skip message |
| `src/config/cmd/import/CHANGES.md` | New — documents all changes |
| `__tests__/config/cmd/import/import.handler.unit.test.ts` | 10 new unit tests |
| `__tests__/__integration__/.../cli/config/import/cli.imperative-test-cli.config.import.integration.subtest.ts` | 5 new integration tests + updated help assertion |
| `__tests__/__integration__/.../cli/config/import/__resources__/test.config.for.merge.json` | New integration test fixture |

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/imperative/src/imperative/src/config/cmd/import/import.handler.ts
+++ b/packages/imperative/src/imperative/src/config/cmd/import/import.handler.ts
@@ -16,7 +16,8 @@ import * as JSONC from "comment-json";
 import { ICommandHandler, IHandlerParameters } from "../../../../../cmd";
 import { ImperativeError } from "../../../../../error";
 import { ImperativeConfig, TextUtils } from "../../../../../utilities";
-import { IConfig } from "../../../../../config";
+import * as lodash from "lodash";
+import { ConfigConstants, IConfig, IConfigLayer } from "../../../../../config";
 import { AuthOrder, RestClient, Session, SessConstants } from "../../../../../rest";
 
 /**
@@ -41,9 +42,25 @@ export default class ImportHandler implements ICommandHandler {
         config.api.layers.activate(params.arguments.userConfig, params.arguments.globalConfig, configDir);
         const layer = config.api.layers.get();
 
-        if (layer.exists && !params.arguments.overwrite) {
-            params.response.console.log(`Skipping import because ${layer.path} already exists.\n` +
-                `Rerun the command with the --overwrite flag to import anyway.`);
+        const isDryRun = params.arguments.dryRun as boolean;
+        const isMerge = params.arguments.merge as boolean;
+        const isOverwrite = params.arguments.overwrite as boolean;
+
+        if (isMerge && isOverwrite) {
+            throw new ImperativeError({
+                msg: "The --merge and --overwrite options are mutually exclusive. " +
+                    "Use --merge to add missing properties to an existing config, " +
+                    "or --overwrite to replace it entirely."
+            });
+        }
+
+        // Guard: skip when the file already exists and neither --overwrite, --merge, nor --dry-run is given
+        if (layer.exists && !isOverwrite && !isMerge && !isDryRun) {
+            params.response.console.log(
+                `Skipping import because ${layer.path} already exists.\n` +
+                `Rerun the command with --overwrite to replace it, --merge to add missing ` +
+                `properties, or --dry-run to preview what would happen.`
+            );
             return;
         }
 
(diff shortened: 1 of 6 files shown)


chipset:
I tested this on my own system based upon a large customer's need to manage multiple configuration files for their developers.  Reach out to me if you need the names.    

Test fixtures are included.

packages/imperative/src/imperative/src/config/cmd/import/import.handler.ts
```diff
@@ -53,10 +70,66 @@ export default class ImportHandler implements ICommandHandler {
...
+        let previewLayer: IConfigLayer | undefined;
+        if (isMerge && layer.exists) {
+            // Safe merge: existing target values always win over incoming imported values.
+            // We operate on a clone for dry-run so nothing is mutated on disk.
+            const target: IConfigLayer = isDryRun
+                ? JSONC.parse(JSONC.stringify(layer, null, ConfigConstants.INDENT)) as unknown as IConfigLayer
+                : config.api.layers.get();
+
+            // profiles: deep-merge with existing winning — lodash.mergeWith(existing, incoming)
+            target.properties.profiles = lodash.mergeWith(
```
gejohnston, line 84: It appears that this logic will work for nexted configs. Has it been tested with nested configs?
Like having both of these profiles in a config:
```
target.properties.profiles.QA.profiles.zosmf
target.properties.profiles.DEV.profiles.zosmf
```

packages/imperative/src/imperative/src/config/cmd/import/import.handler.ts
```diff
@@ -53,10 +70,66 @@ export default class ImportHandler implements ICommandHandler {
...
+                {},
+                configJson.profiles,       // imported (lower priority, applied first)
+                target.properties.profiles, // existing (higher priority, applied second — wins)
+                (existingVal: any, _importedVal: any) => {
+                    // For arrays, keep the existing array as-is
+                    if (lodash.isArray(existingVal)) { return existingVal; }
```
gejohnston, line 90: If an imported config adds a new secure property (like `port`), that secure `port` property will not added. Further, if `port` was previously a plain text property, the plain text `port` property would not be removed. These choices are understandable and consistent, but:
- Would a site think that this behavior is a bug?
- Can we explain our behavior somewhere?

packages/imperative/src/imperative/src/config/cmd/import/import.handler.ts
```diff
@@ -53,10 +70,66 @@ export default class ImportHandler implements ICommandHandler {
...
+                }
+            );
+
+            // defaults: existing key/value pairs win; only add keys missing from target
+            target.properties.defaults = {
+                ...configJson.defaults,        // imported base
+                ...target.properties.defaults  // existing overrides (wins on conflict)
+            };
+
+            // plugins: add any new plugins from the import not already present
```
gejohnston, line 100: We add plugin names from the imported config to the target config.
My guess is that the function which we call later:
```
config.api.layers.set(target.properties)
```
stores the plugin names into plugins.json. Would that action have any real effect if we do not actually install the new plugins?

Further, since the value of the imported configJson.plugins property probably comes from the local plugins.json file (not from anything from the imported config file itself), does copying and storing the plugins really do anything at all? If not, we might choose to remove the entire plugin paragraph.

packages/imperative/src/imperative/src/config/cmd/import/import.handler.ts
```diff
@@ -69,10 +142,22 @@ export default class ImportHandler implements ICommandHandler {
...
 
-        // Write the active created/updated config layer
-        config.api.layers.write();
-
-        params.response.console.log(`Imported config${schemaImported ? " and schema" : ""} to ${layer.path}`);
+        if (isDryRun) {
+            // Print a preview of the would-be result without touching disk
+            params.response.console.log(
+                TextUtils.chalk.yellow("[Dry Run]") +
+                ` The following config would be written to ${layer.path}:\n`
```
gejohnston, line 149: We attempt to update plugins at line 100, but dryRun displays no indication that we will update the set of plugins. We should add a statement that we will update plugins. Of course, if we remove our attempt to update plugins (see earlier comment), No change is needed here.

gejohnston (commented):
Thanks for this effort. It looks like a viable, useful enhancement. With a pull request from a forked branch, the effort to confirm or fix the full set of unit, integration, and system tests, and to reach code coverage test objectives falls on the Zowe team. Our product management team will prioritize such an effort. Completion of this pull request will be undertaken based on that prioritization.

In the meantime, I posted a number of comments/questions that can be considered when this work is undertaken.

packages/imperative/src/imperative/src/config/cmd/import/import.handler.ts
McQuitty-Broadcom: I tested it with nested configs on my box.

zFernand0:
I think we still want to keep this PR 🙏

PR #2714, opened 2026-04-13
Remove error message displayed before valid JSON by the npm pack command

### LINKED ISSUE #2713: Fix npm pack parsing error
**Describe the bug**

The Zowe `NpmFunctions.getPackageInfo` function calls the `npm pack` command to get information from package.json. Due to the parallel execution of tests in integration and system test pipelines , `git` commands can lock git resources. This causes an error to be displayed by the `npm pack` before it outputs the valid JSON contents of package.json. The error text causes a JSON parse error when Zowe attempts to process the output of the `npm pack` command. The output looks something like this:
```
   error: could not lock config file .git/config: File exists
    [
      {
        "id": "@zowe/sample-plugin-for-zowe-cli@4.0.0",
        "name": "@zowe/sample-plugin-for-zowe-cli",
        "version": "4.0.0",
        "size":
    <More stuff omitted>
```

Zowe should discard such error text and process the valid JSON text.

**Describe your environment**

Run the build pipeline for the Zowe sample plug-in. The problem occurs very often on that pipeline. It appears that any other Zowe pipeline could encounter such a problem.

**Additional context**

gejohnston:
**What It Does**

In test pipelines, parallel tests use Git to retrieve source to compile and to run CodeQL analysis. Some conflict causes <YourSourceCodeRootDirectory>/.git/config.lock to be left around. As a result, when the pipeline spawns 'npm pack' to get the contents of package.json, an error message that looks like the following may be displayed before the correct contents of package.json.
```
"error: could not lock config file .git/config: File exists"
    or
"warning: unable to access '.git/config': Permission denied"
```
JSON.parse() fails to parse the output because of that error message. The typical result is that Zowe reports that it cannot install a plugin.

A successful 'npm pack' command always displays its results in an array. This pull request looks for the start of an array and removes any text that occurs before the array of valid JSON. This allows the JSON to be successfully parsed. The removed text is logged.

Fixes  #2713

**How to Test**

Because this problems occurs as an edge-case involving parallel operations, the only reliable way to create the problem is to hack the code in NpmFunctions.getPackageInfo to display an arbitrary message before the expected JSON output. The logic of this pull request will then remove that arbitrary message before the output of the `npm pack` command, allowing that output to be successfully parsed as JSON.

The real-world observation of this fix would be in the build pipeline of the Zowe sample plug-in, in which three to five jobs routinely fail. Currently, you can only make the sample plug-in pipeline succeed by repeatedly re-running each pipeline job. Unfortunately, the changes of this PR will not be included in the sample plug-in pipeline until this PR has been merged into the master branch of Zowe CLI.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Removed text which is sometimes displayed by the `npm pack --dry-run --json` command before its valid JSON output, thus avoiding a Zowe error stating that a plugin cannot be installed. This problem occurs mostly in build pipelines. [#2713](https://github.com/zowe/zowe-cli/issues/2713)
+
 ## `8.31.3`
 
 - BugFix: Increased the max buffer size of output for `npm pack` command run during plug-in installation. [#2708](https://github.com/zowe/zowe-cli/pull/2708)
--- a/packages/imperative/src/imperative/__tests__/plugins/utilities/NpmFunctions.unit.test.ts
+++ b/packages/imperative/src/imperative/__tests__/plugins/utilities/NpmFunctions.unit.test.ts
@@ -126,7 +126,6 @@ describe("NpmFunctions", () => {
         });
 
         it("should discard error message text before valid JSON output", () => {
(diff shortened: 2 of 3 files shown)


t1m0thyj (approved):
LGTM, thanks @gejohnston!

zFernand0 (approved):
LGTM! 😋 
---
Thanks for the fix! 🙏

PR #2732, opened 2026-05-22
added export functionality with automatic config redaction

jace-roell:
**What It Does**
Added `export-redacted` to the `config` command group

When troubleshooting issues, users frequently need to share their configuration files but do not want to expose any sensitive data. The process of redacting this info can be tedious and time consuming.

Redacted info will be replaced with angle brackets and the type of value it was. i.e `<string1>, <profile1>, <num1>, <bool1>`

Nested profiles in the defaults will be replaced with the same structure with the profile names matching the structure
i.e `"<profile1>.<profile2>"`

The command is customizable to adjust the amount of information that gets redacted. Run `zowe config export-redacted -h` for more information

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**


packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.handler.ts
```diff
@@ -0,0 +1,359 @@
...
+            throw new ImperativeError({
+                msg: `Failed to read config file '${activeLayer.path}': ${error.message}`,
+                causeErrors: error
+            });
+        }
+
+        const redacted: any = {};
+
+        for (const [rawKey, value] of Object.entries(originalConfig)) {
+            const key = rawKey.toLowerCase();
```
zFernand0, line 134: `autoStore` gets exported as `autostore` 😋 
should we avoid doing this?

packages/imperative/src/imperative/src/config/ConfigManagementFacility.ts
```diff
@@ -74,7 +75,8 @@ export class ConfigManagementFacility {
                 importDefinition,
                 convertProfilesDefinition,
                 ReportEnvDefinition,
-                updateSchemasDefinition
+                updateSchemasDefinition,
+                exportRedactedDefinition
```
zFernand0, line 79: should we add a new changelog entry for the new command group?

packages/imperative/src/imperative/src/config/cmd/exportRedacted/exportRedacted.handler.ts
```diff
@@ -0,0 +1,359 @@
...
+                                this.getOrCreateKey(nestedProfileName, "profile") : nestedProfileName;
+                            redactedProfile.profiles[redactedNestedName] = this.redactProfileObject(nestedProfileData, args);
+                        }
+                    }
+
+                    if (profile.secure && !args.hideSecureFields) {
+                        redactedProfile.secure = profile.secure;
+                    }
+
+                    const redactedProfileName = args.redactProfileNames ?
```
zFernand0: can we call the `redactProfileObject` function instead of duplicating most of it here?

packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.handler.ts
zFernand0, line 1: for consistency with other directories in this folder, can we rename the directory to `export-redacted`?

packages/imperative/src/imperative/src/config/cmd/exportRedacted/exportRedacted.handler.ts
```diff
@@ -0,0 +1,359 @@
...
+        }
+
+        if (profile.secure && !args.hideSecureFields) {
+            redactedProfile.secure = profile.secure;
+        }
+
+        return redactedProfile;
+    }
+
+    private redactValue(propertyName: string, value: any, schema: any, args: any): any {
```
zFernand0: nit: do we need `schema: any` here?

packages/imperative/src/imperative/src/config/cmd/exportRedacted/exportRedacted.handler.ts
```diff
@@ -0,0 +1,359 @@
...
+                });
+            }
+        }
+
+        return exportedFiles;
+    }
+
+    private async createRedactedConfig(layer: any, args: any): Promise<any> {
+        const teamConfig = this.profileInfo.getTeamConfig();
+        const activeLayer = layer || teamConfig.layerActive();
```
zFernand0: nit: since the method is private, and we are always checking for `layer.exists`, I think having a fallback to `layerActive() might be redundant. Meaning, we should be fine if we remove the fallback
```suggestion
        const activeLayer = layer;
```

zFernand0 (commented):
LGTM! 😋 
---
Functionality looks great! 🙏 

The redacting seems to happen on every property including those that are not secure.
Could we look at the `secureFields` and only redact the values of those properties?

packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.definition.ts
```diff
@@ -0,0 +1,85 @@
...
+    name: "export-redacted",
+    aliases: ["er"],
+    type: "command",
+    summary: "Export Zowe configuration with redacted values",
+    description: "Export your Zowe configuration with sensitive values redacted and common properties given matching keys for " +
+        "troubleshooting and sharing purposes.",
+    handler: join(__dirname, "exportRedacted.handler"),
+    options: [
+        {
+            name: "redact-strings",
```
traeok, line 25: Since this is a `boolean` type, this also registers a `--no-redact-strings` option. Does this imply that all strings are left unredacted? 

My concern is that leaving all strings unredacted may be perceived by security scans as an insecure parameter - should we consider always redacting the set of `DEFAULT_CENSORED_OPTIONS` and then skipping the other strings if this option is provided?

packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.handler.ts
```diff
@@ -0,0 +1,327 @@
...
+            const config = ImperativeConfig.instance.config;
+
+            if (!config.exists) {
+                throw new ImperativeError({
+                    msg: "No Zowe configuration found. Please initialize a configuration first using 'zowe config init'."
+                });
+            }
+
+            this.profileInfo = new ProfileInfo("zowe");
+            await this.profileInfo.readProfilesFromDisk();
```
traeok: This information should already be accessible by the time this command is run (see `ImperativeConfig.instance.config.layers`). Can we use this to avoid an extra load of the config?

packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.handler.ts
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    private stripJsonComments(content: string): string {
+        return content.replace(/("([^"\\]|\\.)*")|(\/\/.*|\/\*[\s\S]*?\*\/)/g, (match, g1) => {
+            if (g1) return g1;
+            return "";
+        });
+    }
+
+    private async writeToFile(content: string, filePath: string): Promise<void> {
```
traeok: There are no async functions called within this function, so `async` can be dropped from its signature:

```suggestion
    private writeToFile(content: string, filePath: string): void {
```

packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.handler.ts
```diff
@@ -0,0 +1,327 @@
...
+                    source: path.join(path.basename(path.dirname(layer.path)), path.basename(layer.path)),
+                    target: filePath
+                });
+            }
+        }
+
+        return exportedFiles;
+    }
+
+    private async createRedactedConfig(layer: any, args: any): Promise<any> {
```
traeok: `async` is not required here. Also, I'd recommend using more specific types than explicit `any` because this implies that any value can be passed in as `layer` or `args` by a caller within this class.

I've updated the suggestion to recommend narrowing the return type so it is casted to a valid type before its used (TypeScript best practice):

```suggestion
    private createRedactedConfig(layer: any, args: any): unknown {
```

packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.handler.ts
```diff
@@ -0,0 +1,327 @@
...
+
+        } catch (error) {
+            throw new ImperativeError({
+                msg: `Failed to export configuration: ${error.message}`,
+                causeErrors: error
+            });
+        }
+    }
+
+    private async exportToDirectory(exportDir: string, args: any): Promise<Array<{ source: string, target: string }>> {
```
traeok: Once `async` is removed from the other 2 functions in this file, it can be removed here too:


```suggestion
    private exportToDirectory(exportDir: string, args: any): Array<{ source: string, target: string }> {
```

traeok (changes requested):
Thanks for working on this Jace! I requested changes regarding the extra config load in the new command handler. Also, had some concerns about unredacted strings but still thinking about whether its truly an issue to offer an insecure parameter.

Working on some manual testing but the logic LGTM ^^

packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.handler.ts
```diff
@@ -0,0 +1,320 @@
...
+            const isDryRun = params.arguments.dryRun;
+
+            if (isDryRun) {
+                const layers = ImperativeConfig.instance.config.layers;
+                const dryRunOutputs: any = {};
+                let hasOutput = false;
+
+                for (const layer of layers) {
+                    if (layer.exists) {
+                        const redactedConfig = await this.createRedactedConfig(layer, params.arguments);
```
traeok: Can drop this `await` now 👍 
```suggestion
                        const redactedConfig = this.createRedactedConfig(layer, params.arguments);
```

packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.handler.ts
```diff
@@ -0,0 +1,320 @@
...
+                        }
+                        params.response.console.log(`--- ${sourceName} ---`);
+                        params.response.console.log(formattedOutput);
+                        dryRunOutputs[sourceName] = redactedConfig;
+                        hasOutput = true;
+                    }
+                }
+                params.response.data.setObj(dryRunOutputs);
+            } else {
+                const exportedFiles = await this.exportToDirectory(exportDir, params.arguments);
```
traeok:
```suggestion
                const exportedFiles = this.exportToDirectory(exportDir, params.arguments);
```

traeok (commented):
Two minor suggestions now that functions are no longer async 👍 
Also, tests are failing & patch coverage is below the threshold.

traeok (approved):
Tested and changes LGTM. Thanks Jace for the new command!

packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.handler.ts
zFernand0: @traeok @jace-roell 
I haven't tested this from a ZE perspective, but if we ever want to expose this functionality there...
Do we think the same holds true?

Theoretically it should work since ZE does create a minimal `ImperativeConfig.isntance.config` 🙏

packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.definition.ts
```diff
@@ -0,0 +1,85 @@
...
+            defaultValue: false
+        },
+        {
+            name: "hide-secure-fields",
+            description: "Hide secure field names from output (secure field names are shown by default since they don't contain sensitive values)",
+            type: "boolean",
+            defaultValue: false
+        },
+        {
+            name: "redact-profile-names",
```
zFernand0, line 53: Nit: Can we add aliases to all these properties? 🙏

packages/imperative/src/imperative/src/config/cmd/export-redacted/__tests__/exportRedacted.handler.unit.test.ts
```diff
@@ -0,0 +1,278 @@
...
+        const logCall = (mockParams.response.console.log as jest.Mock).mock.calls[1][0];
+        const result = JSON.parse(logCall);
+
+        const profileKey = Object.keys(result.profiles)[0];
+        const properties = result.profiles[profileKey].properties;
+
+        // Array elements should be mapped and redacted
+        expect(properties.arrayOfStrings).toBeInstanceOf(Array);
+        expect(properties.arrayOfStrings[0]).toMatch(/<str\d+>/);
+        expect(properties.arrayOfStrings[1]).toMatch(/<str\d+>/);
```
zFernand0, line 215: nit: 
We distinguish `user`, `basePath`, and `host`.
Can we do the same for `password`? 😋 

This is really not necessary.

packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.definition.ts
```diff
@@ -0,0 +1,85 @@
...
+    summary: "Export Zowe configuration with redacted values",
+    description: "Export your Zowe configuration with sensitive values redacted and common properties given matching keys for " +
+        "troubleshooting and sharing purposes.",
+    handler: join(__dirname, "exportRedacted.handler"),
+    options: [
+        {
+            name: "redact-strings",
+            description: "Redact string values with consistent keys",
+            type: "boolean",
+            defaultValue: true
```
zFernand0, line 29: Nit: 
Can we add a `--redact-host-path` option? 🙏 
Which will show the host and basePath values if set to false 😋

zFernand0 (approved):
LGTM! 😋 
---
none of my comments are required. I think this is great as-is. 🙏

packages/imperative/src/imperative/src/config/cmd/export-redacted/exportRedacted.handler.ts
traeok: In ZE, we would probably use a different helper to access the Config API, although as you mentioned we could potentially use the minimal shim that it sets up. But in the context of this function, we should be using `ImperativeConfig.instance.config` to access team config because its a command handler and only registered within Zowe CLI. If any code is exposed around exporting w/ redactions, we should probably configure those functions so that it takes in an instance of the `Config` API (dependency injection) to avoid hardwiring around a specific instance of the config class.

anaxceron (approved):

PR #2737, opened 2026-06-09
fix: inconsistent error messages when credentials fail to save

### LINKED ISSUE #1907: Inconsistent error messages when credentials fail to save error
Sometimes we use the `secureSaveError` helper method to show a pretty error message.

Other times when calling `config.save`, we only show "Failed to load Keytar module" which is not a particularly helpful message.

Maanvi212006:
fixes #1907

**What It Does**
- Found two call sites where `config.save()` was called without using the `secureSaveError` helper
      `ConfigAutoStore.ts` and `ConvertV1Profiles.ts`
` Users were seeing the raw unhelpful message `Failed to load Keytar module` instead of the consistent, user-friendly `Unable to securely save credentials message` with possible solutions
- Wrapped both `config.save()` calls in try/catch blocks that throw `ConfigUtils.secureSaveError()` on failure

**How to Test**
- Set up an environment where keytar fails to load (e.g. missing @zowe/secrets-for-zowe-sdk module)
- Trigger a credential save operation via zowe config secure or any command that auto-stores credentials

Before fix: error message shows Failed to load Keytar module: ...
After fix: error message shows Unable to securely save credentials. Possible Solutions: ...

**Review Checklist**
I certify that I have:
- [ ] updated the changelog

**Additional Comments**
`ConfigUtils` import was also added to `ConvertV1Profiles.ts` as it was not previously imported there.


zFernand0 (commented):
Changes LGTM! 😋 
---
I believe we are missing a few more instances of config.save where we could leverage this same try/catch + `throw ConfigUtils.secureSaveError();`

Image

This PR addresses the first two instances of the picture above 🙏 

👀 @t1m0thyj

zFernand0 (dismissed):
Forgot to request changes 🙏

Maanvi212006:
> ## Changes LGTM! 😋
> I believe we are missing a few more instances of config.save where we could leverage this same try/catch + `throw ConfigUtils.secureSaveError();`
> 
> Image
> This PR addresses the first two instances of the picture above 🙏
> 
> 👀 @t1m0thyj

@zFernand0 Thanks for the review!
I too dug into this. `BaseAuthHandler.ts` has 2 more unguarded `config.save()` calls. There's also the same pattern in `BaseAutoInitHandler.ts` (2), `init.handler.ts` (1), `secure.handler.ts` (1), and `set.handler.ts` (1) — none of them wrap save() in try/catch + secureSaveError().
Happy to extend this PR to cover all of them if that's preferred

t1m0thyj (changes requested):
Thanks for your work on this @Maanvi212006! Please resolve failing unit tests:
```
Errors thrown in packages/imperative/src/config/__tests__/ConfigAutoStore.unit.test.ts
    ● Test suite failed to run
  
      packages/imperative/src/config/__tests__/ConfigAutoStore.unit.test.ts:577:53 - error TS2304: Cannot find name 'handlerParams'.
  
      577             await ConfigAutoStore.storeSessCfgProps(handlerParams as any, {
                                                              ~~~~~~~~~~~~~
```

packages/imperative/src/config/src/ConfigAutoStore.ts
```diff
@@ -200,7 +200,11 @@ export class ConfigAutoStore {
             });
         }
 
-        await config.save();
+        try {
+            await config.save();
+        } catch (err) {
+            throw ConfigUtils.secureSaveError();
+        }
```
t1m0thyj, line 210: Not sure if we want to handle all errors returned by `config.save` as `secureSaveError` since there could be other errors unrelated to secure properties such as failing to write to  `zowe.config.json`. @zFernand0 Thoughts about how we should detect errors specifically related to saving secure props?

Maanvi212006: > Not sure if we want to handle all errors returned by `config.save` as `secureSaveError` since there could be other errors unrelated to secure properties such as failing to write to `zowe.config.json`. @zFernand0 Thoughts about how we should detect errors specifically related to saving secure props?

@t1m0thyj makes sense! Rather than catching all errors, we could check `config.api.secure.loadFailed` inside the catch block and only throw secureSaveError() then — otherwise rethrow the original error. 

Happy to update as per the decided approach!

zFernand0: Agree on the premise that we shouldn't assume that all errors are just  "loading keytar".

Perhaps we should check for the presence of the wording:
- "Failed to initialize secure credential manager"

If present in the error message, then we can call the `secureSaveError`. otherwise, just throw the error as-is

---

@Maanvi212006 
Please note that reviews might be delayed 😢 
apologies in advance 🙏

zFernand0 (commented):
We should discuss/agree during standup before the PR continues on 🙏

Maanvi212006:
> We should discuss/agree during standup before the PR continues on 🙏

Hi @zFernand0 As a regular contributor, I was wondering if I’m allowed to join the standup meetings? If so, I’d love to jump in, sync up with the team. No worries if it's internal-only, just wanted to check as I do not have idea how this works. Thanks!

zFernand0:
> > We should discuss/agree during standup before the PR continues on 🙏
> 
> Hi `@zFernand0` As a regular contributor, I was wondering if I’m allowed to join the standup meetings? If so, I’d love to jump in, sync up with the team. No worries if it's internal-only, just wanted to check as I do not have idea how this works. Thanks!

You should be able to attend any meetings posted in the Zowe Calendar:
- https://zoom-lfx.platform.linuxfoundation.org/meetings/zowe?view=week

Note: Zowe CLI standup happens on Mondays and Wednesdays 😋

packages/imperative/src/config/src/ConfigAutoStore.ts
zFernand0: Hey, 
we had a chance to discuss during standup, and we believe it makes sense to check for the above error message and call the `secureSaveError` method in case the error matches 😋 
otherwise, we should just let it throw the original error 🙏

Maanvi212006: @t1m0thyj @zFernand0 changes addressed. Thanks!

t1m0thyj:
@Maanvi212006 There is a unit test suite failing to run. Before requesting review again, please run the unit tests locally with `npm run test:unit` to verify they all pass.
```
 FAIL   Imperative  packages/imperative/src/config/__tests__/ConfigAutoStore.unit.test.ts
  ● Test suite failed to run

    packages/imperative/src/config/__tests__/ConfigAutoStore.unit.test.ts:563:21 - error TS7018: Object literal's property 'optional' implicitly has an 'any[]' type.

    563                     optional: []
                            ~~~~~~~~~~~~
```

packages/imperative/src/config/src/ConfigAutoStore.ts
```diff
@@ -200,7 +200,14 @@ export class ConfigAutoStore {
             });
         }
 
-        await config.save();
+        try {
+            await config.save();
+        } catch (err) {
+            if (err.message?.includes("Failed to initialize secure credential manager")) {
```
zFernand0, line 206: My apologies for the misunderstanding. 

The actual error strings we should be looking/matching for are: 
- `Failed to load Keytar module`
- `Keytar was not properly loaded due to an unknown cause`

Those are the errors that come from the `DefaultCredentialManager` class.

The `Failed to initialize secure credential manager` is the user-friendly version only used by the `ProfileInfo` APIs which is primarily used in Zowe Explorer.

packages/imperative/src/config/src/ConvertV1Profiles.ts
```diff
@@ -365,7 +366,14 @@ export class ConvertV1Profiles {
         newConfig.api.layers.merge(convertedConfig);
         ConvertV1Profiles.loadV1Schemas();
         ConfigSchema.updateSchema();
-        await newConfig.save();
+        try {
+            await newConfig.save();
+        } catch (err) {
+            if (err.message?.includes("Failed to initialize secure credential manager")) {
```
zFernand0, line 372: Same as previous comment, can we look for these errors instead:
- `Failed to load Keytar module`
- `Keytar was not properly loaded due to an unknown cause`

zFernand0 (changes requested):
Have a minor request. Apologies for the misunderstanding 🙏

PR #2740, opened 2026-06-11
[v3] Trust custom editors

jace-roell:
**What It Does**

 Updated `zowe zos-files (ds/uss) edit` to prompt the user to trust custom editors.

**How to Test**

Run `zowe zos-files edit uss <file> --editor <pathToEditor>`
**OR**
Run `zowe zos-files edit uss <file>` with the `editor` property set on the profile

This may also be ran against `zowe zos-file edit ds`

Expected result:

image

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- **Breaking** BugFix: Updated `zowe zos-files (ds/uss) edit` to prompt the user to trust custom editors. [#2740](https://github.com/zowe/zowe-cli/pull/2740)
+
 ## `8.32.2`
 
 - BugFix: Updated the error handling for plug-in installation to report the first failure that occurs. [#2722](https://github.com/zowe/zowe-cli/pull/2722)
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Zowe CLI package will be documented in this file.
 
 ## Recent Changes
 
-- **Breaking** BugFix: Updated `zowe zos-files (ds/uss) edit` to prompt the user to trust custom editors. [#2740](https://github.com/zowe/zowe-cli/pull/2740)
+- **Breaking** BugFix: Updated the `zowe zos-files (ds/uss) edit` command to prompt the user to trust custom editors. [#2740](https://github.com/zowe/zowe-cli/pull/2740)
 
 ## `8.32.2`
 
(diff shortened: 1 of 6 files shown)


packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- **Breaking** BugFix: Updated `zowe zos-files (ds/uss) edit` to prompt the user to trust custom editors. [#2740](https://github.com/zowe/zowe-cli/pull/2740)
```
anaxceron: We should explain what `zowe zos-files (ds/uss) edit` is -- it is a class, a command, a kitty cat, what?

anaxceron (approved):
🦁

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -131,6 +132,11 @@ export class EditUtilities {
             case Prompt.continueToUpload:
                 promptText = promptPrefix + 'Ignore conflicts and overwrite remote? y/n';
                 break;
+            case Prompt.trustEditor: {
+                const editorName = promptTexts?.length > 0 ? promptTexts[0] : "the custom editor";
+                promptText = `You are about to edit using '${editorName}'. Would you like to trust this editor? y/n`;
```
traeok, line 137: Is there (*or* should there be) a default value for this prompt (e.g. if the user just presses "Enter")?
If so, should we capitalize it in the text to denote the default action?

```suggestion
                promptText = `You are about to edit using '${editorName}'. Would you like to trust this editor? y/N`;
```

traeok (commented):
Changes make sense, testing now. Left a question regarding prompt behavior

packages/cli/src/zosfiles/edit/Edit.utils.ts
jace-roell: at the moment, the `do while` loop will just reprompt in the case you just press enter (`""`). For any of the prompts from this function, there is no default behavior, they seem to require an explicit `y` or `n`

traeok: Gotcha. This seems fairly unorthodox for a CLI, but since the pre-existing prompts do this, I'm okay with the behavior for now 😋

traeok (approved):
Arigato Jace-san

CBforZ (approved):
Someday it would be cool to persist which  editors you've already trusted, but looks good overall

gejohnston (approved):
Is this "breaking" change in response to a potential vulnerability?

If so, I have no objection. I just wanted to confirm our thinking.

PR #2741, opened 2026-06-11
Add IO helper functions and filename checks (v2)

awharn:
**What It Does**

Implements filename checks for output files from download and import commands.

**How to Test**

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (provide build number if applicable) - 2290

**Additional Comments**

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -59,7 +59,7 @@ module.exports = {
         "@typescript-eslint/no-explicit-any": "off",
         "@typescript-eslint/no-inferrable-types": "off",
         "@typescript-eslint/no-magic-numbers": ["warn", {
-            "ignore": [-2, -1, 0, 1, 2],
+            "ignore": [-1, 0, 1, 2],
             "ignoreDefaultValues": true,
             "ignoreEnums": true,
             "ignoreReadonlyClassProperties": true
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -59,7 +59,7 @@ module.exports = {
         "@typescript-eslint/no-explicit-any": "off",
         "@typescript-eslint/no-inferrable-types": "off",
         "@typescript-eslint/no-magic-numbers": ["warn", {
-            "ignore": [-1, 0, 1, 2],
+            "ignore": [-2, -1, 0, 1, 2],
             "ignoreDefaultValues": true,
             "ignoreEnums": true,
             "ignoreReadonlyClassProperties": true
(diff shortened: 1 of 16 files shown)


anaxceron (approved):

gejohnston (approved):

packages/imperative/src/io/__tests__/IO2.unit.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        it("should flag a subpath that is absolute", () => {
+            const rootDir = __dirname.split(path.sep)[0];
+            const parent = "fakedir";
+            const child = rootDir;
+            const result = IO.isSubPath(parent, child);
+            expect(result).toEqual(false);
+        });
+    });
+    describe("containsBacktrack", () => {
+        it("should return true if path contains backtracking", () => {
```
MarkAckert, line 42: how about a case of nested backtracking, like: `a/b/c/../../../../backtracked`

I believe your implementation covers this, so just making it explicit

traeok (approved):
Thanks Andrew, LGTM

MarkAckert (approved):
Thanks LGTM!

zFernand0 (approved):
LGTM! 😋 
---
This seems to align with what was discussed 🎩

packages/zosfiles/src/methods/download/Download.ts
```diff
@@ -222,6 +222,9 @@ export class Download {
             // Otherwise use default directory generated from data set name.
             const baseDir = (() => {
                 if (options.directory) {
+                    // Prevent double slashes
+                    if (options.directory.endsWith(IO.FILE_DELIM)) { return options.directory.slice(0, -1); }
+                    if (options.directory.endsWith("\\")) { return options.directory.slice(0, -2); }
```
t1m0thyj: Shouldn't we only slice off 1 character in this case too? If the first backslash is an escape character, then there is only one literal character that `"\\"` will evaluate to.

.eslintrc.js
```diff
@@ -59,7 +59,7 @@ module.exports = {
         "@typescript-eslint/no-explicit-any": "off",
         "@typescript-eslint/no-inferrable-types": "off",
         "@typescript-eslint/no-magic-numbers": ["warn", {
-            "ignore": [-1, 0, 1, 2],
+            "ignore": [-2, -1, 0, 1, 2],
```
t1m0thyj: (nit) If previous comment is valid, then we may not need `-2` in this list anymore?

t1m0thyj (commented):
Thanks for working on this Andrew - left a comment around trimming backslashes.

t1m0thyj (approved):
LGTM, thanks @awharn!

PR #2743, opened 2026-06-11
(v3): Enforce file permissions on daemon related directories and files

zFernand0:
**What It Does**

- Restricts access to daemon-related files and directories to only the current user on all platforms. 
- The following items were given owner-only permissions (`0o700`/`0o600` on POSIX, owner-only ACL on Windows) to prevent other local users from accessing them.
  - The daemon directory: `~/.zowe/daemon/`
  - The zowe-bin directory: `~/.zowe/bin`
  - The extracted native executable: `~/.zowe/bin/zowe`
  - The daemon PID file: `~/.zowe/daemon/daemon_pid.json`
  - The Unix domain socket: `~/.zowe/daemon/daemon.sock`
- The following items were re-restricted when they already exist, so that artifacts created before this fix with looser permissions are corrected on the next `zowe daemon enable` or `zowe daemon restart`.
  - The daemon directory: `~/.zowe/daemon/`
  - The zowe-bin directory: `~/.zowe/bin`
  - The native executable: `~/.zowe/bin/zowe`

**How to Test**

1. Prepare the binary:
    - `cd` to the root of the project
    - `cd zowex`
    - `cargo build -r`
    - `cd target/release`
    - `gtar -cvzf zowe-macos.tgz zowe` (`zowe-linux.tgz` or `zowe-windows.tgz`)
    - `mv zowe-macos.tgz ../../../packages/cli/prebuilds/`
2. Build the project:
    - `cd` to the root of the project
    - `npm run build` (or just `npm run watch` 😋 )
3. Enable the daemon using the built version of the CLI:
    - `cd` to the root of the project
    - `node packages/cli/lib/main.js daemon enable`
4. Run a command using the binary
    - `~/.zowe/bin/zowe --version`
5. Verify file permissions
    - `ls -al ~/.zowe/bin`
    - `ls -al ~/.zowe.daemon`

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (provide build number if applicable) `Build: #2291`

**Additional Comments**

TODO:
- [x] Look into why the socket keeps going back to `0o755` on `os.platform==darwin`
- [x] Add coverage for potential errors coming from `IO.giveAccessOnlyToOwner`
- [x] Review system tests
  - Expected APIML test files/suites to fail if not using APIML
    - `packages/cli/__tests__/config/auto-init/__system__/cli.config.auto-init.system.test.ts`
    - `packages/cli/__tests__/auth/__system__/cli.auth.login.apiml.system.test.ts`
    - `packages/core/__tests__/auth/__system__/Logout.system.test.ts`
    - `packages/core/__tests__/apiml/__system__/Services.system.test.ts`
    - `packages/core/__tests__/auth/__system__/Login.system.test.ts`
  - These also fail on the `master` branch (with and without the daemon)
    - `packages/cli/__tests__/zosmf/__system__/list/systems/zosmf.list.systems.system.test.ts`
    - `packages/cli/__tests__/zosmf/__system__/check/status/zosmf.check.status.system.test.ts`
    - `__tests__/__system__/cli/logging/LoggingCredentials.system.test.ts`
    - `packages/cli/__tests__/zosfiles/__system__/edit/uss/Edit.uss.system.test.ts`
    - `packages/cli/__tests__/zosfiles/__system__/edit/ds/Edit.ds.system.test.ts`
    - `packages/zosfiles/__tests__/__system__/methods/copy/Copy.system.test.ts`

### DIFF
--- a/packages/cli/__tests__/daemon/__unit__/enable/Enable.handler.unit.test.ts
+++ b/packages/cli/__tests__/daemon/__unit__/enable/Enable.handler.unit.test.ts
@@ -67,11 +67,22 @@ describe("Handler for daemon enable", () => {
         }
     });
 
+    let giveAccessOnlyToOwnerSpy: any;
+
     beforeEach(() => {
         // remove enableDaemon spy & mock between tests
         enableDaemonSpy?.mockRestore();
         unzipTgzSpy?.mockRestore();
         logMessage = "";
+
+        // Prevent real chmod/icacls calls against our fake (non-existent) paths.
+        giveAccessOnlyToOwnerSpy = jest.spyOn(IO, "giveAccessOnlyToOwner").mockImplementation(() => {
+            return;
+        });
+    });
+
+    afterEach(() => {
+        giveAccessOnlyToOwnerSpy?.mockRestore();
     });
 
     afterAll(() => {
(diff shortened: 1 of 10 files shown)


packages/cli/__tests__/daemon/__unit__/enable/Enable.handler.unit.test.ts
```diff
@@ -11,7 +11,7 @@
 
 jest.mock("child_process"); // using child_process from the __mocks__ directory
 
-import { ImperativeConfig, ImperativeError, IO, ProcessUtils, ISystemInfo } from "@zowe/imperative";
+import { ImperativeConfig, ImperativeError, IO, ProcessUtils, ISystemInfo, CliUtils } from "@zowe/imperative";
```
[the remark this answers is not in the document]
zFernand0: I have this resolved locally, will wait for some feedback before I push.
commit (c2ddd67) coming soon

packages/cli/src/daemon/DaemonUtil.ts
```diff
@@ -38,11 +37,20 @@ export class DaemonUtil {
         if (!IO.existsSync(daemonDir)) {
```
gejohnston: SonarCloud makes a valid point. Originally this negated 'if' statement had no 'else' statement, which was legit. Now that we are adding an 'else' statement. we could switch the 'if' and 'else' block and we would not have to negate the 'if' statement.

Someone reading the code would not have to use the double negative thinking of "if not io.existsSync is not true".

Probably a good idea to change, but not a deal-breaker.

gejohnston (dismissed):
I really like the descriptive comments. The next developer to update this code will understand why things are being done.

I also liked the removal of unnecessary boiler-plate mocks in the tests.

I made one small requested change.

packages/cli/src/daemon/DaemonUtil.ts
zFernand0: resolved in b09262470

gejohnston (approved):

traeok (approved):
LGTM, thanks @zFernand0

packages/cli/src/daemon/DaemonDecider.ts
```diff
@@ -113,7 +113,33 @@ export class DaemonDecider {
...
+
+            // On macOS and some BSDs, `chmod` has no effect on Unix domain sockets
+            // after they are created. The only way to secure the socket file is to
+            // set a restrictive umask before the `listen` (bind) call creates it.
+            let oldUmask: number | undefined;
+            if (process.platform !== "win32") {
+                oldUmask = process.umask(0o077);
+            }
+            // Slight concern that the umask will apply to other files being created
+            // by this process at the exact same milliseconds, but I this is acceptable.
```
t1m0thyj: Agree this is probably acceptable, but the concern about thread safety also makes me curious about alternative ideas 😋  I'm curious if this step is really necessary - if the parent directory of the socket has restrictive enough permissions, then it shouldn't be possible to traverse the directory and access files inside of it regardless of their permission level. To demonstrate this:
```
mkdir testDir
echo "Hello, World" > testDir/hello.txt
ls -la testDir  # See that dir is 0o755 and file is 0o644
cat testDir/hello.txt
chmod 500 deleteme
cat testDir/hello.txt  # Fails even though file is still 0o644
```

jace-roell (approved):

packages/cli/src/daemon/DaemonDecider.ts
zFernand0: resolved in 22cf691

PR #2747, opened 2026-06-15
Add IO helper functions and output validation

awharn:
**What It Does**

Adds extra IO functions and validates paths prior to downloads

**How to Test**

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (provide build number if applicable) - 2304

**Additional Comments**

### DIFF
--- a/packages/imperative/src/io/__tests__/IO.noFsMock.unit.test.ts
+++ b/packages/imperative/src/io/__tests__/IO.noFsMock.unit.test.ts
@@ -87,7 +87,7 @@ describe("IO tests", () => {
             }
         });
         skipOnPosix("should return true for posix on windows", () => {
-            expect(IO.isRootDir("/")).toEqual(false);
+            expect(IO.isRootDir("/")).toEqual(true);
         });
     });
 });
--- a/packages/imperative/src/io/__tests__/IO.noFsMock.unit.test.ts
+++ b/packages/imperative/src/io/__tests__/IO.noFsMock.unit.test.ts
@@ -13,6 +13,7 @@ import { IO } from "../../io";
 import * as path from "path";
 
 const skipOnWin = process.platform === 'win32' ? it.skip : it;
+const skipOnPosix = process.platform !== 'win32' ? it.skip : it;
 
 describe("IO tests", () => {
     describe("isSubPath", () => {
@@ -67,6 +68,26 @@ describe("IO tests", () => {
         });
         skipOnWin("should not flag an element containing a Windows path seperator", () => {
             expect(IO.fileEvaluatesToDir("some" + path.win32.sep + "path")).toEqual(false);
+        });
+    });
+    describe("isRootDir", () => {
+        skipOnWin("should return true (posix)", () => {
+            expect(IO.isRootDir("/")).toEqual(true);
+        });
+        skipOnWin("should return false for windows on posix", () => {
+            const alphabet = [...'ABCDEFGHIJKLMNOPQRSTUVWXYZ'];
+            for (const letter of alphabet) {
+                expect(IO.isRootDir(`${letter}:\\`)).toEqual(false);
+            }
+        });
+        skipOnPosix("should return true (windows)", () => {
+            const alphabet = [...'ABCDEFGHIJKLMNOPQRSTUVWXYZ'];
+            for (const letter of alphabet) {
+                expect(IO.isRootDir(`${letter}:\\`)).toEqual(true);
+            }
+        });
+        skipOnPosix("should return true for posix on windows", () => {
+            expect(IO.isRootDir("/")).toEqual(false);
         });
     });
 });
--- a/packages/zosfiles/src/methods/download/Download.ts
+++ b/packages/zosfiles/src/methods/download/Download.ts
@@ -248,7 +248,7 @@ export class Download {
             const baseDir = (() => {
                 if (options.directory) {
                     // Prevent double slashes
-                    if (options.directory.endsWith(posix.sep) || options.directory.endsWith(win32.sep)) {
+                    if ((options.directory.endsWith(posix.sep) || options.directory.endsWith(win32.sep)) && !IO.isRootDir(options.directory)) {
                         return options.directory.slice(0, -1);
                     }
                     return options.directory;
--- a/packages/zosfiles/src/methods/download/Download.ts
+++ b/packages/zosfiles/src/methods/download/Download.ts
@@ -247,6 +247,8 @@ export class Download {
             // Otherwise use default directory generated from data set name.
             const baseDir = (() => {
                 if (options.directory) {
+                    // Prevent double slashes
+                    if (options.directory.endsWith("/")) { return options.directory.slice(0, -1); }
                     return options.directory;
                 }
 
@@ -305,6 +307,9 @@ export class Download {
                 }
 
                 const memberFilePath = posix.join(baseDir, fileName + IO.normalizeExtension(extension));
+                if (!IO.isSubPath(baseDir, memberFilePath) || IO.fileEvaluatesToDir(mem.member)) {
+                    throw new ImperativeError({msg: "The generated data set file path contains illegal characters."});
+                }
                 const memberExistedBefore = IO.existsSync(memberFilePath);
 
                 // Check if file exists and should not be overwritten
@@ -444,6 +449,9 @@ export class Download {
             const mutableOptions: IDownloadOptions = { ...options, task: undefined };
 
             for (const dataSetObj of zosmfResponses) {
+                if (IO.fileEvaluatesToDir(dataSetObj.dsname)) {
+                    throw new ImperativeError({msg: "The data set name contains illegal characters."});
+                }
                 let llq = dataSetObj.dsname.substring(dataSetObj.dsname.lastIndexOf(".") + 1, dataSetObj.dsname.length);
                 if (!options.preserveOriginalLetterCase) {
                     llq = llq.toLowerCase();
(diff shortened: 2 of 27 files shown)


packages/imperative/src/io/__tests__/IO.noFsMock.unit.test.ts
zFernand0, line 1: Curious why not just add these to the `IO.unit.test.ts` file 😋

packages/zosfiles/src/methods/download/Download.ts
```diff
@@ -247,6 +247,10 @@ export class Download {
             // Otherwise use default directory generated from data set name.
             const baseDir = (() => {
                 if (options.directory) {
+                    // Prevent double slashes
+                    if (options.directory.endsWith(posix.sep) || options.directory.endsWith(win32.sep)) {
+                        return options.directory.slice(0, -1);
+                    }
```
zFernand0, line 253: Is this something we need to do a bit more broadly?

Doing a quick search for `slice(0, -` shows a few more places where we check for trailing slashes, and remove them.

> packages/zosfiles/src/methods/lise/List.ts

Image

packages/zosfiles/src/methods/download/Download.ts
```diff
@@ -846,6 +856,10 @@ export class Download {
                         // If .zosattributes says to ignore the file, skip it
                         continue;
                     }
+                    if (IO.containsBacktrack(item.name) ||
+                        !IO.isSubPath(workingDirectory, join(workingDirectory, item.name))) {
+                        throw new ImperativeError({msg: "The generated file path contains illegal backtracking."});
```
zFernand0, line 861: Do we have to tell the user what the exact cause was?

If not, can we combine this with the other (more generic) message?

> `The data set name contains illegal characters`

Maybe something like:

> "The data set or file path contains illegal characters"


---

Ideally we should add these messages to the `ZosFilesMessages`

packages/zosfiles/src/utils/ZosFilesUtils.ts
```diff
@@ -44,13 +44,20 @@ export class ZosFilesUtils {
...
+            throw new ImperativeError({msg: "The data set name contains illegal characters."});
+        }
         let localDirectory = dataSet.replace(new RegExp(`\\${this.DSN_SEP}`, "g"), path.posix.sep).toLowerCase();
         if (localDirectory.indexOf("(") >= 0 && localDirectory.indexOf(")") >= 0) {
             localDirectory = localDirectory.replace(/\(/, path.posix.sep);
             localDirectory = localDirectory.slice(0, -1);
         }
+        if (IO.containsBacktrack(localDirectory) || localDirectory.includes(path.posix.sep + path.posix.sep)) {
+            throw new ImperativeError({msg: "The generated data set file path contains illegal backtracking."});
+        }
```
zFernand0, line 60: Same as previous comments.

We should consider moving these to a constant in `ZosFilesMessages` 🙏

packages/zosjobs/src/DownloadJobs.ts
```diff
@@ -196,8 +197,13 @@ export class DownloadJobs {
         }
 
         const extension = parms.extension ?? DownloadJobs.DEFAULT_JOBS_OUTPUT_FILE_EXT;
+        const finalPath = directory + path.posix.sep + parms.jobFile.ddname + extension;
```
zFernand0, line 200: Do we need to validate the user/developer provided the extension parameter?
`.ext` (dot extension)

Otherwise, we might just be adding a string to the end of the `ddname` 😋

gejohnston (approved):

zFernand0 (commented):
Changes LGTM! 😋 
---

Left some comments/questions for consideration 🙏

packages/imperative/src/io/__tests__/IO.noFsMock.unit.test.ts
awharn: Because it does a module mock on path, and I can't get it to require the actual implementation for these functions in that file. I went so far as to try `jest.unmock` on the module mock but that wasn't working.

packages/zosfiles/src/methods/download/Download.ts
awharn: We could. Would we backport that to V2, since this was also on that PR, but wasn't brought up.

packages/zosjobs/src/DownloadJobs.ts
awharn: I will leave this up for discussion. Extension is controlled by the user, so I am inclined to leave it as is, but I can be persuaded.

packages/zosfiles/src/methods/download/Download.ts
awharn: I mean, in theory, if we know that it is a directory and we are going to explicitly be appending another slash to it later (or using join) then I think we should be removing it.

packages/zosfiles/src/utils/ZosFilesUtils.ts
awharn: Same as previous response.

anaxceron (approved):

packages/imperative/src/io/__tests__/IO.noFsMock.unit.test.ts
traeok: I saw some of the issues around the module mock and I think the separate test makes sense - but I'm wondering if we should call this new test file `IO.noFsMock.unit.test.ts`? This would easily differentiate between the two suites based on filenames alone, and also makes it easy to discern as a user looks for the appropriate test file to contribute to.

packages/imperative/src/io/__tests__/IO.noFsMock.unit.test.ts
```diff
@@ -0,0 +1,72 @@
...
+* SPDX-License-Identifier: EPL-2.0
+*
+* Copyright Contributors to the Zowe Project.
+*
+*/
+
+import { IO } from "../../io";
+import * as path from "path";
+
+const skipOnWin = process.platform === 'win32' ? it.skip : it;
```
traeok, line 15: No requests here, just wanted to comment that I'm a fan of the way the tests are being skipped here for Windows. It's easy to read and makes the test suite clear at a glance.

packages/zosjobs/src/DownloadJobs.ts
traeok: While this sounds nice to have, I would consider this request out-of-scope because this constitutes a separate change in behavior. Before this PR, we always returned `directory + path.posix.sep + parms.jobFile.ddname + extension` - so I'm inclined to leave it as-is for now to keep the scope narrowed.

packages/zosfiles/src/methods/download/Download.ts
traeok: Thinking through this a bit more, there are a few edge cases around this check which this may surface issues:

- If `options.directory` is `/`, the first part of the condition is true. Slicing off the last character returns `""`, which is then silently redirected to the current working directory: `posix.join("", fileName...)`
- If `options.directory` is `C:\` on Windows, the second part of the condition is true. Slicing off the last character returns `C:`, `baseDir` becomes `"C:"` and the final path is constructed as `C:/fileName`. This is fine until `IO.isSubPath` is called - its subtle, but the forward slash vs. the back slash causes the function to return false, thus displaying the backtrack message and failing the request.

Personally, I wouldn't recommend that a user saves directly into the root directory on POSIX, or the root of the C drive on Windows, but it is a possible use case. That said, since I didn't this in v2, I'm wondering how we should proceed here... maybe a separate issue?

zFernand0: Seems like saving at root was address in e634b2d

---

I was also wondering why we are concerned about `win32.sep` for download, and not for other places (e.g. List, Upload)
- `packages/zosfiles/src/methods/list/List.ts#fileList`
- `packages/zosfiles/src/methods/upload/Upload.ts#createDirectoriesRecursively`

packages/zosfiles/src/methods/download/Download.ts
zFernand0: No need.
We rarely change messages, and if we do, these ones should be a simple find-and-replace in multiple places 🙏

packages/zosfiles/src/utils/ZosFilesUtils.ts
zFernand0: no need

packages/zosjobs/src/DownloadJobs.ts
zFernand0: 10-22

zFernand0 (commented):
Changes LGTM! 😋 
---
Thanks for considering the comments/questions 🙏 
Apologies for causing delays on something that was already merged into V2.
I should've paid closer attention. 😓 

Feel free to move forward with the PR as-is and merge once the windows unit tests are fixed 🙏 
In theory you could just revert the previous commit and keep things moving along 🙏

packages/zosfiles/src/methods/download/Download.ts
awharn: In this case, I think it was a difference between V2 and V3. I just cherry picked my changes, but in V2, we were building out the path, whereas in V3, we use join. So in V2, it was possible to have a path contain `\/`, whereas that will never happen here. I can remove the check if we want, since join will resolve it properly.

zFernand0 (approved):

t1m0thyj (approved):
LGTM, thanks @awharn!

MarkAckert (approved):

PR #2750, opened 2026-06-17
Reduce excessive URI encoding [V2]

gejohnston:
**What It Does**

To prepare for future changes to URI encoding, this PR reduces the encoding used for URIs in REST requests by the following Zowe client SDK operations. These operations cover all known REST URIs that are currently encoded by Zowe clients.

- For datasets and uss files:
    - copy
    - create
    - delete
    - download
    - list
    - rename
    - upload
- For only datasets
    - mount
    - unmount
- For jobs
    - cancelJobs
    - deleteJobs
    - downloadJobs
    - getJobs
    - modifyJobs
    
URI encoding has been reduced to the minimum that still allows the operations to work successfully in the current z/OS environment.

URI encoding has been consolidated into one file (EncodeUri.ts) to enable quick adjustment when new URI restrictions come online.

Different levels of URI encoding is implemented for z/OSMF and API-ML, because API-ML requires more characters to be encoded.

Two characters " and \\ are valid characters in USS file names. During testing I discovered that both z/OSMF and API-ML fail with vague server errors (like HTTP 500). As part of this change, the Zowe client SDK throws an error when either character is included in a USS file name. That error message identifies the offending character which cannot be used with REST APIs in the current environment.

These changes have been successfully tested manually with Zowe CLI and Zowe Explorer. They have also been successfully regression-tested by automated pipelines in the CLI.

**How to Test**

Select a valid special character. Place that character into dataset, USS file name, or job. Perform a Zowe CLI or Zowe Explorer command on such an item. USS file names allow more special characters, so they are a good choice for testing. For example, ? is a valid character in a USS file name, but it has special meaning in a URI as the start of a query parameter.

```
> zowe-v2 files create uss-file '/your/chosen/dir/test?XXX.txt'
USS file or directory created successfully.

> zowe-v2 files list uss-files '/your/chosen/dir/test?XXX.txt'
name                          mode       size uid user
/your/chosen/dir/test?XXX.txt -rwxr-xr-x      14  YOURID
```

You can also perform any of the other operations listed above with special characters.

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (provide build number if applicable) - 2341, 2342, 2343, 2344

**Additional Comments**

The same set of changes will be applied to V3 after the testing of V2 is complete.


MarkAckert (commented):
Read through all the changes and they look good to me. I'll approve once I get some manual testing done.

If I'm reading the PR right: does this fix what might've been existing bugs around special-character support for USS-files when communicating through APIML? If so that's a nice bonus 😄

packages/imperative/src/rest/src/client/EncodeUri.ts
```diff
@@ -0,0 +1,194 @@
...
+     *
+     * @returns {boolean} - True if we must encode for APIML. False otherwise.
+     */
+    private static shouldEncodeForApiml(session: AbstractSession): boolean {
+        if (session?.isUsingApiml ) {
+            if (typeof session?.isUsingApiml === "function") {
+                return session.isUsingApiml();
+            }
+        }
+        return false;
```
traeok: This could be collapsed using optional chaining + logical AND:

```suggestion
        return typeof session?.isUsingApiml
   === "function" && session.isUsingApiml();
```

packages/imperative/src/rest/src/client/EncodeUri.ts
```diff
@@ -0,0 +1,194 @@
...
+     *
+     * @param {string} charToEncode - The character to encode.
+     * @param {string} session - The session used to connect to the server.
+     *
+     * @returns {string} - For API-ML, a URI-encoded character string.
+     *                     For z/OSMF, the unencoded character.
+     */
+    private static encCharOnlyForApiml(charToEncode: string, session: AbstractSession): string {
+        let encodedChar = charToEncode;
+        if (EncodeUri.shouldEncodeForApiml(session)) {
```
traeok: This function calls `shouldEncodeForApiml(session)` for every character of the path. Could we compute `isApiml` once and pass it into this function to avoid the per-character overhead?

traeok (commented):
Thanks for digging into this Gene. I left a suggestion regarding a helper function and found an opportunity to optimize per-character encoding logic.

anaxceron (approved):

gejohnston:
> Read through all the changes and they look good to me. I'll approve once I get some manual testing done.
> 
> If I'm reading the PR right: does this fix what might've been existing bugs around special-character support for USS-files when communicating through APIML? If so that's a nice bonus 😄

The original code used a high-level of encoding provided by the JavaScript function encodeUriComponent.
My first implementation used the JavaScript function EncodeURI() that does less encoding. During testing I found that some characters still had to be encoded after calling EncodeURI(). At that point I felt it was most coherent to explicitly encode every character that needed encoding.

The original JavaScript encodeUriComponent() function would have encoded everything needed by APIML and more. Since we are trying to not encode things unnecessarily, and z/OSMf requires fewer characters be encoded than APIML, that is why you are seeing more encoding specific to APIML.

The only existing problem that I think was corrected is that we no longer send a URI path containing \ or ". They just outright fail whether connecting to zosmf or apiml. I assume that nobody ever placed those characters into a file name, so nobody ever noticed.

packages/imperative/src/rest/src/client/EncodeUri.ts
gejohnston: From available literature (as interpreted by AI), the following expression:
```
typeof session?.isUsingApiml
```
will return "undefined" if session is null.

Further claims are that
```
undefined === "function"
```
will throw a compile-time TypeScript Error 2367: "This comparison appears to be unintentional because the types 'undefined' and '"function"' have no overlap."

Admittedly this is an edge-case, but I am leery of making this change for these reasons. I may try a local experiment to confirm these claims.

I did notice that the ? is unnecessary in the second if statement:
```
if (typeof session?.isUsingApiml === "function") {
```
At the very least, I plan to remove that ?.

awharn:
System tests passed, going to approve once comments are resolved and the branch is updated. Thanks, @gejohnston! 🚀

packages/imperative/src/rest/src/client/EncodeUri.ts
gejohnston: I thought about doing this during implementation. The trade-off was calling a function to test a property or still test a property for every character. This is where I landed, but I am open to the idea. Since you and others feel that the change would be a good idea, I will alter the behavior.

packages/imperative/src/rest/src/client/EncodeUri.ts
traeok: Good catch - I didn't consider the possibility of a TypeScript error. You might be able to avoid the TS error with `==` instead of strict equality (`===`), but this is more of a stylistic suggestion - so I'm okay with whatever approach makes the most sense 😋

gejohnston: I tried this technique locally. It compiled successfully. A unit test already existed that supplied a null value for session. That test was also successful. Thanks AI for scaring me!  I committed this change.

traeok (approved):
Thanks Gene! LGTM

packages/imperative/src/rest/src/client/EncodeUri.ts
zFernand0, line 1: For V2, could we make these new APIs protected (or private)?

zFernand0 (commented):
Changes LGTM! 😋 
---
Thanks for porting the V3 encoding utilities. I was wondering if we could make most of these private or protected since they could be considered enhancements on zowe-v2-lts 😋 
Still testing V3, but happy to approve this one as well when I'm done testing over there 🙏

packages/imperative/src/rest/src/client/EncodeUri.ts
gejohnston: The EncodeUri class is in the imperative package, but is called from the zosfiles and zosjobs packages. Zosfiles and zosjobs files get compile errors trying to import functions that are private or protected in an imperative class.

Removing EncodeUri from the rest/index.ts file and making zosfiles and zosjobs import EncodeUri directly from its .../imperative/.../EncodeUri source path gets a compile error that EncodeUri is not within the root directory of zosfiles.

Removing EncodeUri from the rest/index.ts file and making zosfiles and zosjobs import EncodeUri directly from lib path does compile in our environment, but can cause problems for extenders who try to build a their product with a zosfiles (or zosjobs) package that attempts to import from a relative path to a lib directory instead of from the @zowe/imperative package.

Adding @internal to EncodeUri will compile and run successfully for us and extenders. However, the existing public function ZosFilesUtils.sanitizeUssPathForRestCall calls the JavaScript encodeURIComponent function, which does maximum encoding on a USS URI path. This function will likely produce overly-encoded URIs which will fail if URI restrictions are put into place at a customer site. For example, forward slash will be encoded. Because sanitizeUssPathForRestCall was previously public, we cannot remove it. To allow any extender (who previously chose to call sanitizeUssPathForRestCall) to react to new URI-encoding restrictions, I deprecated sanitizeUssPathForRestCall and recommended using EncodeUri.encodeUriPathForUss as a  replacement. Unless we keep EncodeUri public, extenders will not have this course of action available to them if they need to react to URI-encoding restrictions.

Because each of the efforts above reach a dead-end, I believe that we must keep EncodeUri and its functions public.

awharn (approved):

t1m0thyj (approved):
LGTM, thanks @gejohnston! No comments besides the minor ones I left on the V3 PR 😋

zFernand0 (approved):

zFernand0:
Merging as soon as builds pass:
- `release-patch`

Issue #537, opened 2019-10-03
Uploading files to FBA dataset shifts lines one char to the right


tjohnsonBCM:
If there are any lines in the file that are 80 characters long, the last character gets truncated which results in the following error from the z/OSMF API:
```
{
    "category": 6,
    "rc": 8,
    "reason": 520,
    "message": "fwrite() error",
    "stack": "  1 3B903F38+00000096 DatasetWriter::throwRWException(int,int,const char*)\n  2 3B8F7D20+000004BA DatasetWriter::copyTextFrom(ServletInputStream&)\n  3 3B8F2510+00000A3C PutDatasetServlet::doPut()\n  4 3B90AA98+000000DC TsoServlet::service()\n  5 3B937E40+0000017E TsoServlet::run()\n  6 3B937818+00000386 main\n  7 0DA3AF08+0000127E CEEVROND\n  8 3B0F1390+000000B4 EDCZHINV\n  9 0D8D4C68+000001C6 CEEBBEXT\n",
    "details": [
        "EDC5003I Truncation of a record occurred during an I/O operation. (errno2=0xC0460022)",
        "__amrc: __last_op=151 __code=0x00000000"
    ]
}
```

leofang-zOSMF:
Could you please upload your file here so that I can download it and recreate the issue in house?

uzust01:
**Long story short:** Either:
- Use a variable block format.
- Check the length of the line before submitting it.
- Use longer LRECL.
- Avoid the control character record format (FB instead of FBA).

**Additional explanation and links**
FBA means fixed length (blocked, ANSI format). There does not seem to be any conversion of the data you are passing, but an ANSI/ISO control character is added to the first byte of the record for printing purposes (or tapes and punchcards, which are not relevant here). This generates an extra space, sometimes more than one or less, depending on the control character. It can be added by using some JES2 or JES3 defaults or set for each data set.
([MVS JCL Reference - Page 467: CONTROL parameter](https://www-01.ibm.com/servers/resourcelink/svc00100.nsf/pages/zOSV2R3SA231385/$file/ieab600_v2r3.pdf)
[IBM Doc - ISO/ANSI control characters](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.4.0/com.ibm.zos.v2r4.idad500/d5028.htm)
[...]

PR #2753, opened 2026-06-19
fix(zosfiles): strip ANSI control character on FBA dataset download

Maanvi212006:
## What It Does

Addresses the round-trip corruption issue described in #537, where downloading a member from an FBA (Fixed Block ANSI) dataset and re-uploading it without modification causes z/OSMF to add an extra ANSI control character to column 1 of every record. For datasets at max LRECL, this silently truncates the last character of the line and can cause `fwrite()` errors on upload.

This PR addresses approach **# 1** from the issue discussion:
- Before downloading, the dataset's `recfm` is fetched via `List.dataSet(..., { attributes: true })` (same pattern already used in `Copy.isPDS`).
- If `recfm` starts with `FBA`, the first character of each line is stripped from the downloaded file after the download completes.
- Non-FBA datasets are unaffected — no extra calls, no behavior change.

## What this PR does NOT do yet

- Approach **#2** (pre-upload LRECL boundary check) is not included here and will be a separate PR, since it requires an opt-out flag at the SDK level for consumers like Zowe Explorer that already do this check.

## Note for reviewers
If the metadata lookup fails, it fail safe — skip the fix rather than break the download — which means the original bug can silently persist in that edge case

## Related issue

Closes (partially) #537

**How to Test**
NA

**Review Checklist**
I certify that I have:
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/zosfiles/__tests__/__unit__/methods/download/Download.unit.test.ts
+++ b/packages/zosfiles/__tests__/__unit__/methods/download/Download.unit.test.ts
@@ -53,7 +53,7 @@ describe("z/OS Files - Download", () => {
         const zosmfGetFullSpy = jest.spyOn(ZosmfRestClient, "getExpectFullResponse");
         const fakeResponseWithEtag = {
             data: Buffer.from(ussFileContent),
-            response: {headers: { etag: etagValue } }
+            response: { headers: { etag: etagValue } }
         };
 
         beforeEach(() => {
@@ -124,7 +124,7 @@ describe("z/OS Files - Download", () => {
             const destination = dsFolder + extension;
 
             try {
-                response = await Download.dataSet(dummySession, dsname, {volume, extension});
+                response = await Download.dataSet(dummySession, dsname, { volume, extension });
             } catch (e) {
                 caughtError = e;
             }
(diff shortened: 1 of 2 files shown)


zFernand0 (changes requested):
Hey @Maanvi212006,

While this approach might work for most cases, it's always scary to arbitrarily manipulate raw bytes in data. 

For that reason, I'm afraid I don't see this implementation moving forward, unless IBM recommends client-side applications (and users of the z/OSMF REST API) to manipulate such (FBA) data sets.

For now, I believe we should close this PR, and wait for IBM to address the issue on their end (as stated on the original issue - IBM IDEA) 
https://ibm-z-hardware-and-operating-systems.ideas.ibm.com/ideas/ZOS-I-2696

zFernand0:
Closing in favor of waiting on IBM to address the following ticket:
- https://ibm-z-hardware-and-operating-systems.ideas.ibm.com/ideas/ZOS-I-2696

Maanvi212006:
Thanks for the clear explanation @zFernand0 — that makes complete sense. Fixing this client-side without IBM's guidance would be making assumptions about bytes that legitimately mean different things to different users, and that's not the right place to own it.
Also, I've upvoted the IBM IDEA ticket!

PR #2756, opened 2026-06-23
Add alias resolution

### LINKED ISSUE #2728: Add functionality to locate the data set that an alias points to

This would be most useful from the standpoint of a Zowe Explorer user navigating through data sets. Currently, we show aliases in the Zowe Explorer data sets tree, but users cannot open them or interact with them.

Zowe CLI and Zowe Explorer should be able to find the target of an alias, and transparently to the user, be able to show the contents of the data set or list members of the pds that an alias points to.

Add an SDK method to locate the data set associated with an alias, with the goal being that Zowe Explorer can make use of this method to allow users to interact with aliased data sets as they would in ISPF.

I am not sure if there would be demand in changing the behavior of the Zowe CLI commands that get contents of a data set, list members, etc. to automatically go to the target of the alias when the provided dsn is an alias.

Currently, users have to do their own research to locate the target of an alias.

See https://github.com/zowe/zowe-explorer-vscode/issues/4213 for more info.

31puneet:
**What It Does**

Adds a new `zowe zos-files list alias` CLI command and `List.resolveAlias()` SDK method that resolves a data set alias to its target data set name using IDCAMS LISTCAT. 
Closes #2728.

**How to Test**

Build SDK: `cd packages/zosfiles && npm run build`
Run tests: `npx jest packages/zosfiles/__tests__/__unit__/methods/list/List.resolveAlias.unit.test.ts`
Run CLI tests: `npx jest packages/cli/__tests__/zosfiles/__unit__/list/alias/`

**Review Checklist**
I certify that I have:
- [x] updated the changelog
  - `2453`

**Additional Comments**


packages/zosfiles/src/methods/list/List.ts
```diff
@@ -525,6 +527,66 @@ export class List {
...
+    public static async resolveAlias(
+        session: AbstractSession,
+        aliasName: string,
+        options?: IZosFilesOptions
+    ): Promise<IZosFilesResponse> {
+        ImperativeExpect.toNotBeNullOrUndefined(aliasName, ZosFilesMessages.missingDatasetName.message);
+        ImperativeExpect.toNotBeEqual(aliasName, "", ZosFilesMessages.missingDatasetName.message);
+
+        try {
+            const listcatStmt = `LISTCAT ENTRIES('${aliasName.toUpperCase()}') ALIAS`;
```
adam-wolfe: The command `LISTCAT ENTRIES([alias name]) ALIAS` does not produce output that lists the associated data set.

I think the command should be `LISTCAT ENTRIES([alias name]) ALL` otherwise the zowe files list alias command will always return `Target: 0`

31puneet: Thanks, I've updated the parameter to ALL.

adam-wolfe:
I'm still getting `Target: 0`.

The regex seems to be catching another line from the output:
```
outputLines = 
[
  "LISTCAT ENTRIES('ADAM1234.MYALIAS') ALL",
  "ALIAS --------- ADAM1234.MYALIAS",
  "     IN-CAT --- ICF.USERID.USERCAT",
  "     HISTORY",
  "       RELEASE----------------2     CREATION--------2026.180",
  "     ENCRYPTIONDATA",
  "       DATA SET ENCRYPTION-----(NO)",
  "     ASSOCIATIONS",
  "       NONVSAM--ADAM1234.PUBLIC.JCL",         <-- The target. Missed by regex due to only 2 dashes
  "         THE NUMBER OF ENTRIES PROCESSED WAS:",
  "                   AIX -------------------0",
  "                   ALIAS -----------------1",
  "                   CLUSTER ---------------0",
  "                   DATA ------------------0",
  "                   GDG -------------------0",
  "                   INDEX -----------------0",
  "                   NONVSAM ---------------0", <-- Getting caught by regex, resulting in Target: 0
  "                   PAGESPACE -------------0",
  "                   PATH ------------------0",
  "                   SPACE -----------------0",
  "                   USERCATALOG -----------0",
  "                   TAPELIBRARY -----------0",
  "                   TAPEVOLUME ------------0",
  "                   TOTAL -----------------1",
  "         THE NUMBER OF PROTECTED ENTRIES SUPPRESSED WAS 0",
  "IDC0001I FUNCTION COMPLETED, HIGHEST CONDITION CODE WAS 0",
  "IDC0002I IDCAMS PROCESSING COMPLETE. MAXIMUM CONDITION CODE WAS 0",
]
```
I think the regex could be modified to something like `const match = line.match(/(?:NONVSAM|VSAM)-{2,}\s*(\S+)/i);` to remove the optional space after NONVSAM|VSAM and to allow for 2 or more dashes after NONVSAM|VSAM.

31puneet:
Updated the regex to `(NONVSAM|VSAM)-{2,}\s*(\S+)` as suggested. Thanks for catching that!
Also, is there a test system I could get access to for verifying changes like this against real LISTCAT output?

zFernand0 (changes requested):
Please add a changelog to the corresponding packages 🙏

31puneet:
> Please add a changelog to the corresponding packages 🙏

Done!

31puneet:
PTAL @zFernand0, Thanks!

packages/zosfiles/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to the Zowe z/OS files SDK package will be documented in thi
 
 ## `8.32.3`
 
+- Enhancement: Added the ability to resolve data set aliases to their true target names using the `zowe zos-files list alias` command or the `List.resolveAlias` SDK method. [#2728](https://github.com/zowe/zowe-cli/issues/2728)
```
zFernand0: This should go under a new `## Recent Changes` section above the ```## `8.33.1` ```

zFernand0 (changes requested):

packages/zosfiles/CHANGELOG.md
31puneet: Done!

zFernand0 (changes requested):
Changes LGTM! 😋 
---
I plan on doing some manual testing throughout the week.
Running system tests (2400), although I don't expect anything to break.

---

One final request:
- Could you add a system test?

That's the last check in the PR 🙏

31puneet:
>  Could you add a system test?

Done! I have added the system test

31puneet:
PTAL @zFernand0 Thanks.

packages/cli/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Zowe CLI package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added the ability to resolve data set aliases to their true target names using the `zowe zos-files list alias` command or the `List.resolveAlias` SDK method. [#2728](https://github.com/zowe/zowe-cli/issues/2728)
 ## `8.33.3`
```
zFernand0, line 10:
```suggestion

## `8.33.3`
```

packages/cli/src/zosfiles/list/alias/Alias.definition.ts
```diff
@@ -0,0 +1,47 @@
...
+// Does not use the import in anticipation of some internationalization work to be done later.
+const strings = (require("../../-strings-/en").default as typeof i18nTypings).LIST.ACTIONS.ALIAS;
+
+/**
+ * List alias command definition containing its description, examples and/or options
+ * @type {ICommandDefinition}
+ */
+export const AliasDefinition: ICommandDefinition = {
+    name: "alias",
+    aliases: ["al"],
```
zFernand0: Aliases are often the first letter of each word. 🙏 

```suggestion
    aliases: ["a"],
```

packages/cli/src/zosfiles/list/alias/Alias.handler.ts
```diff
@@ -0,0 +1,33 @@
...
+export default class AliasHandler extends ZosFilesBaseHandler {
+    public async processWithSession(commandParameters: IHandlerParameters, session: AbstractSession): Promise<IZosFilesResponse> {
+        const response = await List.resolveAlias(session, commandParameters.arguments.aliasName, {
+            responseTimeout: commandParameters.arguments.responseTimeout
+        });
+
+        // Display the resolved target data set name
+        commandParameters.response.console.log(
+            `Alias: ${response.apiResponse.alias}\nTarget: ${response.apiResponse.targetDsn}`
+        );
```
zFernand0: Please remove this redundant log.

The API `resolveAliases` returns a nice message in the `commandResponse` which gets printed out by the `ZosFileBaseHandler`.

```suggestion
```

packages/zosfiles/src/methods/list/List.ts
```diff
@@ -524,6 +526,66 @@ export class List {
...
+            const outputLines: string[] = amsResponse.apiResponse?.output ?? [];
+            let targetDsn: string | undefined;
+
+            for (const line of outputLines) {
+                const match = line.match(/(NONVSAM|VSAM)-{2,}\s*(\S+)/i);
+                if (match && match[2]) {
+                    targetDsn = match[2].trim();
+                    break;
+                }
+            }
```
zFernand0, line 592: Should we also look at the Symbolic and Resolved values?

For instance I have a data set (`HLQ.ALIAS.PDS`), and it just expands to `HLQ.MY.REAL.PDS` with a symbolic value of `HLD.&MYVAR..PDS`. 
Perhaps as a fallback, we could potentially check to see if the values are different, meaning the LISTCAT command was able to resolve the dataset name 😋 

```typescript
[
  "LISTCAT ENTRIES('HLQ.ALIAS.PDS') ALL",
  'ALIAS --------- HLQ.ALIAS.PDS',
  '     IN-CAT --- MY.COOL.CAT',
  '     HISTORY',
  '       RELEASE----------------2     CREATION--------2026.001',
  '     ENCRYPTIONDATA',
  '       DATA SET ENCRYPTION-----(NO)',
  '     ASSOCIATIONS',
  '       SYMBOLIC-HLQ.&MYVAR..PDS,
  '       RESOLVED-HLQ.MY.REAL.PDS,
  '         THE NUMBER OF ENTRIES PROCESSED WAS:',
  '                   AIX -------------------0',
  '                   ALIAS -----------------1',
  '                   CLUSTER ---------------0',
  '                   DATA ------------------0',
  '                   GDG -------------------0',
  '                   INDEX -----------------0',
  '                   NONVSAM ---------------0',
  '                   PAGESPACE -------------0',
  '                   PATH ------------------0',
  '                   SPACE -----------------0',
  '                   USERCATALOG -----------0',
  '                   TAPELIBRARY -----------0',
  '                   TAPEVOLUME ------------0',
  '                   TOTAL -----------------1',
  '         THE NUMBER OF PROTECTED ENTRIES SUPPRESSED WAS 0',
  'IDC0001I FUNCTION COMPLETED, HIGHEST CONDITION CODE WAS 0',
  'IDC0002I IDCAMS PROCESSING COMPLETE. MAXIMUM CONDITION CODE WAS 0'
]
```

zFernand0 (changes requested):
Changes LGTM! 😋 
---
left a few comments and I found data sets that we should be able to resolve and the command just returns an error 😅

packages/cli/CHANGELOG.md
31puneet: Done

packages/cli/src/zosfiles/list/alias/Alias.definition.ts
31puneet: Done

packages/cli/src/zosfiles/list/alias/Alias.handler.ts
31puneet: Done

packages/zosfiles/src/methods/list/List.ts
31puneet: Done, I've updated the regex to check for the RESOLVED- pattern as well, so symbolic aliases should now be fully supported.

zFernand0: I don't think we can blindly accept the value in the `resolved` "field".

I think we should compare against the `symbolic-` value and only return that as matched target if the values are not the same.

The main reason for this is that the `HLQ.&MYVAR..PDS` won't be an allocated data set.

zFernand0 (changes requested):
Thanks for addressing the other comments.

I still think there is room for improvement on the fallback approach. 🙏

packages/zosfiles/src/methods/list/List.ts
31puneet: Done, I've updated the logic to compare both values.  I also added some unit tests for this scenario.

zFernand0 (approved):
LGTM! 😋 
---
Thanks for addressing the feedback 🙏

31puneet:
@traeok Please approve workflows 🙏

packages/zosfiles/src/methods/list/List.ts
```diff
@@ -524,6 +526,88 @@ export class List {
...
+
+            if (!targetDsn) {
+                throw new ImperativeError({
+                    msg: ZosFilesMessages.aliasTargetNotFound.message
+                });
+            }
+
+            return {
+                success: true,
+                commandResponse: util.format(ZosFilesMessages.aliasResolvedSuccessfully.message, aliasName, targetDsn),
```
CBforZ, line 602: I had a successful manual test of this command. However looking at the output,  I wonder if we should just output the resolved data set name rather than a message with the quoted DSN? Or at least provide an option to this command to do that?  I'm thinking of the case where you may want to assign the output of this command to a shell variable for interoperability with other commands. cc @zFernand0 @t1m0thyj 

```
chris cli $ zowe files list alias chris.test.jcl.alias   
Alias 'chris.test.jcl.alias' resolved to data set 'CHRIS.PUBLIC.CNTL'.
```

CBforZ: Maybe this is not something we tend to worry about as much anymore, since you can always parse JSON when using the --rfj option as  @jace-roell pointed out

I was thinking of the case of writing a script like this: 
```sh
 resolvedAlias="$(zowe files list alias chris.test.alias --print-resolution-only)"
```

traeok: I know we have the `--response-format-json` option (`--rfj` alias) on commands like `zowe zosmf check status` - I haven't verified the output in this PR yet, but would that option suffice for automation?

_Edit:_ We posted the same suggestion at the same time 😅  That said we might want to set the expected object from within the command handler so the output looks good in JSON, rather than it just capturing `stdout` as a key

CBforZ: Here's what it looks like with --rfj . So you could get it out of data.apiResponse.targetDsn 

```json
chris@machine cli $ zowe files list alias chris.test.jcl.alias --rfj
{
  "success": true,
  "exitCode": 0,
  "message": "",
  "stdout": "Alias 'chris.test.jcl.alias' resolved to data set 'CHRIS.PUBLIC.CNTL'.\n",
  "stderr": "",
  "data": {
    "success": true,
    "commandResponse": "Alias 'chris.test.jcl.alias' resolved to data set 'CHRIS.PUBLIC.CNTL'.",
    "apiResponse": {
      "alias": "chris.test.jcl.alias",
      "targetDsn": "CHRIS.PUBLIC.CNTL"
    }
  }
}
```

31puneet: I have updated the handler to use `setObj()`. The JSON output should now be flattened.

CBforZ: I think  your latest push doesn't change any behavior, but given the ability for folks to parse out --rfj, I think you can just revert your last commit and I can resolve this convo. Thanks

traeok: Can you please revert this call to `setObj`? It has no effect, the object is identical before/after this code change.

31puneet: done

jace-roell (commented):
(Deleted)

jace-roell (approved):

traeok (approved):

Issue #2684, opened 2026-02-24
feat(config): Add `preferredLoginMethod` to Team Config


zFernand0:
**Is your feature or enhancement request related to a problem or limitation? Please describe**

User may want to specify a default login method when accessing services via the API Mediation Layer

**Describe your enhancement idea**

Add a new property to Team Config as a sibling of `authOrder` that will allow users to set a preferred login method (`preferredLoginMethod`)

```typescript
const enum PreferredLoginMethod {
  BASIC = 'basic',
  API_ML_TOKEN = `apiMlToken',
  API_ML_CERT = `apiMlCertificate`
}
```

**Describe alternatives you've considered**

N/A

**Provide any additional context**

This enhancement request came up when discussing the issue below:
- https://github.com/zowe/zowe-explorer-vscode/issues/4078

---

Added label `priority-low` tentatively until we have more info.

zFernand0:
Originally:
- Added label `needs-more-info` to indicate that we are waiting for after-SHARE discussions to take place 🙏 

---

UPDATE: 
When this is addressed, we are thinking that this property should be a per-profile property (similar to `authOrder`).

```typescript
const enum PreferredLoginMethod {
  BASIC = 'basic',
  API_ML_CERT = `apiMlCertificate`
}
```

We should not allow for `apimlToken` to be in this enum. One use case for `apimlToken` could be to refresh a token before it expires.

zFernand0:
This is almost community upvoted 😋 
so perhaps we should consider bumping it up 🙏

PR #2758, opened 2026-06-23
Reduce excessive URI encoding [V3]

gejohnston:
**What It Does**

To prepare for future changes to URI encoding, this PR reduces the encoding used for URIs in REST requests by the following Zowe client SDK operations. These operations cover all known REST URIs that are currently encoded by Zowe clients.

- For datasets and uss files:
    - copy
    - create
    - delete
    - download
    - list
    - rename
    - upload
- For only datasets
    - mount
    - unmount
- For jobs
    - cancelJobs
    - deleteJobs
    - downloadJobs
    - getJobs
    - modifyJobs
    
URI encoding has been reduced to the minimum that still allows the operations to work successfully in the current z/OS environment.

URI encoding has been consolidated into one file (EncodeUri.ts) to enable quick adjustment when new URI restrictions come online.

Different levels of URI encoding is implemented for z/OSMF and API-ML, because API-ML requires more characters to be encoded.

Two characters " and \\ are valid characters in USS file names. During testing I discovered that both z/OSMF and API-ML fail with vague server errors (like HTTP 500). As part of this change, the Zowe client SDK throws an error when either character is included in a USS file name. That error message identifies the offending character which cannot be used with REST APIs in the current environment.

**How to Test**

Select a valid special character. Place that character into a data set name, USS file name, or job name. Perform a Zowe CLI or Zowe Explorer command on such an item. USS file names allow more special characters, so they are a good choice for testing. For example, ? is a valid character in a USS file name, but it has special meaning in a URI as the start of a query parameter.

```
> zowe-v2 files create uss-file '/your/chosen/dir/test?XXX.txt'
USS file or directory created successfully.

> zowe-v2 files list uss-files '/your/chosen/dir/test?XXX.txt'
name                          mode       size uid user
/your/chosen/dir/test?XXX.txt -rwxr-xr-x      14  YOURID
```

You can also perform any of the other operations listed above with special characters.

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (provide build number if applicable) - 2345, 2346, 2347, 2348

**Additional Comments**

Because of the many test variations for encoding special characters in a URI, I placed sample manual tests, and a semi-automated PowerShell script into the following directory:  `prototypes/uri_encoding_tests`


anaxceron (approved):

CBforZ:
Didn't get a chance to try this yet, but it would be good to test this version with the VS Code extension

gejohnston:
> Didn't get a chance to try this yet, but it would be good to test this version with the VS Code extension

I manually tested ZE operations successfully using special characters in data set names, USS file names, and job names.

To build ZE with a local encode-uri branch, you will want to do that in V3. A V2 build with a local SDK branch is impractical due to older V2 tech and firewall restrictions on repositories. In a V3 ZE package.json you must override the following:
```
"overrides": {
  "@zowe/imperative": "file:../zowe-cli/packages/imperative",
  "@zowe/zos-files-for-zowe-sdk": "file:../zowe-cli/packages/zosfiles",
  "@zowe/zos-jobs-for-zowe-sdk": "file:../zowe-cli/packages/zosjobs"
},
```

If you build ZE with a local copy of the encode-URI SDK branch it is difficult to tell if you are using the new SDK, since things are supposed to work as before. Try to create a USS file with one of the special characters " or \ in the name of the file. You will get a new popup message saying that the character will never work and that the SDK did not sent the request. Now you know that you are using the right code, and you can try other operations with other special characters (which should work as before).

packages/imperative/src/rest/src/session/AbstractSession.ts
```diff
@@ -174,6 +174,33 @@ export abstract class AbstractSession {
...
+     */
+    public isUsingApiml(): boolean {
+        if (this.ISession.tokenType && this.ISession.tokenType === SessConstants.TOKEN_TYPE_APIML) {
+            return true;
+        }
+        if (this.ISession.basePath) {
+            return true;
+        }
+        return false;
+    }
```
zFernand0, line 202: The only two cases I can think of are:
- Someone not using APIML, but going through a proxy and they have `basePath` setup.
- Someone using APIML, not having tokenType (because of `logout`), not having `basePath` specified for either of the following reasons:
  - `1.` Because developers hardcode the `basePath` in their resource URLs - bad practice
  - `2.` Because system admins have configured a service as the default endpoint (`/`) - uncommon

---

We should revisit this once we add the `allowedLoginMethod` functionality:
- https://github.com/zowe/zowe-cli/issues/2684

packages/zosfiles/__tests__/__system__/methods/create/Create.system.test.ts
```diff
@@ -758,7 +758,7 @@ describe("Create uss file - encoded", () => {
 
         REAL_SESSION = TestEnvironment.createZosmfSession(testEnvironment);
         basePath = defaultSystem.unix.testdir; // `${defaultSystem.zosmf.basePath.trim()}`;
-        filename = `${basePath}/enco#edtest.txt`.replace(/\./g, "");
+        filename = `${basePath}/encodedFileNm-=~,.!:'&$@*_% +?#;<>[]^{|}.txt`.replace(/\./g, "");
```
zFernand0, line 761: I haven't tested this PR thoroughly, but the way these new utilities are implemented inspire a lot of confidence due to the fact that: 
- We **are** able to update these with a more complex set of characters
- And we do **not** need to update expected results of our system tests
- And system tests continue to work  🥳

zFernand0 (commented):
Changes LGTM! 😋 
---
I'll approve as soon as I'm done testing 🙏

packages/imperative/src/rest/src/session/AbstractSession.ts
gejohnston: If the new logic decides that some other service is actually APIML, I believe that this will be the ramifications:

1. We will encode fewer characters than was done previously by the SDK. That might adversely affect the service. This should still better position the encoding for that service for future security restrictions which may reduce allowed encoding. The service might be well advised to make its own adjustments.

2. It is possible that our testing uncovered an unencoded special character that would have previously failed with APIML. We might now be encoding some character that was not encoded before. That might adversely affect the service. I think we will just have to react if and when such a situation occurs.

I agree that we should revisit the isUsingApiml function for better detection of APIML when new capabilities come online.

traeok (approved):
Thanks Gene for working on this, I took some more time reviewing this one given the context I already had from reviewing v2. No real concerns from testing so far and based on the refactored unit tests & the system tests passing, I think this is good to approve.

packages/imperative/src/rest/src/client/EncodeUri.ts
```diff
@@ -0,0 +1,192 @@
...
+                break;
+            case "{":
+                encodedChar = "%7B";
+                break;
+            case "|":
+                encodedChar = "%7C";
+                break;
+            case "}":
+                encodedChar = "%7D";
+                break;
```
t1m0thyj, line 174: (nit) This could be condensed into a single block that runs for all the `case` statements. Not requesting changes as the current approach may be more readable and performant 😋 

```suggestion
            case "#":
            case ";":
            case "<":
            case ">":
            case "[":
            case "]":
            case "^":
            case "{":
            case "|":
            case "}":
                encodedChar = "%" + Buffer.from(charToEncode).toString("hex");
                break;
```

packages/imperative/src/rest/src/session/AbstractSession.ts
```diff
@@ -174,6 +174,33 @@ export abstract class AbstractSession {
...
+     * as before these changes, so we do not expect any breaking changes as a result.
+     *
+     * In the future we may want to replace this implementation with a more certain
+     * confirmation that we are connecting to APIML.
+     *
+     * @returns {boolean} - True if connecting with APIML. False otherwise.
+     * @memberof AbstractSession
+     */
+    public isUsingApiml(): boolean {
+        if (this.ISession.tokenType && this.ISession.tokenType === SessConstants.TOKEN_TYPE_APIML) {
```
t1m0thyj: (nit) The first check is unnecessary since the equality check will fail if `tokenType` is null or undefined.

```suggestion
        if (this.ISession.tokenType === SessConstants.TOKEN_TYPE_APIML) {
```

awharn (approved):

t1m0thyj (approved):

packages/imperative/src/rest/src/session/AbstractSession.ts
gejohnston: Makes sense to remove an unnecessary condition.

packages/imperative/src/rest/src/client/EncodeUri.ts
gejohnston: I also found it much easier to debug resulting encoded URIs when I could look-up the translation directly in the code. For this and for your reasons, I will leave this as it is.

zFernand0 (approved):

PR #2739, opened 2026-06-10
fix: address security concern

zFernand0:
**What It Does**

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] updated the changelog
- [ ] manually tested my changes
- [ ] added/updated automated unit/integration tests
- [ ] created/ran system tests (provide build number if applicable)
- [ ] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

**Additional Comments**

PR #2762, opened 2026-06-25
fix(daemon): Add generated token that only the user can read

zFernand0:
**What It Does**

The daemon generates a fresh 256-bit random token at startup and stores it in the owner-only `daemon_pid.json` file. Every client request must echo this token back. Because only the owner can read the pid file, being able to provide the token proves that you own the process.

**How to Test**

- Restart the daemon
- Check the `daemon_pid.json` it should contain a random token
- Run a command with the daemon binary

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (provide build number if applicable) (2386)

**Additional Comments**

This was the original implementation back in this PR, however we decided to be less strict at the time 🤔 
- https://github.com/zowe/zowe-cli/pull/2739


packages/cli/src/daemon/DaemonClient.ts
```diff
@@ -37,9 +38,12 @@ export class DaemonClient {
      * @param {net.Socket} mClient
      * @param {net.Server} mServer
      * @param {string} mOwner
+     * @param {string} mDaemonToken Secret token that the client must echo back to
+     *      prove it could read the owner-only PID file. When undefined, token
+     *      validation is skipped (used only in tests).
      * @memberof DaemonClient
      */
-    constructor(private mClient: net.Socket, private mServer: net.Server, private mOwner: string) {
+    constructor(private mClient: net.Socket, private mServer: net.Server, private mOwner: string, private mDaemonToken?: string) {
```
MarkAckert: nit: Should we have `mDaemonToken` as a required parameter and update tests to they intentionally pass a `null` or `undefined` value in?

packages/cli/src/daemon/DaemonDecider.ts
```diff
@@ -144,7 +160,8 @@ export class DaemonDecider {
     private recordDaemonPid() {
         const pidForUser: IDaemonPidForUser = {
             user: this.mUser,
-            pid: process.pid
+            pid: process.pid,
+            token: this.mDaemonToken
```
MarkAckert, line 165: This structure causes the token to be logged in cleartext in the trace @ 181. We should obfuscate or redact the token in the log.

packages/cli/src/daemon/DaemonClient.ts
zFernand0: resolved in ad0b0edfa

packages/cli/src/daemon/DaemonDecider.ts
zFernand0: resolved in 55fcba447

traeok (commented):
Thanks for your work on this Fernando. I've only done a bit of manual testing and nothing in-depth yet, but a couple of things jumped out on a first read that I'd like to talk through next week before approving:

1. If the server token is ever null, the check reads as skip-and-allow rather than deny. It can't happen today because of the constructor throw, but the token check is the only thing actually verifying the user, so I'd rather it deny by default.
2. On the Rust side, reading the token re-derives the daemon dir inside the retry loop / per prompt reply, and `util_get_daemon_dir()` re-runs the `icacls`/`chmod` ACL operation each time. It's probably worth reading it once.

Nothing blocking, just want to discuss. Will follow up 😋 
Also, it looks like Node 24 tests are failing for macOS, but it may be a transient failure. I've retried the check to verify.

zFernand0:
> If the server token is ever null, the check reads as skip-and-allow rather than deny. It can't happen today because of the constructor throw, but the token check is the only thing actually verifying the user, so I'd rather it deny by default.

it's unlikely, but I don't see a problem with duplicating the same check from he constructor down in the private `data`  function 🙏 

See 0c58785df

traeok:
> it's unlikely, but I don't see a problem with duplicating the same check from he constructor down in the private `data` function 🙏
> 
> See [0c58785](https://github.com/zowe/zowe-cli/commit/0c58785df6ac3f2a991d7ba11ec485e501f177c6)

Sounds good to me - thanks @zFernand0 for addressing feedback, will revisit 😋

zFernand0:
> On the Rust side, reading the token re-derives the daemon dir inside the retry loop / per prompt reply, and util_get_daemon_dir() re-runs the icacls/chmod ACL operation each time. It's probably worth reading it once.

Took me a sec to figure out where I wanted to do the optimization.
Now no more unnecessary ACL operations are happening over and over inside loops 🙏 
See 5a5b405 and e4d6788

packages/cli/__tests__/daemon/__unit__/DaemonClient.unit.test.ts
```diff
@@ -495,4 +503,205 @@ describe("DaemonClient tests", () => {
         expect(parse).not.toHaveBeenCalled();
         expect(client.end).toHaveBeenCalled();
     });
+
+    it("should failt to initialize the DaemonClient if no token is provided", () => {
```
t1m0thyj:
```suggestion
    it("should fail to initialize the DaemonClient if no token is provided", () => {
```

packages/cli/src/daemon/DaemonClient.ts
```diff
@@ -37,9 +38,15 @@ export class DaemonClient {
      * @param {net.Socket} mClient
      * @param {net.Server} mServer
      * @param {string} mOwner
+     * @param {string} mDaemonToken Secret token that the client must echo back to
+     *      prove it could read the owner-only PID file. When undefined, token
+     *      validation is skipped (used only in tests).
```
t1m0thyj: > When undefined, token validation is skipped (used only in tests).

Should we remove this statement? I believe it's no longer correct, given there is a check below which fails when `mDaemonToken` is undefined.

packages/cli/src/daemon/DaemonClient.ts
```diff
@@ -37,9 +38,15 @@ export class DaemonClient {
...
      * @memberof DaemonClient
      */
-    constructor(private mClient: net.Socket, private mServer: net.Server, private mOwner: string) {
+    constructor(private mClient: net.Socket, private mServer: net.Server, private mOwner: string, private mDaemonToken: string) {
```
t1m0thyj: (nit) Agree with Sonar's suggestion to mark `mDaemonToken` as readonly. Althought this won't prevent JS code from mutating the value, it would indicate that the token value isn't supposed to change for the lifetime of the daemon server.

zowex/src/defs.rs
```diff
@@ -64,12 +64,19 @@ pub struct DaemonResponse {
...
 }
 
 #[derive(Deserialize)]
 pub struct DaemonPidForUser {
     pub user: String,
     pub pid: i32,
+    // Secret token written by the daemon into the owner-only pid file. Older
+    // daemons do not write this field, so it defaults to None when absent.
+    #[serde(default)]
+    pub token: Option<String>,
```
t1m0thyj, line 79: Should we support these older daemon servers that don't write the `token` field? I'm not sure of a case where a new daemon client would need to be compatible with an old daemon server. Chances are the server would be newer than the client, if someone has updated their CLI package but not re-run the `zowe daemon enable` command. In the rare case where there is an incompatibility, we may want to fail to prevent insecure communication.

packages/cli/__tests__/daemon/__unit__/DaemonClient.unit.test.ts
zFernand0: I should just not write test myself and let AI do it 😋  
jk, thanks for catching this 🙏

packages/cli/src/daemon/DaemonClient.ts
zFernand0: Sounds good!
See a18ab9aa060244dd664e0dc7a5714e5a0991436a

packages/cli/src/daemon/DaemonClient.ts
zFernand0: Whoops! 😅 
Resolved in a18ab9aa060244dd664e0dc7a5714e5a0991436a

zowex/src/defs.rs
zFernand0: This is something I struggled with, and went back and forth on. 😓 

I think we should deny those old servers that don't write the token.
The deciding factor for me was the Zowe Docs:
https://docs.zowe.org/stable/user-guide/cli-using-using-daemon-mode/#restarting-daemon-mode

We mention that people should restart the daemon when installing a new version of the CLI 🙏

packages/cli/src/daemon/DaemonClient.ts
```diff
@@ -37,9 +38,14 @@ export class DaemonClient {
...
      * @param {net.Server} mServer
      * @param {string} mOwner
+     * @param {string} mDaemonToken Secret token that the client must echo back to
+     *      prove it could read the owner-only PID file.
      * @memberof DaemonClient
      */
-    constructor(private mClient: net.Socket, private mServer: net.Server, private mOwner: string) {
+    constructor(private mClient: net.Socket, private mServer: net.Server, private mOwner: string, private readonly mDaemonToken: string) {
+        if (!this.mDaemonToken) {
+            throw new ImperativeError({msg: "Unable to initialize the Daemon Client wihtout a proper token"});
```
traeok:
```suggestion
            throw new ImperativeError({msg: "Unable to initialize the Daemon Client without a proper token"});
```

packages/cli/src/daemon/DaemonClient.ts
```diff
@@ -191,6 +197,27 @@ export class DaemonClient {
...
         }
 
+        // The user comparison above is advisory only: the client asserts its own
+        // user name, which any local process can forge. The real authentication is
+        // the token below. Because the token is stored in the owner-only PID file,
+        // a client that echoes it back has proven it could read that file, and is
+        // therefore the owner. This is what protects the daemon on Windows, where
+        // the named pipe can be opened by local users other than the owner.
+        if (!this.mDaemonToken) {
+            throw new ImperativeError({msg: "Unable to initialize the Daemon Client wihtout a proper token"});
```
traeok:
```suggestion
            throw new ImperativeError({msg: "Unable to initialize the Daemon Client without a proper token"});
```

traeok (dismissed):
Small request for changes (just spelling in an error msg) - everything else LGTM! 😋

packages/cli/__tests__/daemon/__unit__/DaemonClient.unit.test.ts
CBforZ: I like "failt"

zFernand0:
Thanks for the reviews 🙏 
And sorry for the typos 😓

traeok:
```
 FAIL  packages/cli/__tests__/daemon/__unit__/DaemonClient.unit.test.ts
  ● DaemonClient tests › should fail to initialize the DaemonClient if no token is provided
```

Tests are failing because they still expect the typo - but once that is resolved I'm ready to approve 😋

packages/cli/src/daemon/doc/IDaemonPidForUser.ts
```diff
@@ -16,4 +16,13 @@
...
     pid: number;
+
+    /**
+     * Secret token that authenticates a daemon client as the owner of this
+     * daemon. The daemon generates a fresh random token each time it starts
+     * and writes it into this owner-only file. Because only the owner can read
+     * the file, a client that can echo this token back has proven it is the
+     * owner. The client sends it on every request (see {@link IDaemonResponse}).
+     */
+    token?: string;
```
t1m0thyj, line 27: I believe the same question applies here as to the Rust code regarding whether the `token` parameter should be optional or required 😋 

Ok with either approach - I don't think it would be a breaking change to make it required since this interface is not exported. But I assume the reason it's currently optional is because old versions of the daemon client may not define it?

zowex/src/run.rs
```diff
@@ -292,25 +295,11 @@ pub async fn run_daemon_command(
...
-            if response.stdinLength.unwrap() > 0 {
-                _resp.push(b'\x0c');
-                _resp.append(&mut stdin);
-            }
-        }
-        Err(err_val) => {
-            eprintln!("Failed to convert response to JSON\nDetails = {}", err_val);
-            return Err(EXIT_CODE_CANT_CONVERT_JSON);
-        }
-    }
```
t1m0thyj, line 312: Curious why this code is being removed? We still need to handle multipart JSON bodies to support piping input to commands over stdin. The parsing logic is here:
https://github.com/zowe/zowe-cli/blob/8f1820bac28b566a5949bff0791b7337e8b9a27f/packages/cli/src/daemon/DaemonClient.ts#L145-L146

t1m0thyj (changes requested):

zowex/src/run.rs
t1m0thyj: Sorry please disregard, I see that this logic was moved.

t1m0thyj (approved):
LGTM, thanks @zFernand0!

packages/cli/src/daemon/DaemonClient.ts
```diff
@@ -209,4 +236,33 @@ export class DaemonClient {
...
+     *
+     * @private
+     * @param {string} requestToken The token supplied by the daemon client.
+     * @returns {boolean} True if the token is present and matches our token.
+     * @memberof DaemonClient
+     */
+    private isValidToken(requestToken: string): boolean {
+        if (this.mDaemonToken == null || typeof requestToken !== "string" || requestToken.length === 0) {
+            return false;
+        }
```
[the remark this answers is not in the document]
zFernand0: Right in the function signature, one can see that the parameter `requestToken` is a required parameter. 🤔

packages/cli/src/daemon/DaemonClient.ts
```diff
@@ -191,6 +197,27 @@ export class DaemonClient {
...
+            Imperative.api.appLogger.warn("A connection was attempted with a missing or invalid daemon token.");
+            const responsePayload: string = DaemonRequest.create({
+                stderr: "The daemon client did not supply a valid daemon token. " +
+                    "Try restarting the daemon with 'zowe daemon restart'.\n",
+                exitCode: 1
+            });
+            this.mClient.write(responsePayload);
+            this.mClient.end();
+            return;
+        }
```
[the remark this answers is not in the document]
zFernand0: This check seem redundant since the method is private and the constructor is already handling this. 

And clearing the token right after validating is legit! 🙏

zowex/src/test.rs
```diff
@@ -110,6 +110,43 @@ fn unit_test_util_get_socket_string() {
...
+fn unit_test_util_get_daemon_token_from_dir() {
+    use std::fs;
+    use std::path::PathBuf;
+
+    // Use an isolated daemon dir and read it directly (not via ZOWE_DAEMON_DIR)
+    // so this test does not race with other tests that mutate that env variable.
+    let mut token_test_dir: PathBuf = env::temp_dir();
+    token_test_dir.push("zowe_daemon_token_test");
+    let _ = fs::remove_dir_all(&token_test_dir);
+    fs::create_dir_all(&token_test_dir).unwrap();
```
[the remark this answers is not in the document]
zFernand0: fair point! 🙏

packages/cli/src/daemon/DaemonClient.ts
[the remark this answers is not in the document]
zFernand0: resolved in 1f4bcf73e

zowex/src/test.rs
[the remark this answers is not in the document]
zFernand0: resolved in 1f4bcf73e

traeok (approved):

PR #2771, opened 2026-06-30
fix: encode diff content before embedding it

traeok:
**What It Does**

Encodes diff content before its embedded within web diffs to prevent stored cross-site scripting.

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (2368)


zFernand0 (commented):
FWIW, LGTM! 😋

zFernand0 (approved):

packages/imperative/src/utilities/src/diff/WebDiffManager.ts
```diff
@@ -112,6 +112,23 @@ export class WebDiffManager implements IWebDiffManager {
...
+     * Encodes a string so it can be safely embedded inside an inline `<script>`
+     * element. `JSON.stringify` produces a valid JavaScript string literal, and
+     * the additional replacements prevent the value from terminating the script
+     * element (e.g. a `</script>` sequence in the diff content) or introducing
+     * line terminators that are illegal inside JS string literals (U+2028/U+2029).
+     * @private
+     * @param str - the raw string to embed in a script context
+     * @returns A safely-encoded JavaScript string literal
+     */
+    private encodeForScript(str: string): string {
```
t1m0thyj, line 125: Changes LGTM, but Sonar appears to have some comments 😋 

Also unsure whether we want to pick an approach (`sanitizeHtml` vs manual encoding) to use in both places where we need to escape text for web diffs.

traeok: Discussed off of GitHub and determined that the approaches are complimentary to one another 👍

anaxceron (approved):

awharn (approved):

PR #2773, opened 2026-07-01
[v2] strict modes for temp files on shared systems

jace-roell:
**What It Does**

- updated temp file creation for workflow commands to create with 700 permissions
- updated `zos-files ds edit` command to store temp files in a dedicated sub directory with owner-only permissions

**How to Test**
Run `npx jest -- packages/workflows/__tests__/__system__/api/Create.system.test.ts`
Run `npx jest -- packages/workflows/__tests__/__unit__/Create.unit.test.ts`

Check local/remote temp file permissions when running `zos-files ds edit` (local) or `workflow (wf) create`

**Review Checklist**
I certify that I have:
- [x] updated the changelog


anaxceron (approved):

packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -85,7 +85,9 @@ export class EditUtilities {
             hash = hash.slice(0, hashLen);
             return path.join(tmpdir(), path.parse(lfFile.fileName).name + '_' + hash + ext);
         }
-        return path.join(tmpdir(), lfFile.fileName + ext);
+        const dsDir = path.join(tmpdir(), "zowe-edit-ds");
+        mkdirSync(dsDir, { recursive: true, mode: 0o700 });
```
traeok: What if a user has already created the `zowe-edit-ds` dir in their temp folder with less restrictive permissions? (e.g. 777)

At that point, the `mkdirSync` call would succeed because the directory already exists, but the permissions don't match. Should we check if the dir exists first and throw an error if the permissions are invalid? For example:

```ts
if (existsSync(dsDir)) {
    const st = lstatSync(dsDir);
    if (!st.isDirectory() || (st.mode & 0o777) !== 0o700 || st.uid !== process.getuid?.()) {
        throw new ImperativeError({ msg: `Unsafe temp directory detected at ${dsDir}` });
    }
} else {
    mkdirSync(dsDir, { recursive: true, mode: 0o700 });
}
```

traeok (commented):
Changes make sense to me, but I left a suggestion regarding the logic for creating the temp `zowe-files-edit` dir 👍

traeok (approved):
LGTM, thanks for adding multiple tests around the updated logic 👍

zFernand0 (approved):
LGTM! 😋

PR #2778, opened 2026-07-01
refactor: censor equals-form command options in logs

traeok:
**What It Does**

Refactors censoring of command-line options to cover equals-form (`--pw=example`) as well as options containing embedded whitespace.

**Note:** I've scoped out `censorCLIArgs` from this PR: it has the same `indexOf`-based `=`-blindness, but it operates on already-split arg arrays and is only reachable through deprecated `CliUtils`/`LoggerUtils` wrappers. I left it untouched to keep the fix focused. If we want to harden this logic, please leave it as a suggestion and I can pursue it as part of this work.

**How to Test**

- Set your log level to trace for Zowe CLI
- Execute a Zowe CLI command with a "censored" option passed as equals-form (`zowe files list ds ... --pw=example`)
- Notice that the option's value is now properly censored within the log file

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] manually tested my changes
- [x] added/updated automated unit/integration tests
- [x] created/ran system tests (run 2375)
- [x] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

PR #2782, opened 2026-07-07
refactor(v2): censor equals-form command options in logs

traeok:
**What It Does**

Refactors censoring of command-line options to cover equals-form (`-u=example`) as well as options containing embedded whitespace.

**How to Test**

- Set your log level to trace for Zowe CLI
- Execute a Zowe CLI command with a censored option passed as equals-form (`zowe files list ds ... --pw=example`)
- Notice that the option's value is now properly censored within the log file

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (job number 2401)

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Masked sensitive command-line options that are supplied in the equals-separated form (for example, `--user=example` or `-u=example`) before the command line is written to the log, and also mask secure values that contain embedded whitespace.
+
 ## `5.27.19`
 
 - BugFix: Added the `base64EncodedAuth` session property to the list of session properties redacted from Imperative debug logs. [#2780](https://github.com/zowe/zowe-cli/pull/2780)
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
-- BugFix: Masked sensitive command-line options that are supplied in the equals-separated form (for example, `--user=example` or `-u=example`) before the command line is written to the log, and also mask secure values that contain embedded whitespace.
+- BugFix: Masked sensitive command-line options that are supplied in the equals-separated form (for example, `--user=example` or `-u=example`) before the command line is written to the log, and also mask secure values that contain embedded whitespace. [#2782](https://github.com/zowe/zowe-cli/pull/2782)
 
 ## `5.27.19`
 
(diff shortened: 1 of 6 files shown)


packages/imperative/src/logger/src/LoggerUtils.ts
```diff
@@ -74,6 +84,64 @@ export class LoggerUtils {
...
+        if (commandLine == null || commandLine.length === 0) { return commandLine; }
+        let censoredLine = commandLine;
+        const censoredOptions = LoggerUtils.COMMAND_LINE_CENSORED_OPTIONS;
+
+        // Value-based censoring first, using the parsed arguments. Because we
+        // know the exact value, this reliably masks values containing embedded
+        // whitespace that the token-based regex below would otherwise truncate.
+        if (args) {
+            for (const optName of Object.keys(args)) {
+                if (optName === "_" || optName === "$0") { continue; }
```
CBforZ: I assume this is to skip positional arguments. It's not a huge deal, but at first I was wondering about adding a comment here...but then also realized _ or $0 are not in censoredOptions so strictly speaking this check isn't necessary right, since the next .includes check would already skip _ and $0? 

Also can't we use LoggerUtils.COMMAND_LINE_CENSORED_OPTIONS directly rather than re-declaring a constant? Being nitpicky... not a big deal if you don't want to change it

CBforZ (approved):
Looks good and well tested. Thank you

packages/imperative/src/logger/src/LoggerUtils.ts
traeok: The `args` is the parsed command-line arguments - at this point they may not contain `$0` or `_` but the type suggests that they can be present, and the caller could technically include them:

image

But I agree on the suggestion for using the constant directly instead of re-declaring (even though they refer to the same array) - refactored in upcoming commit 👍

CBforZ: > The args is the parsed command-line arguments - at this point they may not contain $0 or _ but the type suggests that they can be present:

I just mean, in the case that `optName === "_" || optName === "$0"` ,  then the next line `          if (LoggerUtils.COMMAND_LINE_CENSORED_OPTIONS.includes(optName)) {` will already be false right?

traeok: Ah, I now understand what you mean - this `if` condition is redundant, since the next `if` condition is false anyway for both strings. Fair point, I'll drop this `if` to minimize LOC 😋

traeok: Removed in 0cdd23d

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Masked sensitive command-line options that are supplied in the equals-separated form (for example, `--user=example` or `-u=example`) before the command line is written to the log, and also mask secure values that contain embedded whitespace.
```
anaxceron: Can we include a link w/ this entry?

traeok: Sorry for the delay - added in b96c67a

pujal0909 (approved):
LGTM

zFernand0 (approved):
LGTM! 😋 
---
Thanks for porting #2778

PR #2787, opened 2026-07-09
Fix: Restrict team config profile properties to schema declared fields

pujal0909:
**What It Does**

Restricts team config profile properties to schema declared fields

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+- BugFix: Restricted profile properties sourced from team config to only those declared in the profile type's schema, preventing config injection of arbitrary command options.
+
 ## `5.27.20`
 
 - BugFix: Masked sensitive command-line options that are supplied in the equals-separated form (for example, `--user=example` or `-u=example`) before the command line is written to the log, and also mask secure values that contain embedded whitespace. [#2782](https://github.com/zowe/zowe-cli/pull/2782)
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
-- BugFix: Restricted profile properties sourced from team config to only those declared in the profile type's schema, preventing config injection of arbitrary command options.
+- BugFix: Restricted profile properties sourced from team config to only those declared in the profile type's schema, preventing config injection of arbitrary command options. [#2787](https://github.com/zowe/zowe-cli/pull/2787)
 
 ## `5.27.20`
 
--- a/packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
+++ b/packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
@@ -110,6 +110,30 @@ const SAMPLE_COMMAND_REAL_HANDLER_WITH_OPT: ICommandDefinition = {
     }
 };
 
+// Command definition that declares both a schema-defined option ("color") and a
+// command-specific option ("editor") to exercise config injection protection.
+const SAMPLE_COMMAND_WITH_EDITOR_OPT: ICommandDefinition = {
+    name: "edit",
+    description: "The edit command",
+    type: "command",
+    handler: __dirname + "/__model__/TestArgHandler",
+    options: [
+        {
+            name: "color",
+            type: "string",
+            description: "The banana color.",
+        },
+        {
+            name: "editor",
+            type: "string",
+            description: "The editor to use.",
+        }
+    ],
+    profile: {
+        optional: ["banana"]
+    }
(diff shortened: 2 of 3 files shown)


packages/imperative/src/cmd/src/CommandProcessor.ts
```diff
@@ -960,7 +960,14 @@ export class CommandProcessor {
                     fulfilled.push(profileType);
                     p = this.mConfig.api.profiles.defaultGet(profileType);
                 }
-                fromCnfg = { ...p, ...fromCnfg };
+                // Restrict profile properties to those declared in the profile type's
+                // schema so that an untrusted zowe.config.json cannot inject arbitrary
+                // command options (e.g. "editor") by placing them inside a profile.
+                const allowedProps = this.schemaPropsForProfileType(profileType);
```
CBforZ, line 966: So, I recall a few folks mentioning recently that the `authOrder` property is not currently defined in the schema. If that's still the case, would that mean authOrder would be thrown away here?  And if so are there any other properties currently undocumented in the schema?

zFernand0: The only two properties I can think of are:
- `zosmf:jobEncoding`
- `ssh:serverPath`: this one may have the user ID in the path, but it's only used by devs 🙏

packages/imperative/src/cmd/__tests__/CommandProcessor.unit.test.ts
```diff
@@ -2730,4 +2754,101 @@ describe("Command Processor", () => {
...
+
+            const processor = new CommandProcessor({
+                envVariablePrefix: ENV_VAR_PREFIX,
+                definition: SAMPLE_COMMAND_WITH_EDITOR_OPT,
+                helpGenerator: FAKE_HELP_GENERATOR,
+                profileManagerFactory: FAKE_PROFILE_MANAGER_FACTORY_WITH_PROPS,
+                rootCommandName: SAMPLE_ROOT_COMMAND,
+                commandLine: "",
+                promptPhrase: "dummydummy",
+                config: ImperativeConfig.instance.config
```
zFernand0, line 2785: Could we share this object?
mostly trying to reduce code duplication where it makes some sense 😅

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+- BugFix: Restricted profile properties sourced from team config to only those declared in the profile type's schema, preventing config injection of arbitrary command options. [#2787](https://github.com/zowe/zowe-cli/pull/2787)
```
zFernand0, line 6: nit:
```suggestion
## Recent Changes

- BugFix: Restricted profile properties sourced from team config to only those declared in the profile type's schema, preventing config injection of arbitrary command options. [#2787](https://github.com/zowe/zowe-cli/pull/2787)
```

packages/imperative/src/cmd/src/CommandProcessor.ts
```diff
@@ -960,7 +960,14 @@ export class CommandProcessor {
...
+                const safeP = allowedProps != null
+                    ? Object.fromEntries(Object.entries(p).filter(([k]) => allowedProps.has(k)))
+                    : p;
```
zFernand0, line 969: Thanks for not erroring out if there is no schema defined for a given profile 🙏

zFernand0 (dismissed):
LGTM! 😋

packages/imperative/src/cmd/src/CommandProcessor.ts
```diff
@@ -1279,4 +1286,35 @@ export class CommandProcessor {
...
+     * @private
+     */
+    private schemaPropsForProfileType(profileType: string): Set<string> | null {
+        try {
+            const profiles = ImperativeConfig.instance.loadedConfig?.profiles;
+            if (profiles == null) {
+                return null;
+            }
+            const typeConfig = profiles.find((c) => c.type === profileType);
+            const schemaProps = typeConfig?.schema?.properties;
```
traeok, line 1305: (nit) We can skip declaration of `typeConfig` since its only used to get `schemaProps`:


```suggestion
            const schemaProps = profiles.find((c) => c.type === profileType)?.schema?.properties;
```

packages/imperative/src/cmd/src/CommandProcessor.ts
traeok: Do we need an allowlist for options like `authOrder`, `jobEncoding`, `serverPath` if we do not want to remove these properties under a profile? If present in the allowlist, we can keep these values intact

traeok (commented):
While the changes make sense, I'm wondering if its safe to make this change right now. I believe there are other plug-ins that support options not explicitly specified in the schema. If this change is pushed out, it could break commands from those plug-ins. (Not to mention, the discussion folks are having above about `authOrder` and other properties excluded from the schema)

PR #2788, opened 2026-07-09
fix: redact sensitive environment vars and CLI args

traeok:
**What It Does**

Redacts sensitive environment variable and command-line arguments when a command handler fails to load or throws an error.

**How to Test**

- Run a command with input that causes an error to be thrown (or a bad command handler - harder to repro)
- Notice that any credentials supplied via env. vars are redacted in the log

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (job number 2413)

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
-- BugFix: Redacted sensitive environment variables and command-line arguments from the diagnostic information written to the log when a command handler fails to load or throws an error, preventing credentials supplied via environment variables (for example, `ZOWE_OPT_PASSWORD`) from being persisted to disk in plain text. [#2788](https://github.com/zowe/zowe-cli/pull/2788)
+- BugFix: Redacted sensitive environment variables and command-line arguments from the diagnostic information written to the log when a command handler fails to load or throws an error. This prevents credentials supplied via environment variables (for example, `ZOWE_OPT_PASSWORD`) from being persisted to disk in plain text. [#2788](https://github.com/zowe/zowe-cli/pull/2788)
 - BugFix: Fixed an issue where passing an object with `null` values to the `Censor.mCensorObject` function caused a runtime error. Now, only non-null values are recursively handled in this function. [#2784](https://github.com/zowe/zowe-cli/pull/2784)
 
 ## `8.33.2`
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Redacted sensitive environment variables and command-line arguments from the diagnostic information written to the log when a command handler fails to load or throws an error, preventing credentials supplied via environment variables (for example, `ZOWE_OPT_PASSWORD`) from being persisted to disk in plain text. [#2788](https://github.com/zowe/zowe-cli/pull/2788)
+
 ## `8.33.2`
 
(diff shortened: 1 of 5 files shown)


packages/imperative/src/censor/src/Censor.ts
```diff
@@ -401,6 +410,36 @@ export class Censor {
...
+     *
+     * @param {NodeJS.ProcessEnv} env - the environment to censor (defaults to process.env)
+     * @returns {string} - a censored, pretty-printed JSON string of the environment
+     */
+    public static censorEnvVariables(env: NodeJS.ProcessEnv = process.env): string {
+        const censored: Record<string, string> = {};
+        for (const [name, value] of Object.entries(env)) {
+            censored[name] = this.isSecureEnvName(name) ? this.CENSOR_RESPONSE : value;
+        }
+        return this.censorRawData(JSON.stringify(censored, null, 2));
```
t1m0thyj, line 440: Is it redundant to call `censorRawData` here? I think in the places where this new method is called, the resulting string is written to an Imperative logger object which will handle censoring data.

zFernand0: It does feel redundant, but I guess we should assume that we may use this function for something other than writing to logs.

packages/imperative/src/censor/src/Censor.ts
```diff
@@ -41,6 +41,15 @@ export class Censor {
...
     private static readonly MAIN_SECURE_PROMPT_OPTIONS = ["keyPassphrase", "password", "passphrase", "tokenValue", "user"];
 
+    /*
+    * Environment-variable NAME patterns that indicate a sensitive value.
+    * Environment variables never pass through the CLI-argument censoring path, so their values must be redacted by
+    * name before the environment is serialized into diagnostic output. This catches both the credential env vars that
+    * Zowe explicitly supports (e.g. ZOWE_OPT_PASSWORD, ZOWE_OPT_TOKEN_VALUE) and unrelated secrets that commonly
+    * share the same environment (e.g. AWS_SECRET_ACCESS_KEY, GITHUB_TOKEN).
+    */
+    private static readonly SECURE_ENV_NAME_PATTERN = /PASS|TOKEN|SECRET|KEY|CRED|AUTH/i;
```
zFernand0, line 51: This should catch most cases.
And we are calling censorRawData to cover for other things (e.g. path-related, like certificate files)

zFernand0 (approved):
LGTM! 😋

packages/imperative/src/censor/src/Censor.ts
traeok: The function is `public` and exposed through the `Censor` export in the Imperative package, so I'm not sure its safe to assume that all callers are redirecting this output to the log. In this case, it might be better to leave the double-censoring in place 😅

packages/imperative/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
+- BugFix: Redacted sensitive environment variables and command-line arguments from the diagnostic information written to the log when a command handler fails to load or throws an error, preventing credentials supplied via environment variables (for example, `ZOWE_OPT_PASSWORD`) from being persisted to disk in plain text. [#2788](https://github.com/zowe/zowe-cli/pull/2788)
```
anaxceron:
```suggestion
- BugFix: Redacted sensitive environment variables and command-line arguments from the diagnostic information written to the log when a command handler fails to load or throws an error. This prevents credentials supplied via environment variables (for example, `ZOWE_OPT_PASSWORD`) from being persisted to disk in plain text. [#2788](https://github.com/zowe/zowe-cli/pull/2788)
```

anaxceron: Thinking that splitting it up into two sentences might make it easier to read.

anaxceron (approved):

CBforZ (approved):
😎

packages/imperative/CHANGELOG.md
traeok: Good call, thanks for the suggestion 😸

PR #2789, opened 2026-07-10
Render falsy table values such as 0 and false

### LINKED ISSUE #2698: Numeric values in table rows are not rendered when value is `0`
**Describe the bug**

If I am displaying structured data in a table using Zowe CLI, the formatter does not display `0` for numeric values. The value is `0` when using `--response-format-json` but the table does not render the value.

**Expected:** Zowe CLI should render numeric properties with a value `0` in a table
**Actual:** Zowe CLI does not render numeric properties in a table when the value is `0`

**Describe your environment**

Atishyy27:
**What It Does**

Fixes #2698.

`TextUtils.getTable` built each cell with `obj[header] || ""`, so any falsy-but-meaningful value collapsed to an empty string:

```ts
return TextUtils.wordWrap(obj[header] || "", maxColumnWidth, "", hardWrap);
```

A numeric `0` rendered as a blank cell, while `--response-format-json` (which serialises `mData` directly) still reported the `0` — so the table looked like it was missing data that the JSON output showed.

The cell value is now coerced explicitly rather than tested for truthiness. `null`/`undefined` still render as an empty cell; `0` and `false` render as `"0"` and `"false"`.

Note this also changes `false` from an empty cell to `"false"`, which I believe is the desired behaviour for the same reason `0` is — it matches what `--response-format-json` reports. Happy to restrict the fix to numbers only if reviewers would rather not change the `boolean` rendering.

**How to Test**

Before this change, the `errors` and `active` columns are blank on the first row. After it:

```
name  errors active
alpha 0      false
beta  7      true
```

Two unit tests were added to `TextUtils.unit.test.ts`:
- `should render falsy-but-meaningful values such as 0 and false`
- `should render an empty cell for null and undefined values`

The first fails on `master` (`Expected substring: "0"` / `Received string: "    row1"`) and passes with this change. The existing `getTable` snapshots are unchanged.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

I did not run the system tests — they need a live z/OS connection I don't have. The change is confined to a pure string-formatting helper in `packages/imperative` with no mainframe interaction, so unit coverage should be sufficient here, but do let me know if you'd like a system-test run before merging.

Verified locally on Windows 11 (Node 22): `TextUtils.unit.test.ts` 18 passed / 4 snapshots passed, and `eslint` clean on both touched files. `tsc --noEmit` reports 3 pre-existing errors in `ConvertV1Profiles.ts` and `DefaultCredentialManager.ts` that are present on an unmodified checkout too, and are unrelated to this change.


Atishyy27:
Hi maintainers — this is my first contribution to zowe-cli, so the workflow runs are sitting in `action_required` and CI hasn't been able to start. Could someone approve the workflow run when you get a chance? Happy to rebase onto latest `master` first if that's easier.

For what it's worth, locally on Windows 11 (Node 22) the `packages/imperative` config suite passes and `eslint` is clean on the touched files. No rush — just flagging it, since the required checks can't report until the run is approved.

packages/imperative/src/utilities/src/TextUtils.ts
```diff
@@ -201,7 +201,11 @@ export class TextUtils {
 
         for (const obj of objects) {
             const row = headers.map((header) => {
-                return TextUtils.wordWrap(obj[header] || "", maxColumnWidth, "", hardWrap);
+                const value = obj[header];
+                // Coerce rather than rely on truthiness: falsy-but-meaningful values
+                // such as 0 and false must still be rendered.
+                const text = value == null ? "" : String(value);
+                return TextUtils.wordWrap(text, maxColumnWidth, "", hardWrap);
```
t1m0thyj: A simpler fix would be to replace `||` in the original code with a null-coalescing operator 😋 
```suggestion
                return TextUtils.wordWrap(obj[header] ?? "", maxColumnWidth, "", hardWrap);
```

t1m0thyj (commented):
Thanks for your contribution @Atishyy27! Please resolve merge conflicts, plan to approve after tests are passing.

Atishyy27:
Hi @t1m0thyj, thanks for the review. I've resolved the merge conflicts and pushed the updated branch. The workflow runs now require maintainer approval again because this is my first contribution. When you have a chance, could you approve the workflows so CI can run? One heads-up: the audit job is expected to report the existing RustSec advisory (RUSTSEC-2026-0204 in crossbeam-epoch from Cargo.lock), which isn't related to the TypeScript changes in this PR. Thanks!

packages/imperative/src/utilities/src/TextUtils.ts
CBforZ: I agree I'd prefer this and the comment doesn't seem as necessary with this  shorter version

traeok (commented):
Once the suggested change from @t1m0thyj is implemented and all tests are passing, I plan to approve 👍

Atishyy27:
good call, applied the null-coalescing version and dropped the comment. thanks @t1m0thyj @CBforZ @traeok 🙏

CBforZ (approved):
Thanks for the updates

zFernand0 (approved):
LGTM! 😋 
---
This will also handle `NaN`, `{}` and all funny variations of `0` (`-0.0`, `0o000`, `0x0`) 😋

Issue #2302, opened 2024-10-10
Schema validation is broken for typeless base profiles


t1m0thyj:
**Describe the bug**

When using a typeless profile as a base profile and storing `tokenValue` as a secure field, then schema validation fails.

**Expected and actual results**

Since typeless profiles are often used as base profiles, I would expect that `tokenValue` property is supported in them.

For typeless profiles, the first item in the schema array gets used which is currently `zosmf`. We could fix this by re-ordering the schema array to put base profiles first, but will need to consider the impacts of this change.

**Describe your environment**

Schema created by Zowe CLI v8.0.1
Config edited in VS Code 1.94.1

**Additional context**

![image](https://github.com/user-attachments/assets/b476342e-a2a8-4360-a83b-207cf30fceab)

t1m0thyj:
If we switch the order, this may break other properties defined in a typeless profile such as `encoding`.
Ideally IntelliSense for typeless profiles could show all properties available across all profiles, but this might cause conflicts if multiple profiles define the same property (e.g. zosmf user vs zftp user).

PR #2792, opened 2026-07-10
Require type in config schema conditionals for typeless profiles

### LINKED ISSUE #2602: Zowe schema erroneously flags http as unsupported value for protocol
**Describe the bug**

When editing my nested profiles in Team Config the "protocol" field erroneously marked as not allowed value, when "http" values are used. At the same time the description of the field shows that HTTP or HTTPS can be used.
Image

An example of the nested profile:

```json
    "tso1": {
      "properties": {
        "host": "test.net",
        "port": 50004,
        "protocol": "http"
      },
      "profiles": {
        "ndvr-v2": {
          "type": "endevor",
          "properties": {
            "basePath": "/EndevorService/api/v2/"
          },
          "secure": []
        },
        <...>
    },
}
```

After checking the schema, I noticed that only the RSE profile schema includes a protocol field that is limited to "https" only so it may be a reason it erroneously applied to the nested profiles property section.

```json
{
                            "if": {
                                "properties": {
                                    "type": {
                                        "const": "rse"
                                    }
                                }
                            },
                            "then": {
                                "properties": {
                                    "properties": {
                                        "type": "object",
                                        "title": "IBM RSE API Zowe Profile",
                                        "description": "A profile to issue commands to a z/OS system with a working Zowe REST server,Mediation Layer, or IBM RSE (Remote System Explorer) API server installation.",
                                        "properties": {
                                            <...>,
                                            "protocol": {
                                                "type": "string",
                                                "description": "Only HTTPS supported for a TLS handshake to access REST API.",
                                                "default": "https",
                                                "enum": [
                                                    "https"
                                                ]
                                            },
                                        }
}
```

The notification about limited protocols supported should be applied only to the profiles that have this constraint. 

**Describe your environment**

- Zowe CLI version installed (run command `zowe --version`):
  ```
  CLI Version: 8.26.2
  Zowe Release Version: v3.3.0
  ```
- Installed plug-ins and their version numbers (run command `zowe plugins list --short`):
  ```
  @broadcom/endevor-bridge-for-git-for-zowe-cli@2.7.0
  @broadcom/endevor-for-zowe-cli@8.2.1
  ```
- Node.js and NPM versions installed (run `node --version`, `npm --version`):
  ```
  ❯ node --version
  v18.20.8
  ❯ npm --version
  10.8.2
  ```
- Environment variables in use: N/A
- Operating system and version:
  ```
  ProductName:            macOS
  ProductVersion:         15.6.1
  BuildVersion:           24G90
  ```
- Shell/terminal (bash, cmd, powershell, etc...): zsh
- Daemon mode enabled or disabled: disabled

Atishyy27:
**What It Does**

Fixes #2602 (and addresses the erroneous-rejection symptom in #2302).

`ConfigSchema.buildSchema` emits one `if`/`then` per profile type inside `allOf`:

```ts
andEntries.push({
    if: { properties: { type: { const: profile.type } } },
    then: { properties: this.generateSchema(profile.schema) }
});
```

The `if` constrains the *value* of `type` but never requires `type` to be **present**. Under JSON Schema, a `properties` constraint is vacuously satisfied when the property is absent, so for a profile with no `type` (a base/typeless profile) *every* type's `then` is applied. That is why the RSE-only `protocol: https` rule was imposed on a typeless profile and a valid `protocol: http` was flagged as invalid, as reported in #2602.

Fix: add `required: ["type"]` to each conditional so a type-specific rule only applies to a profile that actually declares that type. The existing `{ if: { properties: { type: false } }, then: … "Missing profile type" }` block is unchanged.

**How to Test**

Unit: `npx jest packages/imperative/src/config/__tests__/ConfigSchema.unit.test.ts`

Added a regression test, "should require the type property in each conditional so type-specific rules are not applied to typeless profiles", which asserts every `type`-const conditional now carries `required: ["type"]`. It fails on `master` (`Expected: ["type"]` / `Received: undefined`) and passes with this change. The three existing `buildSchema` structural assertions and the four snapshots were updated; the snapshot diff is purely the added `required: ["type"]` arrays (no other change).

End to end: build a team config with a typeless base profile that sets `protocol: "http"` (as in the #2602 example) alongside an RSE profile type. Before this change the schema flags `protocol: "http"`; after it, the value validates.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

Relationship to #2302: that issue describes the same root cause from the other direction — a typeless base profile having a valid property (`tokenValue`) rejected because a type-specific schema was applied. This change stops type-specific rules from being applied to typeless profiles at all, so it removes that erroneous rejection too. It does **not** implement #2302's separate suggestion of positively applying the base-profile schema to typeless profiles (reordering the schema array); I've scoped this PR to the conditional-`required` fix and am happy to follow up on #2302 separately if maintainers prefer that direction.

I did not run the z/OS system tests — they require a live mainframe connection I don't have. The change is confined to schema generation in `packages/imperative` with no mainframe interaction, so unit coverage should be sufficient, but let me know if you'd like a system-test run first.

Verified locally on Windows 11 (Node 22): the config suite is 25 passed / 5 snapshots passed, and `eslint` is clean on both touched files.


Atishyy27:
Hi maintainers — this is my first contribution to zowe-cli, so the workflow runs are sitting in `action_required` and CI hasn't been able to start. Could someone approve the workflow run when you get a chance? Happy to rebase onto latest `master` first if that's easier.

For what it's worth, locally on Windows 11 (Node 22) the `packages/imperative` config suite passes and `eslint` is clean on the touched files. No rush — just flagging it, since the required checks can't report until the run is approved.

Atishyy27:
@zFernand0 @t1m0thyj this is blocked on workflow approval, not review 
CI can't even run until that's clicked. flagging since you're both already on it.

zFernand0:
Hey @Atishyy27,

zFernand0 (changes requested):
Adding the `required: ["type"]` to all profiles mitigates the linked issues.
However, there is no description for any of the `properties: {}`

Image

My proposal is to not add the `required: ["type"]` to the base profile which will achieve the results we wanted without having to reorder the profiles

Image

Image

Atishyy27:
thanks @zFernand0, updated per your suggestion: the `base` profile no longer gets `required: ["type"]` in its conditional, so its shared defaults still apply to every profile without needing to reorder the profiles. only the type-specific conditionals require `type` now. regression test, snapshot, and changelog updated to match.

on the missing descriptions: `generateSchema` already carries a property's `description` over from its option definition where one is set, so the empty `properties: {}` are likely properties whose definitions don't declare one. did you want descriptions added to specific ones, or was there a particular conditional you meant? happy to fold it into this PR.

zFernand0 (changes requested):
LGTM! 😋 
---

Please fix failing builds 🙏

PR #2793, opened 2026-07-12
Always allocate cross-LPAR copy targets in tracks

### LINKED ISSUE #2623: Allocation units are not handled correctly when copying data sets between LPARs
**Describe the bug**

Zowe Explorer allocates either too much space or fails silently when attempting to copy a data set between LPARs if that data set was not allocated in tracks.

I moved this to the CLI repo since I'm seeing the issue of data sets allocated in cylinders being copied to 225-track data sets on the target system on both Zowe CLI and Zowe Explorer.

There seem to be other issues with non-track/cylinder allocation units for copying PDSs with Zowe Explorer, so there may need to be an additional fix for Zowe Explorer.

**To Reproduce**
Copying a data set allocated in cylinders:
1. Allocate a 1 cylinder data set (which amounts to 15 tracks).
2. Attempt to copy that data set from one LPAR to another.
3. Check the allocation information for the new copy of the data set.
4. Note that size is now 225 tracks (15^2).

Copying a PDS allocated in MB (attempting to copy a data set allocated in blocks fails similarly):
1. Allocate a 1 MB data set (which seemed to amount to ~23 tracks).
2. Attempt to copy that data set from one LPAR to another.
3. Note that nothing seems to happen with Zowe Explorer.
  a. Both Zowe CLI and Zowe Explorer handle this allocation correctly when given a **sequential data set**.

**Expected behavior**

I would expect the new data set to have the same number of allocated tracks as the original data set (assuming we cannot get more granular allocation information than track count - which is provided to us by z/OSMF).

Presumably, it would be safe to always allocate the copy in tracks since that is the allocation unit described by the `Size` field in the 'show attributes' view. Is there any harm in ignoring the `Space Unit` field?

**Screenshots**

**Desktop (please complete the following information):**

- OS:
- Zowe Explorer Version: 3.3.0
- (Optional) Zowe CLI Version:
- (Optional) Do you have secure credentials enabled?
- (Optional) Are you using a custom credential manager?

**Additional context**

Atishyy27:
**What It Does**

Fixes #2623.

When copying a data set across LPARs, `Copy.generateDatasetOptions` built the target's allocation from the source's listing:

```ts
alcunit: Copy.convertAlcTozOSMF(dsInfo.spacu),
...
primary: parseInt(dsInfo.sizex),
```

As @awharn diagnosed in the issue, the z/OSMF list API always reports `sizex` in **tracks**, but `spacu` still reports the unit the data set was *originally* allocated in. `convertAlcTozOSMF` mapped `CYLINDERS` to `CYL`, so a track count was submitted paired with a cylinder unit — a 15-track source was allocated as 15 cylinders, i.e. 225 tracks, exactly as reported.

`convertAlcTozOSMF`'s own doc comment already stated the correct behaviour ("we will always allocate the new target in tracks"); the implementation contradicted it. This change makes the code match that intent: `alcunit` is now always `TRK`, so it agrees with the `primary` value taken from `sizex`. The now-constant helper is removed and the reasoning kept as a comment at the point of use.

**How to Test**

Unit: `npx jest packages/zosfiles/__tests__/__unit__/methods/copy/Copy.unit.test.ts`

Added a regression test, "should allocate the target in tracks when the source was allocated in cylinders", which lists a source with `spacu: "CYLINDERS"` / `sizex: "15"` against a non-existent target and asserts the options passed to `Create.dataSet`. On `master` it fails with `Expected: "TRK"` / `Received: "CYL"`; with this change it passes.

Worth flagging: the two existing tests named "validate spacu" assert `expect(dataSetPOCYL.spacu).toBe("CYL")` — that re-checks the fixture the test itself declared, and never inspects what reaches `Create`, which is why this slipped through. Those tests also mock the target as already existing, so `Create` is never called in them. I left them alone to keep this diff focused, but I'm happy to tighten them in this PR if you'd like.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

I did not run the z/OS system tests — they need a live mainframe connection I don't have. The change is confined to how `Create` options are derived, with no new mainframe interaction, so unit coverage should cover it; happy to have a system-test run if you'd prefer confirmation on a real LPAR.

Verified locally on Windows 11 (Node 22): the full `zosfiles` unit suite passes (18 suites, 671 tests), and `npm run lint` / `npm run lint:tests` are clean for the package.

One open question for maintainers: this always allocates in tracks, which is what the existing comment intended and what makes `primary` correct. It does mean a source allocated in cylinders yields a target described in tracks (same actual size). If you'd rather preserve the cylinder unit, the alternative would be converting the track count back to cylinders (`Math.ceil(sizex / 15)`) — happy to switch if that's preferred, though it risks rounding up.


Atishyy27:
@zFernand0 @t1m0thyj bumping this - been open since July 12, DCO's green, no review yet. let me know if it needs anything from me.

zFernand0:
Hey @Atishyy27,

PR #2794, opened 2026-07-13
Normalize the Windows project directory to its real path

### LINKED ISSUE #2604: Separate SCS creds are stored for CMD and GitBash
**Describe the bug**

The directory path stored in the secure credential store creates separate entries when paths are identical except for case. On Windows, this can lead to multiple entries when switching between a CMD window and a GitBash window.

1. In CMD:
    - `mkdir ops` (lower case ops)
    - `chdir ops`
    - `zowe config secure`
2. In GitBash:
    - `cd OPS` (Upper case OPS. This is allowed on Windows and it enters the same directory as CMD did)
    - `zowe config secure`

You get two different entries in the Secure Credential Store, which differ by case, even though you only have one directory. The result is that when you run later zowe commands you may use the first set of credentials instead of your latest set of credentials depending on which command window you are in and whether you specified upper or lower case when changing into that directory with a given command window.

Since the information is hidden in the Secure Credential Store, the user has no indication of why the later zowe commands fail.

Here are the contents of such an SCS vault. Notice that the first directory is lowercase 'ops' in the first entry but the first directory is uppercase 'OPS' in the second entry. There is only one ops directory on this Windows system.

```
b'{
    "D:\\\\ops\\\\OPSMVS-REST-API\\\\zowe.config.json":
    {
        "profiles.zosmf.properties.user":"user",
        "profiles.zosmf.properties.password":"password",
        "profiles.tso.properties.user":"user",
        "profiles.tso.properties.password":"password",
        "profiles.ssh.properties.user":"user",
        "profiles.ssh.properties.password":"password",
        "profiles.ops.properties.user":"user",
        "profiles.ops.properties.password":"password",
        "profiles.zftp.properties.user":"user",
        "profiles.zftp.properties.password":"password"
    },
    "D:\\\\OPS\\\\OPSMVS-REST-API\\\\zowe.config.json":
    {
        "profiles.zosmf.properties.user":"user",
        "profiles.zosmf.properties.password":"password",
        "profiles.ssh.properties.user":"user",
        "profiles.ssh.properties.password":"password"
    }
}'
```

**Describe your environment**

**Additional context**

A possible solution may be to make all SCS entries case-insensitive when on Windows.

Atishyy27:
**What It Does**

Fixes #2604.

Windows paths are case insensitive, so `chdir ops` in CMD and `cd OPS` in Git Bash land in the same directory, but `process.cwd()` reports back whatever casing was typed. That value becomes `mProjectDir`, which `layerPath()` turns into the layer path, which in turn is the key the secure credential store is written under. So one real directory ends up with two vault entries, which is what the issue reports.

`Config.reload` already normalized paths on Windows, but only when they were passed in through `opts`:

```ts
if (process.platform === 'win32') {
    if (opts?.projectDir) {
        opts.projectDir = fs.realpathSync.native(opts.projectDir);
    }
    ...
}
...
this.mProjectDir = opts?.projectDir ?? process.cwd();
```

The ordinary CLI invocation doesn't pass `projectDir`, so it falls through to the un-normalized `process.cwd()` and the normalization never runs for the case that actually triggers the bug.

This moves the normalization to after the directories are resolved, so it covers both the explicit and the `process.cwd()` paths.

`realpathSync.native` throws for a path that doesn't exist, which is legitimate for the home directory on a first run, so the helper returns the path unchanged in that case rather than letting `reload` fail.

**How to Test**

Unit: `npx jest packages/imperative/src/config/__tests__/Config.unit.test.ts`

Added two tests:
- "should normalize the current working directory on Windows when projectDir is not supplied" is the regression test. It fails on `master` (`Expected: "C:\users\test\OPS"`, i.e. the raw cwd is used) and passes with this change.
- "should keep the directory as-is on Windows when it cannot be resolved" covers the non-existent directory case, so a first run can't start throwing.

The two existing normalization tests are unchanged and still pass.

End to end on Windows, as in the issue: `mkdir ops`, `chdir ops`, `zowe config secure`, then from Git Bash `cd OPS`, `zowe config secure`. Before this change the vault gains a second entry keyed on the upper-case path; after it, both resolve to the same entry.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

I did not run the z/OS system tests, since they need a live mainframe connection I don't have. This change is confined to local path handling in `packages/imperative` with no mainframe interaction.

Verified on Windows 11 (Node 22), which is the platform the bug is specific to: the `Config` suite is 66 passed / 9 snapshots, the whole config package is 424 passed across 12 suites, and `npm run lint` is clean.

One note on scope: this normalizes the directory used for the layer path, so credentials saved *before* this change under a differently cased path won't be found afterwards. That is the same as any other change of the vault key, and it self-corrects the next time the user runs `zowe config secure`. Happy to add a migration or a warning if you'd rather handle that explicitly.

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
+- BugFix: Fixed an issue on Windows where running a command from the same directory with different casing, such as `ops` in CMD and `OPS` in Git Bash, stored a separate set of credentials in the secure credential store for each. The project directory is now resolved to its real path, so one directory has one set of credentials. [#2604](https://github.com/zowe/zowe-cli/issues/2604)
 - BugFix: Fixed an issue where passing an object with `null` values to the `Censor.mCensorObject` function caused a runtime error. Now, only non-null values are recursively handled in this function. [#2784](https://github.com/zowe/zowe-cli/pull/2784)
 
 ## `8.33.2`
--- a/packages/imperative/src/config/__tests__/Config.unit.test.ts
+++ b/packages/imperative/src/config/__tests__/Config.unit.test.ts
@@ -918,6 +918,59 @@ describe("Config tests", () => {
             Object.defineProperty(process, 'platform', { value: os.platform(), configurable: true });
         });
 
+        it("should normalize the current working directory on Windows when projectDir is not supplied", async () => {
+            // Regression test for #2604. Windows paths are case insensitive, so `chdir ops` and
+            // `cd OPS` land in the same directory but `process.cwd()` reports the casing that was
+            // typed. The project directory becomes the layer path, which is the secure credential
+            // store key, so an unnormalized cwd produces a second vault entry for one directory.
+            Object.defineProperty(process, 'platform', { value: 'win32', configurable: true });
+
+            const typedCwd = "C:\\users\\test\\OPS";
+            const realCwd = "C:\\users\\test\\ops";
+            jest.spyOn(process, "cwd").mockReturnValue(typedCwd);
+            const realPathSpy = jest.spyOn(fs.realpathSync, "native").mockReturnValue(realCwd as any);
+
+            const config = new (Config as any)();
+            config.mApp = MY_APP;
+            config.mLayers = [];
+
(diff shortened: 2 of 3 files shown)


Atishyy27:
@zFernand0 @t1m0thyj bumping this too - open since July 13, DCO's green, no review yet.

zFernand0:
Hey @Atishyy27,

PR #2795, opened 2026-07-13
Strip color escape sequences from the JSON response

### LINKED ISSUE #2615: Disable color when returning JSON formatted output
**Describe the bug**

When requesting JSON output from the CLI (`--response-format-json`) or programmatically, we should disable colors to enusre the JSON output can be successfully parsed.

No unicode-color characters in JSON output

**Additional context**

This came up after discussing:
- https://github.com/zowe/zowe-cli-sample-scripts/discussions/56

Note: 
- Please make sure to port this to V2 when fixing 🙏

Atishyy27:
**What It Does**

Fixes #2615.

Console output is colorized for a terminal, and those same buffers are what `writeJsonResponse` serializes. Nothing strips the escape sequences on the way, so they survive into the JSON string values:

```
"stderr": "[31mSyntax Error:[39m something went wrong\n"
```

The document is still valid JSON, so this doesn't fail a parser, which is probably why it went unnoticed. But anything consuming `--response-format-json` reads the colour codes back as part of the text, so they end up in logs, comparisons and anything rendering the value.

This strips ANSI from `stdout`, `stderr` and `message` in `writeJsonResponse`, which is the point where the response is serialized for JSON output. `buildJsonResponse` is left alone, since its other callers deal with the in-process response object rather than the JSON text.

`strip-ansi` is already a dependency of `@zowe/imperative` and is already used elsewhere in the package (`DefaultHelpGenerator`, `ConvertV1Profiles`), so this adds an import rather than a dependency.

**How to Test**

Unit: `npx jest packages/imperative/src/cmd/__tests__/response/CommandResponse.unit.test.ts`

Added "should strip ANSI color codes from the JSON response". It fails on `master`, where `stderr` comes back as `[31mSyntax Error:[39m something went wrong`, and passes with this change.

The test writes the escape sequences directly rather than producing them through chalk, because this suite calls `jest.mock("chalk")`, so a chalk-driven test would pass vacuously whether or not the fix is present. Writing the escapes explicitly tests the stripping itself and is independent of whether the process has a TTY.

End to end: run any failing command with `--response-format-json` from a terminal where colour is enabled (or with `FORCE_COLOR=1`) and pipe it through a JSON consumer. Before this change the `stderr` value carries the escape codes; after it, it is plain text.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

I did not run the z/OS system tests, since they need a live mainframe connection I don't have. This change is confined to how the response is serialized in `packages/imperative`, with no mainframe interaction.

Verified on Windows 11 (Node 22): the `cmd` package is 303 passed across 14 suites with 276 snapshots, and `npm run lint` is clean.

One scoping question for maintainers: I limited this to `stdout`, `stderr` and `message`, which are the buffers the console APIs colorize. `error` and `data` can in principle carry colorized strings too if a handler puts them there deliberately, but stripping those felt like it could destroy data a caller meant to keep, so I left them. Happy to widen it if you'd prefer.

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
+- BugFix: Fixed an issue where color escape sequences appeared inside the output of commands run with `--response-format-json`, so tools consuming that JSON read the escape codes back as part of the text. [#2615](https://github.com/zowe/zowe-cli/issues/2615)
 - BugFix: Fixed an issue where passing an object with `null` values to the `Censor.mCensorObject` function caused a runtime error. Now, only non-null values are recursively handled in this function. [#2784](https://github.com/zowe/zowe-cli/pull/2784)
 
 ## `8.33.2`
--- a/packages/imperative/src/cmd/__tests__/response/CommandResponse.unit.test.ts
+++ b/packages/imperative/src/cmd/__tests__/response/CommandResponse.unit.test.ts
@@ -1026,6 +1026,32 @@ describe("Command Response", () => {
         expect(response.buildJsonResponse()).toMatchSnapshot();
     });
 
+    it("should strip ANSI color codes from the JSON response", () => {
+        // Regression test for #2615. Console output is colorized for a terminal, and those buffers
+        // are what writeJsonResponse serializes, so without stripping, the escape sequences survive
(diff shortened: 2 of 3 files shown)


t1m0thyj:
Thanks for your contribution @Atishyy27! After further discussion the team decided that this change should be deferred to the next major release V4. Closing the PR for now.

PR #2803, opened 2026-07-14
refactor(v2): Invalidate config cache in LoggerUtils

traeok:
**What It Does**

Invalidates the `mConfig` cache in `LoggerUtils` to ensure that the latest secure property paths are fetched when in daemon mode.

**How to Test**

The unit tests cover the core behavior:

```
npm test --prefix packages/imperative -- LoggerUtils.unit.test.ts
```

For a manual check in daemon mode:

1. Enable daemon mode (`zowe daemon enable`) and confirm it's running with `zowe daemon status`.
2. In your team config, add a custom property to a profile with a recognizable value, e.g. `customToken: "SECRET123"`. Don't use user/password/tokenValue/keyPassphrase, those already get masked separately as special values and won't exercise this path.
3. Run a command that echoes the value back (`zowe config list --rfj` works) and confirm SECRET123 shows up in plain text, since it isn't secure yet.
4. Leaving the daemon running, mark the property secure: `zowe config set "profiles.lpar1.properties.customToken" "SECRET123" --secure`.
5. Run the same command again. On zowe-v2-lts the value keeps leaking until you restart the daemon; with this change it comes back masked as `****`.

Removing the property from the secure array and rerunning should show the value again, confirming de-securing also takes effect on the next command.

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (job number 2429)

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Recomputed the set of secure property paths to redact on every log call instead of caching it at first use. [#2803](https://github.com/zowe/zowe-cli/pull/2803)
+
 ## `5.27.21`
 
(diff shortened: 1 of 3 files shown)


anaxceron (approved):

zFernand0 (approved):
LGTM! 😋 
---

I remember some discussion about potential performance degradation around censoring secure fields. However we deemed it necessary if we wanted to redact any new secure values after the daemon had started. 🙏 

Thanks for clarifying offline 🙏

jace-roell (approved):
LGTM, Thanks @traeok

PR #2806, opened 2026-07-15
refactor(v2): reject unsafe fields in config

traeok:
**What It Does**

Reject unsafe fields in the config API.

**How to Test**

Run the config suite:

```
npx jest packages/imperative/src/config/__tests__/Config.unit.test.ts packages/imperative/src/config/__tests__/Config.api.unit.test.ts packages/imperative/src/config/__tests__/ConfigUtils.unit.test.ts packages/imperative/src/config/__tests__/Config.secure.unit.test.ts
```

Everything should pass, including the new `prototype pollution` and `getProfilePathFromName` cases.

Want to see it actually catch something? Revert just the four source files (`Config.ts`, `ConfigUtils.ts`, `ConfigProfiles.ts`, `ConfigSecure.ts`), keep the test files, and rerun. 18 tests should fail, and one of them shows `polluted` landing directly on `Object.prototype`.

For a manual repro, build the package and try this in a node REPL from the repo root:

```js
const { Config } = require("./packages/imperative/lib/config/src/Config");
const config = await Config.load("test");
config.set("profiles.__proto__.properties.host", "x"); // used to pollute, now throws
({}).properties // undefined - prototype stays clean
config.set("profiles.lpar1.properties.host", "x"); // normal paths unaffected
```

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] manually tested my changes
- [x] added/updated automated unit/integration tests
- [x] created/ran system tests (job number 2437, 2439)
- [x] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

PR #2807, opened 2026-07-15
refactor: reject unsafe fields in config

traeok:
**What It Does**

Same fix as #2806, ported to v3.

**How to Test**

Same test files as the v2 fix, since the config module layout didn't change:

```
npx jest packages/imperative/src/config/__tests__/Config.unit.test.ts packages/imperative/src/config/__tests__/Config.api.unit.test.ts packages/imperative/src/config/__tests__/ConfigUtils.unit.test.ts packages/imperative/src/config/__tests__/Config.secure.unit.test.ts
```

If you're checking out this branch fresh, run `npm install` first.

To confirm the tests aren't just green by luck, revert the four source files (`Config.ts`, `ConfigUtils.ts`, `ConfigProfiles.ts`, `ConfigSecure.ts`) while keeping the tests, then rerun. 18 cases should fail, same as on the v2 branch.

Manual repro after `npm run build`, from a node REPL in the repo root:

```js
const { Config } = require("./packages/imperative/lib/config/src/Config");
const config = await Config.load("test");
config.set("profiles.__proto__.properties.host", "x"); // throws instead of polluting Object.prototype
```

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (job number 2438)

### DIFF
--- a/packages/imperative/src/config/src/api/ConfigProfiles.ts
+++ b/packages/imperative/src/config/src/api/ConfigProfiles.ts
@@ -30,12 +30,14 @@ export class ConfigProfiles extends ConfigApi {
      * @param profile The JSON profile object to set into the specified location,
      */
     public set(path: string, profile: IConfigProfile): void {
-        // Guard the hand-rolled path walker below against prototype pollution.
-        // A path such as "__proto__" would otherwise reparent the profiles
-        // object or write onto Object.prototype.
-        if (ConfigUtils.hasUnsafeProperty(path)) {
+        // Guard the hand-rolled path walker below against prototype pollution
+        // and malformed paths. A path such as "__proto__" would otherwise
+        // reparent the profiles object or write onto Object.prototype, and an
+        // empty segment (e.g. "lpar1..zosmf") would silently create a bogus
+        // "" profile.
+        if (ConfigUtils.hasUnsafeOrEmptyProperty(path)) {
             throw new ImperativeError({
-                msg: `Invalid profile path '${path}': path segments may not use ` +
+                msg: `Invalid profile path '${path}': path segments may not be empty or use ` +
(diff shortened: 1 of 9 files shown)


CBforZ (approved):
Fix looks good to me. Thank you

anaxceron (approved):

packages/imperative/src/config/src/api/ConfigProfiles.ts
```diff
@@ -107,6 +118,18 @@ export class ConfigProfiles extends ConfigApi {
      * @returns The expanded path.
      */
     public getProfilePathFromName(shortPath: string): string {
+        // Defense-in-depth: reject profile names that resolve to an unsafe
+        // JSON path before they reach any downstream path walker. This covers
+        // every source of a profile name (config `defaults`, the CLI
+        // `--<type>-profile` argument, base-profile fallbacks, and nested
+        // profile keys), not just `defaults`.
+        if (shortPath.split(".").some((segment) => segment.length === 0 || ConfigUtils.UNSAFE_PROP_NAMES.includes(segment))) {
```
zFernand0: This is a good check to prevent UB when having two dots `..` in a `shortPath`.

I'm wondering if we should move this check to the `hasUnsafeProperty` and rename it to something like `validateProfilePathSegments`?

Then we can reuse it here, and have the same benefits (of checking for empty profile names) across the board 😋

zFernand0 (approved):
LGTM! 😋 
---

It's good that we are compering a deterministic array against the segments and not the other way around. That way we ensure we still support existing properties like `prototype_mars` 😋

packages/imperative/src/config/src/api/ConfigProfiles.ts
traeok: Thanks, addressed in 3b2c929 👍

PR #2811, opened 2026-07-17
refactor: adjust permissions for macOS TLS socket

traeok:
**What It Does**

Fixes a possible race condition in the macOS TLS socket by restricting permissions.

Also fixes an unrelated CI issue found while working on this: the `armv7-unknown-linux-gnueabihf` cross-build was failing because cross-rs's build image needs to `apt-get install libsecret-1-dev:armhf`, which pulls in `python3.12-minimal:armhf`. Its postinst script execs an armhf binary that the amd64 runner can't run without a registered QEMU/binfmt handler. Added the same `multiarch/qemu-user-static --reset` step already used elsewhere in this workflow, scoped to that one target.

**How to Test**

*Automated:*
- `cd packages/secrets/core && cargo test tls_pipe::tests -- --nocapture` - 4 tests cover: directory created with owner-only (`0700`) permissions, that this holds even under a forced permissive `umask(0)` (the exact hostile precondition from the original report), unique directories per call, and that legitimate same-user bind/connect still works.
- These tests now also run automatically in CI (`secrets-sdk.yml`, `build` job, `macos-latest` legs) via the new `Run tls_pipe security regression tests` step.
- `cd packages/secrets && npm run build:debug` - confirms the real napi addon still builds on macOS.

*Manual (recommended for reviewers on macOS):*
```bash
# Set up a private dir + socket, exactly like create_private_socket_dir() does
SOCK_DIR=$(mktemp -d); chmod 700 "$SOCK_DIR"
ls -ld "$SOCK_DIR"                      # expect: drwx------

nc -lU "$SOCK_DIR/test.sock" &          # start a listener

# Create a throwaway second local user (if you have permissions to do so)
sudo dscl . -create /Users/_zowepentest
sudo dscl . -create /Users/_zowepentest UserShell /usr/bin/false
sudo dscl . -create /Users/_zowepentest UniqueID 510
sudo dscl . -create /Users/_zowepentest PrimaryGroupID 20
sudo dscl . -create /Users/_zowepentest NFSHomeDirectory /var/empty

# This must fail (run before the sanity check below,
# since a single-shot `nc -lU` consumes its one accept on first connect)
sudo -u _zowepentest nc -U "$SOCK_DIR/test.sock"
# expect: nc: connect to test.sock [unix]: Permission denied

# Sanity check - the owner can still connect fine
echo hi | nc -U "$SOCK_DIR/test.sock"

# Cleanup
sudo dscl . -delete /Users/_zowepentest
rm -rf "$SOCK_DIR"
```

*Pre-existing certificate-based auth check (still applies - unaffected by this change):*
- Use a certificate with a non-exportable key for authenticating one of your profiles
- Specify the required certificate fields in the profile to use the native logic for certificate-based auth.
- Notice that the secure certificate support behaves the same as it did before these changes

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [ ] manually tested my changes (not able to as a Windows user; verified changes through added tests in CI)
- [x] created/ran system tests (job number 2448)


.github/workflows/secrets-sdk.yml
```diff
@@ -134,10 +134,21 @@ jobs:
...
+        # host kernel can't run without a registered binfmt_misc handler.
+        if: matrix.settings.target == 'armv7-unknown-linux-gnueabihf'
+        run: docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
       - name: Build
         run: ${{ matrix.settings.build }}
         if: ${{ !matrix.settings.docker }}
         shell: bash
+      - name: Run tls_pipe security regression tests
+        if: ${{ startsWith(matrix.settings.host, 'macos') }}
+        run: cargo test --manifest-path core/Cargo.toml tls_pipe::tests -- --nocapture
```
t1m0thyj: Thanks for adding automated tests 👍

Would it make sense to generalize this to support running other tests we may add in the future?
* Run all tests in the `packages/secrets/core` directory
* Run tests on all platforms and move OS checks to the Rust code

packages/secrets/core/Cargo.toml
```diff
@@ -36,6 +36,9 @@ core-foundation-sys = "0.8.4"
 fmutex = "0.1.0"
 security-framework = "2.9.2"
 
+[target.'cfg(target_os = "macos")'.dev-dependencies]
+libc = "0.2"
```
t1m0thyj, line 40: Is this dev dep needed? Don't see it used anywhere

anaxceron (approved):

t1m0thyj (approved):
LGTM, thanks @traeok! Left a few optional comments.

Tested by adding a log statement to Imperative `KeychainAgent` that prints socket path:
```
$ zdev files ls ds SYS1.PARMLIB --zosmf-p zosmf_cert
/var/folders/lr/nhndr56j4v38z34sgycg0zm00000gn/T/zowe_secrets_core_d5fb56e2/zowe.sock
SYS1.PARMLIB
...
```

Socket communications continue to work after the changes 😋 Although the length limit for socket paths on MacOS is fairly short (104 chars) we're still ok at 85.

packages/secrets/core/Cargo.toml
traeok: Yeah, it's used for tests in `packages/secrets/core/src/os/mac/tls_pipe.rs` 👍

.github/workflows/secrets-sdk.yml
traeok: Refactored the step in CI to be platform-agnostic:
- Merged a pre-existing test into one lifecycle tests to prevent race conditions from running tests in parallel (default for `cargo test`)
- Replaced the macOS-only name-filtered step with unconditional `cargo test ...` - added `--test-threads=1` since tests share mutable state (credential store entries). 

Note that I had to add a check for Linux in the CI step to ensure the keyring is unlocked. If we want to use the pre-existing unlock-keyring action I can look into that as a separate effort, but this was a short pass for unlocking the Linux keyring in CI.

Addressed in 39927b4 👍

.github/workflows/secrets-sdk.yml
```diff
@@ -134,10 +134,39 @@ jobs:
...
+        # No platform check here: which tests exist/compile per OS is
+        # already decided by `cfg(target_os = ...)` in core/src/os/mod.rs.
+        # The only OS-specific bit left is Linux needing a D-Bus session to
+        # even reach its credential store, which isn't something a Rust
+        # cfg attribute can express, so it stays here rather than moving into the crate.
+        run: |
+          if [ "${{ runner.os }}" = "Linux" ]; then
+            dbus-run-session -- bash -c 'echo -n "test" | gnome-keyring-daemon --unlock; cargo test --manifest-path core/Cargo.toml -- --test-threads=1 --nocapture'
+          else
+            cargo test --manifest-path core/Cargo.toml -- --test-threads=1 --nocapture
```
zFernand0, line 169: would it be possible to run tests in a multi-threaded fashion?
Or should we document why we must run them sequentially?

packages/secrets/core/src/os/mac/tls_pipe.rs
```diff
@@ -58,27 +77,14 @@ pub fn create_tls_pipe(
...
-    
-    // Clean up just in case
-    let _ = fs::remove_file(&socket_path);
-    
+    let dir_path = create_private_socket_dir()
+        .map_err(|e| KeyringError::Os(e.to_string()))?;
+    let mut socket_path = dir_path.clone();
+    socket_path.push("zowe.sock");
+
     let listener = UnixListener::bind(&socket_path)
```
zFernand0, line 85: nit:
do we care if we don't remove the socket directory (inside a `temp_dir`) in the unlikely scenario that the `bind` operation fails?

zFernand0 (approved):
LGTM! 😋

packages/secrets/core/src/os/mac/tls_pipe.rs
traeok: Since its in the temp directory and just a directory entry, I think its safe to leave it - it should get wiped as part of system management 😋

.github/workflows/secrets-sdk.yml
traeok: I have a similar comment above one of the new unit tests, but the reason we are running tests sequentially is to avoid potential race conditions in the credential store/keyring. I've added a comment above this code block in aa0fcd0d to clarify the reason for running tests in sequence 👍

PR #2813, opened 2026-07-17
verify ssh host key - breaking

ATorrise:
**What It Does**

Adds host key verification to SSH connections. The CLI now confirms a server's host key before sending credentials. 
First time you connect it shows the fingerprint and asks you to confirm. Say yes the key persists in your ssh profile under the new `hostKey` property. If the key ever changes, the connection stops and warns you. You can skip key verification with `--insecure`.

| CLI Option | Property |  Description |
|---|---|---|
| `--host-key`, `--hk` | `hostKey` | Trusted host key (base64 key blob the server presents). When set, the server's key is verified against it before credentials are sent. |
| `--insecure` |  | Skip host key verification (default `false`). |

**Details**

- **Trust on first use.** First connect gives you the fingerprint and a yes or no prompt. Yes saves the key to `hostKey` on your ssh profile, and later connects match on their own with no prompt.
- **Changed key.** A saved key that no longer matches is refused before credentials go out. It shows the previously trusted and newly presented fingerprints and does not offer to overwrite. To trust the new key, remove `hostKey` from the ssh profile or pass a new `--host-key`, then reconnect.
- **CI and automated runs.** With no terminal to answer, the CLI doesn't hang. It keeps the check on and fails fast, telling you to set the key with `--host-key` or use `--insecure`.
- **Same key type.** With a key saved, the CLI asks the server for that same key type, so a normal type swap isn't mistaken for a changed key.
- **Rotated key type.** If the server changes its host key type, the connect fails before verification and the error names the pinned key type instead of a generic error.

**Host key detail**

The `hostKey` is the base64 blob the server presents, shaped `[length][algorithm name][key material]`. The code only reads the algorithm name to ask for the same key type, and compares the full blob exactly to decide if you connect. The fingerprint is a separate SHA256 hash of the key, shown at the prompt just so you can eyeball it. Verification uses the full key, not the fingerprint.

**What "Pin" Means**

Pin is the TLS cert term for storing one trusted key and locking future connections to it. You set it with `--host-key` or the `hostKey` property, or it's saved when you accept the first connect prompt.

**Breaking**

CLI: Verification is on by default. Use `--insecure` (or `ZOWE_OPT_INSECURE=true`, or the `insecure` profile property) to restore prior behavior.

SDK: `Shell.isConnectionValid` now rejects with a host key error when the key is untrusted or changed, rather than returning `false`, so callers can tell it apart from bad credentials.

**How to Test**
You need an ssh host you can reach and a team config with an ssh profile so the key has somewhere to save. To reset runs, remove your pinned `hostKey` from your config. 

```
# ---- setup ----
alias zoweDev="node ~/<ur path>/zowe-cli/packages/cli/lib/main.js"
zoweDev zos-uss issue ssh --help | grep -i host-key      # confirms you're on the branch build

# quick way to check the saved key at any point (or check zowe.config.json directly)
grep hostKey ~/.zowe/zowe.config.json                    # prints the hostKey line, or nothing


# ---- 1. first connect, saves the key ----
zoweDev zos-uss issue ssh "uname -a"                     # fingerprint prompt, answer yes
grep hostKey ~/.zowe/zowe.config.json                    # hostKey now under your default ssh profile

# ---- 2. reconnect, no prompt ----
zoweDev zos-uss issue ssh "uname -a"                     # runs silently, key matched

# ---- 3. insecure, skip the check ----
zoweDev zos-uss issue ssh "uname -a" --insecure          # warns it's off, connects, no prompt
CI=true zoweDev zos-uss issue ssh "uname -a" --insecure     # warns, connects, no prompt


# ---- 4. no terminal, fails fast instead of hanging ----
# first clear the saved key (see reset below)
CI=true zoweDev zos-uss issue ssh "uname -a"             # no prompt, fails fast with guidance
zoweDev zos-uss issue ssh "uname -a" < /dev/null         # same, fails but does not hang

# ---- 5. changed key ----
# edit ~/.zowe/zowe.config.json and set the ssh profile hostKey to a bogus value like "AAAAbogus"
zoweDev zos-uss issue ssh "uname -a"                     # refuses, no yes/no prompt, prints both fingerprints
# to adopt the new key you clear hostKey or pass --host-key

# ---- 6. decline ----
# clear the saved key first
zoweDev zos-uss issue ssh "uname -a"                     # answer no, rejected, nothing saved
grep hostKey ~/.zowe/zowe.config.json                    # should print nothing


# ---- reset between runs ----
# open ~/.zowe/zowe.config.json and delete the hostKey line under your ssh profile
```

**Review Checklist**

I certify that I have

- [x] updated the changelog


packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -162,6 +167,106 @@ export abstract class SshBaseHandler implements ICommandHandler {
...
+
+        // Write to the layer that actually contains the profile, then restore the previously active layer.
+        const beforeLayer = config.api.layers.get();
+        const foundLayer = config.api.layers.find(profileName);
+        if (foundLayer != null) {
+            config.api.layers.activate(foundLayer.user, foundLayer.global);
+        }
+        config.set(`${profilePath}.properties.hostKey`, hostKey, { secure: false });
+        await config.save();
+        config.api.layers.activate(beforeLayer.user, beforeLayer.global);
```
traeok: If `config.set` throws on line 263 *or* `config.save` throws on line 264, then the `layers.activate` function is never called. This is fine when running Zowe CLI commands w/o the daemon, but this could corrupt the active layers when daemon mode is enabled.

I'd recommend wrapping the activate/set/save logic in a `try { ... } finally { config.api.layers.activate(beforeLayer.user, beforeLayer.global); }` to be safe 👍

packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -162,6 +167,106 @@ export abstract class SshBaseHandler implements ICommandHandler {
...
+     * Attaches an interactive host key verifier to the session so that, when the SSH server presents a
+     * host key that is not already trusted, the user is asked to confirm it before any credentials are sent
+     * (trust on first use). The accepted key is then persisted to the ssh profile so the user is not
+     * prompted again. In non-interactive / CI environments the key is never prompted for: verification
+     * stays on and the connection fails with an explanatory error instead of hanging.
+     *
+     * @param {SshSession} session - the session to attach the verifier to
+     * @param {IHandlerParameters} commandParameters - command parameters, used for prompting and persistence
+     */
+    private attachHostKeyVerifier(session: SshSession, commandParameters: IHandlerParameters): void {
```
traeok, line 180: When `info.changed` is true, the code warns but still lets the user answer "yes" and overwrites the pinned key. 

By contrast, OpenSSH refuses outright and requires manual key removal. An attacker performing MITM on a reconnect can socially‑engineer a "yes." This may be an intentional UX choice, but it is a real reduction in the guarantee; at minimum consider requiring a stronger confirmation (retyping the fingerprint, or a distinct flag) to overwrite an existing pinned key. 

(The non‑TTY/CI path correctly hard‑fails, which is good 😋)

packages/zosuss/src/Shell.ts
traeok, line 1: Should we revisit the `isConnectionValid` function since it is public and exposed to developers? 

With the new logic in the `Shell.connect` function, it always installs the new `hostVerifier`. This means any unpinned host with valid credentials would now return `false` when this function is used. Should we have the `isConnectionValid` function surface the "host key mismatch" error, or do you think that this is an acceptable regression in expected function behavior?

packages/zosuss/src/Shell.ts
```diff
@@ -252,11 +275,147 @@ export class Shell {
...
+            const algorithm = this.getHostKeyAlgorithm(Buffer.from(pinnedKey, "base64"));
+            if (algorithm == null) {
+                return undefined;
+            }
+            // An RSA host key is stored as "ssh-rsa" but modern servers offer it under the SHA-2 names.
+            // Request all three so the RSA key is still offered without the retired SHA-1 negotiation.
+            if (algorithm === "ssh-rsa") {
+                return ["rsa-sha2-512", "rsa-sha2-256", "ssh-rsa"];
+            }
+            return [algorithm];
```
traeok: While this isn't something that would immediately happen after this new code is released, consider what may happen if a server eventually rotates to different key types.

- ssh2 would fail host-key algorithm negotiation before the host verifier runs
- the user would get a generic `unexpected` error instead of a signal that the host key has changed

I suspect that rotating key types may become a more common practice, especially as existing algorithms are weakened through adversarial efforts & development of PQC... but also as new algorithms are introduced to z/OS OpenSSH. Could we detect the change in host keys by scanning the error text and mapping it to a "host key changed" error?

packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -162,6 +167,106 @@ export abstract class SshBaseHandler implements ICommandHandler {
...
+     * @param {string} hostKey - the base64-encoded host key to store
+     */
+    private async persistHostKey(commandParameters: IHandlerParameters, hostKey: string): Promise<void> {
+        const config = ImperativeConfig.instance.config;
+        if (config == null || !config.exists || !config.properties.autoStore) {
+            this.console.error("Host key not saved; you will be prompted again. Pin it with --host-key.\n");
+            return;
+        }
+        const profileName = ConfigUtils.getActiveProfileName("ssh", commandParameters.arguments);
+        // TODO: consider creating an ssh profile here (shared config layer) so the accepted key has somewhere to persist,
```
ATorrise, line 253: @traeok @t1m0thyj @MarkAckert @awharn does it make sense to create an enhancement request for this todo?

traeok (commented):
Thanks Amber, left a few comments to consider as part of implementation

zFernand0:
Hey @ATorrise,
Please make sure to sign-off all commits (even those coming from AI)
- image
- image

packages/zosuss/src/SshBaseHandler.ts
traeok: Yeah I would consider this out of scope, separate enhancement makes sense 👍

packages/cli/__tests__/zosfiles/__unit__/edit/Edit.utils.unit.test.ts
```diff
@@ -76,7 +76,7 @@ describe("Files Edit Utilities", () => {
 
     beforeEach(async () => {
         jest.resetAllMocks();
-        jest.spyOn(ZosFilesUtils, "ensureSafeTempDir").mockImplementation();
+        jest.spyOn(ZosFilesUtils as any, "ensureSafeTempDir").mockImplementation();
```
zFernand0: I don't think this change is related to the SSH host key functionality 😕 

In either case, the function should be public, so no need for the `as any` 😅

packages/zosuss/src/SshSession.ts
```diff
@@ -122,6 +122,30 @@ export class SshSession {
...
+            "by the server. When set, the server's key is verified against it before any credentials are sent.",
+        type: "string",
+        group: SshSession.SSH_CONNECTION_OPTION_GROUP
+    };
+
+    /**
+     * Option used in profile creation and commands to skip host key verification.
+     */
+    public static SSH_OPTION_INSECURE: ICommandOptionDefinition = {
+        name: "insecure",
```
zFernand0, line 141: nit: should we have an alias for this?

packages/zosuss/src/doc/ISshSession.ts
```diff
@@ -66,4 +66,23 @@ export interface ISshSession {
...
+     * @memberof ISession
+     */
+    hostKey?: string;
+
+    /**
+     * Skip verification of the SSH server's host key. Defaults to false, meaning the server's identity
+     * is verified before any credentials are sent. When true, host key verification is skipped entirely
+     * and credentials are sent without confirming the server's identity.
+     * @type {boolean}
+     * @memberof ISession
```
zFernand0: nit: this should probably say `@memberof ISshSession`

We should probably fix all others in this interface as well 🙏

packages/zosuss/src/SshBaseHandler.ts
```diff
@@ -162,6 +167,106 @@ export abstract class SshBaseHandler implements ICommandHandler {
...
+        // instead of only warning
+        if (profileName == null || !config.api.profiles.exists(profileName)) {
```
zFernand0: nit:
the check `profileName == null ||` can be safely removed since the `getActiveProfileName` function ultimately falls back to the string `"ssh"`

packages/zosuss/src/Shell.ts
```diff
@@ -252,11 +275,147 @@ export class Shell {
...
+     * @returns the host key algorithm names to request, or undefined to use the default negotiation
+     */
+    private static getPinnedHostKeyAlgorithms(session: SshSession): string[] | undefined {
+        const pinnedKey = session.ISshSession.hostKey;
+        if (session.ISshSession.insecure === true || pinnedKey == null ||
+            pinnedKey === "" || pinnedKey === "undefined") {
+            return undefined;
+        }
+        try {
+            const algorithm = this.getHostKeyAlgorithm(Buffer.from(pinnedKey, "base64"));
```
zFernand0: nit: I don't think we need this try/catch here since the `getHostKeyAlgorithm` function won't throw an error 🙏 

And if it does, we should probably add it to that function's TSDoc 😋

packages/zosuss/src/Shell.ts
```diff
@@ -252,11 +275,147 @@ export class Shell {
...
+        };
+    }
+
+    /**
+     * Compute the OpenSSH-style SHA256 fingerprint of a host key for display to the user,
+     * e.g. "SHA256:Xn5vB...". Matches the format shown by the standard `ssh` client.
+     * @param keyBuf - the raw host key blob presented by the server
+     * @returns the fingerprint string
+     */
+    public static getHostKeyFingerprint(keyBuf: Buffer): string {
```
zFernand0: nit: Could this function be private?

packages/zosuss/src/Shell.ts
```diff
@@ -252,11 +275,147 @@ export class Shell {
...
+        return `SHA256:${digest}`;
+    }
+
+    /**
+     * Extract the algorithm name from an SSH public key blob. The blob is self-describing: it begins with
+     * a 4-byte big-endian length followed by the algorithm name, e.g. "ssh-ed25519" or "ssh-rsa".
+     * @param keyBuf - the raw host key blob
+     * @returns the algorithm name, or undefined if the blob cannot be parsed
+     */
+    public static getHostKeyAlgorithm(keyBuf: Buffer): string | undefined {
```
zFernand0: Same as previous comment:
nit: Could this function be private?

zFernand0 (commented):
LGTM! 😋 
---
Left some comments for consideration, but nothing that should prevent the PR from getting merged 🙏 
My only real request is to add the try/finally around the `config.save` 🙏

packages/zosuss/src/Shell.ts
ATorrise: good call, made both private. to get getHostKeyFingerprint private I moved the pinned fingerprint computation into Shell, so the handler no longer calls it. Shell now passes pinnedFingerprint through the hostKeyVerifier hook and the handler just reads it

zFernand0:
Hey @ATorrise 
Please update snapshots 🙏 
Also, you might want to combine all DCO items in a single empty remediation commit.
If that doesn't work, I'm happy to do the third party version of it 🙏

ATorrise:
note: waiting on system tests to pass. will add build number once they go through

zFernand0:
> note: waiting on system tests to pass. will add build number once they go through

We had a bit of an issue in internal pipelines that prevented the system tests from running.
- `2486` is the running build.

zFernand0 (dismissed):
LGTM! 😋 
---
Let's wait for system tests to run 🙏 
Also, thanks a lot for adressing all the feedback so quickly 🙏

ATorrise:
new system test: ~2496~

2498

MarkAckert:
Changes LGTM during local testing! And system tests passed ✅

MarkAckert (approved):

packages/zosuss/src/Shell.ts
```diff
@@ -243,6 +285,7 @@ export class Shell {
         if (session.ISshSession.password != null && session.ISshSession.password !== "undefined") {
```
zFernand0, line 285: In multiple places we use the literal string `"undefined"`. And this is something that was not introduced in this PR. 🤔 
Dumb Question:
What's producing that literal string (`"undefined"`) to percolate all the way down to these functions? 🤔

packages/zosuss/src/Shell.ts
zFernand0: I could be wrong, but I think this comment is resolved based on the changes in 6267aa2de3a617891bb3129734bfb9915417d4a6.
The `isConnectionValid` should be rejecting instead of silently returning false.

zFernand0 (approved):
LGTM! 😋

packages/cli/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to the Zowe CLI package will be documented in this file.
 
 ## `8.35.0`
 
+- **Breaking**: `zowe zos-uss` (ssh) commands now verify the z/OS SSH server's host key before sending credentials. On first connect you confirm the fingerprint and the key is saved to your ssh profile; a changed key is rejected. Use `--insecure` to skip verification or `--host-key` to pin a key. [#2813](https://github.com/zowe/zowe-cli/pull/2813)
```
zFernand0:
```suggestion
## Recent Changes

- **Breaking**: `zowe zos-uss` (ssh) commands now verify the z/OS SSH server's host key before sending credentials. On first connect you confirm the fingerprint and the key is saved to your ssh profile; a changed key is rejected. Use `--insecure` to skip verification or `--host-key` to pin a key. [#2813](https://github.com/zowe/zowe-cli/pull/2813)

## `8.35.0`
```

PR #2822, opened 2026-07-21
refactor(v2): censor raw data when logging at trace level

traeok:
**What It Does**

The `Logger.trace` method was the only log level that did not route its final formatted message through `LoggerUtils.censorRawData` before handing it to the appender. Every other level (`debug`, `info`, `warn`, `error`, `fatal`, `simple`) already does this.

This change brings `trace` in line with the rest, so that secure configuration property values are consistently masked at every log level rather than depending on which level a message happens to be emitted at. Both the direct-to-appender path and the in-memory queue path are covered.

**How to Test**

Automated:
- `npm run test:unit -- packages/imperative/src/logger/__tests__/Logger.unit.test.ts`
- New cases under the `Secure data censoring` block confirm `trace` masks a secure config value on both emission paths, plus a regression check that every log level routes through `LoggerUtils.censorRawData`.

Manual:
1. In a team config, mark a property as secure and give it a value.
2. Lower the log level to `TRACE` (e.g. `ZOWE_IMPERATIVE_LOG_LEVEL=TRACE` or `ZOWE_APP_LOG_LEVEL=TRACE`).
3. Run any command that produces trace output.
4. Inspect the resulting trace log file and confirm the secure value appears masked (`****`) rather than in clear text.

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (job number 2461)

**Additional Comments**

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -2,9 +2,12 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
-## `5.27.22`
+## Recent Changes
 
 - BugFix: Routed messages passed to the `Logger.trace` function through the `LoggerUtils.censorRawData` function, so secure config property values are redacted before written to the trace log. [#2822](https://github.com/zowe/zowe-cli/pull/2822)
+
+## `5.27.22`
+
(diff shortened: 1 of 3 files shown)


t1m0thyj (approved):
LGTM, thanks @traeok!

anaxceron (approved):

zFernand0 (approved):

PR #2821, opened 2026-07-21
Ensure safe tmp directory for copy and edit

jace-roell:
**What It Does**
- Ensure safe temporary directory when executing a PDS copy or zosfiles edit

**How to Test**
Run `npx jest packages/zosfiles/__tests__/__unit__/methods/copy/Copy.unit.test.ts`

**Review Checklist**
I certify that I have:
- [ ] updated the changelog
- [x] manually tested my changes
- [x] added/updated automated unit/integration tests
- [x] created/ran system tests (provide build number if applicable)
- [x] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

PR #2798, opened 2026-07-13
Remove node: from import and downgrade sanitize-html

t1m0thyj:
**What It Does**

Fixes the V2 branch to be compatible with Zowe Explorer V2:
* Removes `node:` prefix from import which isn't supported by old versions of TypeScript and Webpack
* Downgrades `sanitize-html` dependency to an older and non-vulnerable version compatible with Node 20

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] updated the changelog
- [ ] manually tested my changes
- [ ] added/updated automated unit/integration tests
- [ ] created/ran system tests (provide build number if applicable)
- [ ] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

**Additional Comments**

PR #2823, opened 2026-07-22
Use random file names for etag-refresh step for `zowe files edit`

zFernand0:
**What It Does**

- Follow-up to #2821. 
  - That PR moved the `zowe zos-files edit` working files into a safe temp location. 
  - This PR extends the same behavior to the temporary file used to refresh the etag during an edit.
- Added a new cross-platform utility to verify if a given file or directory has owner-only access.
- Added a per-user temp directory to avoid DoS for co-tenants

**How to Test**

Run the unit tests: `npx jest packages/cli/__tests__/zosfiles/__unit__/`

**Review Checklist**
I certify that I have:
- [x] updated the changelog
  - `2467`

**Additional Comments**

### DIFF
--- a/packages/cli/src/zosfiles/edit/Edit.utils.ts
+++ b/packages/cli/src/zosfiles/edit/Edit.utils.ts
@@ -162,8 +162,18 @@ export class EditUtilities {
      * @returns {ILocalFile}
      */
     public static async localDownload(session: AbstractSession, lfFile: ILocalFile, useStash: boolean): Promise<ILocalFile>{
-        // account for both useStash|!useStash and uss|ds when downloading
-        const tempPath = useStash ? path.posix.join(tmpdir(), "toDelete.txt") : lfFile.tempPath;
+        // account for both useStash|!useStash and uss|ds when downloading.
+        // When only refreshing the etag (useStash), download to a throwaway scratch file with a
+        // unique, unpredictable name inside the safe temp dir rather than a shared, predictable path.
+        let scratchPath!: string;
+        if (useStash) {
+            const crypto = require("crypto");
+            const randomNameBytes = 16;
+            const scratchDir = path.join(tmpdir(), `zowe-edit-${lfFile.fileType}`);
+            ZosFilesUtils.ensureSafeTempDir(scratchDir);
+            scratchPath = path.join(scratchDir, `.etag-refresh-${crypto.randomBytes(randomNameBytes).toString("hex")}`);
+        }
+        const tempPath = useStash ? scratchPath : lfFile.tempPath;
         const args: [AbstractSession, string, IDownloadOptions] = [
             session,
             lfFile.fileName,
@@ -183,7 +193,7 @@ export class EditUtilities {
         }
 
         if (useStash){
-            await this.destroyTempFile(path.posix.join(tmpdir(), "toDelete.txt"));
+            await this.destroyTempFile(scratchPath);
         }
         return lfFile;
     }
--- a/packages/cli/src/zosfiles/edit/Edit.utils.ts
+++ b/packages/cli/src/zosfiles/edit/Edit.utils.ts
@@ -13,7 +13,7 @@ import { Download, Upload, IZosFilesResponse, IDownloadOptions, IUploadOptions, 
 import { AbstractSession, IHandlerParameters, ImperativeError, ProcessUtils, GuiResult,
     TextUtils, IDiffNameOptions, CliUtils } from "@zowe/imperative";
 import { CompareBaseHelper } from "../compare/CompareBaseHelper";
-import { chmodSync, existsSync, unlinkSync } from "fs";
+import { existsSync, unlinkSync } from "fs";
 import { tmpdir } from "os";
 import * as path from "path";
 import LocalfileDatasetHandler from "../compare/lf-ds/LocalfileDataset.handler";
--- a/packages/cli/src/zosfiles/edit/Edit.utils.ts
+++ b/packages/cli/src/zosfiles/edit/Edit.utils.ts
@@ -206,17 +206,6 @@ export class EditUtilities {
             lfFile.zosResp = await Download.dataSet(...args);
         }
 
-        if (!useStash && tempPath && process.platform !== "win32") {
-            // The stash now holds remote contents; restrict it to the owner (defense-in-depth atop
-            // the 0700 dir). Best-effort: the enclosing directory already blocks other-user access.
-            const ownerReadWrite = 0o600;
-            try {
-                chmodSync(tempPath, ownerReadWrite);
-            } catch (err) {
-                // ignore - the 0700 temp directory is the primary protection
-            }
-        }
-
         if (useStash){
             await this.destroyTempFile(scratchPath);
         }
--- a/packages/cli/src/zosfiles/edit/Edit.utils.ts
+++ b/packages/cli/src/zosfiles/edit/Edit.utils.ts
@@ -13,7 +13,7 @@ import { Download, Upload, IZosFilesResponse, IDownloadOptions, IUploadOptions, 
 import { AbstractSession, IHandlerParameters, ImperativeError, ProcessUtils, GuiResult,
     TextUtils, IDiffNameOptions, CliUtils } from "@zowe/imperative";
 import { CompareBaseHelper } from "../compare/CompareBaseHelper";
-import { existsSync, unlinkSync } from "fs";
+import { chmodSync, existsSync, unlinkSync } from "fs";
 import { tmpdir } from "os";
 import * as path from "path";
 import LocalfileDatasetHandler from "../compare/lf-ds/LocalfileDataset.handler";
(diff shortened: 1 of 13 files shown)


packages/cli/src/zosfiles/edit/Edit.utils.ts
```diff
@@ -162,8 +177,17 @@ export class EditUtilities {
...
-        const tempPath = useStash ? path.posix.join(tmpdir(), "toDelete.txt") : lfFile.tempPath;
+        // account for both useStash|!useStash and uss|ds when downloading.
+        // When only refreshing the etag (useStash), download to a throwaway scratch file with a
+        // unique, unpredictable name inside the safe temp dir rather than a shared, predictable path.
+        let scratchPath!: string;
+        if (useStash) {
+            const crypto = require("crypto");
+            const randomNameBytes = 16;
+            const scratchDir = this.ensureEditTempDir(lfFile.fileType);
+            scratchPath = path.join(scratchDir, `.etag-refresh-${crypto.randomBytes(randomNameBytes).toString("hex")}`);
```
traeok, line 188: (nitpick) `randomNameBytes` is only used once after its declared, is this to avoid magic numbers? 😋

packages/imperative/src/io/src/IO.ts
```diff
@@ -375,6 +375,47 @@ export class IO {
...
+                return false;
+            }
+            if (os.platform() === IO.OS_WIN32) {
+                // On windows, ask icacls for the current ACL and confirm it grants
+                // access to nobody but the current user (mirrors what giveAccessOnlyToOwner sets).
+                const stdout = ExecUtils.spawnAndGetOutput("icacls", [ fileName ], { encoding: "utf8" }).toString();
+                const aces: string[] = stdout.split(/\r?\n/)
+                    .map((line: string) => line.startsWith(fileName) ? line.slice(fileName.length) : line)
+                    .map((line: string) => line.trim())
+                    .filter((line: string) => line.length > 0 && !line.startsWith("Successfully processed"));
```
traeok: This parser filters out lines starting with the English phrase "Successfully processed" - but in a non-English Windows environment, that line is not present. Wouldn't this cause the function to return `false` in those environments? 

Can we key off the exit status and/or some other structure returned by `icacls` instead?

packages/zosfiles/src/utils/ZosFilesUtils.ts
```diff
@@ -61,12 +63,37 @@ export class ZosFilesUtils {
...
+        try {
+            id = userInfo().username;
+        } catch (err) {
+            // userInfo() can throw when the current user has no OS account entry; fall back to uid or a constant
+            if (typeof process.getuid === "function") {
+                id = String(process.getuid());
+            }
+        }
+        const tokenLen = 10;
+        return crypto.createHash("sha256").update(id).digest("hex").slice(0, tokenLen);
```
traeok, line 86: I noticed that the token is derived from the username (defined as the ID). Do we have any concerns that this token is fully derivable by co-tenants that know the username?

I think failing when an unsafe temp directory is detected is correct, but was just a bit confused since the PR mentions avoiding DoS - when the DoS still remains if a co-tenant was to create the directory ahead of time.

traeok (commented):
Thanks for working on this Fernando! I have a request for changes around detecting `icacls` success, and a couple suggestions around the token/random-byte generation.

traeok (dismissed):
Accidentally hit comment instead of request changes - please refer to comments above 😋

packages/cli/src/zosfiles/edit/Edit.utils.ts
zFernand0: yeah, should we just eslint-disbale the rule? 😅

packages/imperative/src/io/src/IO.ts
ATorrise: Went and read up on how icacls prints Windows permissions before commenting. Quick version for anyone else, folders have a permissions list and every line in it is an Access Control Entry (ACE), basically one user or group and what permissions they have

Agree with the concern. Right now the parser reads "Successfully processed" and expects exactly one ACE, so on a non English Windows box that text is different and it ends up rejecting a dir we actually made.

I dont think keying off exit status fixes it though. Exit status only says whether icacls ran, not who has access. This call just prints the current permissions, so it returns 0 as long as it could read the list, doesnt matter if the dir is ours or someone elses. It would work for giveAccessOnlyToOwner since that one actually changes permissions and can pass or fail, but not for a read like this.

Claude suggested keeping only the lines that look like an ACE instead of matching the English text. An ACE line is an identity then a permission set in parens, like DESKTOP-XYZ\amber:(OI)(CI)(F). The summary line doesnt have that shape in any language so it just falls out.

    const stdout = ExecUtils.spawnAndGetOutput("icacls", [fileName], { encoding: "utf8" }).toString();
    const aces = stdout.split(/\r?\n/)
        .map((line) => line.startsWith(fileName) ? line.slice(fileName.length) : line)
        .map((line) => line.trim())
        // keep only ACE lines (identity followed by a parenthesized permission set),
        // which drops the localized summary line without matching its wording
        .filter((line) => line.includes(":(") && line.endsWith(")"));

Everything after that, the one ACE check and the identity match, stays the same. giveAccessOnlyToOwner keys off the same English string so it could get the same fix.

ATorrise (dismissed):
Hey so I left a comment on one of trae's threads with the details, but the short version is the icacls parsing keys off the English "Successfully processed" text and assumes exactly one ACE, so it breaks on a non English Windows box. I dropped a suggested fix in that thread that picks the ACE lines out by shape instead of by the English text.

Requesting changes just for that one part. Rest of it looks good to me. I also pulled the branch and ran the changed unit tests on my machine (Windows) and they all pass.

packages/imperative/src/io/src/IO.ts
traeok: I still think we should check the exit code in addition to this, because the exit code indicates whether an execution failure has occurred. But aside from that I agree with your suggestion

zFernand0: resolved in 643d2f0a4

--- 

As far as exit code goes, the `icacls` call can return exit-code of 0 in some failure scenarios, however the summary line format stays consistent across languages. 

Also, the same English specific error checking existed for quite some time in the `giveAccessOnlyToOwner`
https://github.com/zowe/zowe-cli/blob/35e34b1d12797e09b0d732c09fc352dbf09673b8/packages/imperative/src/io/src/IO.ts#L360

I'm at least attempting to handle languages that use semicolons 😅  

Those messages are often formatted as follows:
The success number is before the `;` and the failure number is after it.
`[<text>] <number> [<text>] ; [<text>] <number> [<text>]`

---

Forgot to mention that if the command itself fails, the spawnAndGetOutput function will throw an error altogether 😅

packages/zosfiles/src/utils/ZosFilesUtils.ts
zFernand0: That's true.
I was mostly trying to avoid the scenario where two regular users on the same machine run a `zowe files edit` and whoever ran it first wins since they get to create the shared temp directory. 😅

zFernand0:
From SonarCloud:

- 3 instances of prefering `node:crypto` (or `node:os`) over just `crypto` (or `os`)
  - I believe there was a PR that we had to revert those suggestions, i.e. remove the `node:` portion for backwards compatibility with older TS versions.
  - It might be ok to do for V3, but probably not for V2
  - Source: https://github.com/zowe/zowe-cli/pull/2798
 - image
   - On the fence. somtimes parameterizing make tests a bit less readable 😅 
- Regex shouldn't cause linear backtracking
  - The summary output of `icacls` is fairly small and standard. 
  - One option could be to reduce the number of digits to one (replace `\d+` with `\d`) 
    - That's not going to help much. And we might want to support recursively changing perms in the future
  - The easy/lazy thing to do is just go back to the english-only check (which I'm not opposed to) 😋  
- Handle error possibly thrown by `spawnAndGetOuput` in `catch` block
  - I kind of wanted to keep the function secure by default, and if there is anything funny preventing us from determining your access level, we assume you don't have it 😅

packages/imperative/src/io/__tests__/IO.unit.test.ts
```diff
@@ -626,35 +650,43 @@ describe("IO tests", () => {
...
-                    "          fernando:(F)\r\n" +
+                    "          rubyTheDog:(F)\r\n" +
                     "Successfully processed 1 files; Failed processing 0 files" as any
                 );
                 expect(IO.hasOwnerOnlyAccess("C:\\Temp\\dir")).toBe(false);
             });
             it("should return false when the single access entry is a different user", () => {
                 jest.spyOn(ExecUtils, "spawnAndGetOutput").mockReturnValue(
-                    "C:\\Temp\\dir someoneelse:(F)\r\nSuccessfully processed 1 files; Failed processing 0 files" as any
+                    "C:\\Temp\\dir monaTheCat:(F)\r\nSuccessfully processed 1 files; Failed processing 0 files" as any
```
zFernand0, line 678: @anaxceron 👀

packages/imperative/src/io/__tests__/IO.unit.test.ts
```diff
@@ -626,35 +650,43 @@ describe("IO tests", () => {
...
+                    "C:\\Temp\\dir MYDOMAIN\\rubyTheDog:(F)\r\nSuccessfully processed 1 files; Failed processing 0 files" as any
                 );
                 expect(IO.hasOwnerOnlyAccess("C:\\Temp\\dir")).toBe(true);
             });
             it("should return false when more than one access entry is present", () => {
                 jest.spyOn(ExecUtils, "spawnAndGetOutput").mockReturnValue(
                     "C:\\Temp\\dir NT AUTHORITY\\SYSTEM:(F)\r\n" +
                     "          BUILTIN\\Administrators:(F)\r\n" +
-                    "          fernando:(F)\r\n" +
+                    "          rubyTheDog:(F)\r\n" +
```
zFernand0, line 671: @t1m0thyj 👀

packages/imperative/src/io/__tests__/IO.unit.test.ts
anaxceron: 😺

traeok (approved):
Changes LGTM, thanks Fernando for the fix & for addressing feedback!

jace-roell (approved):
LGTM! Thanks for looking into this. I definitely prefer using the helper functions rather than omitting the checks for win32. 😋

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added `IO.hasOwnerOnlyAccess` to check whether a file or directory's access is restricted to the current user only, cross-platform (POSIX mode bits/owner; Windows `icacls` ACL). [#2823](https://github.com/zowe/zowe-cli/pull/2823)
```
anaxceron: We should describe what `IO.hasOwnerOnlyAccess` is

zFernand0: resolved in d51105b75

anaxceron (approved):

PR #2827, opened 2026-07-23
expose config export redacting functionality

jace-roell:
**What It Does**
Exposes the config export functionality so it can be used by SDK consumers

**Review Checklist**
I certify that I have:
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/imperative/CHANGELOG.md
+++ b/packages/imperative/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Imperative package will be documented in this file.
 
 ## Recent Changes
 
-- Enhancement: Added a `ConfigRedact` API, accessible via `Config.api.redact`, with methods for programmatically retrieving Zowe configuration layers with sensitive values redacted. [#2827](https://github.com/zowe/zowe-cli/pull/2827)
+- Enhancement: Added a `ConfigRedact` API, accessible via the `Config.api.redact` class instance, with methods for programmatically retrieving Zowe configuration layers with sensitive values redacted. [#2827](https://github.com/zowe/zowe-cli/pull/2827)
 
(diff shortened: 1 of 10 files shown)


zFernand0 (approved):
LGTM! 😋

packages/imperative/src/config/src/api/ConfigRedact.ts
traeok, line 1: Now that this is exposed, I'm wondering if its expected that the code silently reflects on-disk state rather than the using the config API?

This may be where dependency injection comes into play - providing an instance of the API allows us to work with pre-existing config modifications that haven't been persisted to disk.

Or at least, we should document that this reflects on-disk state of the config.

traeok (commented):
I know this is marked as `no-changelog`, but wondering if we want to mention the exporting as an enhancement to the APIs?

Also, had a question about exporting configs (expected vs. actual state) and plan to approve based on the result of that conversation 😋

traeok (approved):
LGTM, thanks Jace!

packages/imperative/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Imperative package will be documented in this file.
 
+## Recent Changes
+
+- Enhancement: Added a `ConfigRedact` API, accessible via `Config.api.redact`, with methods for programmatically retrieving Zowe configuration layers with sensitive values redacted. [#2827](https://github.com/zowe/zowe-cli/pull/2827)
```
anaxceron: We should explain what `Config.api.redact` is

anaxceron (approved):
🦁

CBforZ (approved):
Nice thank you!

PR #2838, opened 2026-08-04
chore(cli): Migrate from shrinkwrap to bundleDependencies

t1m0thyj:
**What It Does**

* Adds `prepack` and `postpack` scripts to CLI package that bundle dependencies instead of using shrinkwrap
* Posts comment on PRs with link to CLI package artifact that can be downloaded for testing

**How to Test**

Verify that CLI is packaged successfully by running `cd packages/cli && npm pack`.
Can also test installing from the TGZ that was created: `npm install -g <pathToTgz>`.

**Review Checklist**
I certify that I have:
- [ ] updated the changelog
- [x] manually tested my changes
- [ ] added/updated automated unit/integration tests
- [ ] created/ran system tests (provide build number if applicable)
- [ ] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

**Additional Comments**

PR #2839, opened 2026-08-06
Add the ability to pass `--allow-scripts` during `plugins install`

zFernand0:
**What It Does**

Added an `--allow-scripts` option to the `zowe plugins install` and `zowe plugins update` commands. 
It follows the same format as `npm` commands (e.g. `preinstall`, `install`, `postinstall`, and `prepare`). 

**How to Test**

1. Create two dummy directories (e.g. `home` and `project`) and `cd project`
2. Run `ZOWE_CLI_HOME=../home <zowe-cli-built-from-source> plugins install @zowe/db2-for-zowe-cli --allow-scripts="ibm_db"`
3. `cd ../home/plugins/installed/lib/node_modules/@zowe/db2-for-zowe-cli` and run `node` inside it
4. `require("ibm_db")`

<details><summary> It shoudl look something like this: (click here) ⬇️ </summary>

image

</details>

_Please repeat the testing steps (above) with all combinations of node [22, 24, 26] and npm [10, 11, 12]_

**NOTE:**
The `ibm_db` version that we bundle with the `@zowe/db2-for-zowe-cli@zowe-v3-lts` plug-in does not support node 26. That why you might see errors like the ones below:

- <details><summary> During `plugins install`: (click here) ⬇️ </summary>
    image
    </details>
- <details><summary>During `require("ibm_db")`: (click here) ⬇️  </summary>
    image
    </details>

**Review Checklist**
I certify that I have:
- [x] updated the changelog
  - `2550`

**Additional Comments**


packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
```diff
@@ -57,12 +57,25 @@ export function installPackages(npmPackage: string, npmArgs: INpmInstallArgs, ve
...
         }
     }
+    if (allowScripts != null) {
+        const allowedPackages = allowScripts.split(",").map((pkg) => pkg.trim()).filter((pkg) => pkg.length > 0);
+        if (allowedPackages.length > 0) {
+            // npm 12 blocks install scripts unless the package is in this list, so we pass the
+            // list to npm and let npm decide. Older versions of npm do not know this option and
+            // ignore it, so they keep working the same way. We add nothing to the command when
+            // the user did not name any package.
+            args.push(`--allow-scripts=${allowedPackages.join(",")}`);
```
MarkAckert, line 76: NPM 12's behavior gets worse: given a package with bundleDependencies, even when `--allow-scripts` is set the transitive dependency lifecycle scripts are not called.

This solution works with the db2 plugin today installed from remote registries where we're using shrinkwrap. When the db2 plugin shifts to bundleDependencies instead of shrinkwrap, this will stop working, as `ibm_db` lifecycle scripts are never called. 

When we see an `allowScript` for a package, I think we need to find the package in the plugin's node_modules and manually execute the NPM lifecycle to maintain expected behavior with bundleDependencies. 

I did some quick tests in my local environment which verified the problematic behavior w/ NPM 12 and Zowe installing a DB2 plugin w/ bundleDependencies. I used a custom build but this should be reproducible with the offline installable distribution as well

CBforZ: That's a pretty frustrating limitation... I wonder if they have written anywhere about why. Maybe they assume it's not needed if you are bundling the dependencies but  as we see with db2 there are situations where that's not the case.  It's not super fun that we would have to crawl the filesystem and  spawn commands. Would we want to search for which lifecycle commands are actually defined in the package.json of the plugin?

CBforZ (commented):
The changes look good, but with Mark's comment, should we address  the need to  manually run scripts for bundled dependencies now? Or split that out into another PR?  Since we have to update to bundledDependencies as part of the npm 12 compatibility (right?) , we'll have to address it either way it seems but I can see splitting that work into another PR if you prefer

packages/imperative/src/imperative/src/plugins/cmd/install/install.definition.ts
```diff
@@ -64,6 +64,19 @@ const verboseDescription =
     "Specifies that verbose output is printed for npm install.\n" +
     "This may be useful for debugging errors during plugin installation.";
 
+const allowScriptsDescription =
+    "A comma-separated list of package names that are allowed to run their npm install scripts " +
+    "(preinstall, install, postinstall, and prepare). The names must match exactly. Zowe passes " +
```
anaxceron: don't need "exactly" as matching = matching exactly

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
MarkAckert: I don't remember seeing doc covering this behavior 😞 

For the lifecycle commands, imo we should issue the exact same lifecycle commands NPM would. I believe that's just `preinstall`, `install`, and `postinstall` from my research.

traeok, line 76: We may want to have some integration or unit tests around using `--allow-scripts` with a comma-separated list. My concern: I noticed weird behavior when passing the packages in this way when testing #2838 - it seemed to ignore every package after the first one before the comma. This may be a bug with `npm` itself, but worth evaluating in action before considering it ready for prod.

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
```diff
@@ -57,12 +57,25 @@ export function installPackages(npmPackage: string, npmArgs: INpmInstallArgs, ve
...
+    // dashes (--allow-scripts) and does not match the property name.
+    const { allowScripts, ...registryArgs } = npmArgs;
+    for (const [k, v] of Object.entries(registryArgs)) {
         if (v != null) {
             // If npm arg starts with @ like @zowe:registry, must use = as separator
             args.push(...k.startsWith("@") ? [`--${k}=${v}`] : [`--${k}`, v]);
         }
     }
+    if (allowScripts != null) {
+        const allowedPackages = allowScripts.split(",").map((pkg) => pkg.trim()).filter((pkg) => pkg.length > 0);
```
traeok, line 70: Consider logging a warning if `allowedPackages` is empty so that the no-op (omitting the option) is visible to the user.

packages/imperative/src/imperative/src/plugins/cmd/update/update.definition.ts
```diff
@@ -32,6 +32,19 @@ const loginDescription =
...
     "https://docs.npmjs.com/cli/adduser for NPM < 9";
 
+const allowScriptsDescription =
+    "A comma-separated list of package names that are allowed to run their npm install scripts " +
+    "(preinstall, install, postinstall, and prepare). The names must match exactly. Zowe passes " +
+    "this list to npm as its own --allow-scripts option.\n" +
+    "\n" +
+    "npm 12 blocks these scripts unless the package is in the list. This breaks plug-ins that " +
+    "have a dependency that must be built during install, such as the Db2 plug-in and its " +
+    "ibm_db dependency. Older versions of npm do not know this option and ignore it, so using " +
```
traeok: We may want to clarify that the older versions of `npm` may still print a warning that the command option `--allow-scripts` is unknown, but that the warning is a false positive. See example output below when using `npm i --allow-scripts` on `npm<12`:

```log
npm warn Unknown cli config "--allow-scripts". This will stop working in the next major version of npm.
```

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
CBforZ: I spoke with Fernando offline and where we landed was we may want to just document that --allow-scripts works when installing over registries but may not work when installing from a bundled tarball, and be sure that the DB2 plugin's post install hook clarifies to the user the steps they can take if the ibm_db package is not installed (how to manually trigger the install) 

I would like to avoid crawling the node_modules looking for all instances of a package and searching for which lifecycle scripts it defines if we can

Let me know if I've summarized accurately, Fernando

traeok (commented):
Thanks for working on this Fernando! I agree with Mark's concern about the open issue with Db2 - it could be resolved in its own PR, but I'm okay with whatever path we choose for fixing that. I left a few comments, one around tests for the new option, an edge-case for `allow-scripts`, and a suggestion around the expected behavior for `npm<12` in the doc 👍

packages/imperative/src/imperative/src/plugins/cmd/install/install.definition.ts
zFernand0: resolved in 08c460882

packages/imperative/src/imperative/src/plugins/cmd/PluginCmdConstants.ts
zFernand0, line 1: Hey @anaxceron,
Could you take a quick look at this file too? 🙏

zFernand0: <details><summary>Here is what help text looks like for the `plugins install` command (click here) ⬇️ </summary>
image
</details>

<details><summary>And here is the help text for the `plugins update` command (click here) ⬇️ </summary>
image
</details>

Images updated as of: e94f1118e

packages/imperative/src/imperative/src/plugins/cmd/update/update.definition.ts
zFernand0: Added the following sentence:

`Older versions of npm may show a warning about the --allow-scripts option being unknown, those warnings can be safely ignored.`

resolved in e94f1118e

packages/imperative/src/imperative/src/plugins/cmd/PluginCmdConstants.ts
anaxceron: My comments -- lmk if this is confusing b/c it probably is:

top image

POSITIONAL ARGUMENTS

Paragraph 1: Just one period after "etc."

2: For example, there is a local plug-in in folder `test-plugin` that you want to install. Specify the ... 

4: If you omit ..., the install command looks for

5: If the `plugin` argument is omitted, the plugins.json file determines which plug-ins are installed. ... see the --file entry in the Options section.

OPTIONS

1: The location of a ... .

2: All plug-ins specified ... are installed to the base CLI and the contents are placed into ... .

3: If you do not .... the default plugin.json file (blahblah) is used. ...

6: (start a new line) When omitted, credentials from the .npmrc file are used.

8: Verbose output is ....

9: This can be useful for debugging errors during plug-in installation.

11: Older versions of npm may show a warning about the --allow-scripts option being unknown. This warning can be safely ignored.

Sometimes you use contractions. Sometimes you don't. Pick one style and apply it consistently. 

bottom image

POSITIONAL ARGUMENTS

2: If omitted, no action is taken.

3: When omitted, the value returned by ...

5: Same as above.

8: Same as above.

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
zFernand0: Agree that it'd be good to have a test for verifying that multiple packages are properly parsed and passed to each plugin's `npm install` command 🙏 

resolved in 421acd7cb

---

image

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
zFernand0: Added a warning for empty `--allow-scripts` 🙏 

resolved in 421acd7cb

packages/imperative/src/imperative/src/plugins/cmd/PluginCmdConstants.ts
zFernand0: <details><summary>Full list of requirements: (click here) ⬇️ </summary>

> My comments -- lmk if this is confusing b/c it probably is:
> 
> top image
> 
> POSITIONAL ARGUMENTS
> 
> ✅ Paragraph 1: Just one period after "etc."
> 
> ✅ 2: For example, there is a local plug-in in folder `test-plugin` that you want to install. Specify the ...
> 
> ✅ 4: If you omit ..., the install command looks for
> 
> ✅ 5: If the `plugin` argument is omitted, the plugins.json file determines which plug-ins are installed. ... see the --file entry in the Options section.
> 
> OPTIONS
> 
> ✅ 1: The location of a ... .
> 
> ✅ 2: All plug-ins specified ... are installed to the base CLI and the contents are placed into ... .
> 
> ✅ 3: If you do not .... the default plugin.json file (blahblah) is used. ...
> 
> ✅ 6: (start a new line) When omitted, credentials from the .npmrc file are used.
> 
> ✅ 8: Verbose output is ....
> 
> ✅ 9: This can be useful for debugging errors during plug-in installation.
> 
> ✅ 11: Older versions of npm may show a warning about the --allow-scripts option being unknown. This warning can be safely ignored.
> 
> ✅ Sometimes you use contractions. Sometimes you don't. Pick one style and apply it consistently.
> 
> bottom image
> 
> POSITIONAL ARGUMENTS
> 
> ✅ 2: If omitted, no action is taken.
> 
> ❓ 3: When omitted, the value returned by ...
> 
> ❓ 5: Same as above.
> 
> ❓ 8: Same as above.


</details>

resolved in bc98dc30e

<details><summary>Help for`plugins install` command  (click here) ⬇️ </summary>
image
</details>

<details><summary>Help for `plugins update` command (click here) ⬇️ </summary>
image
</details>

anaxceron: Re: 3: When omitted, the value returned by ...: Referring to the first paragraph under the --registry entry.

Re: 5: Same as above.: Referring to making the same edits in the same definition (registry) that I called out in the top image.: The npm registry that is used when installing remote packages. **When omitted**, the value returned blahblahblah. 

Re: 8: Same as above.: Referring to making the same edits in the same definition (allow scripts) that I called out in the top image. Seems like you already made the edits to the content in the bottom image.

zFernand0: <details><summary>`plugins install` (click here) ⬇️ </summary>
image
</details>

<details><summary>`plugins update` (click here) ⬇️ </summary>
image
</details>

resolved in aee46ba75

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
zFernand0: This comment should be fully addressed with a warning printed out by the db2 plug-in when the `ibm_db` package cannot be properly required.
- https://github.com/zowe/zowe-cli-db2-plugin/pull/242

packages/imperative/src/imperative/src/plugins/cmd/PluginCmdConstants.ts
traeok, line 1: Thanks for adding this constants file, this is definitely cleaner 😋

traeok (approved):

packages/imperative/src/imperative/src/plugins/cmd/PluginCmdConstants.ts
anaxceron: Noticed one more thing -- NPM vs npm. Decide on one and use it consistently.

zFernand0:
One final commit incoming 🙏 

> Noticed one more thing -- NPM vs npm. Decide on one and use it consistently

---

UPDATE:
resolved in 1c937d7bd

MarkAckert (approved):
LGTM and OK addressing `--allow-scripts` for packages with `bundleDependencies` in another PR.

anaxceron (approved):
🐈

zFernand0:
All system tests related to the changes in this PR passed 🙏 
Failures seen in this PR (see `2550`) were also seen on the `master` branch (see `2552`)

PR #2840, opened 2026-08-06
Fix: npm pack command - output format change

pujal0909:
**What It Does**

As a result of npm 12, the `npm pack` command fails since with the change, it returns output in an object format rather than array. The fix addresses that change.

**How to Test**

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**


packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
```diff
@@ -132,7 +132,8 @@ export function getPackageInfo(pkgSpec: string): { name: string, version: string
 
         // parse the json output of the npm pack command
         try {
-            packageName = JSON.parse(execOutput)[0].name;
+            const parsedOutput = JSON.parse(execOutput);
```
MarkAckert, line 135: nit: I *think* this approach is safe, but slightly prefer an explicit test of NPM's major version and switching our parsing logic based on that. 
It gives us a cleaner hook if NPM ever decides to change their JSON response format again in a future version, and clarifies what this code is solving at the same time.

anaxceron (approved):

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
t1m0thyj: @MarkAckert Apologies, I suggested a different direction which I thought I had run past you but maybe forgot.

I'd prefer to avoid adding extra command invocations if we can help it, since in some environments (particularly on Windows) each command may be slow.

Would adding a comment to explain the reason for the check be sufficient to clarify?

MarkAckert: No worries, I understand the other side as well, and changes to the response *should* be exceedingly rare. Adding a comment is sufficient for clarification. I'm good to approve after that 👍 .

MarkAckert (approved):

packages/imperative/src/imperative/src/plugins/utilities/NpmFunctions.ts
t1m0thyj: Thanks Mark! @pujal0909 Can we add a brief explanatory comment here?

t1m0thyj (approved):

t1m0thyj (changes requested):
Please fix unit tests failing on Windows - seems like we may need to replace hardcoded forward slashes (`/`) in tests with `path.sep`.

t1m0thyj (approved):

zFernand0 (approved):
LGTM! 😋 
---
Looks similar to doing something like `jq '.[keys_unsorted[0]].name'` 😋

PR #2846, opened 2026-08-13
Fix script to bundle CLI deps in npm package

t1m0thyj:
**What It Does**

Refactor the script for bundling CLI deps:
* Copies `node_modules` instead of using hard links which may not get resolved by `npm pack` correctly
* Uses `npm ls` for dependency resolution instead of building the deps tree ourselves
* Excludes platform-specific native binaries compiled by `node-gyp` from the bundle

**How to Test**

Download the TGZ from the comment below. Verify that it can be installed with `npm install -g <pathToTgz>`

You can also verify there are no hard links inside the TGZ with this command: `tar -tvzf <pathToTgz> | grep "^h"`

**Review Checklist**
I certify that I have:
- [ ] updated the changelog

**Additional Comments**


CBforZ (approved):
I was able to verify the testing steps from the PR description.

jace-roell (approved):

PR #2847, opened 2026-08-13
Restrict npm version for Dependabot and CLI devs

t1m0thyj:
**What It Does**

* Defines `packageManager` in top-level `package.json` which should influence the npm version used by Dependabot
  * This property must define an exact version and does not support ranges so we use the latest `npm@10`
* Defines `engines` in top level package.json so that a range of npm versions is allowed for CLI devs
  * If you try to install with Node 24 + npm 11, the command succeeds but displays a warning

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,6 +45,10 @@
         "ts-node": "^7.0.1",
         "typedoc": "^0.26.3",
         "typescript": "^5.5.3"
+      },
+      "engines": {
+        "node": ">=20.9.0",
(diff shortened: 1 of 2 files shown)


zFernand0 (approved):

CBforZ (approved):
Nice find , let's give it a try. Hopefully prevents us from needing to fiddle with the lockfiles

PR #2872, opened 2026-08-24
Restrict access permissions on Windows TLS named pipe

pujal0909:
**What It Does**

Restricts DACL on Windows local named pipes so that it's access owner only.

**How to Test**

1. On Windows machine (or VM), create a non administrator user in settings
2. In Powershell, get the SID of that user by running : `Get-LocalUser | Select-Object Name, SID`
3. Modify `tls_pipe.rs` by replacing the `OW` on line 24 (const OWNER_ONLY_SDDL) with the SID
4. Then build the secrets package and run a command (such as listing datasets) and see that an error occurs restricting access

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/packages/secrets/core/src/os/win/tls_pipe.rs
+++ b/packages/secrets/core/src/os/win/tls_pipe.rs
@@ -46,6 +46,28 @@ fn schannel_error_hint(code: i32) -> &'static str {
     }
 }
 
+fn owner_only_security_descriptor() -> Result<PSECURITY_DESCRIPTOR, KeyringError> {
+    let mut sddl_wide: Vec<u16> = OWNER_ONLY_SDDL.encode_utf16().collect();
+    sddl_wide.push(0);
+
+    let mut sec_desc: PSECURITY_DESCRIPTOR = std::ptr::null_mut();
+    let conv_res = unsafe {
+        ConvertStringSecurityDescriptorToSecurityDescriptorW(
+            sddl_wide.as_ptr(),
+            SDDL_REVISION_1,
+            &mut sec_desc,
+            std::ptr::null_mut(),
+        )
+    };
+
+    if conv_res == 0 {
+        let err = unsafe { GetLastError() };
+        return Err(KeyringError::Os(format!("Failed to convert SDDL to security descriptor. Error: {}", err)));
+    }
+
+    Ok(sec_desc)
+}
+
 pub fn create_tls_pipe(
     remote_host: &String,
     remote_port: u16,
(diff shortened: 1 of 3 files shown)


t1m0thyj (approved):
LGTM, thanks @pujal0909! Followed the steps to test - cert auth works with the default security descriptor (`OW`), but fails with an `EPERM` error when changed to a different user SID 👍

packages/secrets/core/src/os/win/tls_pipe.rs
```diff
@@ -60,19 +69,47 @@ pub fn create_tls_pipe(
...
             0, // default timeout
-            std::ptr::null(), // default security attributes
+            &mut sec_attrs,
         )
     };
 
+    unsafe {
+        LocalFree(sec_desc as _);
+    }
+
```
traeok, line 119: I think this added function call makes sense, but consider the new impact on lines 113-116: the details from `GetLastError` could now come from any `LocalFree` errors, rather than the pipe-related error we intend to display. I'd recommend updating the logic so we call `GetLastError` before `LocalFree`:

```rust
let pipe_handle = unsafe { CreateNamedPipeW(/* ... */, &sec_attrs) };
let create_err = unsafe { GetLastError() };
unsafe { LocalFree(sec_desc as _) };
if pipe_handle == INVALID_HANDLE_VALUE {
    return Err(KeyringError::Os(format!("Failed to create named pipe. Error: {}", create_err)));
}
```

packages/secrets/core/src/os/win/tls_pipe.rs
```diff
@@ -232,3 +269,31 @@ pub fn create_tls_pipe(
     Ok(socket_path_ret)
 }
 
+#[cfg(all(test, target_os = "windows"))]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_owner_only_sddl_conversion() {
```
traeok, line 283: The test is duplicating production code, would it be possible to expose a helper function that we can call from the test? This ensures that the test still passes even if changes are made to the underlying logic.

traeok (changes requested):
Thanks for working on this Pujal, I left a couple requests for changes but the code makes sense

packages/secrets/core/src/os/win/tls_pipe.rs
```diff
@@ -6,13 +6,22 @@ use std::io::{Read, Write};
...
 const PIPE_TYPE_BYTE: u32 = 0x00000000;
 const PIPE_READMODE_BYTE: u32 = 0x00000000;
 const PIPE_WAIT: u32 = 0x00000000;
+const PIPE_REJECT_REMOTE_CLIENTS: u32 = 0x00000008;
 const FILE_FLAG_FIRST_PIPE_INSTANCE: u32 = 0x00080000;
+const SDDL_REVISION_1: u32 = 1;
+
+// SDDL granting full control to the creating process's owner only; all other
+// principals (including Everyone / ANONYMOUS LOGON) get no access.
+const OWNER_ONLY_SDDL: &str = "D:P(A;;GA;;;OW)";
```
CBforZ, line 24: This is my first time encountering SDDL 


D:  specifies that the SDDL is of the "dacl flags" format
P: The SE_DACL_PROTECTED flag is set.

`(A;;GA;;;OW)`  Would be an ACE string which are in the format: `ace_type;ace_flags;rights;object_guid;inherit_object_guid;account_sid;(resource_attribute)`

GA: is rights which is SDDL_GENERIC_ALL 
So the account_sid is : 
"OW" | SDDL_OWNER_RIGHTS

https://learn.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-string-format
https://learn.microsoft.com/en-us/windows/win32/secauthz/ace-strings

So from my reading this seems right, protect except for the owner account SID who has all rights

CBforZ (approved):
Looks like you addressed the feedback and  the code seems good to me

awharn (approved):

traeok (approved):

PR #2875, opened 2026-08-28
refactor: add Secrets SDK message for postinstall

### LINKED ISSUE #2777: Evaluate alternatives for `install` npm lifecycle script in Secrets SDK
The Secrets SDK package has an `install` script which checks for prebuilds and falls back to building from source if they are missing:
https://github.com/zowe/zowe-cli/blob/71e5573e88f235b64360b8697b9cd8d070ac85f7/packages/secrets/package.json#L42

In npm 12, lifecycle scripts will be disabled by default and controlled by the `allowScripts` option:
> **Install scripts are off by default.** `preinstall`, `install`, and `postinstall` from dependencies won't run unless explicitly allowed.

<sup>Source: https://github.com/orgs/community/discussions/198547</sup>

Consider alternatives for the `install` script - in most cases it may not be needed if prebuilds are valid.

traeok:
**What It Does**

Resolves #2777 

Adds a detailed error message if the prebuild resolution for Secrets SDK fails as part of running Zowe CLI's `postinstall` script.

**How to Test**

- Install and build the packages in this branch: `npm install && npm run build`
- Package the CLI using `cd packages/cli && npm pack`
- Using the newly-packaged tarball, `npm install -g <tarball> --allow-scripts=@zowe/cli --foreground-scripts`
- See that the new `validateSecretsPrebuild.js` script runs as part of `postinstall` (the script output for `postinstall` is streamed to console because of `--foreground-scripts`)

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [x] created/ran system tests (job 2598)

**Additional Comments**

This is largely for folks that are installing Zowe CLI on unsupported [target triples](https://wiki.osdev.org/Target_Triplet) 😋


traeok:
@anaxceron Hey Ana, I'd like to get your thoughts on the messaging in the `validateSecretsPrebuild.js` file, as the warnings may appear when users install Zowe CLI on unsupported platforms. I can show an example in-person tomorrow if you are curious to see how they are displayed! Thanks 😁

anaxceron:
> @anaxceron Hey Ana, I'd like to get your thoughts on the messaging in the `validateSecretsPrebuild.js` file, as the warnings may appear when users install Zowe CLI on unsupported platforms. I can show an example in-person tomorrow if you are curious to see how they are displayed! Thanks 😁

Read through the file and, yep, I would like to see how this displays. Thank ya.

pujal0909 (approved):
tested and LGTM!

packages/cli/scripts/validateSecretsPrebuild.js
```diff
@@ -0,0 +1,231 @@
...
+        return {
+            kind: "noPrebuild",
+            problem: `${SECRETS_PKG}@${version} contains no keyring prebuild for ${target}.`,
+            cause: `The keyring loader expects keyring.${target}.node, which is not present ` +
+                "in this installation, so the module has to be built from source here."
+        };
+    }
+
+    try {
+        const { keyring } = require(pkgDir);
```
zFernand0, line 114: nit: should we throw an error if keyring is undefined here?

That will go to the catch bellow and return the unloadable error (which is still accurate in this scenario)

packages/cli/scripts/validateSecretsPrebuild.js
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/* The repair steps the user should follow, tailored to what we found. We do not
+ * attempt the repair ourselves: rebuilding a native module needs prerequisites
+ * we cannot install, and re-entering npm from inside a lifecycle script is not
+ * safe.
+ */
+function repairSteps(problem) {
+    const isGlobal = process.env.npm_config_global === "true";
+    const cliVersion = readJson(path.join(process.cwd(), "package.json")).version;
```
zFernand0, line 139: nit: Should we use `__dirname` (` + "/.."`) instead of `cwd()`?

`cwd` should always be the root of the `@zowe/cli` package when running the post install script. It's only in the scenario where someone wants/needs to trigger the post install (i.e. npm12) by themselves inside node_modules and doesn't navigate to the proper `.../@zowe/cli`, but instead runs the script manually from a custom/random/temp directory.

zFernand0 (approved):
LGTM! 😋 
---
I'm on the fence about the `cwd` v. `__dirname` since this validation script shoudl always run from the root of the CLI, and I'm sure we are guilty of using cwd somewhere else already, so not requesting changes for this 🙏

packages/cli/scripts/validateSecretsPrebuild.js
traeok: Since the `keyring.getPassword` access is inside the try block, if `keyring` is undefined, it will throw a `TypeError`. That gets caught right below and returns the unloadable error anyway, so an explicit check/throw here isn't strictly necessary.

packages/cli/scripts/validateSecretsPrebuild.js
traeok: You made a good point regarding the impact of using `cwd()` instead of `__dirname`, so I will update it to use dirname 😋 Thanks

traeok: Updated to use `__dirname` in 23986d1 - `resolveQuietly` is the only instance left that keeps paths `[__dirname, process.cwd()]` to ensure proper package resolution

traeok: Had to revert this change as I noticed some issues 😓 happy to discuss more offline

traeok:
Sorry for the delay @anaxceron - here are the screenshots of messages shown when Zowe CLI cannot use credential storage:

---

When Secrets SDK is missing:
image

---

When a prebuild is missing for the user's platform for Secrets SDK:
image

---

When a prebuild is broken for the user's platform (missing functions, corrupted/modified build, etc)

image

---

When the prebuild is present but fails to load:

image

anaxceron:
@traeok -- thanks for the screenshots. I have minor edits for only the first screenshot:

When Secrets SDK is missing:

1. Re-run the **WHAT** install … (I think we should clarify what specific install to redo.)

2: Step 1 (capitalize “step”)

traeok:
> @traeok -- thanks for the screenshots. I have minor edits ....

No problem, thanks for the feedback! Addressed in 883ffb9c 👍

anaxceron (approved):
🌴

PR #2882, opened 2026-09-02
proposal(schema): restrict additional fields in profile object

traeok:
**What It Does**

Happy to postpone this to v4 if there are any objections to doing this now, but its something that came to mind based on our discussion 😋 

This does not have any effect on current behavior for interpreting the config structure. Its mostly for improving the user experience (the IntelliSense underline for warnings/errors) when users accidentally place profile properties outside of the `properties` object.

This does not change any pre-existing config/schema pairs right now. This **will** change if/when the functionality is added to update the schema after upgrading to a newer CLI version.

**How to Test**

- `zowe config init` in a project directory should create a new config/schema
- open the config in VS Code and add a property outside of the `properties` object
- notice the squiggly lines now appear for those properties

**Review Checklist**
I certify that I have:
- [x] updated the changelog


CBforZ:
I like it! Tested locally and it had the intended effect. I think this will be useful to clarify where fields like user & password should go.

zFernand0 (approved):

CBforZ (approved):

pujal0909 (approved):
Seeing the intellisense warnings. LGTM!

PR #2889, opened 2026-09-10
fix(zosfiles): handle malformed chtag list response with safe default

### LINKED ISSUE #2888: bug : Utilities.isFileTagBinOrAscii / applyTaggedEncoding crash on malformed chtag list response instead of safe default
### Describe the bug

`Utilities.isFileTagBinOrAscii()` and `Utilities.applyTaggedEncoding()` (`packages/zosfiles/src/methods/utilities/Utilities.ts:119-166`) crash with unhandled `SyntaxError` / `TypeError` when the `chtag list` (`putUSSPayload`) response body is non-JSON, literal `null`, empty, or has a misshapen `stdout` field:

```ts
const jsonObj = JSON.parse(response.toString());   // :127,:156 — throws SyntaxError on HTML/empty
if (Object.prototype.hasOwnProperty.call(jsonObj, "stdout")) {  // throws on JSON "null"
    const stdout = jsonObj.stdout[0];              // undefined on stdout:[] → undefined.indexOf/trim() TypeError
```

`Download.ussFile()` calls `applyTaggedEncoding()` on every download without explicit `--binary`/`--encoding` (`Download.ts:699`), so a single malformed tag response crashes the download instead of falling back to the default z/OSMF conversion behavior.

Searched `#207` (`.zosattributes` feature) and `#740` (respect tagging feature) — neither files this crash; no existing duplicate found.

---

#### 1. Steps to reproduce
Command issued:
```bash
zowe zos-files download uss-file "/u/user/file"
```
*(Issued without `--binary`/`--encoding`, so USS tag discovery is triggered against a host/proxy returning a malformed `chtag list` body).*

Alternatively via SDK test:
```ts
jest.spyOn(Utilities, "putUSSPayload").mockResolvedValue(Buffer.from("<html>error</html>"));
await Utilities.isFileTagBinOrAscii(session, "/u/user/file"); // throws unhandled exception
```

**Full crash matrix (mocked `putUSSPayload` return):**
* `null` / `undefined` &rarr; `TypeError` on `.toString()`
* `""` / `"<html>...</html>"` &rarr; `SyntaxError` on `JSON.parse`
* `"null"` &rarr; `TypeError` on `Object.prototype.hasOwnProperty.call(null, ...)`
* `{ stdout: [] }` &rarr; `TypeError` on `undefined.indexOf()` / `.trim()`
* `{ stdout: "str" }` / `{ stdout: [123] }` &rarr; inverted branching / `TypeError`

#### 2. Expected results
* `isFileTagBinOrAscii` gracefully handles invalid payloads and returns `false`.
* `applyTaggedEncoding` leaves `options.binary` and `options.encoding` unmutated and returns cleanly — matching the existing `{}` (no-`stdout`) safe-default behavior.

#### 3. Actual results
Unhandled `SyntaxError` or `TypeError` propagates up the stack and fails the entire operation, violating the documented `@throws {ImperativeError}` / return contracts.

#### 4. Applicable log files
N/A (unit-reproducible with mocked `putUSSPayload`; no mainframe connection or daemon needed).

---

### Describe your environment

* **Zowe CLI version installed (`zowe --version`):** `8.36.1` (`@zowe/zos-files-for-zowe-sdk` source audit; reproducible on V2 / V3 SDK line)
* **Installed plug-ins and their version numbers (`zowe plugins list --short`):** N/A
* **Node.js and NPM versions installed (`node --version`, `npm --version`):** Node >= 20.9.0, NPM ^10 (per repository `engines`)
* **Environment variables in use:** N/A
* **Operating system and version:** Any (platform-independent defect)
* **Shell/terminal:** N/A
* **Daemon mode enabled or disabled:** N/A (identical SDK path used in daemon mode)

---

### Additional context

**Proposed Fix (minimal, non-breaking):**
Introduce a shared `private static parseChtagStdout(response: any): string | undefined` helper method that encapsulates:
1. Null/undefined response guard.
2. `try/catch` wrapping `JSON.parse`.
3. Safe object check + `Array.isArray(jsonObj.stdout)` + `typeof jsonObj.stdout[0] === "string"` validation.
4. Early return of the safe fallback if parsing fails.

This maintains byte-identical logic for valid tags while preventing unhandled runtime crashes on malformed responses. Happy to submit a PR with this fix, accompanying unit test coverage across the crash matrix, and a CHANGELOG entry.

AdeshDeshmukh:
**What It Does**

Fixes #2888 — and honestly, this one bit me while reading through the USS download path, so I wanted to fix it properly.

**The problem in plain terms:** every time you download a USS file without passing `--binary`/`--encoding`, the CLI quietly asks z/OSMF for the file's tag (`chtag list`) to decide whether to convert it. The two helpers behind that — `Utilities.isFileTagBinOrAscii()` and `Utilities.applyTaggedEncoding()` (`packages/zosfiles/src/methods/utilities/Utilities.ts`) — assumed the response would always be neat JSON with a `stdout` array. If anything else came back (an HTML error page from a proxy, an empty body, a literal `null`, an empty `stdout: []`), the whole download blew up with a raw `SyntaxError`/`TypeError` instead of just carrying on with the default behavior.

**The fix:** a small shared `parseChtagStdout()` helper that null-guards the response, wraps `JSON.parse` in try/catch, and validates the shape (`Array.isArray` + `typeof string`). On anything unparseable, callers now fall back to the safe default — `isFileTagBinOrAscii` returns `false`, `applyTaggedEncoding` leaves your `binary`/`encoding` options untouched — exactly like the existing no-`stdout` (`{}`) case already did. Valid-tag handling is byte-for-byte unchanged, so no behavior change for happy paths, no API change.

**How to Test**

No mainframe needed — everything is unit-testable with a mocked `putUSSPayload`:

1. `npx jest packages/zosfiles/__tests__/__unit__/methods/utilities/Utilities.unit.test.ts --coverage false` → 45/45 pass, including 14 new cases covering the full crash matrix (`null`, empty body, HTML, `"null"`, `stdout: []`, non-array `stdout`, non-string entry) for both methods.
2. `npx eslint src/methods/utilities/Utilities.ts` (from `packages/zosfiles`) → clean.
3. `npx tsc --noEmit -p tsconfig.json` (from `packages/zosfiles`) → clean.
4. Sanity check the unhappy path manually, e.g. mock `putUSSPayload` to resolve `Buffer.from('<html>error</html>')` and confirm `isFileTagBinOrAscii` resolves `false` instead of throwing.

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

- I searched existing issues before filing (#207 and #740 are feature requests about tagging, not this crash) — no duplicate.
- On system tests: not applicable here since the defect is fully covered by mocked unit tests (no mainframe path changed) — happy to run them if you'd like.
- Happy to adjust the changelog placement/wording if your release process prefers it elsewhere. Thanks for maintaining this!


AdeshDeshmukh:
@CBforZ @traeok PTAL.

traeok:
Please update the PR to use the full, unabridged PR checklist.

Thanks

AdeshDeshmukh:
Thanks for the pointers..! 
Checklist is restored verbatim , and I've rebased onto latest master..!

traeok:
Hi @AdeshDeshmukh can you clarify what steps you encountered to reproduce this error? Is this a known issue with the APIs or just some preventative measures?

AdeshDeshmukh:
> Hi @AdeshDeshmukh can you clarify what steps you encountered to reproduce this error? Is this a known issue with the APIs or just some preventative measures?


I found this reading the download path, not from a live mainframe incident. So I did the next most honest thing....

I wrote 14 unit tests driving the real code path ( putUSSPayload → the chtag list parse in both methods) with malformed bodies, and ran them against the unmodified code first. This is what came back :

IMAGE of failing run ...  the 13 red ✕ lines + "Tests: 13 failed, 1 passed"

SCR-20260921-tana


Every one of those is an unhandled throw out of shipped logic — SyntaxError: Unexpected token '<' on HTML bodies, SyntaxError: Unexpected end of JSON input on empty ones, TypeError: Cannot read properties of null on null bodies, and TypeError variants on empty/misshapen stdout. And this isn't some corner helper, either: every zowe zos-files download uss-file without --binary/--encoding walks straight through applyTaggedEncoding, so any proxy returning an error page or any truncated body takes down the whole download.

Then the same run after the fix:

IMAGE of passing run — "Test Suites: 1 passed, Tests: 46 passed, 46 total"

SCR-20260921-tbfq


Same 14 cases, all green, and the other 32 existing tests untouched — valid-tag behavior is byte-identical, since the fallback is just the safe default the code already used for the {} case. Full output above is verbatim from my terminal about ten minutes ago...!

The one thing I'll openly label as inferred rather than proven is how often production serves such bodies .... I can't put a number on that. But the defect itself is on record in the first screenshot, and the fix costs nothing on happy paths. 

Happy to Move in the direction you would like to...!

traeok (approved):
This is definitely more of a defensive measure, but I think the new logic is safer compared to what we had before.

System tests are passing and changes LGTM, thanks @AdeshDeshmukh

packages/zosfiles/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the Zowe z/OS files SDK package will be documented in this file.
 
+## Recent Changes
+
+- BugFix: Fixed `Utilities.isFileTagBinOrAscii` and `Utilities.applyTaggedEncoding` crashing with unhandled `SyntaxError`/`TypeError` on malformed `chtag list` responses. Both methods now fall back to the safe default (return `false` / leave download options unmutated). [#2888](https://github.com/zowe/zowe-cli/issues/2888)
```
```suggestion
- BugFix: Fixed `Utilities.isFileTagBinOrAscii` and `Utilities.applyTaggedEncoding` crashing with unhandled `SyntaxError`/`TypeError` on malformed `chtag list` responses. Both methods now fall back to the safe default (return `false`/leave download options unmutated). [#2888](https://github.com/zowe/zowe-cli/issues/2888)
```

anaxceron (commented):
Left a very minor edit 🐱

packages/zosfiles/CHANGELOG.md
AdeshDeshmukh: Done..!

anaxceron (approved):

